PR #4, opened 2022-12-13
Basic security measures

alextserepov:
Signed-off-by: Aleksandr Tserepov-Savolainen <alextserepov@gmail.com>


src/scs/basics.md
```diff
@@ -0,0 +1,33 @@
+# Basic Security Measures
+
+## Source code / Version control security
+
+Briefly, the source code security is based on the fact that the source code is Two-person reviewed, version controlled, the history is verified, retained indefinitely. 
+
+### Commit signing
+All the commits to repositories must be GPG-signed. This can be achieved by enabling GPG commit signatures in the config:
```
alextserepov, line 8: @vilvo & @jenninikko Should we open this abbreviations as well? I mean GPG, TLS, TCP?

alextserepov:
First thought that I will redo this one from scratch, as I thought it is better to split it a bit, but now the process is more clear after discussing with Jenni. Let's keep it the way it is for now. Comments are welcomed and much appreciated. :)

src/scs/basics.md
brianmcgillion: we should at least once, however, i think we should maintain a glossary in the appendix for all acronyms. then the question is if we should create links to that entry in the glossary when an abbreviation is encountered for the first time on a page/?

jenninikko: The glossary was in my plans. We can stick to the rule as when you use any acronym or a term for the first time, you make a link to the glossary (glossary.md). Then in the glossary itself, in addition to the definition, it will be possible to provide links to third-party sources for additional information, not to overload sections.

I also suggest not making terms out of obvious abbreviations, such as OS -- operating systems.

This information will be added to some guidelines.

src/scs/basics.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Software Signing
+
+Software signing is an important measure to validate the author and ensure that the code has not been altered on the way from the developer to the customer. Nix tooling is offering means to sign the derivations using libsodium with EdDSA, however, as the modular system is assumed, scripts need to be developed to support signing mechanisms in an absence of Nix tooling.
+
+As of today, the software image is singed only on the binary cache per request. Which means that it is possible to request a software signature separately as part of nar-info. 
+
+At the same time, the code is not signed on the way from repository to Hydra and from Hydra to binary cache, leaving those parts insecure.
+
+## Data encryption in transit
```
jenninikko, line 31: Data Encryption in Transit

src/scs/basics.md
```diff
@@ -0,0 +1,33 @@
...
+  + Require status checks before merging
+  + Require conversation resolution before merging
+  + Require signed commits (see commit signing)
+  + Deletions should be forbidden (req: immutable history)
+
+## Software Signing
+
+Software signing is an important measure to validate the author and ensure that the code has not been altered on the way from the developer to the customer. Nix tooling is offering means to sign the derivations using libsodium with EdDSA, however, as the modular system is assumed, scripts need to be developed to support signing mechanisms in an absence of Nix tooling.
+
+As of today, the software image is singed only on the binary cache per request. Which means that it is possible to request a software signature separately as part of nar-info.
```
jenninikko, line 27: singed -> signed

This means that...

... as part of nar-info -- is it slang?

src/scs/basics.md
```diff
@@ -0,0 +1,33 @@
...
+  + Require signed commits (see commit signing)
+  + Deletions should be forbidden (req: immutable history)
+
+## Software Signing
+
+Software signing is an important measure to validate the author and ensure that the code has not been altered on the way from the developer to the customer. Nix tooling is offering means to sign the derivations using libsodium with EdDSA, however, as the modular system is assumed, scripts need to be developed to support signing mechanisms in an absence of Nix tooling.
+
+As of today, the software image is singed only on the binary cache per request. Which means that it is possible to request a software signature separately as part of nar-info. 
+
+At the same time, the code is not signed on the way from repository to Hydra and from Hydra to binary cache, leaving those parts insecure.
```
jenninikko, line 29: ... on the way from **a** repository to Hydra**,** and from Hydra to **the** binary cache...

jenninikko (changes requested):
@alextserepov my comments are below.

src/scs/basics.md
alextserepov: "nar" = NixOS Archive. "nar-info" an API command (GET Request) offered by binary cache to get the information about particular nar file.

jenninikko (approved):

PR #6, opened 2022-12-13
Initial commit of PKI chapter skeleton

alextserepov:
Just the intro. More to be added. 

Signed-off-by: Aleksandr Tserepov-Savolainen <alextserepov@gmail.com>

### DIFF
--- a/src/scs/pki.md
+++ b/src/scs/pki.md
@@ -1,0 +1,31 @@
+# Public Key Infrastructure
+
+In SCS context a public key infrastructure (PKI) term refers to a system for creation, storage, and distribution of digital certificates which are used to verify that a particular key belongs to a certain entity. PKI is creating and managing a hierarchical set of digital certificates which map public keys to certain entities. Storage and revocation methodologies are to be defined.
+
+The PK of SCS should consist of:
+  + A certificate authority (CA) responsible for storing, issuing and signing the digital certificates
+  + A registration authority (RA) responsible for requesting entity identity verification
+  + A central directory responsible for secure storage of the keys
+  + A Certificate Management System (CMS) managing access to stored certificates
+ 
+## Private Certificate Authority (PCA)
+
+PCA enables creation of private certificate authority hierarchies, consisting of root and subordinate CAs. It issues end-entity X.509 certificates, that are used for: 
+
+  + Encrypted TLS communication channels (data encryption in transit)
+  + Code and image signing
+
+PCA can be established in the cloud or on premise. Initially, openssl based on-prem solution is assumed, however, some of the target projects might consider using commercial cloud solutions.
+
+### CA Hierarchy Options
+
+CA usually consists of:
+  + A Root CA - the root of trust of the entire PKI, responsible for issuing and signing the certificates used by other CAs
+  + A Subordinate CA is typically responsible for issuing end-entity certificates
+
+Typically, there are three types of hierarchies. One-tier, two-tier and three-tier. The hierarchy can be chosen based on the target projects needs and complexity, however it is important to keep in mind that one-tier hierarchy is not recommended for any production use due to the low security, as the compromise of a single CA leads to a compromise of the entire PKI, thus one-tier CA hierarchy is not considered in this document.
+
+In a two-tier hierarchy the root CA and issuing (Subordinate) CAs are separated, thus increasing the security level. This is the simplest production level hierarchy allowing to keep root CA at the most secure and restrictive levels, yet making subordinate access slightly more permissive. This hierarchy is most likely sufficient for most of the target projects.
+
+In a three-tier CA an intermediate CA is placed between root CA and Subordinate (issuing) CA. This is done to further separate the root CA from low-level CA operations. The middle layer (intermediate CA) is only used to sign subordinate CAs that issue the end-entity certificates. 
+
--- a/src/scs/pki.md
+++ b/src/scs/pki.md
@@ -1,31 +1,31 @@
 # Public Key Infrastructure
 
-In SCS context a public key infrastructure (PKI) term refers to a system for creation, storage, and distribution of digital certificates which are used to verify that a particular key belongs to a certain entity. PKI is creating and managing a hierarchical set of digital certificates which map public keys to certain entities. Storage and revocation methodologies are to be defined.
+In the SCS context a public key infrastructure (PKI) term refers to a system for the creation, storage, and distribution of digital certificates which are used to verify that a particular key belongs to a certain entity. PKI creates and manages a hierarchical set of digital certificates which map public keys to certain entities. Storage and revocation methodologies are to be defined.
 
 The PK of SCS should consist of:
-  + A certificate authority (CA) responsible for storing, issuing and signing the digital certificates
-  + A registration authority (RA) responsible for requesting entity identity verification
-  + A central directory responsible for secure storage of the keys
-  + A Certificate Management System (CMS) managing access to stored certificates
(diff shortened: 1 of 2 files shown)


src/scs/pki.md
```diff
@@ -0,0 +1,31 @@
...
+  + A Certificate Management System (CMS) managing access to stored certificates
+ 
+## Private Certificate Authority (PCA)
+
+PCA enables creation of private certificate authority hierarchies, consisting of root and subordinate CAs. It issues end-entity X.509 certificates, that are used for: 
+
+  + Encrypted TLS communication channels (data encryption in transit)
+  + Code and image signing
+
+PCA can be established in the cloud or on premise. Initially, openssl based on-prem solution is assumed, however, some of the target projects might consider using commercial cloud solutions.
```
brianmcgillion, line 18: You could mention the use of a HSM either directly connected {e.g. a YubiHSM} or network based should be employed as they offer the best protection for the keys.

jenninikko, line 18: Let's think together on "...openssl based on-prem solution is assumed...".
Considering that on premise -> on-premises and openssl -> OpenSSL, what do we have?
... the OpenSSL solution deployed on-premises is assumed... -- if we write like this, the meaning is not distorted?

src/scs/pki.md
```diff
@@ -0,0 +1,31 @@
...
+
+### CA Hierarchy Options
+
+CA usually consists of:
```
jenninikko, line 22: Same here: for this list remove all 'a' articles in the items. Do not forget to begin each entry in a bulleted list with a capital letter.

src/scs/pki.md
```diff
@@ -0,0 +1,31 @@
...
+  + A Root CA - the root of trust of the entire PKI, responsible for issuing and signing the certificates used by other CAs
+  + A Subordinate CA is typically responsible for issuing end-entity certificates
```
jenninikko, line 24: Subordinate CA that is responsible for issuing end-entity certificates
(typically here the parasite word)

src/scs/pki.md
```diff
@@ -0,0 +1,31 @@
...
+
+Typically, there are three types of hierarchies. One-tier, two-tier and three-tier. The hierarchy can be chosen based on the target projects needs and complexity, however it is important to keep in mind that one-tier hierarchy is not recommended for any production use due to the low security, as the compromise of a single CA leads to a compromise of the entire PKI, thus one-tier CA hierarchy is not considered in this document.
```
jenninikko, line 26: (typically here the parasite word) :D

jenninikko, line 26: Here it is better to join in one sentence:
There are three types of hierarchies: one-tier, two-tier, and three-tier.

jenninikko, line 26: ...**project's** needs and complexity. 

Let's rewrite the next sentence as:
A one-tier hierarchy is not considered for any production usage due to the low security, as the compromise of a single CA leads to a compromise of the entire PKI.

One-tier CA hierarchy is not considered in this document. -- delete this

src/scs/pki.md
```diff
@@ -0,0 +1,31 @@
...
+
+In a two-tier hierarchy the root CA and issuing (Subordinate) CAs are separated, thus increasing the security level. This is the simplest production level hierarchy allowing to keep root CA at the most secure and restrictive levels, yet making subordinate access slightly more permissive. This hierarchy is most likely sufficient for most of the target projects.
```
jenninikko, line 28: In a two-tier hierarchy, the root CA and issuing (subordinate) CAs are separated for increasing the security level. -- Maybe this is better?

src/scs/pki.md
```diff
@@ -0,0 +1,31 @@
...
+
+In a three-tier CA an intermediate CA is placed between root CA and Subordinate (issuing) CA. This is done to further separate the root CA from low-level CA operations. The middle layer (intermediate CA) is only used to sign subordinate CAs that issue the end-entity certificates.
```
jenninikko, line 30: ...the root...
... the Subordinate...

jenninikko (changes requested):
Hei @alextserepov, thanks for this section. There are wordings that are better to think about together. You will find my comments below <3

src/scs/pki.md
```diff
@@ -0,0 +1,31 @@
+# Public Key Infrastructure
+
+In SCS context a public key infrastructure (PKI) term refers to a system for creation, storage, and distribution of digital certificates which are used to verify that a particular key belongs to a certain entity. PKI is creating and managing a hierarchical set of digital certificates which map public keys to certain entities. Storage and revocation methodologies are to be defined.
```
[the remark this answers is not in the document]
alextserepov: Storage and revocation methodologies depend on the way we implement PKI. For example, if the target project decides to use cloud, then it is more or less mandated by the cloud provider, the project just needs to take the methods into use. Should the project decide to have "on prem" solution, then both storage (and securing the storage) and revocation mechanisms are totally open and must be specially crafted for the particular implementation.

src/scs/pki.md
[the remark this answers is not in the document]
alextserepov: Yes, I need to get that out of my head somehow. Been using that for 10+ years, so it will be a hard journey, but I will learn one day. Thanks! :)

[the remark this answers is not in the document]
alextserepov: But can we use "on-prem" or should it be only "on premises"?

alextserepov: "Initially, the OpenSSL based solution deployed on-premises is assumed" will not change the meaning ... But this sentence might soon change as I will add more text on HSM as per comment from Brian.

src/scs/pki.md
alextserepov: removed "is responsible" and "typically". Though, I should probably reconsider the whole list here, as it might seem that we make a strong accent on two-tier implementation.

src/scs/pki.md
alextserepov: Agreed, that is much better!

src/scs/pki.md
alextserepov: Yes.

src/scs/pki.md
alextserepov: Sure. I will upload a fix for Jenni's comments now and then will take HSM part in a separate commit..

src/scs/pki.md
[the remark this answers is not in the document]
jenninikko: Thanks for explaining. Let's leave it as it is.

src/scs/pki.md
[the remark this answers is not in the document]
jenninikko: In CONTRIBUTING.md we have some kind of rule that sounds like:
_Avoid jargon and acronyms, if you can._
and 
_Do not assume that readers know everything you currently know._

In short, no :) And spelled with a hyphen *_*

Since these are not personal notes but official documentation that will be used as a basis for further development, we need to follow the rules. We do not know the readers' background. That's why it is better to use a way of writing that helps users understand a document the first time they read it.

So... I do not want to seem like a bore, but it is better not to use abbreviations and slang that are understandable to those who have been in the subject for a long time. If we assume that the text will be read by a person who is only studying security, it will be easier for him to google articles related to on-premises https://en.wikipedia.org/wiki/On-premises_software.

jenninikko: As we know, there is nothing more permanent than temporary)
Let's use the updated version until the next iteration.

jenninikko (approved):
You are such a great contributor!

PR #10, opened 2022-12-16
Create LICENSE

vilvo:


vilvo:
@brianmcgillion - please review and merge at your convenience

jenninikko:
I'm not sure I can evaluate the contents of the file itself, but I have a layout question: is it supposed to be that the text looks like one big commit?

vilvo:
> I'm not sure I can evaluate the contents of the file itself, but I have a layout question: is it supposed to be that the text looks like one big commit?

Yes. The license is supposed to be exactly like that.

PR #17, opened 2022-12-28
imx8qm ethernet platform device passthrough Lessons learned initial

josa41:
Lessons learned on device passthrough on imx8qm

Lessons learned about ethernet platform device passthrough on
imx8qm board.

Signed-off-by: Jon Sahlberg <jon.sahlberg@unikie.com>


jenninikko:
Thank you! Let's discuss it.

vilvo:
Thanks @josa41  - really good story on platform ethernet passthrough. Needs some fixes to misspellings and formatting but I think we could do those after merge with @jenninikko

jenninikko:
This commit does not meet our commit message guidelines requirements.
@josa41, please see this: https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md#commit-message-guidelines

jenninikko:
I have some concerns about naming here. I understand that the concept of lessons learned is well known. In documentation particularly, this term is usually used for identifying project management process improvements (retrospective). Are we sure that readers will find the necessary information on their own in this chapter?

It is more the question about if it is the chapter where we are going to store our notes on the research topic, so maybe name it Research Notes? It also seems that it would be more logical because we are working with the Secure Systems Research Center (SSRC), which already has the word Research in its name, which reflects the essence of our work.

I am not sure, so I am just sharing (thinking out loud).

jenninikko (changes requested):
@josa41, thanks a lot for this new content. Please find my comments below.

I was told that you and the guys from the other team are waiting for this chapter, so I have quickly proofread your work—small spelling mistakes will not confuse the heck out of our future readers. All corrected variants are already there, you just need to modify your source file.

My comment about the chapter name is not a stop question, it is just a topic for discussion.

If you need any help just let me know.

PR #21, opened 2022-12-31
Add simple flake outputs for NixOS based setup

mikatammi:
Add simple flake outputs for NixOS based setup. Doesn't yet do anything

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,23 +4,23 @@
   inputs.nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
   inputs.flake-utils.url = "github:numtide/flake-utils";
 
-  outputs = { self, nixpkgs, flake-utils }:
-    let
-      systems = [
-        "x86_64-linux"
-        "aarch64-linux"
-      ];
-    in
-      flake-utils.lib.eachSystem systems (system: {
+  outputs = {
+    self,
+    nixpkgs,
+    flake-utils,
+  }: let
+    systems = [
+      "x86_64-linux"
+      "aarch64-linux"
+    ];
+  in
+    flake-utils.lib.eachSystem systems (system: {
+      packages = let
+        pkgs = nixpkgs.legacyPackages.${system};
+      in {
+        doc = pkgs.callPackage ./doc.nix {};
+      };
 
-        packages =
-          let
-            pkgs = nixpkgs.legacyPackages.${system};
-          in {
-            doc = pkgs.callPackage ./doc.nix {};
-          };
-
-        formatter = nixpkgs.legacyPackages.${system}.alejandra;
-
-      });
+      formatter = nixpkgs.legacyPackages.${system}.alejandra;
+    });
 }
--- a/flake.nix
+++ b/flake.nix
@@ -20,5 +20,7 @@
             doc = pkgs.callPackage ./doc.nix {};
           };
 
+        formatter = nixpkgs.legacyPackages.${system}.alejandra;
+
       });
 }
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,15 @@
 {
   description = "Ghaf - documentation for TII SSRC Common Tech";
 
+  nixConfig = {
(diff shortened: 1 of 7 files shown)


flake.nix
```diff
@@ -1,24 +1,74 @@
...
-            doc = pkgs.callPackage ./doc.nix {};
+          in
+          {
+            doc = pkgs.callPackage ./doc.nix { };
           };
 
-      });
+      }))
+      # Final target images
+      {
```
vilvo, line 53: Really nice work with the initial flake. The outputs for the final target images is fine for now but moving forward we could outline host and guest VMs structure as packages (e.g. like doc.nix) so that the top level would only have reference HW packages and those would further build their own hosts but take guest VMs from both common and reference VM specific packages.

vilvo (approved):
Looks good to me as initial baseline.

flake.nix
```diff
@@ -1,24 +1,74 @@
 {
-  description = "Ghaf - documentation for TII SSRC Common Tech";
+  description = "Ghaf - documentation and implementation for TII SSRC Common Tech";
 
-  inputs.nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
-  inputs.flake-utils.url = "github:numtide/flake-utils";
+  nixConfig = {
```
unbel13ver, line 4: The binary cache link is outdated. Please, use this one instead:
https://cache.vedenemo.dev
cache.vedenemo.dev:RGHheQnb6rXGK5v9gexJZ8iWTPX6OcSeS56YeXYzOcg=

mikatammi: Done

flake.nix
mikatammi: I put the final target images to a separate file called targets.nix (Is this a good name?). Also added default targets

vilvo: `targets.nix` is a good name.
Please add a commit to include the supported build commands in `ghaf\README.md` like:
https://github.com/mikatammi/ghafhack/blob/master/README.md#x86-64

I tested that:
`nix build` 
now builds the default target fine and results in vm that can be started with `result/bin/run-nixos-vm`

Having this at least in instructions and optionally also in the commit message would be good. I think you can drop `-o result-vm` as the `result` is common with nix.

mikatammi: README.md needs some revamp also otherwise, but I can add the section.

vilvo: Also, Ctrl-C does not seem to kill the QEMU process:
```
❯ result/bin/run-nixos-vm
Formatting '/home/vilvo/mikatammi/ghaf/nixos.qcow2', fmt=qcow2 cluster_size=65536 extended_l2=off compression_type=zlib size=1073741824 lazy_refcounts=off refcount_bits=16
^Cqemu: terminating on signal 2
^C
```
Probably worth mentioning in guide if you have a clean method (other than `kill -9 <pid>`) to terminate the VM (e.g. login, `shutdown -h now`). These will be needed by anyone not familiar with the setup and testing it for the first time.

vilvo: > README.md needs some revamp also otherwise, but I can add the section.

It does but we can work that out with @jenninikko

jenninikko: > README.md needs some revamp also otherwise, but I can add the section.

You can add just before this https://github.com/tiiuae/ghaf#directory-structure section a new section Build Instructions.

unbel13ver (approved):

jenninikko:
As it seems, for now, I need to rewrite the README file, as first, our baseline was that this repository was only for documentation. Now we are merging code source files, and Ghaf returned to be some kind of hardened OS distribution.

flake.nix
vilvo: I don't think I had those with sway :)

vilvo: There was something else messed with my setup. Now both Ctrl-C work and QEMU window has drop down menu bar. After reboot 🎉

mikatammi: I added Build Instructions section into the README.md

PR #25, opened 2023-01-04
dev: login to VM with preset development user account

vilvo:
* Default account implemented as module. Easy to remove/harden.
* Tested with: nix build .#packages.x86_64-linux.vm result/bin/run-nixos-vm login with module configured username and password
* Module included only for "x86_64-linux" for now

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,9 @@ or
 
     result/bin/run-nixos-vm
 
+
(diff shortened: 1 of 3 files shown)


brianmcgillion:
Can you add a line to the front README so folks can see the default user:passwd

brianmcgillion (approved):
verified to work in qemu x86

PR #39, opened 2023-01-11
doc: refactored instructions to ghaf GH pages

vilvo:
* Main motivation is to ensure the relevant instructions are kept up-to-date with other system documentation instead of maintaining README outside of GH Pages
* Reference implementations added and structured based on testing with NVIDIA Orin AGX. This content is mainly from the old README.md Added prequisities with NVIDIA Orin AGX.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,74 +2,22 @@
 
 This repository contains the source code of Ghaf Framework — an open-source project for enhancing security through compartmentalization on edge devices.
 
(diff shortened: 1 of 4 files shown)


mikatammi (approved):
As a refactor looks good to me.

There are some inconsistencies regarding cross-compilation but it's fine for me if those are fixed later.

PR #41, opened 2023-01-11
Host minimal profile

vilvo:
* imported and tested minimal profile on host for NVIDIA Orin AGX
* documented the ADR proposal for the minimal host

### DIFF
--- a/configurations/host/configuration.nix
+++ b/configurations/host/configuration.nix
@@ -1,5 +1,8 @@
-{ ... }:
+{ modulesPath, ... }:
 {
+  imports = [
+    (modulesPath + "/profiles/minimal.nix")
+  ];
   networking.hostName = "ghaf-host";
   system.stateVersion = "22.11";
 }
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -3,6 +3,7 @@
 - [About Ghaf](about/overview.md)
 - [Architecture](architecture/architecture.md)
     - [Decision Record](architecture/adr.md)
+       - [Minimal host](architecture/adr/minimal-host.md)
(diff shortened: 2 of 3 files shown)


brianmcgillion (commented):
Is this really the minimal we want? This is minimal for a generic headless device, I envisaged us creating minimal-minimal.nix, or barely-enough-to-boot.nix profiles, which we could contribute back.

vilvo:
> Is this really the minimal we want? This is minimal for a generic headless device, I envisaged us creating minimal-minimal.nix, or barely-enough-to-boot.nix profiles, which we could contribute back.

This is only an iteration towards that target most_minimal+hardened with testing of the profile import on NVIDIA Orin AGX target. There's `hardened` profile already in Nix but I did not dare to propose that yet as I'm expecting `hardened`-profile to break more.

Not sure if you noticed the other commit message on details of reducing the image size and # packages. That would give already some benefits for early development from this iteration.
```
* Use NixOS minimal profile on host instead of default
  to match the design target of minimal trusted computing
  base (TCB).
* Metrics on nvidia-jetson-orin:
  * reduces the image size and root-partition used size
    with 500M from 4.2G to 3.5G and 2.6G to 2.1G respectively
  * Number of pkgs in /nix/store from 707 t0 526
* Numbers not seen on x86 VM target as the nix-store of the
  host is mounted to the VM. Not imported on x86 QEMU target.
```

Given this is only a minimal iteration towards the target we have options of:
- taking this into use early (merge) and enable other profiles for other development needs with additional build targets
- postpone this PR, continue analysis of extra content in minimal profile and based on that continue development of the most_minimal+hardened-profile we could contribute back.

PR #52, opened 2023-01-18
dev: enable nix configuration as module

vilvo:
* import and configure as module
* enables flakes and nix-command to build ghaf image on host. NVIDIA Orin AGX is powerful after all.


vilvo:
Seems to work fine with flakes on NVIDIA Orin AGX
image

unbel13ver (approved):

PR #53, opened 2023-01-20
host: support nested virtualization on x86 vm

vilvo:
* requires vnclient to access the host - ::5900
* includes microvm and it's default VMMs to "host" VM
* enables microvm.nix defined guest VMs

### DIFF
--- a/targets/default.nix
+++ b/targets/default.nix
@@ -8,6 +8,10 @@
     system = "x86_64-linux";
     modules = [
       microvm.nixosModules.host
+      # .microvm for vm-format "host" to nest vms for system development on x86 (Intel)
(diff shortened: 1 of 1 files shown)


vilvo:
Tested with 
image

unbel13ver (approved):

targets/default.nix
```diff
@@ -8,6 +8,10 @@
     system = "x86_64-linux";
     modules = [
       microvm.nixosModules.host
+      # .microvm for vm-format "host" to nest vms for system development on x86 (Intel)
+      # NOTE: Ghaf nested virtualization is not assumed nor tested yet on AMD
+      microvm.nixosModules.microvm
```
mikatammi, line 13: By my understanding, the microvm-module should be only imported on the guest-microvm itself

PR #59, opened 2023-01-27
x86-nuc: enable flakes via nix-module

vilvo:
* enables use of microvm.nix

### DIFF
--- a/targets/default.nix
+++ b/targets/default.nix
@@ -25,6 +25,7 @@
       microvm.nixosModules.host
       ../configurations/host/configuration.nix
       ../modules/development/intel-nuc-getty.nix
+      ../modules/development/nix.nix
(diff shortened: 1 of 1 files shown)


mikatammi (approved):
We should probably have a sets of "common modules" for different targets

vilvo:
> We should probably have a sets of "common modules" for different targets

Sure, we kind of do already - these are for `development` as I see we would drop dev modules from any later `deployment|production`-scenario. For example, e.g. flakes nor hard-coded trivial authentication are not be needed in latter common modules.

unbel13ver:
Let's add this for now, but in near future as @mikatammi noticed we have to get rid of code duplication and probably create separate file for every target. In the same way as it is already done for Jetson.

unbel13ver (approved):

PR #62, opened 2023-01-30
Doc: update nvidia_agx_pt_uart.md

ruizjuanpablo:
Fix the platform@c000000 ranges, and fix serial@c000000 reg  cells size to 1. Without this fix, the VM was not able to load the virtio disk (vda1), because serial@c000000 registers where  overlapping.


ruizjuanpablo:
Done, also I added a missing zero.

jenninikko (approved):

PR #63, opened 2023-01-31
netvm: basic structure to include the netvm to ghaf

vilvo:
- declaration of the netvm using flake
  - optionally could be in-tree flake in subdir with dir
- netvm passed to the targets but not used (yet)
- structure update tested to build with x86 target

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,6 +28,12 @@
       url = "github:anduril/jetpack-nixos";
       inputs.nixpkgs.follows = "nixpkgs";
     };
+    # ghaf reference guest virtual machines
+    # - optionally ghaf can be composed of virtual machines from the same repo with dir
+    # that defines the subdirectory
+    netvm = {
+      url = "github:tiiuae/netvm?rev=b7103ce47b17cd61403b459e72839f38d68204f7";
+    };
   };
 
   outputs =
@@ -37,6 +43,7 @@
     , nixos-generators
     , microvm
     , jetpack-nixos
+    , netvm
     ,
     }:
     let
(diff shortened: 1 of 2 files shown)


vilvo:
Please note that even the placeholderthis netvm enables me to work on the matching architecture documentation (ADR), and further work on network interface passthrough and netvm contents (e.g. reverse proxy for multicast, default firewall rules etc.)

All comments welcome.

flake.nix
```diff
@@ -28,6 +28,12 @@
       url = "github:anduril/jetpack-nixos";
       inputs.nixpkgs.follows = "nixpkgs";
     };
+    # ghaf reference guest virtual machines
+    # - optionally ghaf can be composed of virtual machines from the same repo with dir
+    # that defines the subdirectory
+    netvm = {
+      url = "github:tiiuae/netvm?rev=b7103ce47b17cd61403b459e72839f38d68204f7";
```
mikatammi, line 35: There should be `follows` attributes for: nixpkgs, flake-utils, and microvm.

Otherwise, as you can see from the flake.lock file changes, it added second instances of them all, which are be different versions

flake.nix
```diff
@@ -78,7 +85,7 @@
       }
 
       # Final target images
-      (import ./targets {inherit self nixos-generators microvm jetpack-nixos;})
+      (import ./targets {inherit self nixos-generators microvm jetpack-nixos netvm;})
```
mikatammi, line 88: The function in targets/default.nix does not take netvm as an argument in the argument set

mikatammi: Maybe some change is missing from this commit?

vilvo:
56cb5e5defaae5e843ee1cd2b3bd028b1c301016 implements the initial netvm with much extended and tested functionality.

Issue #45, opened 2023-01-13
most minimal host


vilvo:
Is this really the minimal we want? This is minimal for a generic headless device, I envisaged us creating minimal-minimal.nix, or barely-enough-to-boot.nix profiles, which we could contribute back.

_Originally posted by @brianmcgillion in https://github.com/tiiuae/ghaf/pull/41#pullrequestreview-1245043832_

jenninikko:
Documented here: https://tiiuae.github.io/ghaf/architecture/adr/minimal-host.html

remimimimimi:
As it would be inconvenient to do all those experiments in the ghaf repository, I have yet to create a draft pull request. All of my experiments (which were committed) are located [here](https://github.com/remimimimimi/ghaf-minimal/tree/gist) and [here](https://github.com/remimimimimi/ghaf-not).

# Experiments

## Ghaf minimal

The first repository consists of several branches, each of which represents some experiment:

### gist

Implements [this gist](https://gist.github.com/chrisdone/02e165a0004be33734ac2334f215380e) in nix.

### gist-glibc

Refines branch above with the use of glibc instead of musl.

### gist-systemd

Uses systemd for initialization instead of a hand-written script.

### nixos-toplevel

Keeps the essential packages for NixOS. Removes almost everything, and the remaining system builds but not working.

It shows how everything indirectly depends upon top-level derivation. Dependencies could be tracked buy only through source code. But when I looked into it, I saw that many stuff are hard-coded.

For example, there's no easy way to replace coreutils and util-linux with busybox because nixpkgs depend on it heavily. Also, Perl used all over NixOS for different scripts: install grub, activation scripts, and switch to configuration.

![graph](https://github.com/tiiuae/ghaf/assets/33205215/0f31b44c-a8a6-42a5-a340-bad220d8d6f7)

## Ghaf not

Here I tried to fork not-os to essentially reimplement low-level components of NixOS and then plug high-level modules in.

# Next steps

We have the following requirements: easy-maintainable, small TCB in the resulting image. The problem is that it is currently impossible for both of them. The maximum that we can have right now without forking NixOS or reimplementing low-level components is to modify the systemd package.

If we choose to maintain our code base, then the most sensible steps will be the following:

- Analyze modules to separate them into low-level and high-level.
- Reimplement low-level modules and plug high-level ones on top of them.

We'll have the smallest possible TCB, but we'll need to support our code base, which may break over time.

[...]

PR #64, opened 2023-01-31
Minifiy nvidia jetson orin image

remimimimimi:
Related to #45.
I reduced the image size from 1.9GB to 1.2GB by configuring qemu, systemd, and git. Next step will be apply @vadika kernel config.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,10 +21,7 @@
       url = "github:nix-community/nixos-generators";
       inputs.nixpkgs.follows = "nixpkgs";
     };
-    nixos-hardware = {
-      url = "github:nixos/nixos-hardware";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
+    nixos-hardware.url = "github:nixos/nixos-hardware";
     microvm = {
       # TODO: change back to url = "github:astro/microvm.nix";
       url = "github:mikatammi/microvm.nix/wip_hacks_2";


targets/default.nix
```diff
@@ -37,6 +38,52 @@
...
+            gtkSupport = false;
+            vncSupport = false;
+            smartcardSupport = false;
+            spiceSupport = false;
+            ncursesSupport = false;
+            libiscsiSupport = false;
+            tpmSupport = false; # XXX: Probably shouldn't be disabled
+            hostCpuOnly = true;
+          };
+          git = super.git.override {
```
vilvo, line 58: I'd like to see git on image via `modules/development/essentials.nix` but this is good for now.

We can mark it as `# XXX` (or `# TODO:`) for now. 

On most minimal host profile git should not be needed - even for updates.

targets/default.nix
```diff
@@ -37,6 +38,52 @@
...
+            withPortabled = false; # XXX
+            # withResolved = false; # XXX
+            withShellCompletions = false;
+            withTpm2Tss = false;
+            withUserDb = false;
+          };
+        };
+      in
+        import nixpkgs {
+          localSystem = "aarch64-linux";
```
vilvo, line 84: Please also add `# TODO: enable also for x86-linux targets` here. We could do that in the next iterations but I'd like to test most minimal host on x86 targets as well - those are plenty more available and have generally better virtualization support on different VMMs than `aarch64`. From that perspective those are good benchmark for the `aarch64` reference work.

vilvo:
> Related to #45. I reduced the image size from 1.9GB to 1.2GB by configuring qemu, systemd, and git. Next step will be apply @vadika kernel config.

Super nice work. I think we can remove "Draft" from this and go and review and merge this. More minimal kernel config warrants a PR of its' own. Same for the `# XXX`-ideas.

targets/default.nix
remimimimimi: I’ve done overlay for git because of microvm. It uses it internally in the host nixos module, and with this overlay, image size was reduced by 200mb. So it is actually unrelated to the development environment, but if you want to add git to devenv you should be aware of this change.

It also can be done with a little hack for microvm module, by overlaying packages especially for it. But for now, let it be this way.

targets/default.nix
remimimimimi: With the latest change, this is no more a problem.

vilvo:
It definitely reduces the image size:
```
ghaf (54521a6) [?] # this PR
❯ ls -lah result/nixos.img
Permissions Size User Date Modified Name
.r--r--r--  5.1G root  1 Jan  1970  result/nixos.img

ghaf (197f9dc) [?] # main
❯ ls -lah result/nixos.img
Permissions Size User Date Modified Name
.r--r--r--  5.7G root  1 Jan  1970  result/nixos.img
```

As of now, "This branch cannot be rebased due to conflicts"
so please rebase your fork branch with the ghaf-repo main and resolve conflicts.

While at it, I'd prefer you squash the commits in your PR and also please add DCO (Developer Certificate of Origin) with `signed-off-by:` to the PR commit(s).

vilvo (approved):
Looks good to me. Basics work - boots, ssh, netvm starts. Reduces the image and host size.
None disabled features are critical for now but we may want to come back to TPM support and such later.

mikatammi:
In general I'd like to say good work and nice investigation

Now that the_great_refactor has been merged, I'd put the import of ../configurations/host/minimal.nix into the targets/common-release.nix

mikatammi:
One thing I've noticed during my builds, is that both qemu and qemu-host-cpu-only -packages get built, even though the ghaf currently doesn't even use QEMU. So it shouldn't be building QEMU at all but for some reason it does. I haven't checked the dependency tree to see what pulls it in

vilvo:
> One thing I've noticed during my builds, is that both qemu and qemu-host-cpu-only -packages get built, even though the ghaf currently doesn't even use QEMU. So it shouldn't be building QEMU at all but for some reason it does. I haven't checked the dependency tree to see what pulls it in

I would not optimize QEMU out from development supporting packages. It's still the swiss-knife of VMMs to test KVM supported functionality when somethings do not work out-of-the box with other VMMs.

vilvo:
> In general I'd like to say good work and nice investigation
> 
> Now that the_great_refactor has been merged, I'd put the import of ../configurations/host/minimal.nix into the targets/common-release.nix

Please go ahead. @remimimimimi 's work is still early and I have not heard updates from it for a while. `../configurations/host/minimal.nix` can be a placeholder for that work and support the general ghaf macro structure.

PR #71, opened 2023-02-07
flake.nix: Fix extra-trusted-substituters

mikatammi:
The attribute that should be used is extra-trusted-substituters, not extra-subtituters


vilvo (approved):

mikatammi:
Noticed the warning while the caches weren't part of my system config, that the caches weren't actually in use because they weren't trusted.

PR #73, opened 2023-02-10
Enable Weston on host

mikatammi:


vilvo (approved):
Looks legit to enable the approach:
1. gui on host to enable further development
2. move gui to vm with passthrough

PR #74, opened 2023-02-10
Fix microvm.nix on x86_64

mikatammi:
microvms should now work both on x86_64-linux and aarch64-linux.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -45,16 +45,16 @@
         ]
       },
       "locked": {
-        "lastModified": 1675244568,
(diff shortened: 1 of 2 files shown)


vilvo:
Nice work and discussion with https://github.com/astro/microvm.nix/pull/72

Would it make sense to open discussion in the upstream for solution on HW architecture specifics in `runners/qemu.nix` c527054302a9a0cae206abfe1c8b282fba1d464b ?

And the same for 4e1d7cdca4097852646ed3ce0d4707aaa6c790c8 ? Makes sense for the upstream also to use less privileges ownership for `vfio`?

PR #77, opened 2023-02-17
The Great Refactor

mikatammi:
- Debug and Release variants
- Cross-compilation targets for NVIDIA (doesn't build successfully yet)
- Major clean-up and refactoring


mikatammi:
Finally had time to refactor this stuff. Now the good base to start the actual project starts to be in place

mikatammi:
kuva

docs/doc.nix
```diff
@@ -1,3 +1,5 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
```
vilvo, line 2: Docs are actually under different license - CC - but I'll add a commit to this branch to update this.

vilvo (approved):
Looks good to me.
A general structure note is that NUC as a specific target likely unnecessary for several reasons: 
* we can support x86-based device with more general support
* NUCs are not specific targets - there are bunch of HW variants even within a generation of NUC (11 Pro) in this case (i5, i7, etc.) - only the variant ID specifies the device exactly
* we've discussed of supporting HW specifics with `nixos-hardware` modules
* we'll face the same discussion with coming Jetson-family of device variants that are very close to Orin AGX but have less HW compute resources (in fact - it is already possible to test the variants on Orin AGX with NVIDIA BSP provided device trees)

Summary: more general at high/common level would be preferred and device/variants specifics to nixos-hardware modules - preferably contributed to https://github.com/NixOS/nixos-hardware.
That way we could target more general support for aarch64 ghaf and x86 ghaf but still support specific hardware better.

docs/doc.nix
vilvo: Done with [414e807](https://github.com/tiiuae/ghaf/pull/77/commits/414e807e94e695e6ec21c7946f4f4bc2af250ced)

vilvo:
Please note that even with review approval, I haven't done any testing to this PR branch yet.

vilvo:
>  A general structure note is that NUC as a specific target likely unnecessary for several reasons:

This is already a great (refactor) step towards right direction. With my notes I hope we will take further steps. Really nice work in general. Thanks a lot!

targets/common.nix
```diff
@@ -0,0 +1,9 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Common modules for all targets
+{
+  imports = [
+    ../modules/graphics/weston.nix
```
unbel13ver, line 7: I would propose to keep common target headless and have graphics enabler separately.

unbel13ver:
Great work!
I love this new structure - logical and clean. Feel free to merge if you disagree with my comment about headless config.

unbel13ver (approved):

targets/common.nix
vilvo: Very good point!

mikatammi: I will do this

mikatammi: I was about to do the change when the first headless target comes in, but we can of course do that beforehand

PR #96, opened 2023-03-30
architecture: requirements update on minimal host

vilvo:


unbel13ver:
Seems quite clear and logical for me. Do I understand correctly that the order of reequirements in the table is not really strict?

vilvo:
TODO: address how the host allocates dedicated filesystems for guest VMs.

PR #99, opened 2023-04-04
Do not enable systemd-boot in host by default

mikatammi:
* Do not enable systemd-boot in modules/host/minimal.nix
* This commit fixes imx8qm-mek targets, which use GRUB


mikatammi:
@remimimimimi

PR #101, opened 2023-04-04
systemd-boot-dtb: Load device tree using systemd-boot

mikatammi:
- Fix formatting
- Use systemd-boot to load device tree
- Add new module to copy device tree into right place
- Make NVIDIA Jetson Orin target use the new module


mikatammi:
I still want to test tomorrow this does not break the flashing

mikatammi:
Tested that everything is in right place even after the flashing

unbel13ver (approved):

PR #102, opened 2023-04-09
Change back to astro's microvm.nix main-branch

mikatammi:
* All of my wip_hacks_2 branch things now have proper implementations in microvm.nix main-branch, so change back to it


mikatammi:
@panufi

PR #113, opened 2023-04-18
DRAFT: netvm: writableStoreOverlay

vilvo:
* mounts the ramdisk correctly on host
* mounts the overlay on netvm
* runs nix without read-only error
* fails with nix search path entries and nixpkgs not found:

[ghaf@netvm:~]$ nix-shell -p screen
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels/nixos' does not exist, ignoring warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I)

### DIFF
--- a/microvmConfigurations/netvm/default.nix
+++ b/microvmConfigurations/netvm/default.nix
@@ -63,6 +63,8 @@ nixpkgs.lib.nixosSystem {
       };
 
       microvm.qemu.bios.enable = false;
+
+      microvm.writableStoreOverlay = "/nix/netvm/store";
     })
   ];
 }
--- a/modules/hardware/dev-filesystems.nix
+++ b/modules/hardware/dev-filesystems.nix
@@ -1,0 +1,11 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{...}: {
+  fileSystems."/nix/netvm/store" = {
+    device = "none";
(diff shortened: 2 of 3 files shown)


vilvo:
microvm.nix generated overlay mount on `netvm`:
```
[ghaf@netvm:~]$ mount | grep overlay
overlay on /nix/store type overlay (rw,relatime,lowerdir=/nix/store,upperdir=/mnt-root//nix/netvm/store/store,workdir=/mnt-root//nix/netvm/store/work)
```
host tmpfses:
```
[ghaf@ghaf-host:~]$ mount|grep tmpfs
devtmpfs on /dev type devtmpfs (rw,nosuid,size=1565920k,nr_inodes=3880104,mode=755)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev,size=15659180k)
tmpfs on /run type tmpfs (rw,nosuid,nodev,size=7829592k,mode=755)
tmpfs on /run/wrappers type tmpfs (rw,nodev,relatime,size=15659180k,mode=755)
none on /nix/netvm/store type tmpfs (rw,relatime,size=2097152k,mode=755)
tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,size=3131832k,nr_inodes=782958,mode=700,uid=1000,gid=100)
```

PR #116, opened 2023-04-20
Some steps towards usable graphic image

unbel13ver:
* Chromium and Element are added to the image;
* Launchers for ^ are added to weston panel;
* New fonts are added to the system;
* Screen blanking is disabled by default.

### DIFF
--- a/modules/graphics/weston.nix
+++ b/modules/graphics/weston.nix
@@ -11,6 +11,11 @@
     weston
     # Seatd is needed to manage log-in process for weston
     seatd
+
+    # Graphical applications
+    # Probably, we'll want to re/move it from here later
+    chromium
+    element-desktop
   ];
 
   # Next 4 services/targets are taken from official weston documentation:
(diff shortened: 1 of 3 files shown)


modules/graphics/weston.nix
```diff
@@ -6,11 +6,21 @@
...
+
   environment.noXlibs = false;
   environment.systemPackages = with pkgs; [
     weston
     # Seatd is needed to manage log-in process for weston
     seatd
+
+    # Graphical applications
+    # Probably, we'll want to re/move it from here later
+    chromium
```
mikatammi, line 22: Maybe we should move them quite soon, I just wanted to check few things from this PR by building it myself and testing, before giving approval. And now I'm probably going to have to wait at least ~4 hours building chromium from source

mikatammi (approved):

nesteroff (approved):

modules/graphics/weston.nix
unbel13ver: I hope you used vedenemo cache, since chromium is already there=)

unbel13ver:
Added comment on why I've added /etc/xdg to the XDG_CONFIG_DIRS one more time for weston service

PR #118, opened 2023-04-21
Add external GPU passthrough and guivm

Nerox9:
* Guivm and network are created,
* External GPU passthrough is applied.
 
Signed-off-by: Berk Arslan <berk.arslan@unikie.com>


flake.nix
```diff
@@ -15,7 +15,7 @@
   };
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs/nixos-22.11";
+    nixpkgs.url = "github:mikatammi/nixpkgs/nvidia_x11_aarch64_nixos-unstable";
```
unbel13ver, line 18: This worries me a lot! Is there really no other way of doing this without forking nixpkgs?

microvmConfigurations/guivm/default.nix
```diff
@@ -0,0 +1,60 @@
...
+
+    ({pkgs, ...}: {
+      networking.hostName = "guivm";
+      # TODO: Maybe inherit state version
+      system.stateVersion = "22.11";
+
+      microvm.hypervisor = "qemu";
+
+      nixpkgs.overlays = [
+        (self: super: {
```
unbel13ver, line 25: As a proposal: let's use `final: prev` instead of `self: super`?

flake.nix
Nerox9: The Nvidia drivers work with unstable version but it doesn't work with aarch64 yet, unfortunately.

microvmConfigurations/guivm/default.nix
Nerox9: The names are replaced.

flake.nix
mikatammi: Could we just wait a month before merging this, then the driver will be in the new nixos-23.05 stable branch. The other way around this would be to add the package through an overlay

microvmConfigurations/guivm/default.nix
```diff
@@ -0,0 +1,60 @@
...
+
+      boot.kernelParams = [
+        "pci=nomsi"
+      ];
+
+      microvm.interfaces = [
+        {
+          type = "tap";
+          id = "vm-guivm";
+          mac = "02:00:00:01:01:01";
```
mikatammi, line 55: This MAC-address collides with the one from NetVM. Can you please set this to eg 02:00:00:01:01:01

modules/hardware/nvidia-jetson-orin.nix
```diff
@@ -32,13 +32,16 @@
...
+  hardware.opengl.driSupport = true;
+
   imports = [
     ../boot/systemd-boot-dtb.nix
   ];
 
   # Passthrough Jetson Orin WiFi card
   boot.kernelParams = [
-    "vfio-pci.ids=10ec:c82f"
+    "vfio-pci.ids=10ec:c82f,10de:1cb1,10de:0fb9"
```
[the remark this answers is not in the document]
mikatammi: I think these might not be even needed, the microvm.nix generates a script (systemd unit) to plug the device into VFIO. I think we originally enabled these because we thought it might make the passthrough easier if the device is bound to VFIO already boot time. The microvm.nix seems to generate the scripts anyway

flake.nix
mikatammi: but the overlay is so lot of work so I'd just rather not do extra work but just wait for a month :)

microvmConfigurations/guivm/default.nix
```diff
@@ -0,0 +1,60 @@
...
+      networking.firewall.allowedTCPPorts = [22];
+
+      environment.systemPackages = with pkgs; [
+        weston
+      ];
+
+      hardware.nvidia.open = false;
+      hardware.nvidia.modesetting.enable = true;
+      nixpkgs.config.allowUnfree = true;
+      services.xserver.videoDrivers = ["nvidia"];
```
mikatammi, line 43: I actually wouldn't like to enable the NVIDIA's closed source drivers by default

mikatammi (changes requested):
Overall looks good and nicely shows what you need for GUIVM development, I'd just like to postpone this until nixos-23.05

mikatammi:
I just wanted to test this but the git tree is somehow messed, probably because two commits have the same subject then git rebase breaks.

mikatammi:
Also can you please squash the commits into one commit? (It looks like the two commits on top are just fixing things to be the same as in mainline, and the top one is a merge-commit)

PR #124, opened 2023-04-27
Add a basic background

brianmcgillion:
Add a more welcoming wallpaper


modules/graphics/weston.ini.nix
```diff
@@ -9,6 +9,8 @@
 
       [shell]
       locking=false
+      background-image=${./assets/wallpaper.jpg}
+      background-type=scale
```
vilvo, line 13: I recommend `scale-crop` instead. It will handle different sized displays better.

vilvo (approved):

modules/graphics/weston.ini.nix
brianmcgillion: i tried that, the problem is it was a bit overzealous in cropping
:)

PR #128, opened 2023-05-04
Create a script to launch Windows VM

nesteroff:
As Brian requested this pull request adds a script that can launch a Windows 11 virtual machine VHDX image downloaded from the [Windows Insider Preview ARM64 page](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewARM64).
Example:
`windows-launcher ./Windows11_InsiderPreview_Client_ARM64_en-us_25324.VHDX`

In addition to that it adds an icon to the Weston toolbar. When clicked for the first time it opens a file selection dialog. Once Windows 11 image has been selected it saves the path to the ~/.config/windows-launcher-ui.conf file and launches the virtual machine. When clicked next time it reads the path from the configuration file and immediately launches the VM. This way the user can launch a Windows virtual machine with a single click.

Signed-off-by: Yuri Nesterov <yuriy.nesterov@unikie.com>

### DIFF
--- a/modules/graphics/weston.ini.nix
+++ b/modules/graphics/weston.ini.nix
@@ -1,40 +1,50 @@
 # Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
-{pkgs, ...}: {
+{
+  pkgs,
+  lib,
+  ...
+}: {
   environment.etc."xdg/weston/weston.ini" = {
-    text = ''
-      # Disable screen locking
-      [core]
-      idle-time=0
+    text =
+      ''
+        # Disable screen locking
+        [core]
+        idle-time=0
 
-      [shell]
-      locking=false
-      background-image=${./assets/wallpaper.jpg}
-      background-type=scale-crop
+        [shell]
+        locking=false
+        background-image=${./assets/wallpaper.jpg}
+        background-type=scale-crop
 
-      # Enable Hack font for weston-terminal
-      [terminal]
-      font=Hack
-      font-size=16
+        # Enable Hack font for weston-terminal
+        [terminal]
+        font=Hack
+        font-size=16
 
-      # Add application launchers
-      # Adding terminal launcher because it is overwritten if other launchers are on the panel
-      [launcher]
-      path=${pkgs.weston}/bin/weston-terminal
-      icon=${pkgs.weston}/share/weston/icon_terminal.png
+        # Add application launchers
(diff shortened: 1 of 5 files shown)


jenninikko:
After this pull request this one https://github.com/tiiuae/ghaf/pull/129 can be also updated =)

unbel13ver (approved):
Great work and great PR! Thanks!

mikatammi (approved):
Didn't test, but overall looks good 👍

mikatammi:
Does this work on NUC / x86_64?

user-apps/windows-launcher/default.nix
```diff
@@ -0,0 +1,89 @@
...
+
+      IMG_DIR="$(dirname "$1")"
+      AAVMF_VARS="$IMG_DIR/AAVMF_VARS.fd"
+
+      if [ ! -f $AAVMF_VARS ]; then
+        cp ${pkgs.OVMF.fd}/FV/AAVMF_VARS.fd $AAVMF_VARS
+        chmod 644 $AAVMF_VARS
+      fi
+
+      ${pkgs.qemu}/bin/qemu-system-aarch64 \
```
mikatammi, line 30: Here you could put something like

let arch = builtins.head (builtins.split "-" pkgs.stdenv.hostPlatform.system);

...

${pkgs.qemu}/bin/qemu-system-${arch}

modules/graphics/weston.ini.nix
```diff
@@ -34,6 +34,10 @@
       [launcher]
       path=${pkgs.gala-app}/bin/gala --enable-features=UseOzonePlatform --ozone-platform=wayland
       icon=${pkgs.weston}/share/weston/icon_terminal.png
+
+      [launcher]
```
mikatammi, line 38: Could you please make this conditional? Only add this launcher icon if on aarch64-linux or NVIDIA

mikatammi (changes requested):

user-apps/windows-launcher/default.nix
mikatammi: I think this needs even more work to make work this work on x86_64, so could we make this conditional so that this feature only appears while using NVIDIA

modules/graphics/weston.ini.nix
nesteroff: This is a good point, will do.

nesteroff: Ok, I made it conditional and rebased on top of main.

vilvo (approved):

PR #133, opened 2023-05-10
Rename NUC target to somewhat generic x86

unbel13ver:
`debug` build is tested with Dell rugged laptop.


mikatammi (approved):
I think we can go with this for now. The original idea I had we would have both more generic x86 image and more specific NUC image which relies on NUC's hardware configuration, but we can work on that later if needed

jenninikko (approved):

PR #138, opened 2023-05-17
Configure weston dependencies

remimimimimi:
Remove remote desktop dependencies (freerdp, pipewire), and xwayland from weston.

### DIFF
--- a/modules/overlays/custom-packages.nix
+++ b/modules/overlays/custom-packages.nix
@@ -37,19 +37,16 @@
         # First, weston package is overridden (passing colord = null)
         (prev.weston.override {
           colord = null;
+
+          pipewire = null;
+          freerdp = null;
+          xwayland = null;
         })
         # and then this overridden package's attributes are overridden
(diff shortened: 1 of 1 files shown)


modules/overlays/custom-packages.nix
```diff
@@ -37,19 +37,16 @@
         # First, weston package is overridden (passing colord = null)
         (prev.weston.override {
           colord = null;
+
+          pipewire = null;
+          freerdp = null;
+          xwayland = null;
```
vilvo, line 43: This will break legacy X apps - which I think is fine in the long run considering the security of Ghaf GUI stack to be isolated in the GUIVM. If/when legacy X apps are used with ghaf, those will go outside GUIVM to AppVMs of their own and must then include their own xwayland. Whether remote xwayland will work cross-domain (from AppVM with X app to GUIVM) is another story and trial to be tested beyond this PR.

vilvo (approved):

baz2142 (approved):

PR #140, opened 2023-05-17
Host minification module

remimimimimi:
Draft implementation of minification ADR in form of nixos module that reflects it's structure.

Some ADR items are not reflected in the module options, since they are not directly related to minification and will be added as a result of code refactoring. Can be refactored to module that will provide options for high-level configuration of ghaf host system (e.g enable/disable gui, update system, read only filesystem). 

Currently implementation is incomplete because of infinite recursion caused by conditional modules importing.

### DIFF
--- a/modules/host/default.nix
+++ b/modules/host/default.nix
@@ -4,18 +4,60 @@
   self,
   microvm,
   netvm,
-}: {lib, ...}: {
-  imports = [
-    (import ./minimal.nix)
+}: {
+  lib,
+  config,
+  pkgs,
+  ...
+} @ args: let
+  cfg = config.ghaf.host.minification;
+in
+  with lib; {
+    options.ghaf.host.minification = with lib; {
+      reduceProfile = mkEnableOption "reduced profile";
+      disableNetwork = mkEnableOption "disableing of network";
+      disableGetty = mkEnableOption "disabling of getty";
+      removeNix = mkEnableOption "nix tooling";
+      minimalKernelConfig = {
+        enable = mkEnableOption "minimal kernel config";
+        structuredExtraConfig = mkOption {
+          default = types.attrs;
+          example = literalExpression ''
+            with lib.kernel; {
+              SCHED_MUQSS = yes;
+            }'';
+          description = "Extra config for the kernel. See https://nixos.wiki/wiki/Linux_kernel#Custom_configuration.";
+          type = types.attrs;
+        };
+      };
+    };
 
-    microvm.nixosModules.host
+    # HACK: import other modules with explicit passing of arguments, otherwise import
+    # cause infinite recursion.
+    config = lib.mkMerge [
+      {
+        networking.hostName = "ghaf-host";
+        system.stateVersion = lib.trivial.release;
+      }
+      (import ../overlays/custom-packages.nix {inherit lib pkgs;})
+      (lib.mkIf cfg.reduceProfile (import ./minimal.nix {inherit pkgs lib;}))
+      (lib.mkIf (!cfg.disableNetwork) (import ./networking.nix {}))
+      (lib.mkIf cfg.disableGetty {
+        # XXX: Incompatable with `services.getty` options.
+        systemd.services."getty@".enable = lib.mkForce false;
+      })
+      (lib.mkIf cfg.removeNix {
+        # TODO: Patch raw-efi image build script to support building without relying on
+        # image nix tools (replace with packages from nixpkgs).
+        nix.enable = false;
+      })
+      (lib.mkIf cfg.minimalKernelConfig.enable {
+        boot.kernelPatches = lib.singleton {
+          name = "minimal";
+          patch = null;
+          extraStructuredConfig = cfg.minimalKernelConfig.structuredExtraConfig;
+        };
+      })
+    ];
+  }
 
-    ../overlays/custom-packages.nix
-
-    (import ./microvm.nix {inherit self netvm;})
-    ./networking.nix
-  ];
-
-  networking.hostName = "ghaf-host";
-  system.stateVersion = lib.trivial.release;
-}
--- a/modules/host/default.nix
+++ b/modules/host/default.nix
@@ -42,10 +42,7 @@ in
       (import ../overlays/custom-packages.nix {inherit lib pkgs;})
       (lib.mkIf cfg.reduceProfile (import ./minimal.nix {inherit pkgs lib;}))
       (lib.mkIf (!cfg.disableNetwork) (import ./networking.nix {}))
-      (lib.mkIf cfg.disableGetty {
-        # XXX: Incompatable with `services.getty` options.
-        systemd.services."getty@".enable = lib.mkForce false;
-      })
+      (lib.mkIf cfg.disableGetty (import ./nogetty.nix lib))
       (lib.mkIf cfg.removeNix {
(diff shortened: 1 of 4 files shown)


remimimimimi:
Infinite recursion is fixed by inlining module in 4273b8ed3362e8becf40c126df371c81c9de32a0

vadika:
Would you please squash the commits?

vilvo (changes requested):
Good progress! Thanks for sharing the draft PR for review.
OK
+ the PR branch builds and flashes to Orin AGX and boots to analyze and validate the minimal host work status in a run-time system
+ debug-access over network and serial with `-debug`-target ~ I think we could enable some module options in this development branch for better minimal host structural validation. See proposed areas in the end.
NOK
- no microvm -> no netvm ~ this was probably not intentional?

Proposing next development areas/improvements:
- microvm and netvm enabled as defaults - for both `-debug` and `-release` targets.
- everything gui related (apps and even weston) should be removed by default in the minimal host development branch. GUI stuff can be left out in development for now or optionally with configurable option via gui.nix or something. The GUI stack is to be isolated to guivm with GPU passthrough.
- nix-tooling, networking and read-write should be enabled by default in `-debug` targets. This will support development on minimal-host in other areas. 
- please set the stateVersion to `23.05 / unstable` - we are moving to it soon
- the first iteration of kernel config minification - e.g. drop graphics with the GUI stuff

I'll make sure to clarify the ADR further with diagrams.

vadika:
Another minification approach to consider https://gist.github.com/chrisdone/02e165a0004be33734ac2334f215380e

remimimimimi:
For now I just added options, without toggling behavior that were before. I think that their specialization for debug/release targets should be made during refactoring.

vilvo:
> For now I just added options, without toggling behavior that were before. I think that their specialization for debug/release targets should be made during refactoring.

Nice work on enabling the configurable options for minimal host. I think we need to add usage of the module to the flake.nix `-release`-targets - just for testing it in this branch.

modules/host/default.nix
```diff
@@ -4,18 +4,41 @@
...
+  ...
+} @ args: let
+  cfg = config.ghaf.host.minification;
+in
+  with lib; {
+    imports = [microvm.nixosModules.host];
 
-    microvm.nixosModules.host
+    options.ghaf.host.minification = with lib; {
+      reduceProfile = mkEnableOption "reduced profile";
```
vilvo, line 19: Here we could describe the reduced profile a bit more vocally - e.g. "Profile to reduce host size with NixOS supported options - see modules/host/minimal.nix and https://nixos.org/manual/nixos/stable/options.html for details"

modules/host/default.nix
```diff
@@ -4,18 +4,41 @@
...
+      disableNetwork = mkEnableOption "disableing of network";
```
vilvo, line 20: e.g. "Disable network by leaving the modules/host/networking.nix declarations out. Does not yet have effect on limiting host kernel networking support or user space networking tools."

modules/host/default.nix
```diff
@@ -4,18 +4,41 @@
...
+      disableGetty = mkEnableOption "disabling of getty";
+      removeNix = mkEnableOption "nix tooling";
```
vilvo, line 22: e.g. "Removes nix tooling from the host. Disables interaction with the host /nix/store/ and host configuration."

modules/host/default.nix
```diff
@@ -4,18 +4,41 @@
...
+      disableMicrovm = mkEnableOption "microvm";
```
vilvo, line 23: I'd propose we change this to `disableGuests` at this point. That way this will provide better semantics for the external users of the ghaf host module. It's a small change at this point but if we rename it later, it'll cause backward compatibility breaking change. We can also rename `mkEnableOption "microvm"` to `mkEnableOption "guestvms"` That will hide the internals of our microvm-based declarations and allow us to change the implementation if/when we have to use other implementations than `microvm.nix`.

Proposing also more vocal description - e.g. "Disables all ghaf reference guest virtual machines"

vilvo (changes requested):
Please address:
- enabling all minimal host options for `-release`-target - e.g. `.#packages.aarch64-linux.nvidia-jetson-orin-release` for testing - enabling testing this work is next critical step.
- minor updates to the descriptions of options

modules/host/default.nix
remimimimimi: I don't think that it's valuable to create elaborate descriptions for those options right now, because the will be changed in near refactoring.

modules/host/default.nix
```diff
@@ -4,18 +4,41 @@
...
+  cfg = config.ghaf.host.minification;
+in
+  with lib; {
+    imports = [microvm.nixosModules.host];
 
-    microvm.nixosModules.host
+    options.ghaf.host.minification = with lib; {
+      reduceProfile = mkEnableOption "reduced profile";
+      disableNetwork = mkEnableOption "disableing of network";
+      disableGetty = mkEnableOption "disabling of getty";
```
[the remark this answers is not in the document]
remimimimimi: Same as above. Option name should be changed to `enableInteractiveLogin`.

vilvo:
Thanks, I tested this with:
```
ghaf (1cd5d80) [$!?] took 2m59s
❯ git log --oneline -1
1cd5d80 (HEAD, remi/host-config) Enable minification for release targets

ghaf (1cd5d80) [$!?]
❯ nixos-rebuild --flake .#nvidia-jetson-orin-release --target-host root@192.168.1.149 --fast boot
```
and it's towards right direction. No GUI (nothing on Orin AGX DisplayPort out), and login is disabled in the release mode (apparently I have to do full reflash to `-debug` build to be able to use the device again)
screencapture
Technically we still have work to do regarding `Networking Setup`, `Ghaf graphical session` and `Getty`s in the next iterations as we can see from screencapture

vilvo:
Giving some more thought, `-release`-mode only disables the authentication module so there's no apparent impact from that to `Getty` which was started.

I'll do some more testing. I'll flash this with `sudo nix build .#packages.aarch64-linux.nvidia-jetson-orin-release-flash-script` just to check if the `getty` configuration has any impact.

modules/host/default.nix
```diff
@@ -4,18 +4,41 @@
...
+      removeNix = mkEnableOption "nix tooling";
+      disableGuestVms = mkEnableOption "guest vms";
+    };
```
vilvo, line 24: I meant adding `description`s to `mkEnableOption`s. Seems you chose to ignore the request and resolved those instead. Ok, postponed for later.

vilvo (approved):
Integrates to `-release`-targets and tested to do what it says.

mikatammi:
You merged this too quickly, more work is needed with regard to modules imported by modules/host/default.nix

Instead of conditionally importing each of the module, each of the modules should be imported, and the conditions should be put inside each module

vilvo:
That work could have been continued with further work after this iteration now that we have got more detailed examples from the modules of your work. There was no reason to revert this as we were careful to integrate this with the `-release`-target (of minor disturbance to any other work) for now. 
The problem with the random reverts based in just-in-time improved specs available only on example modules is that it flushes the parallel work down the toilet instead of improving it. That parallel work had DoD which was met but f* that. Sure, system can be developed this way as well. Let's base coming work to people reading your minds.

PR #153, opened 2023-05-27
Update to NixOS 23.05

mikatammi:


mikatammi:
:white_check_mark: The 23.05 change is green to go in from my side. Testers wanted to test this first, so please give your approvals from the testing side when ready

brianmcgillion:
just leaving notes:

nix build .#nvidia-jetson-orin-debug-from-x86_64 

I can see it connecting to awsarm for some tasks so not purely xcompile.

next test is to disable remote builders and binfmt to see what happens.

brianmcgillion:
```
error: builder for '/nix/store/lvx0rqdn59lvwv9z59s48lxkrq019ym1-keytar-aarch64-unknown-linux-gnu-7.9.0.drv' failed with exit code 1;
       last 10 log lines:
       > npm ERR! gyp ERR! System Linux 6.3.3
       > npm ERR! gyp ERR! command "/nix/store/hdwy8j1jvj2d68p449vvpf5j4mrgzldi-nodejs-18.16.0/bin/node" "/nix/store/hdwy8j1jvj2d68p449vvpf5j4mrgzldi-nodejs-18.16.0/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
       > npm ERR! gyp ERR! cwd /build/source
       > npm ERR! gyp ERR! node -v v18.16.0
       > npm ERR! gyp ERR! node-gyp -v v9.3.1
       > npm ERR! gyp ERR! not ok
       > 
       > npm ERR! Log files were not written due to an error writing to the directory: /nix/store/0f0f5wq5xiqsphzl9nq3basbznpc4plh-npm-deps/_logs
       > npm ERR! You can rerun the command with `--loglevel=verbose` to see the logs in your terminal
       > 
       For full logs, run 'nix log /nix/store/lvx0rqdn59lvwv9z59s48lxkrq019ym1-keytar-aarch64-unknown-linux-gnu-7.9.0.drv'.
error: 1 dependencies of derivation '/nix/store/nqsq22irw4f5rhm069cf780gzd581i2f-element-desktop-1.11.31-aarch64-unknown-linux-gnu.drv' failed to build
error (ignored): error: cannot unlink '/tmp/nix-build-libvpx-aarch64-unknown-linux-gnu-1.13.0.drv-0/source': Directory not empty
error: 1 dependencies of derivation '/nix/store/8d0hsvx9g2d6hbb96z3clmxp4wq3v7as-etc-weston.ini.drv' failed to build
error: 1 dependencies of derivation '/nix/store/djwwdgxavlfyakmwzlcq46fyn5sz05ik-system-path.drv' failed to build
error (ignored): error: cannot unlink '/tmp/nix-build-texinfo-aarch64-unknown-linux-gnu-7.0.3.drv-0/texinfo-7.0.3/tp': Directory not empty
error: 1 dependencies of derivation '/nix/store/1h0djmmjqc1nbi4ym5549368rx9yx98y-nixos-system-ghaf-host-23.05.20230525.3e01645.drv' failed to build
error (ignored): error: cannot unlink '/tmp/nix-build-linux-config-aarch64-unknown-linux-gnu-6.1.29.drv-0/linux-6.1.29/drivers/hid': Directory not empty
error (ignored): error: cannot unlink '/tmp/nix-build-lkl-aarch64-unknown-linux-gnu-2022-08-08.drv-0/source/drivers/rtc': Directory not empty
error: 1 dependencies of derivation '/nix/store/gg6zgavxw3760aswzikj69nczgg7mnhh-nixos-disk-image.drv' failed to build
```
So i think we can say that npm (gala app dependency) is still broken for xcompile.


this raises the issue that we should probably have a target for base system and system vms. then separate targets for app vms / apps in the host. with a meta package target "build all" so that we can isolate these issues better

for issues like this we should then create a tmp overlay that forces native compilation for components that are not yet xcompile friendly while we work with upstream nixpkgs to fix them

mikatammi:
> just leaving notes:
> 
> nix build .#nvidia-jetson-orin-debug-from-x86_64
> 
> I can see it connecting to awsarm for some tasks so not purely xcompile.
> 
> next test is to disable remote builders and binfmt to see what happens.

This is a problem that is neither fixed or caused by this PR. It is a known issue that the cross-compilation targets are not proper, the problem is caused by the fact that the microvm-configurations are still natively compiled. Separate nixosConfigurations that have enabled cross-compilation should be created

vilvo:
> > just leaving notes:
> > nix build .#nvidia-jetson-orin-debug-from-x86_64
> > I can see it connecting to awsarm for some tasks so not purely xcompile.
> > next test is to disable remote builders and binfmt to see what happens.
> 
> This is a problem that is neither fixed or caused by this PR. It is a known issue that the cross-compilation targets are not proper, the problem is caused by the fact that the microvm-configurations are still natively compiled. Separate nixosConfigurations that have enabled cross-compilation should be created

Note is valid but I agree we should keep xcompile-support separate from this PR as an objective. In addition, native compilation and cross-compilation must have separate targets - they will produce different derivations because of different toolchains. Then we can later on decide which one we support as primary target, keeping in mind the community priority has been to support native. My 2 cents is to support both and follow longer term how well the cross-compilation will work with the nix community support.

mikatammi:
There will be one more fix for this PR, to align the flash-script with the stuff from jetpack-nixos

unbel13ver (approved):

vilvo (approved):
Seems legit. I'm still building the target to test it but if it builds, it builds. Let's move on to build Ghaf on 23.05.

vilvo:
> Seems legit. I'm still building the target to test it but if it builds, it builds. Let's move on to build Ghaf on 23.05.

screencapture

Builds and runs 23.04.

mikatammi:
> There will be one more fix for this PR, to align the flash-script with the stuff from jetpack-nixos

I aligned the mk-flash-script.nix with the how it is called from jetpack-nixos/device-pkgs.nix and now from my side we are good to go with this

mikatammi:
I tested with AGX Orin and Intel NUC that there are no serious errors in system logs and I'm also able to log into NetVM

cazfi:
hydra.vedenemo.dev died with this. That's unsurprising, and likely not a big deal - just a lot faster deprecation of that hydra instance than what we expected.

However, also hydra2 builds fail: https://hydra2.vedenemo.dev/jobset/ghaf/ghaf
I've had no time to analyze the failures - so can't yet tell if they are something to fix on hydra or on the ghaf side.

mikatammi:
Based on the error, it tried to connect to localhost using ssh. Even if the keys would be in place it would probably fail because of the locking.

In my local setup I solved this by having separate machines declaration for the hydra service, and leaving the hydra host's main nix configuration (/etc/nix/machines) with only the remote builders.

Example of how I did the configuration for Hydra:

image

mikatammi:
What I still see with Hydra is that it fails to build the x86_64-linux flash script the first time with error "Unsupported system aarch64-linux", because the image it depends should be built on aarch64-linux. However if trying second time, it succeeds because the aarch64-linux image is already in the cache

cazfi:
> Based on the error, it tried to connect to localhost using ssh. Even if the keys would be in place it would probably fail because of the locking.

First of all, you saw the error that was now next to resolve, when I iteratetively fix (rather: work around, but we can still do that with hydra2; rules will be stricter on next instance) problems & retry.

I find it unlikely that building over ssh -connection would suddenly be a cause of problem.

> In my local setup I solved this by having separate machines declaration for the hydra service, and leaving the hydra host's main nix configuration (/etc/nix/machines) with only the remote builders.
 
That's what hydras' configuration has basically always been.

Did you look at the actual build failure at all?

cazfi:
> What I still see with Hydra is that it fails to build the x86_64-linux flash script the first time with error "Unsupported system aarch64-linux", because the image it depends should be built on aarch64-linux. However if trying second time, it succeeds because the aarch64-linux image is already in the cache

So no change there. Still fails the same way when a build gets triggered (that manual intervention to retrigger the build isn't really a solution).

PR #155, opened 2023-05-30
Add docker

jkuro-tii:
Added the docker module. It adds to Ghaf docker package and changes system configuration (systemd, groups, socket, etc.) to support docker.

### DIFF
--- a/modules/development/authentication.nix
+++ b/modules/development/authentication.nix
@@ -11,7 +11,7 @@ in {
     users."${user}" = {
       isNormalUser = true;
       password = password;
-      extraGroups = ["wheel" "video"];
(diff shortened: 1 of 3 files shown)


vilvo:
Thanks Jarek. Would you please refactor docker support under common `-debug`-targets and make it development module instead of host?
Like move import from `modules/host/default.nix` to https://github.com/tiiuae/ghaf/blob/main/targets/common-debug.nix#L14
and move `modules/host/docker.nix` to `modules/development/docker.nix`. The latter would also indicate that the docker-module for development is also available in guests instead of host only.

vilvo (approved):

Issue #122, opened 2023-04-27
`features.md` file needs status update based on the the current progress.


vilvo:
This whole `features.md` file needs status update based on the the current progress.

_Originally posted by @vilvo in https://github.com/tiiuae/ghaf/pull/121#discussion_r1178891084_

jenninikko:
https://tiiuae.github.io/ghaf/features/features.html

vilvo:
Closed with de09aa382e4405861779bb9c67d2c73ee3ae5dae

PR #157, opened 2023-05-30
doc: features update

vilvo:
* features update in progress on popular demand


jenninikko:
I would prefer to edit the text after merging, so just make sure that the information here is up-to-date and true.

vilvo:
Addresses https://github.com/tiiuae/ghaf/issues/122

PR #164, opened 2023-06-03
Add templates to allow easy inheritance of Ghaf

brianmcgillion:
Ghaf is a framework to allow the easy creation of
virtualized edge devices. As such it should be
used to derive projects that are configured to
the specific use cases.

These templates allow for that easy creation.

### DIFF
--- a/docs/src/ref_impl/custom_product.md
+++ b/docs/src/ref_impl/custom_product.md
@@ -1,0 +1,27 @@
+<!--
+    Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+    SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Derive a Custom Project From Ghaf
+
+Ghaf is a framework for creating virtualized edge devices, it is therefore expected that projects wishing to use Ghaf should import it to create a derived work
+for the specific use case.
+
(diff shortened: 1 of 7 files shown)


vilvo:
Could we develop this approach so that we'd avoid `<PROJ NAME>` renaming separately here https://github.com/tiiuae/ghaf/pull/164/files#diff-cb02caf321218f41518829b1217328404d838b7caf4832abdf5cb46e9b148881R25

`nix flake` supports templates with both `new` and `init` allow project folder (name) as parameter and further
```


--arg name expr

Pass the value expr as the argument name to Nix functions.

--argstr name string

Pass the string string as the argument name to Nix functions
```
which could probably be used for this renaming instead.

vilvo:
> ```
> --arg name expr
> 
> Pass the value expr as the argument name to Nix functions.
> 
> --argstr name string
> 
> Pass the string string as the argument name to Nix functions
> ```
> 
> which could probably be used for this renaming instead.

I checked that my proposal seems not supported https://github.com/NixOS/nix/issues/4017#issuecomment-942192727

I also tested the PR templates with:
`nix flake new --template github:brianmcgillion/ghaf/templates#target-x86_64-generic ~/generic_ghaf`
and it works like it says in the documentation.

vilvo (approved):

PR #170, opened 2023-06-14
make the ssh daemon configurable module

brianmcgillion:


mikatammi (approved):
A further development idea would be to make the SSH keys itself configurable through an option, but this is good for now

PR #172, opened 2023-06-15
Revert "Host minification module"

brianmcgillion:
Reverts tiiuae/ghaf#140


brianmcgillion:
revert for now until reviewed and the enabling is enabled within the modules

PR #173, opened 2023-06-15
refactor host module

brianmcgillion:


modules/host/networking.nix
```diff
@@ -1,31 +1,46 @@
...
-        {
-          addressConfig.Address = "192.168.101.2/24";
-        }
-      ];
+{
+  config,
+  lib,
+  ...
+}: let
+  cfg = config.ghaf.networking.host;
```
mikatammi, line 8: Maybe this should be ghaf.host.networking?

templates/default.nix
```diff
@@ -1,6 +1,6 @@
 # Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
-{self}: {
+{...}: {
```
mikatammi, line 3: A nit-pick but can't we just remove the function from this altogether, i.e. the `{...}:` part, and then just remove the `{self}` from the flake. I generally don't like using this `...`-syntax because it hides things.

If we need to pass self later we can then add it again.

mikatammi (approved):

PR #179, opened 2023-06-16
Fix building Weston against nixos-unstable

mikatammi:
In NixOS 23.05, the Weston package still takes colord as a depdendency, but it has been removed in nixos-unstable. Define pipewireSupport, rdpSupport and xwaylandSupport to be false if the arguments are present in the package.


brianmcgillion (approved):
LGTM

PR #181, opened 2023-06-16
Move towards the profile based definitions

brianmcgillion:
Start the migration to using profiles to define
the feature set that is included in any particular target.

Many TODO and cleanup still needed.


modules/profiles/applications.nix
```diff
@@ -0,0 +1,27 @@
...
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  cfg = config.ghaf.profiles.applications;
+in
+  with lib; {
+    options.ghaf.profiles.applications = {
+      enable = mkEnableOption "Enable some sample applications";
```
remimimimimi, line 14: Description should be changed to "some sample applications" according to https://github.com/NixOS/nixpkgs/blob/nixos-23.05/lib/options.nix#L105

PR #183, opened 2023-06-19
Add autogenerated documentation for ghaf options

remimimimimi:
Allows easier discoverability of ghaf modules options for users of the framework.

There are two main cases of modification.

The first is in case of changes in the documentation. In this case, you need to move the empty file to the appropriate location and also change the path in `ghaf/docs/default.nix`.

The second is in case of changing the code of modules. Some modules may cause errors and therefore may be added to exceptions in the list of modules for which documentation is generated. Edit `flake.nix`.

Screenshot 2023-06-19 at 11 53 00


remimimimimi:
> 1. If it is something for the developers only, maybe we can move it under 5.2.2?
>        It would be:
> 
> 
>     * Development
>       -- Cross-Compilation
>       -- Options

I guess this not for developers only, because ghaf is a framework.

>     2. And maybe we can add some additional information to the section name: options of what is it?

I think we can name this appendix section "Modules options". That would be intuitive enough if user have experience with nix, which is required to use ghaf AFAIK.

jenninikko:
> I guess this not for developers only, because ghaf is a framework.

Ok, let's try from another side.
Is this information needed only for the development process, or, for example, some guys from CI/CD can use it? Or, if I want to set up the demo environment, do I need this?

vilvo:
> > I guess this not for developers only, because ghaf is a framework.
> 
> Ok, let's try from another side. Is this information needed only for the development process, or, for example, some guys from CI/CD can use it? Or, if I want to set up the demo environment, do I need this?

This would be useful for Usage - to customize Ghaf in an external project - here https://github.com/tiiuae/ghaf/blob/main/docs/src/ref_impl/usage.md#customize-your-ghaf-based-project

flake.nix
```diff
@@ -61,7 +61,13 @@
       in {
         packages.doc = pkgs.callPackage ./docs {
           revision = lib.version;
-          options = self.nixosConfigurations.vm-debug.options;
+          options = let
+            cfg = nixpkgs.lib.nixosSystem {
+              system = "aarch64-darwin";
```
brianmcgillion, line 66: Darwin?

modules/virtualization/docker.nix
```diff
@@ -1,27 +0,0 @@
...
-{
-  config,
-  pkgs,
-  modulesPath,
-  ...
-}: let
-  cfg = config.ghaf.virtualization.docker.daemon;
-in
-  with lib; {
-    options.ghaf.virtualization.docker.daemon = {
```
mikatammi, line 12: Did you accidentally remove this module in this PR?

Although I think it is not very useful module, basically one can just set the virtualisation.docker options directly if one wants

jenninikko:
@remimimimimi do not forget to change the section name from "Options" to "Module Options" and options.md -> module_options.md correspondingly, as I was planning to do this after merge but right now you have a chance to fix some references or whatever links you have to the file =)

jenninikko:
And please add a commit message to this pull request. Maybe this can help https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md#commit-message-guidelines

modules/virtualization/docker.nix
remimimimimi: No, it was on purpose. The module didn't work and caused infinite recursion during evaluation because of conditional import in the config attribute.

vilvo (approved):
Reviewed upon author request. Looks nice and good to me. Link from usage works as expected.
screencapture

jenninikko (changes requested):
Could you please add the following information:

- what precisely this pull request adds, what is the purpose of this autogenerated documentation
- how to maintain the process further if I change the structure or rename the section that you added with this pull request

jenninikko (approved):
Thank you honey for the changes! From the documentation side, it looks good.

modules/virtualization/docker.nix
mikatammi: Although I think it is a quite pointless module, I'd still like to explore the route of trying to fix it first in another PR, instead of completely removing it in this PR.

mikatammi (changes requested):

mikatammi:
https://github.com/tiiuae/ghaf/pull/217/files

remimimimimi:
> https://github.com/tiiuae/ghaf/pull/217/files Fix Docker module PR here

Hah, I haven't seen your pr, so I fixed it by myself also 😃 .

Also, paths to source files in the generated documentation now refer to the GitHub repository.

modules/virtualization/docker.nix
remimimimimi: Fixed in [https://github.com/tiiuae/ghaf/pull/217](https://github.com/tiiuae/ghaf/pull/217/files)

mikatammi (approved):

jenninikko:
Make sure there is a blank line at the end of the file.

PR #185, opened 2023-06-20
Fix date and time synchronization issues

nesteroff:
In Ghaf date and time are synchronized using systemd-timesyncd service. In order to get the correct date and time it requires a working internet connection. However, internet connection doesn't work when the current date and time are not correct because we have Domain Name System Security Extensions (DNSSEC) enabled. This leads to a chicken and egg problem. It's especially bad for Nvidia Orin devices since they don't have batteries for RTC.

Systemd-timesyncd resolves this issue by setting SYSTEMD_NSS_RESOLVE_VALIDATE=0 environmental variable in the unit configuration file. It disables DNSSEC validation for the timesyncd service so that it can connect to the NTP server even if current date and time is incorrect. However, this option doesn't work in Nix because it uses NSCD instead of calling NSS libraries directly. NSCD is a caching service which communicates with glibc over a Unix socket and it doesn't have SYSTEMD_NSS_RESOLVE_VALIDATE=0 set.

This patch disables NSCD in systemd-timesyncd using an internal glibc function when DNSSEC validation is disabled and it also sets LD_LIBRARY_PATH so that the service can find the libnss_resolve.so.2 library in the Nix store.

Testing procedure:
- Stop timesyncd: sudo systemctl stop systemd-timesyncd
- Delete the clock file: sudo rm /var/lib/systemd/timesync/clock
- Set a wrong date and time: sudo date 01010101
- Start timesyncd or reboot the system: sudo systemctl start systemd-timesyncd
- Older versions of Ghaf will not be able to synchronize time and DNS will not work but with this patch the system should get the correct date and time from the NTP server and DNS requests with DNSSEC should work fine.

### DIFF
--- a/overlays/custom-packages.nix
+++ b/overlays/custom-packages.nix
@@ -66,6 +66,20 @@
             # Only override mesonFlags if colord argument is accepted
             mesonFlags = prevAttrs.mesonFlags ++ ["-Ddeprecated-color-management-colord=false"];
           });
+      systemd = prev.systemd.overrideAttrs (prevAttrs: {
+        patches = prevAttrs.patches ++ [./systemd-timesyncd-disable-nscd.patch];
+        postPatch =
+          prevAttrs.postPatch
+          + ''
+            substituteInPlace units/systemd-timesyncd.service.in \
+              --replace \
+              "Environment=SYSTEMD_NSS_RESOLVE_VALIDATE=0" \
+              "${lib.concatStringsSep "\n" [
+              "Environment=LD_LIBRARY_PATH=$out/lib"
+              "Environment=SYSTEMD_NSS_RESOLVE_VALIDATE=0"
+            ]}"
+          '';
+      });
     })
   ];
 }
--- a/overlays/systemd-timesyncd-disable-nscd.patch
+++ b/overlays/systemd-timesyncd-disable-nscd.patch
@@ -1,0 +1,46 @@
+From bb150a178bebb88ce3bfe8c726c75d495423b4a2 Mon Sep 17 00:00:00 2001
+From: Yuri Nesterov <yuriy.nesterov@unikie.com>
+Date: Wed, 21 Jun 2023 17:17:38 +0300
+Subject: [PATCH] timesyncd: disable NSCD when DNSSEC validation is disabled
+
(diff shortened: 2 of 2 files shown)


mikatammi:
I wonder if this will this cause massive rebuild of almost every package in nixpkgs when this gets merged

nesteroff:
Yeah, systemd is in every house on every street now. When building Ghaf for Nvidia Orin AGX with default apps disabled it rebuilds about 133 packages on my system.

overlays/custom-packages.nix
```diff
@@ -66,6 +66,23 @@
             # Only override mesonFlags if colord argument is accepted
             mesonFlags = prevAttrs.mesonFlags ++ ["-Ddeprecated-color-management-colord=false"];
           });
+      systemd =
+        (prev.systemd.override {
```
mikatammi, line 70: By the way I think you can just write prev.systemd.overrideAttrs if you are not going to override any of the callPackage's inputs

brianmcgillion:
> Yeah, systemd is in every house on every street now. When building Ghaf for Nvidia Orin AGX with default apps disabled it rebuilds about 133 packages on my system.

Just get the hydra folks to prebuild a branch and populate the cache before this hits mainline, or we can take the rest of the week off :D

nesteroff:
> > Yeah, systemd is in every house on every street now. When building Ghaf for Nvidia Orin AGX with default apps disabled it rebuilds about 133 packages on my system.
> 
> Just get the hydra folks to prebuild a branch and populate the cache before this hits mainline, or we can take the rest of the week off :D

It's not that bad :) It takes just a couple of hours to rebuild everything that depends on systemd on the arm server.

overlays/custom-packages.nix
nesteroff: Fixed it, thanks!

nesteroff:
BTW, there was an interesting discussion about this issue [here](https://github.com/systemd/systemd/pull/28110)

mikatammi (approved):
I'm fine with putting this in now so we can start tp get the caches populated. We can then revert it if it causes problems

brianmcgillion:
Sounds good

PR #187, opened 2023-06-21
Github actions: Use KTUSAWRK_GHAF_ACCESS_TOKEN for docs

cazfi:


ktusawrk (approved):
Token name is correct.

PR #197, opened 2023-06-27
doc: crosvm x86 pcie passthrough

josa41:
Pcie passthrough example for x86 crosvm

### DIFF
--- a/docs/src/technologies/passthrough.md
+++ b/docs/src/technologies/passthrough.md
@@ -14,3 +14,4 @@ Whether the device platform is x86 or ARM, the passthrough device needs to be bo
 Our current supported passthrough devices implementations:
 - [NVIDIA Jetson AGX Orin: UART Passthrough](nvidia_agx_pt_uart.md)
 - [NVIDIA Jetson AGX Orin: PCIe Passthrough](nvidia_agx_pt_pcie.md)
+- [Generic x86: PCIe Passthrough on crosvm](x86_pcie_crosvm.md)
--- a/docs/src/technologies/x86_pcie_crosvm.md
+++ b/docs/src/technologies/x86_pcie_crosvm.md
@@ -1,0 +1,53 @@
+<!--
+    Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+    SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# x86 PCIe Device Passthrough with crosvm
+
+
+## Enabling PCIe Devices for VFIO with driverctl
+
+As with other passthroughs, first, we need to set the target device to use VFIO
+ driver. This can be done manually or by using the [driverctl](https://gitlab.com/driverctl/driverctl) tool as below.
+ 
+> Running driverctl requires root permissions.
+
+```
+export BUS="0000:01:00.0"
+driverctl --nosave set-override ${BUS} vfio-pci
+```
+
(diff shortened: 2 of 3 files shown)


brianmcgillion (approved):
LGTM

docs/src/technologies/x86_pcie_crosvm.md
```diff
@@ -0,0 +1,44 @@
...
+## Enabling PCIe Devices for VFIO with using driverctl tool
+As with other passthroughs we first need to set the target device to use vfio
+ driver. This can be done manually or by using `driverctl` tool as below. Note
+ that running driverctl requires root permissions.
+```
+export BUS="0000:01:00.0"
+driverctl set-override ${BUS} vfio-pci
+```
+
+Starting crosvm as an example. In some cases crosvm may need some privileged
```
jenninikko, line 18: Replace with:

Let's consider the example of starting crosvm.

In some cases, crosvm may need privileged permissions to work properly. This applies especially for passthrough hardware devices as VFIO devices are generally owned by the root user. For simplicity, it may be easier to run crosvm as the root user but it should be possible to set up correct permissions so that running as root is not needed.

crosvm expects the device's system path as its `--vfio` argument. The argument can differ depending on how the passthrough device is referenced in QEMU. Using the  `guest-address` option is not strictly required by the source documentation but it gives a bit more control for handling the passthrough device on the guest side.

docs/src/technologies/passthrough.md
```diff
@@ -10,4 +10,5 @@ and their memory access in one or several VMs. This reduces the Trusted Code Bas
 
 Our current supported passthrough devices implementations:
 - [NVIDIA Jetson AGX Orin: UART Passthrough](nvidia_agx_pt_uart.md)
-- [NVIDIA Jetson AGX Orin: PCIe Passthrough](nvidia_agx_pt_pcie.md)
\ No newline at end of file
+- [NVIDIA Jetson AGX Orin: PCIe Passthrough](nvidia_agx_pt_pcie.md)
+- [X86: PCIe Passthrough on crosvm](x86_pcie_crosvm.md)
```
jenninikko, line 14: X86 -> Generic x86

jenninikko (changes requested):
Please sorry that it took me so long T_T
You can find my comments below.

Also, please update the SUMMARY file so that a new section can be found from the ToC on the left:

```
- [Technologies](technologies/technologies.md)
    - [Passthrough](technologies/passthrough.md)
        - [NVIDIA Jetson AGX Orin: UART Passthrough](technologies/nvidia_agx_pt_uart.md)
        - [NVIDIA Jetson AGX Orin: PCIe Passthrough](technologies/nvidia_agx_pt_pcie.md)
        - [Generic x86: PCIe Passthrough on crosvm](technologies/x86_pcie_crosvm.md)
    - [Hypervisor Options](technologies/hypervisor_options.md)
```

For more information, see https://github.com/tiiuae/ghaf/blob/main/docs/README-docs.md#adding-new-files, point 3.

docs/src/technologies/x86_pcie_crosvm.md
josa41: "The argument can differ depending on how the passthrough device is referenced in QEMU." Good rephrasing but also changes the meaning a bit. I rewrote it a bit in edit.

jenninikko (approved):

PR #200, opened 2023-06-29
Reference configuration support for orin nx

emrahbillur:
Orin Nx support related work is added in this branch.


brianmcgillion:
Please, create a branch from main with just the 1 commit that you are adding

PR #209, opened 2023-07-03
Initial changes for Orin NX support

emrahbillur:
Orin NX support
Changes for Orin Nx support related work. 

Signed-off-by: Emrah Billur emrah.billur@unikie.com

### DIFF
--- a/modules/hardware/nvidia-jetson-orin-nx/default.nix
+++ b/modules/hardware/nvidia-jetson-orin-nx/default.nix
@@ -1,0 +1,12 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Top-level module entry point for the Orin family of chips
+{lib, ...}:
+with lib; {
+  imports = [
+    ./partition-template.nix
+    ../../boot/systemd-boot-dtb.nix
+    ./jetson-orin-nx.nix
+  ];
+}
--- a/modules/hardware/nvidia-jetson-orin-nx/jetson-orin-nx.nix
+++ b/modules/hardware/nvidia-jetson-orin-nx/jetson-orin-nx.nix
@@ -1,0 +1,89 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Configuration for NVIDIA Jetson AGX Orin Nx
+{
+  lib,
+  config,
+  ...
+}: let
+  cfg = config.ghaf.hardware.nvidia.orin;
+in
+  with lib; {
+    options.ghaf.hardware.nvidia.orin = {
+      # Enable the Orin boards
+      enable = mkEnableOption "Orin hardware";
+
+      flashScriptOverrides.preFlashCommands = mkOption {
+        description = "Commands to run before the actual flashing";
+        type = types.str;
+        default = "";
+      };
+
+      somType = mkOption {
+        description = "SoM config Type (NX|AGX|Nano)";
+        type = types.str;
+        default = "nx";
+      };
+
+      carrierBoard = mkOption {
+        description = "Board Type";
+        type = types.str;
+        default = "devkit";
+      };
+    };
+
+    config = mkIf cfg.enable {
+      hardware.nvidia-jetpack = {
+        enable = true;
+        som = "orin-${cfg.somType}";
+        carrierBoard = "${cfg.carrierBoard}";
+        modesetting.enable = true;
+
+        flashScriptOverrides = {
+          flashArgs = lib.mkForce ["-r" config.hardware.nvidia-jetpack.flashScriptOverrides.targetBoard "nvme0n1p1"];
+        };
+      };
+
+      nixpkgs.hostPlatform.system = "aarch64-linux";
+
+      ghaf.boot.loader.systemd-boot-dtb.enable = true;
+
+      boot.loader = {
+        efi.canTouchEfiVariables = true;
+        systemd-boot.enable = true;
+      };
+
+      boot.kernelPatches = [
+        {
+          name = "passthrough-patch";
+          patch = ./pci-passthrough-test.patch;
+        }
+        {
+          name = "vsock-config";
+          patch = null;
+          extraStructuredConfig = with lib.kernel; {
+            VHOST = yes;
+            VHOST_MENU = yes;
+            VHOST_IOTLB = yes;
+            VHOST_VSOCK = yes;
+            VSOCKETS = yes;
+            VSOCKETS_DIAG = yes;
+            VSOCKETS_LOOPBACK = yes;
+            VIRTIO_VSOCKETS_COMMON = yes;
+          };
+        }
+      ];
+
+      hardware.deviceTree = {
+        enable = true;
+        name = "tegra234-p3767-host-passthrough.dtb";
+      };
+
+      # Passthrough Jetson Orin WiFi card
+      boot.kernelParams = [
+        "vfio-pci.ids=10ec:8168"
+        "vfio_iommu_type1.allow_unsafe_interrupts=1"
+      ];
+    };
+  }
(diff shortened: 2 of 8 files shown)


emrahbillur:
Here this version has ethernet working as the drone team is working with but still a change is needed for netvm.

targets/nvidia-jetson-orin-nx.nix
vilvo, line 1: We should avoid copying this file as commented in previous PR by @mikatammi and @brianmcgillion
Have a look: `git diff emrah/support-for-nx:targets/nvidia-jetson-orin-nx.nix origin/main:targets/nvidia-jetson-orin.nix`
The diff is essentially only the PCI path.
What if we'd rename the file in `main` to `targets/nvidia-jetson.nix` and pass the name `[nvidia-jetson-orin-agx|nvidia-jetson-orin-nx]` from imports in `targets/default.nix` for respective targets? Maintenance of the nvidia-jetson family reference configurations would be easier and less error prone.

targets/default.nix
```diff
@@ -13,6 +13,7 @@
 }:
 lib.foldr lib.recursiveUpdate {} [
   (import ./nvidia-jetson-orin.nix {inherit self lib nixpkgs nixos-generators microvm jetpack-nixos;})
+  (import ./nvidia-jetson-orin-nx.nix {inherit self lib nixpkgs nixos-generators microvm jetpack-nixos;})
```
vilvo, line 16: See comment https://github.com/tiiuae/ghaf/pull/209/files#r1255589951 for the context:

Could we do instead:
```
  (import ./nvidia-jetson-orin.nix {inherit self lib nixpkgs nixos-generators microvm jetpack-nixos nvidia-jetson-orin-agx;})
  (import ./nvidia-jetson-orin.nix {inherit self lib nixpkgs nixos-generators microvm jetpack-nixos nvidia-jetson-orin-nx;})
```
See passing arguments with `import` at https://nixos.org/guides/nix-pills/functions-and-imports.html#idm140737320462832

modules/hardware/nvidia-jetson-orin-nx/jetson-orin-nx.nix
```diff
@@ -0,0 +1,89 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Configuration for NVIDIA Jetson AGX Orin Nx
```
vilvo, line 4: `# Configuration for NVIDIA Jetson Orin NX`

modules/hardware/nvidia-jetson-orin-nx/default.nix
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Top-level module entry point for the Orin family of chips
```
vilvo, line 4: Is the comment statement related to a future intent? Now this is very much NX specific rather than entry point for the "Orin family of chips" - both the file path and the imports.

modules/hardware/nvidia-jetson-orin-nx/default.nix
vilvo, line 1: If this is to be entry point for the "Orin family of chips", the file path should reflect that.
Please rename to `modules/hardware/nvidia-jetson-orin/default.nix` and import the SoM variant specifics based on named argument (`[NX|AGX|Nano]`)

modules/hardware/nvidia-jetson-orin-nx/jetson-orin-nx.nix
vilvo, line 1: Please rename to `modules/hardware/nvidia-jetson-orin/jetson-orin-nx.nix` with https://github.com/tiiuae/ghaf/pull/209/files#r1255612184

modules/hardware/nvidia-jetson-orin-nx/jetson-orin-nx.nix
```diff
@@ -0,0 +1,89 @@
...
+            VIRTIO_VSOCKETS_COMMON = yes;
+          };
+        }
+      ];
+
+      hardware.deviceTree = {
+        enable = true;
+        name = "tegra234-p3767-host-passthrough.dtb";
+      };
+
```
vilvo, line 82: Could we define these https://github.com/tiiuae/ghaf/pull/209/files#r1255612184 `NX` PCI paths here under `hardware`?

targets/nvidia-jetson-orin-nx.nix
emrahbillur: True. I'll review the code to that way.

vilvo:
> Here this version has ethernet working as the drone team is working with but still a change is needed for netvm.

We need ethernet working via passthrough to netvm - this recent documentation from @josa41 may help https://tiiuae.github.io/ghaf/technologies/nvidia_agx_pt_pcie.html#binding-device-for-vfio

Also, the drone team will not need any of the GUI stuff - setting those `false` will reduce the target build time significantly.

vilvo:
> Orin NX support Changes for Orin Nx support related work.

Does the flashing to `nvme0n1p1` via NVIDIA APX mode work now? 
Or should I still test via external USB boot media?

emrahbillur:
No  nvme0n1p1 flashing yet. I just replaced it to make myself to update that script. Still the USB media. I created my linux script to manually flash the nx but it is still not in ghaf.

emrahbillur:
While waiting for review I made a small change and tested to pass         
modules = [
          {
            microvm.devices = [
              {
                bus = "pci";
                path = "0008:01:00.0";
              }
            ];
nvidia-jetson-orin-nx.nix file in the targets section .
Where this change allowed passing the ethernet device to netvm. Unfortunately the even though netvm itself detects the r8168 network card as r8169 and enables the device, it did not run due to not detecting the carrier. 
I tested with mii-tool  where it detects cable and negotiates with switch according to speed. But ethtool detects no cable.

emrahbillur:
These above changes of pass of the pci device makes netvm  run at least. But I only staged that change locally and made a build on local flake.

vilvo:
> While waiting for review I made a small change and tested to pass modules = [ { microvm.devices = [ { bus = "pci"; path = "0008:01:00.0"; } ]; nvidia-jetson-orin-nx.nix file in the targets section . Where this change allowed passing the ethernet device to netvm. Unfortunately the even though netvm itself detects the r8168 network card as r8169 and enables the device, it did not run due to not detecting the carrier. I tested with mii-tool where it detects cable and negotiates with switch according to speed. But ethtool detects no cable.

This sounds familiar - please sync with @josa41 - he did a significant work with a prev reference device (NXP i.MX8) and has insight into what's required for ethernet passthrough.

modules/hardware/nvidia-jetson-orin-nx/jetson-orin-nx.nix
emrahbillur: Re organizing all the structure to accept parametric orin type and the related modules also.

emrahbillur:
After Brian's rework through the Orin to support AGX and NX structure this PR is not needed anymore. I'll only add a small section to Brian's work as

```
    somDevices = {
    "agx" = {
      bus = "pci";
      path = "0001:01:00.0";
    };
    "nx" = {
      bus = "pci";
      path = "0008:01:00.0"; 
    };
  };
  formatModule = nixos-generators.nixosModules.raw-efi;
  nvidia-jetson-orin = som: variant: extraModules: let
    netvmExtraModules = [
      {
        # This part should be conditional for AGX 01:01 for NX 08:01
        microvm.devices = [
          {
            bus = somDevices."${som}".bus;
            path = somDevices."${som}".path;
          }
        ];
```
to add support for different devices to be passed through in changing orin versions.

vilvo:
> After Brian's rework through the Orin to support AGX and NX structure this PR is not needed anymore. I'll only add a small section to Brian's work as

This can be closed and you can create a new PR for the small section addition.

emrahbillur:
After rework of Brian Nx support is already merged to some level. I'll add a small fix for that in a new pull request.

PR #213, opened 2023-07-05
Declarative NetVM

mikatammi:
* Move from flake-based microvm-configuration to declarative
      configuration.
    * Cross-compilation enabled for NetVM, when host has cross-compilation
      enabled.
    * Refactored modules/host from function to a nixosModule.
    * While building documentation, microvm-host.nix does not need to be
      filtered out anymore.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,17 +64,12 @@
           options = let
             cfg = nixpkgs.lib.nixosSystem {
               inherit system;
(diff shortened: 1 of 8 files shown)


mikatammi:
Ready for review and merge now!

brianmcgillion (approved):

PR #215, opened 2023-07-07
Polarfire Icicle-kit reference platform

codehub8:
Added services
- avahi
- dendrite
- gpsd
- postgresql

Fixed docker enablement issue
Added module for camera streaming using RTSP
Patching of postgresql for MESH
Enabled template for Polarfire


modules/host/bare-host.nix
mbssrc, line 1: We can use the default.nix here as well, and move the additional configurations into riscv target. Note that options for Ghaf have changed, so the `host.networking.netvm.enable` should not be required anymore.

modules/services/avahi.nix
mbssrc, line 1: All project related packages and configurations should be in the particular project. So let's move these services there.

modules/virtualization/docker.nix
mbssrc, line 1: This has been fixed now and can be removed.

overlays/custom-packages.nix
mbssrc, line 1: Both `wpa_supplicant` patch for the mesh and camera support should be moved to the projects as well, as they are specific to it.

mbssrc (commented):
Good stuff, thank you. After changes we should be able to pull this in.

flake.lock
brianmcgillion, line 1: Can this be removed as the nixpkgs has recently been bumped to a newer version

modules/host/default.nix
```diff
@@ -26,6 +26,7 @@
       ghaf = {
         virtualization.microvm-host.enable = true;
         host.networking.enable = true;
+        host.networking.netvm.enable = true;
       };
```
brianmcgillion, line 30: This is covered in the targets/board configs

modules/host/networking.nix
brianmcgillion, line 1: Should rebase to see and use updated structure

modules/module-list.nix
brianmcgillion, line 1: These should be removed and placed in a project derived from a template

modules/host/bare-host.nix
codehub8: In default.nix, following configurations are irrelevant for Polarfire:
```
    (import ../virtualization/microvm/microvm-host.nix {inherit self microvm netvm;})
    ./networking.nix

    {
      ghaf = {
        virtualization.microvm-host.enable = true;
        host.networking.enable = true;
        host.networking.netvm.enable = true;
      };
    }

```

PR #218, opened 2023-07-10
Minimal systemd test configuration

mbssrc:
This is a first draft for minimal systemd implementation in Ghaf, for discussion only. Do not merge!

Changes:

- Added minimal systemd package configuration
- Added 2 new modules: host systemd configuration and boot (systemd in initrd) configuration, tested on vm and generic-x86
- Added netvm configuration, but currently does nothing. Requires discussion how to integrate into microvm

Notes:

- This draft only showcases the options for the systemd configuration. The general idea is to have a minimal configuration, and then add required features in the different parts of the system (i.e., boot/initrd, host, system-vms, etc.)
- The many commented options are to ease testing 

Points for discussion:

- General consent on the way to modularize and configure systemd in Ghaf
- Way forward to integrate systemd confs into microvm / system vms 

-- Update 31.07.23:

General:
- Package configuration now based on upstream `minimalSystemd` with additional parameters (e.g., removed kexec) 
- Minimal systemd is now default host configuration for all systems. Tested on vm, generic-x86, and icicle.
 
 Host configuration:
- Added Ghaf configuration options to account for 1) features to be removed from host in the future, 2) features to be enabled in certain configurations, and 3) solve some interdependencies. These options determine the systemd build flags
- Suppress and thus remove system unit files. This currently seems to be the better option compared trying to disable them (e.g.,  `services.systemd-kexec.enable = false;` ), which works for many and would mask the service, but leads to build failures. The current configuration removes some units unconditionally for all configurations, a number conditionally depending on the ghaf options, and others in debug mode. Some user unit files are also conditionally masked.

Stage1 configuration:
- As of now, stage1 systemd only works for vm builds

Future work:
- Currently looking into removing binaries as part of `postInstall`
- Systemd service hardening work can start once this is merged
- Once we build a 'full' system image for VMs we can build individual systemd configurations


modules/systemd/boot.nix
```diff
@@ -0,0 +1,109 @@
...
+    withKmod = true;
+    withTpm2Tss = true;
+  };
+    
+  # Excluded systemd units for initrd
+  excludedInitrdUnits = [
+      
+    # Targets
+    # "basic.target"
+    # "swap.target"
```
vilvo, line 34: We might want `swap`. Minimal host could be allocated minimal RAM mitigated with `swap`. `swap` can be encrypted.
No swap can lead to OoO and poor perf in general.

modules/systemd/boot.nix
```diff
@@ -0,0 +1,109 @@
...
+    "sys-kernel-config.mount"
+    "systemd-modules-load.service"        
+    "systemd-halt.service"
+    "sigpwr.target"
+    "rescue.service"
+    "systemd-journald.service"
+    "systemd-poweroff.service"
+    "systemd-reboot.service"
+    "systemd-sysctl.service"
+    "systemd-ask-password-console.service"
```
vilvo, line 74: Would 
`(if config.ghaf.profiles.debug.enable then "systemd-ask-password-console.service" else "")`
work here?

modules/systemd/boot.nix
mbssrc: The elements not commented are here the surpressed units, so the ones commented out are included. Also note that this is only for the systemd in initrd (so stage 1 of the boot), not the host systemd configuration or other.

modules/systemd/boot.nix
mbssrc: Same as above for the options. My approach here was to disable all shell access except for emergency access in debug, as in stage 1 there should be no need to access other than debugging. For the host configuration, something along the lines of what you suggested should be implemented though, also keeping in mind the host only configuration. As of now, I only excluded `getty.target ` as I think this was causing the glitch in vm configuration...

modules/systemd/boot.nix
```diff
@@ -0,0 +1,109 @@
...
+    # Targets
+    # "basic.target"
+    # "swap.target"
+    # "initrd.target"
+    # "initrd-fs.target"
+    # "initrd-switch-root.target"
+    # "initrd-root-device.target"
+    # "initrd-root-fs.target"
+    # "sysinit.target"
+    (if config.ghaf.profiles.debug.enable then "emergency.target" else "")
```
remimimimimi, line 41: It would be cleaner if you rewrite this way
```nix
[
  # targets
] ++ lib.optional config.ghaf.profiles.debug.enable "emergency.target";
```

remimimimimi (changes requested):

modules/systemd/boot.nix
mbssrc: yes will do that - thanks

modules/systemd/minimal.nix
```diff
@@ -0,0 +1,58 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  pkgs,
+}: {
+
+  # Minimal Systemd as per /nixpkgs/pkgs/top-level/all-packages.nix
+  # + additional options as per /nixpkgs/pkgs/os-specific/linux/systemd/default.nix
+  minimalSystemd = pkgs.systemd.override {
```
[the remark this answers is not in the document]
mbssrc: I have experimented with that - it makes sense if we would override options that are not exposed, which is not the case at the moment. Do you have any optimization in mind where we could utilize an overlay?

Personally I'd like using an overlay, but this is not necessarily shared by everyone. I propose we introduce it when we have a reason to. Opinions welcome.

modules/systemd/netvm.nix
```diff
@@ -0,0 +1,69 @@
...
+    pname = "netvm-systemd";
+    withAcl = true;
+    # withAnalyze = true;
+    # withApparmor = true;
+    # withAudit = true;
+    withCompression = true;
+    withCoredump = true;
+    # withCryptsetup = true;
+    # withDocumentation = true;
+    withEfi = pkgs.stdenv.hostPlatform.isEfi;
```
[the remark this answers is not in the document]
mbssrc: The general idea was to turn all options off, and enable where needed. So turning it off by default does exactly what you suggest? Or am I missing your point?

modules/systemd/host.nix
```diff
@@ -0,0 +1,345 @@
...
+    withTimedated = true;
+    withTimesyncd = cfg.withNetwork;
+    withTpm2Tss = cfg.withTpm;
+    withUtmp = cfgDebug || (cfg.withAudit && cfg.withLogin);
+  };
+
+  # Definition of suppressed system units in systemd configuration. This removes the units and has priority.
+  # Required to avoid build failures when only disabling the units, and removes unit files in '/etc'. Note that
+  # errors will be silently ignored.
+  systemUnitConfiguration =
```
mikatammi, line 48: I think `suppressedSystemUnits` would be more descriptive name. Then below you could just say `inherit suppressedSystemUnits`

modules/systemd/host.nix
```diff
@@ -0,0 +1,345 @@
...
+
+        # @TODO: Add system service configurations
+        # Service configuration
+        services = {
+          systemd-logind.enable = cfg.withLogin;
+          timesyncd.enable = cfg.withNetwork;
+        };
+
+        # Disable default system units
+        suppressedSystemUnits = systemUnitConfiguration;
```
mikatammi, line 339: Related to comment on line 48:
Here you could then just say `inherit suppressedSystemUnits;`

modules/systemd/host.nix
```diff
@@ -0,0 +1,345 @@
...
+  cfgDebug = config.ghaf.profiles.debug.enable;
+  cfgGui = config.ghaf.graphics.weston.enable;
+
+  # Override minimal with host configuration set
+  packageConfiguration = pkgs.systemdMinimal.override {
+    pname = "host-systemd";
+    withAcl = true;
+    withAnalyze = cfg.withAnalyze;
+    withAudit = cfg.withAudit;
+    withCompression = cfgDebug || cfg.withVirtualization;
```
mikatammi, line 20: Why disable compression of journal file blocks at all? Maybe we should also think what do we do with systemd-journalling on release-build?

mikatammi (commented):
I will report more things as I go forward with the testing of the functionalities.

I want to say generally this patch looks very good and promising work. There's just still work to do before it is merge-ready

mikatammi (changes requested):
Few things I noticed while testing this with AGX Orin. I'm currently working remotely so I cannot verify does graphical stuff (Weston) work, and I'm using AGX Orin because it's the easiest platform to develop with when working remotely.

On AGX Orin: Even though I enable `ghaf.systemd.host.withSerial = true;` , I don't get the login prompt on serial port.



Then my system logs get filled with this kind of message:
```
Aug 03 18:43:32 ghaf-host systemd[1]: systemd-timesyncd.service: Got notification message from PID 890 (WATCHDOG=1)
Aug 03 18:43:42 ghaf-host systemd[1]: systemd-oomd.service: Got notification message from PID 887 (WATCHDOG=1)
Aug 03 18:43:46 ghaf-host systemd[1]: systemd-networkd.service: Got notification message from PID 1120 (WATCHDOG=1)
Aug 03 18:44:00 ghaf-host weston[1199]: [18:44:00.100] atomic: couldn't commit new state: Permission denied
Aug 03 18:44:00 ghaf-host weston[1199]: [18:44:00.100] repaint-flush failed: No such file or directory
Aug 03 18:44:12 ghaf-host systemd[1]: systemd-udevd.service: Got notification message from PID 815 (WATCHDOG=1)
Aug 03 18:44:12 ghaf-host systemd[1]: systemd-resolved.service: Got notification message from PID 888 (WATCHDOG=1)
Aug 03 18:44:28 ghaf-host systemd[1]: systemd-oomd.service: Got notification message from PID 887 (WATCHDOG=1)
Aug 03 18:44:28 ghaf-host systemd[1]: systemd-journald.service: Got notification message from PID 794 (WATCHDOG=1)
Aug 03 18:44:31 ghaf-host systemd[1]: systemd-networkd.service: Got notification message from PID 1120 (WATCHDOG=1)
Aug 03 18:44:42 ghaf-host systemd[1]: systemd-timesyncd.service: Got notification message from PID 890 (WATCHDOG=1)
Aug 03 18:45:00 ghaf-host weston[1199]: [18:45:00.104] atomic: couldn't commit new state: Permission denied
Aug 03 18:45:00 ghaf-host weston[1199]: [18:45:00.104] repaint-flush failed: No such file or directory
Aug 03 18:45:12 ghaf-host systemd[1]: systemd-logind.service: Got notification message from PID 1021 (WATCHDOG=1)
Aug 03 18:45:13 ghaf-host systemd[1]: systemd-oomd.service: Got notification message from PID 887 (WATCHDOG=1)
Aug 03 18:45:16 ghaf-host systemd[1]: systemd-networkd.service: Got notification message from PID 1120 (WATCHDOG=1)
```

I don't know if this is bad or not, the same happens also on generic-x86 build. IMO the serial login should be made working on debug build though.

modules/systemd/host.nix
mbssrc: Good point - was only thinking of coredump compression at the time, and also was unsure about how to deal with journaling in this configuration, as I anticipated we currently need it even in release. 
Let's discuss a strategy for that - probably we want encrypted logs on release, and auditd by default for the time being.

mbssrc:
> Few things I noticed while testing this with AGX Orin. I'm currently working remotely so I cannot verify does graphical stuff (Weston) work, and I'm using AGX Orin because it's the easiest platform to develop with when working remotely.
> 
> On AGX Orin: Even though I enable `ghaf.systemd.host.withSerial = true;` , I don't get the login prompt on serial port.
> 
> I don't know if this is bad or not, the same happens also on generic-x86 build. IMO the serial login should be made working on debug build though.

You also need the `getty.target ++`, which I disabled whenever GUI is present, because of the gaphics glitch in the vm build. Seeing that it's used, can change this to `++ (lib.optionals (cfgGui && (!cfgDebug)) [ `, so the getty would be present in all headless configurations and all GUI configurations in debug. mode

PR #223, opened 2023-07-14
Enable audio on host

nesteroff:
Currently we have multiple applications that are running on host and require audio support. This PR adds the ability to use a USB headset as an audio device.

How to test:
1. Connect a USB headset to the device and reboot it.
2. Make sure the headset has 0 index in the cards list `cat /proc/asound/cards`.
3. Open Chromium and play any video with audio.
4. Run `alsamixer` in the terminal in order to change the volume.

Tested on Nvidia Orin AGX with Sennheiser PC 8 USB Headset.


nesteroff:
Closed this one since we don't need audio on host anymore.

PR #211, opened 2023-07-03
switch to nixpkgs branch, which contains recent fixes

avnik:
This branch contain fixes for clang (not yet accepted into stable), and fixes for perl-Module-Build

PR #224, opened 2023-07-17
Backport fixes for perl cross-build

avnik:
This is another take on perl fixes, avoiding forked nixpkgs. Replace #211 
Builds with `enableDemoApplications = false` with  `nix build -L ".#nvidia-jetson-orin-agx-debug-from-x86_64-flash-script" --cores 16 `.  Fails with final image generation (as generator not fixed yet)


mikatammi (approved):
LGTM, and even tested I get into the phase where it tries to build the nixos-disk-image

overlays/cross-compilation.nix
```diff
@@ -5,6 +5,41 @@
 #
 {...}: {
   nixpkgs.overlays = [
-    (final: prev: {})
+    (final: prev: let
+      crossCompiling = prev.stdenv.buildPlatform != prev.stdenv.hostPlatform;
```
mikatammi, line 9: I think we might not necessarily need to do this since we are anyway including this overlay only on cross-compiled builds

overlays/cross-compilation.nix
```diff
@@ -5,6 +5,41 @@
...
+      filterOutByName = name: builtins.filter (x: (builtins.baseNameOf x) != name);
+      # FIXME: should be prev.buildPackages.fetchpatch2, why I can't use fetchpatch2 here
```
mikatammi, line 11: Does final.buildPackages.fetchpatch2 work? Or final.fetchpatch2?

mikatammi (commented):
Actually few questions

overlays/cross-compilation.nix
avnik: doesn't worked for me, but I believe GH yield stable commits, so we can hash it

overlays/cross-compilation.nix
avnik: Correct, but we don't need this patch applied for buildPackages (which happens when we apply patch unconditionally here), but only for target build

mikatammi:
We will need to find another solution for the `fetchPatch`, my Hydra is full of evaluation errors now, similar to:

```
in job ‘nvidia-jetson-orin-nx-debug-from-x86_64.x86_64-linux’:
error: access to URI 'https://raw.githubusercontent.com/ck3d/nixpkgs/2d6f287f403f11f48bba19e2b2f2a7050592d51a/pkgs/development/interpreters/perl/cross.patch' is forbidden in restricted mode

```

overlays/cross-compilation.nix
mikatammi: I don't know if I will soon hit some other problem, but using both final.buildPackages.fetchpatch2 and prev.buildPackages.fetchpatch2 at least evaluates for me, and starts building. Maybe there's error to come back to

PR #225, opened 2023-07-17
TMP reduce image size

brianmcgillion:


brianmcgillion:
https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/config/system-path.nix defaultPackages set here

brianmcgillion:
Out of date

PR #234, opened 2023-07-27
Hart Software Services for Microchip Polarfire Icicle Kit (RISC-V)

mikatammi:
Build Hart Software Services (HSS) to be able to ~~run RISC-V images on QEMU~~ build own HSS binary from source and run it on Icicle-Kit board

### DIFF
--- a/packages/flake-module.nix
+++ b/packages/flake-module.nix
@@ -1,6 +1,8 @@
 # Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
 {self, ...}: {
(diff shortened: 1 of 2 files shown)


mikatammi:
This is work in progress. The Hart Software Services build, but I was unable to boot QEMU with the resulting binaries with instructions from here: https://www.qemu.org/docs/master/system/riscv/microchip-icicle-kit.html#running-hss

I'm wondering what image file to use, 2 types of image files are built, and both .elf and .bin versions of those

mikatammi:
Ready for review now. Checked that it builds

brianmcgillion (approved):

PR #235, opened 2023-07-27
Use sdcard-module for building NVIDIA Jetson Orin images

mikatammi:
Use sdcard-module for building NVIDIA Jetson Orin images instead of make-disk-image and then extracting filesystems from that

This enables cross-compiled builds of disk images AND flashing scripts.

Also packages for nodemoapps-version of the builds are added to NVIDIA targets. Those are also added to hydrajobs.


mikatammi:
This should enable the flash-script generation through cross compilation, when ready.

@avnik had an idea to use the new bootspec feature of NixOS 23.05, which sounds like a good idea.
https://github.com/NixOS/rfcs/blob/master/rfcs/0125-bootspec.md

mikatammi:
This is now finally ready and tested. Let's merge this first and then @avnik could rebase his PR #248 on top of the change.

brianmcgillion (approved):

PR #243, opened 2023-08-21
Compartmentalize GPU and graphic apps into VMs

unbel13ver:
!!! Please read this first !!!
!!! Do not merge this PR, even in case there are enough approvals. These changes are part of another work, and I need to align both parts before merging this into mainline. Also there is no documentation yet about how to use GPU and APP vms. The documentation is in process right now. !!!

This commit introduces several new entities such as GPU VM and Application VMs. To use those, you need to properly set up devices passthroug first. Refer to the project documentation on how to achieve that.

**Lenovo X1 Carbon Gen11 is introduced as a new target and the
reference device for implementing compartmentalization.**

It is still possible to build standard "mainline" ghaf with weston and demo apps working on host.

* User apps are now added into the scope of nixpkgs;
* minimal.nix module is disabled since it introduced some weird metadata changes.


brianmcgillion:
great start. Do we have jira's created to remind us that e.g. the key managment for the ssh would need to reviewed once the TEE/TPM is in use?

unbel13ver:
> key managment for the ssh would need to reviewed

We get rid of SSH in Wayland proxying very soon, no need to manage these keys in future.

unbel13ver:
The complementary part of these changes: https://github.com/unbel13ver/ghaf-lib
Just to catch the whole idea.

modules/virtualization/microvm/appvm.nix
```diff
@@ -0,0 +1,176 @@
...
+            networks."10-ethint0" = {
+              matchConfig.MACAddress = vm.macAddress;
+              addresses = [
+                {
+                  # IP-address for debugging subnet
+                  addressConfig.Address = vm.ipAddress;
+                }
+              ];
+              routes = [
+                {routeConfig.Gateway = "192.168.101.1";}
```
vilvo, line 91: What if we'd declare the `netvm` gateway address, and the whole internal network, following this idea https://github.com/unbel13ver/ghaf-lib/blob/main/my-hardware/lenovo-t14.nix 
It would make declaring the subnetwork and VM-specific IP addresses clearer and centralized instead of distributed across the ghaf VMs declarations.

modules/virtualization/microvm/appvm.nix
```diff
@@ -0,0 +1,176 @@
...
+            packages = mkOption {
+              description = ''
+                Packages that are included into the AppVM
+              '';
+              type = types.listOf package;
+              default = [];
+            };
+            ipAddress = mkOption {
+              description = ''
+                AppVM's IP address in the inter-vm network
```
vilvo, line 132: This addresses the VMs part of my earlier comment https://github.com/tiiuae/ghaf/pull/243/files#r1302862794
Ghaf inter-vm network subnet/mask comment still applies - for example when one would like to change the subnet from `192.168.101.0` to say `10.0.0.0` (or something else)

modules/virtualization/microvm/guivm.nix
```diff
@@ -0,0 +1,128 @@
...
+
+        microvm = {
+          mem = 2048;
+          hypervisor = "qemu";
+          qemu.bios.enable = false;
+          storeDiskType = "squashfs";
+          interfaces = [
+            {
+              type = "tap";
+              id = "vm-guivm";
```
vilvo, line 62: This is minor but should we just DRY this to `id =  "vm-${networking.hostname}"` with this declaration https://github.com/tiiuae/ghaf/pull/243/files#diff-48f3927dbfa0f525b6dd5c7bc57cc323d0b54bdbd60cf71264f84e0714cdd5f5R40

modules/virtualization/microvm/guivm.nix
```diff
@@ -0,0 +1,128 @@
...
+          networks."10-ethint0" = {
+            matchConfig.MACAddress = "02:00:00:02:02:02";
+            addresses = [
+              {
+                # IP-address for debugging subnet
+                addressConfig.Address = "192.168.101.3/24";
+              }
+            ];
+            routes = [
+              {routeConfig.Gateway = "192.168.101.1";}
```
vilvo, line 90: Earlier comment on subnet applies here - DRY with `Gateway` and when we want to change the VM addresses, one place for network declarations could help. For new, out of ghaf (ghaf as lib), VM declarations, users could still use the platform subnet + gateway but still set their VM static address in their own VMs.

targets/lenovo-x1-carbon.nix
```diff
@@ -0,0 +1,181 @@
...
+    };
+  in {
+    inherit hostConfiguration;
+    name = "${name}-${variant}";
+    package = hostConfiguration.config.system.build.${hostConfiguration.config.formatAttr};
+  };
+  debugModules = [../modules/development/usb-serial.nix {ghaf.development.usb-serial.enable = true;}];
+  targets = [
+    (generic-x86 "debug" debugModules)
+    (generic-x86 "release" [])
```
brianmcgillion, line 172: is it still generic-x86 or should it be x1?

targets/lenovo-x1-carbon.nix
```diff
@@ -0,0 +1,181 @@
...
+                    cores = 2;
+                  }
+                  {
+                    name = "zathura";
+                    packages = [pkgs.zathura];
+                    ipAddress = "192.168.101.7/24";
+                    macAddress = "02:00:00:03:03:07";
+                    ramMb = 512;
+                    cores = 1;
+                  }
```
brianmcgillion, line 127: can be fixed in a later patch, but should these definitions be in modules/vms/ or something similar, so that they could be more easily shared ?

targets/lenovo-x1-carbon.nix
```diff
@@ -0,0 +1,181 @@
...
+            path = "${pkgs.waypipe}/bin/waypipe ssh -i ${pkgs.waypipe-ssh}/keys/waypipe-ssh -o StrictHostKeyChecking=no 192.168.101.6 gala --enable-features=UseOzonePlatform --ozone-platform=wayland";
+            icon = "${pkgs.weston}/share/weston/icon_editor.png";
+          }
+
+          {
+            path = "${pkgs.waypipe}/bin/waypipe ssh -i ${pkgs.waypipe-ssh}/keys/waypipe-ssh -o StrictHostKeyChecking=no 192.168.101.7 zathura";
+            icon = "${pkgs.weston}/share/weston/icon_editor.png";
+          }
+        ];
+      })
```
brianmcgillion, line 66: can these live in the same definition as the modules/vms/... as discussed below? Can be a follow on patch

modules/graphics/weston.nix
```diff
@@ -64,14 +64,20 @@ in {
         StandardOutput = "journal";
         StandardError = "journal";
         ExecStart = "${pkgs.weston}/bin/weston";
+        #GPU pt needs some time to start - weston fails to restart 3 times in avg.
+        ExecStartPre = "${pkgs.coreutils}/bin/sleep 3";
```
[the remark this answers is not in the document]
unbel13ver: SP-2838 is created to track this

targets/lenovo-x1-carbon.nix
unbel13ver: SP-2840 is created to track this

targets/lenovo-x1-carbon.nix
unbel13ver: SP-2840 is created to track this

targets/lenovo-x1-carbon.nix
unbel13ver: Fixed

modules/virtualization/microvm/guivm.nix
unbel13ver: SP-2839 is created to track this

modules/virtualization/microvm/guivm.nix
unbel13ver: Fixed

modules/virtualization/microvm/appvm.nix
unbel13ver: SP-2839 is created to track this

docs/src/ref_impl/creating_appvm.md
```diff
@@ -0,0 +1,90 @@
...
+    SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Creating an Application VM
+
+## What is AppVM?
+
+AppVM is a virtual machine, which isolates an application from a host OS. An application seems like it is running inside an ordinary OS, yet it runs in a completely different system. An application inside renders, shows and controls from a host OS via GUI VM, network access is provided by NetVM.
+
+In the perfect scenario, there is a separate AppVM for every application. But it is possible to run multiple applications inside one AppVM.
```
vilvo, line 12: We could add here also the rationale for the AppVM - something like 

```
## Why AppVM?

AppVMs are used to improve trust in system components by isolating the apps from both host OS and other applications. This way user can use applications of different trust levels within the same system and without compromising system security. This is because virtualization with hardware backed mechanisms provides better resource protection than traditional OS. While the VMs have overhead, it's acceptable via improved security and usability that makes the application seem like it is running inside an ordinary OS.

As a result - both highly trusted applications and untrusted applications can be hosted in the same secure system when the concerns are separated in their own AppVMs.
```

docs/src/ref_impl/creating_appvm.md
```diff
@@ -0,0 +1,90 @@
...
+  }
+];
+```
+
+Each VM has the following properties:
+
+
+| **Property** | **Type**                  | **Unique** | **Description**                                                                                               | **Example**         |
+| -------------- | --------------------------- | ------------ | --------------------------------------------------------------------------------------------------------------- | --------------------- |
+| name         | str                       | yes        | This name will be shown in microvm list, a VM will have the same hostname                                     | “chromium”        |
```
vilvo, line 58: Should be (automatically) postfix the `name` with `-vm`? That would make it clearer that the application name is a virtual machine of its' own.

docs/src/ref_impl/creating_appvm.md
```diff
@@ -0,0 +1,90 @@
...
+| ramMb        | int, [1, …, host memory] | no         | Memory in MB                                                                                                  | 3072                |
+| cores        | int,  [1, …, host cores] | no         | Virtual CPU cores                                                                                             | 4                   |
+
+
+### 2. Add an app launcher in GUI VM
+
+To add an app launcher, add an element in the [guivm.nix](../../../modules/virtualization/microvm/guivm.nix) file to the **graphics.weston.launchers** list.
+A launcher element has 2 properties:
+
+1. **path** – at the moment, we launch an app via **waypipe** . In general, the path string will be as follows: "\${pkgs.waypipe}/bin/waypipe ssh -i ${waypipe-ssh}/keys/waypipe-ssh -o StrictHostKeyChecking=no **vm.ipAddress** **vm.executable**", where:
```
vilvo, line 71: I would leave the detailed path string example out because it's an implementation detail and we've already discussing (and implementing) leaving the ssh tunneling out with waypipe due to its' unnecessary overhead in inter VM comms.
The problem is that when it's left out, the example will deprecate and will require that someone remembers to update it with the code changes.

docs/src/ref_impl/creating_appvm.md
```diff
@@ -0,0 +1,90 @@
...
+   **vm.ipAddress** – ipAddress you have defined in the target’s AppVM configuration,
+   **vm.executable** – an executable you want to execute, like a graphical application.
+2. **icon** – path to an icon to show.
+
+
+#### Example of the current launchers
+
+```
```
vilvo, line 79: Could we just link to the implementations of launchers instead?

docs/src/ref_impl/example_project.md
```diff
@@ -0,0 +1,36 @@
...
+    Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+    SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Example Project
+
+The compartmentalization could be applied to many specific x86_64 computers and laptops with some customization applied to the Ghaf. The best way of the Ghaf customization is using Ghaf templates.
+
+1. Create a template project as described in [Ghaf as Library](../ref_impl/ghaf-based-project.md) section
+2. Adjust your system configuration with accordance to your HW specification. Determine all VIDs and PIDs of the devices that are passed to the VMs
```
vilvo, line 11: s/VIDs/vendor IDs (VIDs)/
s/PIDs/product IDs (PIDs)/

docs/src/technologies/compartment.md
```diff
@@ -0,0 +1,13 @@
+<!--
+    Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+    SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Compartmentalization
+Compartmentalization is the technique of separating parts of a system to decrease attack surface and prevent malfunctions from cascading in the system. In Ghaf architecture, there is a separate Virtual Machine (VM) for every vital function of the system.
+
+Current implementation supports Graphic User Interface (GUI) VM, Networking VM and a couple of Application VMs, such as Chromium web-browser, GALA-app and Zathura pdf reader.
```
vilvo, line 9: There was discussion to leave GALA-app out as a Ghaf reference app and move it to a ghaf-based template project.
There's good reason for it from Ghaf-perspective - non-source delivery, requires back-end infra with access rights to  Cuttlefish virtual devices. 

At this point, I would not mention it here.

docs/src/ref_impl/creating_appvm.md
vilvo: It's actually prefixed here - https://github.com/tiiuae/ghaf/pull/243/files#diff-b7c4d8b1828ec4e55c3b85b348f4c180560550100c49b79a826d4cacfd1afaedR14
Maybe mention in the `Description` field of `name` that `This name is prefixed with vm- and will be shown in microvm list. The prefixed name - e.g. `vm-chromium` will be also the VM hostname.

modules/virtualization/microvm/appvm.nix
```diff
@@ -0,0 +1,176 @@
...
+    hostname = "vm-" + vm.name;
+    appvmConfiguration = {
+      imports = [
+        ({
+          lib,
+          config,
+          ...
+        }: {
+          ghaf = {
+            users.accounts.enable = lib.mkDefault configHost.ghaf.users.accounts.enable;
```
vilvo, line 23: Should we actually not share the `configHost.ghaf.users.accounts` enabling with the appvms but have enabling appVMs accounts separately from host? 

I know for `-debug` targets it makes development easier but the problem is that `-release` target should not allow interactive login - like `getty` - to host at all. In that scenario, the host systemd services are run with unprivileged user (similar to `ghaf`) but we likely do not want the share the accounts declarations between the host and the appVMS by default. In some special, unforseen scenario(s) sharing the host and appVMs user account(s) might be needed but that should be exception.

modules/virtualization/microvm/guivm.nix
```diff
@@ -0,0 +1,128 @@
...
+  waypipe-ssh = pkgs.callPackage ../../../user-apps/waypipe-ssh {};
+  guivmBaseConfiguration = {
+    imports = [
+      ({
+        lib,
+        pkgs,
+        ...
+      }: {
+        ghaf = {
+          users.accounts.enable = lib.mkDefault configHost.ghaf.users.accounts.enable;
```
vilvo, line 19: Referring to this comment - https://github.com/tiiuae/ghaf/pull/243/files#r1308262741
GUIVM should share the user account with the AppVMs but should NOT share the user account with host.
I would prefer to have this user account separation either:
- part of this refactoring PR or
- ticket to track that it'll get implemented

modules/virtualization/microvm/appvm.nix
```diff
@@ -0,0 +1,176 @@
...
+  lib,
+  pkgs,
+  ...
+}: let
+  configHost = config;
+  cfg = config.ghaf.virtualization.microvm.appvm;
+  waypipe-ssh = pkgs.callPackage ../../../user-apps/waypipe-ssh {};
+
+  makeVm = {vm}: let
+    hostname = "vm-" + vm.name;
```
vilvo, line 14: For consistency with system VMs (`netvm`, `guivm`, `adminvm`) we could also consider postfixing (`-vm`) appVMs instead of prefixing (`vm-`).

modules/virtualization/microvm/netvm.nix
```diff
@@ -38,7 +38,7 @@
         microvm.interfaces = [
           {
             type = "tap";
-            id = "vm-netvm";
+            id = "vm-${config.networking.hostName}";
```
vilvo, line 41: My 2 cents for aligning the `netvm` filename and documentation name with the system declared `id` - so `netvm` instead of this line.

vilvo (changes requested):
Mainly minor documentation related but please choose how the issue on sharing the user account between host and VMs should be addressed in respect to this PR. I'm fine with follow-up work as well.

modules/virtualization/microvm/appvm.nix
baz2142: Either way, let's keep '-' sign somewhere. For my taste, 'someappnamevm' or 'vmsomeappname' looks not so obvious, than 'someappname-vm' or 'vm-someappname'

docs/src/ref_impl/creating_appvm.md
baz2142: I agree, that it isn't the perfect solution with problems like keeping it up to date.

Still, at least a reader knows, how to add an application link here. Without wondering, why it is not shown in an application list -- which it the primary goal for this article.

docs/src/ref_impl/creating_appvm.md
unbel13ver: Added

docs/src/ref_impl/creating_appvm.md
unbel13ver: Added

docs/src/ref_impl/creating_appvm.md
unbel13ver: Done

docs/src/ref_impl/example_project.md
unbel13ver: Discussed this with @jenninikko  - worth adding those to the Glossary section.

docs/src/technologies/compartment.md
unbel13ver: Removed the mention about GALA

modules/virtualization/microvm/guivm.nix
unbel13ver: SP-2869 is tracking this

modules/virtualization/microvm/appvm.nix
unbel13ver: Will be a part of SP-2839 task

modules/virtualization/microvm/netvm.nix
unbel13ver: I thought a bit about this and decided this should be also out of scope for this PR since it touches other files, and the logic of that work does not correspond with the aim of this PR.
Moreover, changing `netvm` id somehow breaks the netvm.
I'll track this issue as the part of SP-2839 ticket.

vilvo (approved):

vadika:
Lets merge it :)

brianmcgillion (approved):

jenninikko:
@brianmcgillion I was just in the middle of reviewing the documentation T_T

PR #249, opened 2023-08-29
Add simple installer

remimimimimi:
Adds simple CLI based installer that runs at system startup (installer image) and flushes image with which we configured it to the target device. Also allows anyone to build an installer for their configuration if it builds raw-efi image.


docs/src/ref_impl/build_and_run.md
```diff
@@ -188,3 +188,75 @@ In the case of the Icicle Kit, Ghaf deployment consists of creating an SD image
...
+      fi
+    '';
+  };
+}
+```
+
+To bulid installer image write:
+
+```
+nix build .#nixosConfiguration.<CONFIGURATION>.config.system.build.installer
```
vilvo, line 250: `s/nixosConfiguration/nixosConfigurations` - for anyone copy-pasting the command to build the installer they have to change both `<CONFIGURATION>` and fix the typo

vilvo (approved):
In general - looks good to me as the first iteration of installer. Building this now for testing.

vilvo:
Installer does not work for generic target.
IMG_2223

vilvo (changes requested):
Please fix installer for generic.

modules/installer/default.nix
```diff
@@ -0,0 +1,104 @@
...
+
+        All code must be written for the current pkgs.runtimeShell.
+      '';
+      type = lib.types.lines;
+      default = "";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    ghaf.installer.installerModules.flushImage = {
```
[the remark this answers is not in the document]
remimimimimi: `flush` module mainly were written as an example. And should be deleted in further versions of Ghaf when we'll have more complex installation process.

vilvo (approved):
Verified to work on X1 with `.#nixosConfigurations.generic-x86_64-debug.config.system.build.installer`. 

The basic structure looks good baseline for further work.
@mikatammi @brianmcgillion - please consired reviewing / merging this soon so further work can be based on it.

vilvo:
IMG_2225

jenninikko (changes requested):
A couple of comments:
-- Make a separate subsection 5.1.2.
-- Change the main title + change Built-in Installer Modules
-- Make a list of actions instead of plain text.
-- Add reference to https://tiiuae.github.io/ghaf/ref_impl/modules_options.html
-- nix -> Nix

jenninikko:
> IMG_2225

@remimimimimi, do not forget to change ghaf -> Ghaf ( ` ω ´ )

vilvo (changes requested):
After retesting this together with @remimimimimi - this is not working correctly. The image written to the internal memory is not booting correctly to the desktop.

The problem is that we only get very few early boot logs to the display - then just black display. No IP via DHCP like with the target image booted from SSD.

All I can think now for debug is to print the partitions of `nvme0n1` after `sync` and sleep 20s - here https://github.com/tiiuae/ghaf/pull/249/files#diff-789892fdf4e4c650d81ce42dbc84586807ced5e73a8d53c06775acf7df2d0876R172
Then add a message like: `echo "Rebooting automatically in 20 seconds.`

If that does not help, we need to check if we can get early kernel logs either to display or to USB debugging - https://docs.kernel.org/driver-api/usb/usb3-debug-port.html

vilvo:
@remimimimimi - this fixes the issue on X1
```
diff --git a/targets/lenovo-x1-carbon.nix b/targets/lenovo-x1-carbon.nix
index 6bf7bcf..6f00e8f 100644
--- a/targets/lenovo-x1-carbon.nix
+++ b/targets/lenovo-x1-carbon.nix
@@ -194,6 +194,7 @@
               # Passthrough Intel Iris GPU 8086:a7a1
               "vfio-pci.ids=8086:51f1,8086:a7a1"
             ];
+            boot.initrd.availableKernelModules = [ "nvme" ];
           }
         ]
         ++ (import ../modules/module-list.nix)
```
kudos @mikatammi for thinking about loud about possibly missing nvme kernel module - I remember I had to add `nvme` to initrd way back on my Lenovo P43s so it was easy fix to test.

vilvo:
Ghaf on X1 Carbon gen3 - booted from nvme with no SSD attached.
IMG_2256

vilvo (approved):

docs/src/ref_impl/installer.md
```diff
@@ -0,0 +1,73 @@
+<!--
+    Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+    SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Installer
+
+## Configuring And Building Installer For Ghaf
```
jenninikko, line 8: Configuring and Building Installer for Ghaf

docs/src/ref_impl/installer.md
```diff
@@ -0,0 +1,73 @@
...
+
+You can obtain the installation image for your Ghaf configuration using this guide. To check possible configuration options see (ghaf documentation for modules options)[https://tiiuae.github.io/ghaf/ref_impl/modules_options.html#ghafinstaller].
```
jenninikko, line 10: Replace with: (pay attention to the link)

`You can obtain the installation image for your Ghaf configuration. To check possible configuration options, see [Modules Options](../ref_impl/modules_options.md).`

docs/src/ref_impl/installer.md
```diff
@@ -0,0 +1,73 @@
...
+      else
+          echo "Some error occured during flushing process, exit code: $?."
+          exit
+      fi
+    '';
+  };
+}
+```
+
+After that you can build an installer image using this command:
```
jenninikko, line 42: Replace with:

Then you can build an installer image using the command:

docs/src/ref_impl/installer.md
```diff
@@ -0,0 +1,73 @@
...
+
+```sh
+nix build .#nixosConfigurations.<CONFIGURATION>.config.system.build.installer
+```
+
+## Adding Installer Modules
+
+To add installer module replace the corresponding placeholders with your code and add this to your configuraiton:
```
jenninikko, line 50: To add an installer module, replace...

docs/src/ref_impl/installer.md
```diff
@@ -0,0 +1,73 @@
...
+  providedVariables = {
+    # Notice the dollar sign before the actual variable name in Bash.
+    <VARIABLE_NAME> = "$<VARIABLE_NAME>";
+  };
+};
+```
+
+## Built-in Installer Modules
+
+Provided variables show variable names in nix, for actual names of variables in bash see sources of the module.
```
jenninikko, line 66: Provided variables show variable names in Nix. For actual names of variables in Bash, see the sources of the module.

jenninikko (changes requested):
Nice job! Let's play a little bit more with letters :D

jenninikko (approved):

docs/src/ref_impl/installer.md
```diff
@@ -0,0 +1,73 @@
...
+## Configuring and Building Installer for Ghaf
+
+You can obtain the installation image for your Ghaf configuration. To check possible configuration options, see [Modules Options](../ref_impl/modules_options.md#ghafinstallerenable).
+
+1. Set `ghaf.installer.enable` to `true`.
+2. Add nixos-generators module to `ghaf.installer.imgModules` list to configure installer image type.
+3. Choose installer modules from `ghaf.installer.installerModules` and set `ghaf.installer.enabledModules` to list of their names.
+4. Write code for the installer in `ghaf.installer.installerCode`.
+
+Then you can build an installer image using the command:
```
mikatammi, line 17: Is this extra?

vilvo: Good catch. Should be above line 44, not here.

mikatammi:
I tested the installer and there are some rough edges, for example I was staring away from laptop while it was installing, then it rebooted, and I wasn't fast enough to disconnect the USB-drive. So the system somehow strangely booted up into ghaf desktop but still something was mounted from USB-drive.

But I'm fine with merging this now and improving things later.

mikatammi:
Note for the future: Also the resulting image was 26 gigabytes, so adding (zstd) compression even to the image embedded, would save many gigabytes

mikatammi (approved):

PR #256, opened 2023-09-07
Setup vsock devices for APPVMs and use it for Waypipe

nesteroff:
This PR makes Waypipe work over VSOCK instead of UNIX domain sockets. It increases performance by getting rid of the unnecessary overhead caused by forwarding UNIX sockets over SSH. Testing with iperf show that UNIX sockets over SSH can transfer about 4 Gbits/sec on Lenovo X1 Carbon but with VSOCK it can handle about 15 Gbits/sec between GUIVM and APPVMs.

### DIFF
--- a/modules/virtualization/microvm/appvm.nix
+++ b/modules/virtualization/microvm/appvm.nix
@@ -10,8 +10,15 @@
   cfg = config.ghaf.virtualization.microvm.appvm;
   waypipe-ssh = pkgs.callPackage ../../../user-apps/waypipe-ssh {};
 
-  makeVm = {vm}: let
+  makeVm = {
+    vm,
+    index,
+  }: let
     hostname = "vm-" + vm.name;
+    cid =
+      if vm.cid > 0
+      then vm.cid
+      else cfg.vsockBaseCID + index;
     appvmConfiguration = {
       imports = [
         ({
@@ -72,6 +79,12 @@
                 id = hostname;
                 mac = vm.macAddress;
               }
+            ];
+            qemu.extraArgs = [
+              "-M"
+              "q35,accel=kvm:tcg,mem-merge=on,sata=off"
+              "-device"
+              "vhost-vsock-pci,guest-cid=${toString cid}"
             ];
           };
 
@@ -167,6 +180,14 @@ in {
               '';
               default = [];
             };
+            cid = mkOption {
+              description = ''
+                VSOCK context identifier (CID) for the AppVM
+                Default value 0 means auto-assign using vsockBaseCID and AppVM index
+              '';
+              type = int;
+              default = 0;
+            };
           };
         });
         default = [];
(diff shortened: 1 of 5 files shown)


targets/lenovo-x1-carbon.nix
```diff
@@ -120,6 +120,7 @@
                     macAddress = "02:00:00:03:03:05";
                     ramMb = 3072;
                     cores = 4;
+                    cid = 4;
```
brianmcgillion, line 123: is there anyway to script this as e.g. ++cid so as to avoid hard coding for each VM?

nesteroff: Sure, I'll check how to do it.

vadika:
It's basically bad idea cause we need to know this addresses exactly for each VM to be able to address vsock endpoints. —vadik ***@***.*** 8. Sep 2023, at 13.27, Brian McGillion ***@***.***> wrote:﻿
@brianmcgillion commented on this pull request.



In targets/lenovo-x1-carbon.nix:
> @@ -120,6 +120,7 @@
                     macAddress = "02:00:00:03:03:05";
                     ramMb = 3072;
                     cores = 4;
+                    cid = 4;

is there anyway to script this as e.g. ++cid so as to avoid hard coding for each VM?

nesteroff:
In this particular case APPVMs connect to GUIVM but not the other way around. It means that they only need to know the CID of the GUIVM which is always 3 (first available number).

targets/lenovo-x1-carbon.nix
nesteroff: Added auto-assigning but at the same time kept the ability to override with a static value.

user-apps/vsockproxy/default.nix
```diff
@@ -0,0 +1,33 @@
...
+  lib,
+  ...
+}:
+stdenv.mkDerivation {
+  name = "vsockproxy";
+
+  nativeBuildInputs = with pkgs; [meson ninja];
+
+  src = pkgs.fetchFromGitHub {
+    owner = "nesteroff";
```
mikatammi, line 15: I think we should create a repo inside TIIUAE for this

mikatammi (changes requested):
Please create repo under `tiiuae` for `vsockproxy`. Otherwise LGTM

user-apps/vsockproxy/default.nix
nesteroff: Ok, now it's in tiiuae.

mikatammi:
Needs rebase now

modules/virtualization/microvm/appvm.nix
```diff
@@ -64,11 +71,21 @@
                 mac = vm.macAddress;
               }
             ];
-            # Use qboot BIOS on x86_64-linux as workaround
-            qemu.extraArgs = lib.optionals (config.nixpkgs.hostPlatform.system == "x86_64-linux") [
-              "-bios"
-              "${pkgs.qboot}/bios.bin"
-            ];
+            qemu.extraArgs =
```
mikatammi, line 74: No need for this stuff anymore, as the machine type has now changed back to q35

mikatammi (approved):
LGTM. I will test this first

PR #260, opened 2023-09-11
Update microvm in flake.lock

mikatammi:
The behavior onwards from now, is the same as if the microvm.qemu.bios.enable is always set to false.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -48,11 +48,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1690637422,
-        "narHash": "sha256-hSEkpMhbQUcYLkKwOM5NiaSQ78NdPkosqcYc1TyMog8=",
+        "lastModified": 1694526290,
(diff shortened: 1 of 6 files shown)


mikatammi:
The AppVMs don't start after this change :(

mikatammi:
Okay, now works on the NUC at least. Please review again!

mikatammi:
I fixed the issue by not updating to latest, but latest commit that works which is https://github.com/astro/microvm.nix/commit/c00af667690ac551764da0268c994e46bead24cd

mikatammi:
The problematic commit is this: https://github.com/astro/microvm.nix/commit/c422e7207a0225f949dc975acb5b7902ac342578

The `romfile=`gives errors, it doesn't work even on the `nixos-unstable`

mikatammi:
There is follow up here:
https://github.com/tiiuae/ghaf/pull/261

brianmcgillion (approved):

brianmcgillion:
verified

PR #263, opened 2023-09-13
Update Lenovo X1 udev rules to use evdev input devices

nesteroff:
This is another fix for the Lenovo X1 touchpad passthrough.

I noticed that sometimes Lenovo X1 boots to black screen and the GUIVM fails to start. It turned out that the touchpad has two devices under the same name: PS/2-style mousedev (/dev/input/mouse*) and evdev (/dev/input/event*). When the legacy mousedev device comes first the GUIVM fails to start with the following error: "/dev/mouse: is not an evdev device".

In this patch I added an additional filter to the udev rules to make sure we always use event devices for the passthrough.


nesteroff:
Added udev rules for the touchpad from the UAE revision.

brianmcgillion (approved):

PR #264, opened 2023-09-13
Add integrated Lenovo X1 webcam for Chromium AppVM

josa41:
Separate extra modules config for appVMS
Lenovo X1 Integrated webcam support for chromium appVM as reference

### DIFF
--- a/modules/virtualization/microvm/appvm.nix
+++ b/modules/virtualization/microvm/appvm.nix
@@ -106,7 +106,7 @@
     };
   in {
     autostart = true;
-    config = appvmConfiguration // {imports = appvmConfiguration.imports ++ cfg.extraModules ++ [{environment.systemPackages = vm.packages;}];};
+    config = appvmConfiguration // {imports = appvmConfiguration.imports ++ cfg.extraModules ++ vm.extraModules ++ [{environment.systemPackages = vm.packages;}];};
     specialArgs = {inherit lib;};
   };
 in {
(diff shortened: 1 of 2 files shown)


modules/virtualization/microvm/appvm.nix
```diff
@@ -156,6 +156,13 @@ in {
               '';
               type = int;
             };
+            appvmExtraModules = mkOption {
```
mikatammi, line 159: I think we could simply call this extraModules, and maybe we could also remove the common extraModules which configures settings for all AppVMs

mikatammi: I'm not sure why the extraModules were left as the way is it is in line 164, but it is probably remnant of copy-pasting code from other VMs in the past

josa41: I have no idea what the common extraModules does. It was added/taken into use after my original pull request of this feature. I had to redo and retest this patch because of that change.
I cannot remove the common extraModules as i don't know what it is supposed to do and how to test its possible effects.

mikatammi: Could you please change the variable name from `appvmExtraModules`to `extraModules`. I can then write another patch, which removes the common extraModules.

mikatammi (changes requested):

modules/virtualization/microvm/appvm.nix
josa41: Done.

PR #265, opened 2023-09-13
Add the whitepaper for zathura

brianmcgillion:
Add a pdf to showcase zathura in work. using the whitepaper from the demo image

### DIFF
--- a/targets/lenovo-x1-carbon.nix
+++ b/targets/lenovo-x1-carbon.nix
@@ -67,7 +67,9 @@
           }
         ];
       }
-      ({pkgs, ...}: {
+      ({pkgs, ...}: let
+        demo-pdf = pkgs.callPackage ../user-apps/demo-pdf {};
+      in {
         ghaf.graphics.weston.launchers = [
           {
             path = "${pkgs.openssh}/bin/ssh -i ${pkgs.waypipe-ssh}/keys/waypipe-ssh -o StrictHostKeyChecking=no chromium-vm.ghaf ${pkgs.waypipe}/bin/waypipe --border \"#ff5733,5\" --vsock -s ${toString guivmConfig.waypipePort} server chromium --enable-features=UseOzonePlatform --ozone-platform=wayland";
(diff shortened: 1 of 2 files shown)


user-apps/demo-pdf/default.nix
```diff
@@ -0,0 +1,28 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  stdenv,
+  pkgs,
+  lib,
+  ...
```
mikatammi, line 7: You are not using `pkgs` and `lib`, and you don't need `...`.

However, you could add `zathura` here instead of the above

user-apps/demo-pdf/default.nix
```diff
@@ -0,0 +1,28 @@
...
+stdenv.mkDerivation {
+  name = "demo-pdf";
+
+  src = ./Whitepaper.pdf;
+  phases = ["installPhase"];
+
+  installPhase = ''
+    mkdir -p $out/bin
+    cp ${./Whitepaper.pdf} $out/Whitepaper.pdf
+    echo "zathura $out/Whitepaper.pdf" > $out/bin/run-zathura
```
mikatammi, line 18: Then if you add `zathura` to the "callPackage-style function call attribute set" above, you can say `${zathura}/bin/zathura` here, instead of zathura.

user-apps/demo-pdf/default.nix
```diff
@@ -0,0 +1,28 @@
...
+}:
+stdenv.mkDerivation {
+  name = "demo-pdf";
+
+  src = ./Whitepaper.pdf;
+  phases = ["installPhase"];
+
+  installPhase = ''
+    mkdir -p $out/bin
+    cp ${./Whitepaper.pdf} $out/Whitepaper.pdf
```
mikatammi, line 17: As you are doing full mkDerivation, you could say `cp $src $out/Whitepaper.pdf` instead

targets/lenovo-x1-carbon.nix
```diff
@@ -134,7 +136,7 @@
                   }
                   {
                     name = "zathura";
-                    packages = [pkgs.zathura];
+                    packages = [pkgs.zathura (pkgs.callPackage ../user-apps/demo-pdf {})];
```
mikatammi, line 139: You could also just put `demo-pdf` here instead of the pkgs.callPackage.

Also Note that you are using pkgs from the host and not from the appvm guest.

targets/lenovo-x1-carbon.nix
```diff
@@ -28,7 +28,7 @@
         networking.wireless = {
           enable = true;
 
-          #networks."ssid".psk = "psk";
+          #networks."ssid".psk = "passwd";
```
mikatammi, line 31: Why change this?

brianmcgillion:
Just rebased for now

brianmcgillion:
dump outdated patch

PR #266, opened 2023-09-14
Docker fix

emrahbillur:
This is added to resolve the issue https://ssrc.atlassian.net/browse/SP-2697 where originated from https://github.com/NixOS/nixpkgs/issues/244159. We can describe this bug  as while running docker containers got a `http: invalid Host header`
response failing to run.
This is due to a bug in Go package in NixOS where the previous package solves this issue. Same solution can be applied Go compiler related net access bugs till next NixOS release.
When moving to NixOS 23.11, this fix will be removed as it is already fixed in the unstable channel.


emrahbillur:
Just need to resolve a conflict before creating the pull request .

PR #267, opened 2023-09-14
Docker bug fix

emrahbillur:
This is added to resolve the issue https://ssrc.atlassian.net/browse/SP-2697 where originated from https://github.com/NixOS/nixpkgs/issues/244159. We can describe this bug as while running docker containers got a http: invalid Host header
response failing to run.
This is due to a bug in Go package in NixOS where the previous package solves this issue. Same solution can be applied Go compiler related net access bugs till next NixOS release.
When moving to NixOS 23.11, this fix will be removed as it is already fixed in the unstable channel.


emrahbillur:
Note: Unless docker is enabled this fix will not cause extra build time.

vilvo (approved):

mikatammi:
Do we really need this? This says backport has been merged to NixOS 23.05 and it has already hit channel few months ago:
https://github.com/NixOS/nixpkgs/pull/244310

https://nixpk.gs/pr-tracker.html?pr=244310

mikatammi:
https://github.com/tiiuae/ghaf/commit/0e14b9c6104564bf9d098d3b97ae076df9c829f0
```
• Updated input 'nixpkgs':
    'github:NixOS/nixpkgs/48e82fe1b1c863ee26a33ce9bd39621d2ada0a33' (2023-07-28)
  → 'github:NixOS/nixpkgs/e5f018cf150e29aac26c61dac0790ea023c46b24' (2023-09-12)
```

Of course there's slight possibility it still wasn't on the channel 9 days after the merge which happened 2023-07-19. But it should be at least now

mikatammi (changes requested):
Please double check the need, after the flake input has now been updated

emrahbillur:
> Please double check the need, after the flake input has now been updated

Of course. This was just a temporary fix to be removed till it was on the channel. Since it was an issue with the Go compiler, this PR was just to notify other packages that may be affected by this issue. Otherwise it was to be removed in near future.

PR #270, opened 2023-09-15
Re-enable qboot on AppVMs

mikatammi:
Workaround the issue where vm's somewhat randomly hang at the last kernel output line being "ACPI: Core revision 20220331"


nesteroff (approved):
Tested this on Lenovo X1 Carbon. After the previous flake.lock update all APPVMs failed to start but with this workaround everything works fine again.

brianmcgillion (approved):

PR #271, opened 2023-09-15
Refactor demo apps inclusion

avnik:
Allow precise control to include/exclude GUI demo application, as well as add firefox as second browser.

### DIFF
--- a/modules/graphics/demo-apps.nix
+++ b/modules/graphics/demo-apps.nix
@@ -1,0 +1,57 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}: let
+  cfg = config.ghaf.graphics.demo-apps;
+  weston = config.ghaf.graphics.weston;
+
+  /*
+  Generate launchers to be used in weston.ini
+
+  Type: mkProgramOption ::  string -> bool -> option
+
+  */
+  mkProgramOption = name: default:
+    with lib;
+      mkOption {
+        inherit default;
+        type = types.bool;
+        description = "Include package ${name} to menu and system environment";
+      };
+in {
+  options.ghaf.graphics.demo-apps = with lib; {
+    chromium = mkProgramOption "Chromium browser" false;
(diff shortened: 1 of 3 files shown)


modules/graphics/demo-apps.nix
```diff
@@ -0,0 +1,72 @@
...
+      mkOption {
+        inherit default;
+        type = types.bool;
+        description = "Include package ${name} to menu and system environment";
+      };
+in {
+  options.ghaf.graphics.demo-apps = with lib; {
+    chromium = mkProgramOption "Chromium browser" false;
+    firefox = mkProgramOption "Firefox browser" weston.enableDemoApplications;
+    gala-app = mkProgramOption "Gala App" false;
```
mikatammi, line 38: Why is gala false by default?

mikatammi (changes requested):
Please fix formatting, eg. run `nix fmt *` on all commits and enable gala-app by default. Otherwise looks good and I like the idea

modules/graphics/demo-apps.nix
```diff
@@ -0,0 +1,72 @@
...
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}: let
+  cfg = config.ghaf.graphics.demo-apps;
+  weston = config.ghaf.graphics.weston;
+
+  /* Scaled down firefox icon */
```
mikatammi, line 12: I don't want to be too much of a style police, but this is the only place in this project where C-style comments are used, I actually didin't know before this Nix supports such comment style. Elsewhere the commenting style with `#` is used

avnik: https://github.com/tiiuae/ghaf/blob/main/modules/graphics/weston.ini.nix#L20 here also C-style comments, actually I split this file, and tried to keep same style.

modules/graphics/demo-apps.nix
avnik: gala require an account. And @vilvo suggest me to off it by default. Could be overriden by downstream projects

mikatammi: I guess we can disable it now and then enable it again when someone asks for it

mikatammi (approved):

PR #273, opened 2023-09-19
Pulseaudio lenovo x1

josa41:


mikatammi:
Please rebase. Maybe we could merge this for now, and improve later. I'd like to see audio enabling options in a separate module, and not enabled for all who use x86_64 module by default. Also using Pipewire instead of Pulseaudio could be one thing to experiment in future.

But I'm fine with merging this now if this has been properly tested.

modules/virtualization/microvm/appvm.nix
```diff
@@ -68,6 +74,8 @@
             qemu.extraArgs = lib.optionals (config.nixpkgs.hostPlatform.system == "x86_64-linux") [
               "-bios"
               "${pkgs.qboot}/bios.bin"
+              "-M"
+              "microvm,pcie=on"
```
mikatammi, line 78: Shouldn't this also have rest orf the options from the microvm.nix?
https://github.com/astro/microvm.nix/blob/main/lib/runners/qemu.nix#L35

mikatammi: Like @nesteroff has done here: https://github.com/tiiuae/ghaf/pull/256/files#diff-b7c4d8b1828ec4e55c3b85b348f4c180560550100c49b79a826d4cacfd1afaedR78

modules/hardware/x86_64-linux.nix
```diff
@@ -25,6 +25,16 @@ in
       hardware.enableRedistributableFirmware = true;
       hardware.enableAllFirmware = true;
 
+      # Enable pulseaudio systemwide support for host
```
mikatammi, line 28: Maybe move these into targets/lenovo-x1-carbon.nix for now? (As a temporary solution)

modules/virtualization/microvm/appvm.nix
nesteroff: Actually, It looks like it preserves previously set machine props. For example, it doesn't complain about missing pcie when I pass this:
```
-M microvm,pcie=on \                                                                                                                                                                             
-M microvm,rtc=off 
```
It is still a hack though.

josa41: I agree with you, it should be better indicated which props are active. However adding "all props" to multiple places is not good solution as then modifying some single prop gets complicated.

targets/lenovo-x1-carbon.nix
```diff
@@ -99,6 +99,8 @@
               SUBSYSTEM=="input",ATTRS{name}=="ELAN067C:00 04F3:31F9 Touchpad",KERNEL=="event*",GROUP="kvm",SYMLINK+="touchpad"
               # Laptop TrackPoint
               SUBSYSTEM=="input",ATTRS{name}=="TPPS/2 Elan TrackPoint",GROUP="kvm"
+              # Lenovo X1 integrated webcam
+              SUBSYSTEM=="usb", ATTR{idVendor}=="04f2", ATTR{idProduct}=="b751", GROUP+="kvm"
```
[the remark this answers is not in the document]
josa41: True, tested and works without the "+".
I'll create a separate patch for that since it is related to the webcam rather than pulseaudio.
I have to test the usb passthrough a bit more anyway.
Usb passthrough "works" but qemu usb device passthrough breaks after a while when the device has too much IO.

mikatammi (approved):

PR #274, opened 2023-09-20
DBUS host power control

baz2142:


mikatammi (changes requested):

overlays/custom-packages.nix
```diff
@@ -19,6 +19,7 @@
 {lib, ...}: {
   nixpkgs.overlays = [
     (final: prev: {
+      dbus-powercontrol = final.callPackage ../user-apps/dbus-powercontrol {};
```
mikatammi, line 22: Also maybe this stuff does not need to go via overlay? Instead ... (see below)

targets/lenovo-x1-carbon.nix
```diff
@@ -54,7 +54,10 @@
         ];
       }
       ({pkgs, ...}: {
-        ghaf.graphics.weston.launchers = [
+        ghaf.graphics.weston.launchers = let
+          adwaitaIconsRoot = "${pkgs.gnome.adwaita-icon-theme}/share/icons/Adwaita/32x32/actions/";
```
mikatammi, line 58: Instead you could add a line:
```
 dbus-powercontrol = pkgs.callPackage ../user-apps/dbus-powercontrol {};
```
in this let-section

mikatammi (changes requested):

PR #282, opened 2023-09-26
Ghaf nix build GitHub actions

henrirosten:
Run nix build in Github Actions:

- Uses cross-compile targets, but also allows emulated builds.
- Builds are run on GitHub-hosted runners.  
- Makes use of Cachix to push the build artifacts for later re-use. Using Cachix cache isn't strictly required, but might be useful for many reasons e.g.: to speed-up the PR builds, to allow faster local builds, and to offload the bursts of concurrent builds away from the main cache (vedenemo). See: https://app.cachix.org/cache/ghaf-dev.
- All builds are run concurrently, see the build matrix configuration for more details. From the current targets, 'lenovo-x1-carbon-gen11-debug' seems to take longest. Execution time obviously depends on how much needs to be re-built, currently the best case for the whole build matrix to finish is around 40 minutes. We start-off with limited amount of targets, to limit the github action minutes used by this new workflow.
- Some (not fully developed) ideas to improve the execution time:
  - Would it be possible to have this CI build output (subsets) that would not change unless the relevant inputs changed, meaning all build results would be read from the cache if the target had been built earlier with the same inputs. Currently, the main build results (store-disk.squashf, nixos.img) are re-build every time a new commit is pushed, even if the commit didn't introduce any changes to the inputs - [reference](https://github.com/tiiuae/ghaf/blob/5dd5711ff984df60f2d0e099d40a903577b42566/lib/default.nix#L6).
  
For more information, see: https://ssrc.atlassian.net/wiki/spaces/SP/pages/837648757/Running+nix+build+in+github+actions.

### DIFF
--- a/.github/workflows/nix-build-matrix.yml
+++ b/.github/workflows/nix-build-matrix.yml
@@ -1,0 +1,113 @@
+name: nix 
+
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build_matrix:
+    name: "build"
+    runs-on: ubuntu-latest
+    timeout-minutes: 360
+    strategy:
+      matrix:
+        include:
+          - arch: x86_64-linux
+            target: generic-x86_64-debug
+          - arch: x86_64-linux
+            target: lenovo-x1-carbon-gen11-debug
+          - arch: x86_64-linux
+            target: nvidia-jetson-orin-agx-debug-nodemoapps-from-x86_64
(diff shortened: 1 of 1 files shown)


mikatammi:
I already want to give 👍 at this point

brianmcgillion (approved):

henrirosten:
Closing this PR in favor of: https://github.com/tiiuae/ghaf/pull/293

PR #285, opened 2023-09-27
declarative partitioning

vilvo:
Depends on #249 
See commit message for status and testing.

- [ ] resolve `https://github.com/nix-community/disko/blob/master/lib/make-disk-image.nix#L17-L25` inclusion to ghaf installer image - currently dependencies are installed by disko during run-time with network connection.
- [ ] include documentation


vilvo:
This draft PR is to be superseded by https://github.com/tiiuae/ghaf/pull/340
Reasoning explained in more detail in https://github.com/tiiuae/ghaf/pull/328

PR #295, opened 2023-10-02
Add Lenovo P16 target

nesteroff:
This is a copy of Lenovo X1 Carbon target with updated passthrough devices for P16.


nesteroff:
Closing this one since there are no plans to add Lenovo P16 as a target for now.

PR #297, opened 2023-10-02
Add caml-crush package

avnik:
Draft version, need more customization as well as cross-compilation support


mikatammi:
Could you please rebase?

mikatammi:
We will continue development here: https://github.com/tiiuae/ghaf-caml-crush

PR #298, opened 2023-10-03
Add sysbench and speedtest-cli to debug-tools

mikatammi:


leivos-unikie (approved):
- Tested sysbench on Lenovo-X1
- And tested that riscv builds

brianmcgillion (approved):

PR #299, opened 2023-10-04
Add Ghaf icons

unbel13ver:
The assets directory is moved out from the modules to the source tree since it does not contain Nix source code.

The icons are stored in two formats: .png is used by weston and .svg is stored as a source for possible future conversions.

Imagemagick can not convert .svg to .png since it supports mostly raster graphics format. I made it using Inkscape, but I believe trying to convert svgs on the fly every build will carry huge Inkscape as a buildHost dependency and in future will cause troubles with crosscompilation.


jenninikko:
Is it possible to have all icons in the same SVG format?

unbel13ver:
> Is it possible to have all icons in the same SVG format?

Weston DE does not support icons in SVG format.

vilvo (approved):

jenninikko (approved):

vilvo:
Just for the record - imagemagick convert (svg -> png) via nix packaging would’ve been nice but not essential in the context of coming demo

unbel13ver:
> Just for the record - imagemagick convert (svg -> png) via nix packaging would’ve been nice but not essential in the context of coming demo

I've updated the PR description about this.

vilvo:
> > Just for the record - imagemagick convert (svg -> png) via nix packaging would’ve been nice but not essential in the context of coming demo
> 
> I've updated the PR description about this.

How about `rsvg-convert`?

screencapture

brianmcgillion (approved):

PR #309, opened 2023-10-07
windows-launcher: Change spice to enableSpice

mikatammi:
Change spice to enableSpice in package arguments. This is because there exists a package called spice, so when called through callPackage the spice argument gets filled with that package, instead of defaulting to false.

### DIFF
--- a/modules/windows-launcher/default.nix
+++ b/modules/windows-launcher/default.nix
@@ -7,16 +7,14 @@
   ...
 }: let
   cfg = config.ghaf.windows-launcher;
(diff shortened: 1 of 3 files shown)


nesteroff (approved):
Holy Guacamole! Thanks for fixing it.

henrirosten:
After https://github.com/tiiuae/ghaf/pull/307, the github action PR builder workflow also confirms that the builds for both `lenovo-x1-carbon-gen11-debug` and `nvidia-jetson-orin-agx-debug-nodemoapps-from-x86_64` now pass for this PR. 

See the full build workflow at: https://github.com/tiiuae/ghaf/actions/runs/6441216010

PR #290, opened 2023-09-29
Refactor VM networking

unbel13ver:
*Introduce new naming pattern for VMs
All VMs' names/hostnames are created using '<vmname>-vm' pattern. All VMs' network interfaces named using
'tap-<hostname>' pattern. NetVM's bridge is configured to include all 'tap-*' interfaces.
This approach makes no distinction between service VMs and application VMs, but it allows to easily guess the hostname without using a network toolkit.

*Introduce .ghaf local domain.
All VMs except of NetVM now get their IPs from NetVM's DHCP server. The VMs are accessible by their hostnames with .ghaf postfix added, for instance, 'ssh gui-vm.ghaf'.

*All VMs use separate subnet from now on, which is different from the debug-subnet.

*The network configuration which is common for every VM is now extracted into the separate 'vm-networking.nix' file.

PR #312, opened 2023-10-10
Fix netdev name in the windows-launcher

nesteroff:
In #290 network interface name pattern was changed from "vm-\*" to "tap-\*". This PR changes it in the windows-launcher as well so that the Windows VM can get network access through the netvm bridge.


mikatammi:
One question came to my mind, do you anywhere setup and stop the tap interfaces, like the way it is done for microvms

https://github.com/astro/microvm.nix/blob/main/nixos-modules/host.nix#L244-L279

PR #324, opened 2023-10-12
hardening: custom kernel config with ghaf option

vilvo:
* config option used with Lenovo X1 Carbon but set to false by default - safe to merge for further development
* builds for X1 and early boots...
* hardening process steps on NixOS documented
* overlay for kernel modules with allowMissing = true ... probably removes too much modules


modules/host/kernel.nix
```diff
@@ -0,0 +1,81 @@
...
+    configfile = ./ghaf_host_hardened_baseline;
+    allowImportFromDerivation = true;
+  };
+
+  cfg = config.ghaf.host.kernel_hardening;
+in
+  with lib; {
+    options.ghaf.host.kernel_hardening = {
+      type = types.bool;
+      default = false;
```
mikatammi, line 67: These two lines do not do anything, it is only the mkEnableOption next line that has any effect on anything

mikatammi (commented):
Overall LGTM, with one minor thing

mikatammi:
I'm questioning a bit is it a good idea to use the term "hardening" so widely in this context? I'd say better words would be "kernel minification" and "kernel customization". Especially usage of term `hardened_kernel` makes one implicitly think it would have something to do with the hardened linux kernel which is present in most distibutions with package named linux-hardened, and it has been name for this project for 20+ years https://github.com/anthraxx/linux-hardened

vilvo:
> I'm questioning a bit is it a good idea to use the term "hardening" so widely in this context? 

I wrote the general ghaf hardening under Architecture to give the context. It's true that this iteration does not yet implement the hardening requirements but only kernel customization. Nevertheless, I would rather keep the hardening for the kernel and also continue adding more info on hardening the other parts of the system - such as host use space, netvm, guests, etc. 

> I'd say better words would be "kernel minification" and "kernel customization". Especially usage of term `hardened_kernel` makes one implicitly think it would have something to do with the hardened linux kernel which is present in most distibutions with package named linux-hardened, and it has been name for this project for 20+ years https://github.com/anthraxx/linux-hardened

Minification/customization could be added to describe the current state of the kernel but the longer term target of ghaf is to harden the system - kernel is only one part of that hardening regardless of whether we use more generic linux-hardened (as in suitable generically for more x86_64 devices) like anthraxx, something else existing hardened or we develop ghaf-specific hardened.

modules/host/kernel.nix
vilvo: Removed redundant (already set) type and default value in [2e984bc](https://github.com/tiiuae/ghaf/commit/2e984bce87be0f2070aff32025f59a3420a21876)

vilvo:
@brianmcgillion - this kernel hardening and customization iteration is good to review for now. @vunnyso tested the kernel to boot to shell on X1. I tested it to build and did quick host kernel size comparison between the Ghaf X1 latest:
```
$ ls -lah /nix/store/4rviyxa4dqkzhchnwprgzgkc5cpvdw6w-linux-6.1.55/bzImage
-r--r--r-- 1 root root 8.4M Jan  1  1970 havealook/nix/store/4rviyxa4dqkzhchnwprgzgkc5cpvdw6w-linux-6.1.55/bzImage
```
and this kernel:
```
ls -lah /nix/store/fayyixcj7cq952ls0cx1j5jh2zl390rr-linux-6.5.5-ghaf-hardened/bzImage
-r--r--r-- 2 root root 1.6M Jan  1  1970 /nix/store/fayyixcj7cq952ls0cx1j5jh2zl390rr-linux-6.5.5-ghaf-hardened/bzImage
``` 
*NOTE*: this PR iteration should be considered a minimal (even host-only) baseline for Ghaf, not complete Ghaf hardened kernel with virtualization support adapted to X1 peripherals.

If you want to test building, enable the option:
``` 
$ git diff
diff --git a/targets/lenovo-x1-carbon.nix b/targets/lenovo-x1-carbon.nix
index 2895505..c942437 100644
--- a/targets/lenovo-x1-carbon.nix
+++ b/targets/lenovo-x1-carbon.nix
@@ -136,7 +136,7 @@
             users.extraUsers.microvm.extraGroups = ["audio" "pulse-access"];

             ghaf = {
-              host.kernel_hardening.enable = false;
+              host.kernel_hardening.enable = true;
```
and build `nix build .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug`

vilvo:
Rebased and tested on X1.

mikatammi (approved):

PR #334, opened 2023-10-26
Render window borders on the guivm side instead of the appvm

nesteroff:
- Windows borders are rendered on the compositor side.
- Border size and color can be configured in the appvm options.
- All waypipe processes are completely separated; each app gets its own instance of waypipe and vsockproxy.

### DIFF
--- a/modules/virtualization/microvm/appvm.nix
+++ b/modules/virtualization/microvm/appvm.nix
@@ -160,6 +160,20 @@ in {
               type = int;
               default = 0;
             };
+            borderColor = mkOption {
+              description = ''
+                Border color of the AppVM window
+              '';
+              type = nullOr str;
+              default = null;
+            };
+            borderSize = mkOption {
+              description = ''
+                Border size of the AppVM window
+              '';
+              type = int;
+              default = 5;
+            };
           };
         });
         default = [];
--- a/modules/virtualization/microvm/guivm.nix
+++ b/modules/virtualization/microvm/guivm.nix
@@ -92,21 +92,6 @@
 
         imports = import ../../module-list.nix;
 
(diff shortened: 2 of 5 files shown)


mikatammi:
@nesteroff Please rebase the branch, otherwise LGTM. Didn't test yet as the patch didn't apply cleanly

leivos-unikie:
Could the border size increase outside the actual window?

IMG_2297

leivos-unikie:
In zathura default border size (5) is already disturbing, the border is blocking view to the command line at the bottom.

IMG_2298
Here chromium with border size 3 and zathura with border size 5.

nesteroff:
Drawing a border outside of the actual window is quite hard to implement in Waypipe. We will need to keep track of the original window size, send increased size to the compositor, reallocate buffers, copy data and etc. One of the things that you can do is to use transparency. For example this is a semi-transparent red color: #FF000080.

Overall, drawing window borders in Waypipe is a temporary solution. Dmitry is currently implementing this feature in the compositor where it will be possible to change the color of the original window decorations. We plan to switch to this implementation once it's ready.

leivos-unikie:
> Drawing a border outside of the actual window is quite hard to implement in Waypipe. We will need to keep track of the original window size, send increased size to the compositor, reallocate buffers, copy data and etc. One of the things that you can do is to use transparency. For example this is a semi-transparent red color: #FF000080.
> 
> Overall, drawing window borders in Waypipe is a temporary solution. Dmitry is currently implementing this feature in the compositor where it will be possible to change the color of the original window decorations. We plan to switch to this implementation once it's ready.

I agree, no need to put more effort on this. The new implementation looks promising.

nesteroff:
Closing this one since the plan is to switch to labwc and change the colors of native window decorations.

PR #337, opened 2023-10-30
Module for caml-crush

mikatammi:
This module provides systemd service for caml-crush, and the options to configure it in various ways.


mikatammi:
We will continue developement here:
https://github.com/tiiuae/ghaf-caml-crush

PR #341, opened 2023-11-02
Introduce LabWC compositor/desktop environment

unbel13ver:
This commit introduces LabWC and making it a hardcoded option. (let's discuss).

This commit adds third option for DE, which can be enabled with `profiles.graphics.compositor = "labwc"` option.

Also at this moment I am not quite sure about how to implement painless configuration of labwc and its components. This topic needs some additional research.

### DIFF
--- a/modules/graphics/labwc.nix
+++ b/modules/graphics/labwc.nix
@@ -1,0 +1,67 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}: let
+  cfg = config.ghaf.graphics.labwc;
+in {
+  options.ghaf.graphics.labwc = {
+    enable = lib.mkEnableOption "labwc";
+  };
+
+  config = lib.mkIf cfg.enable {
+    ghaf.graphics.window-manager-common.enable = true;
+
+    environment.systemPackages = with pkgs; [
+      labwc
+    ];
+
+    # Next 2 services/targets are taken from official weston documentation
+    # and adjusted for labwc
+    # https://wayland.pages.freedesktop.org/weston/toc/running-weston.html
+
+    # labwc socket
+    systemd.user.sockets."labwc" = {
+      unitConfig = {
+        Description = "labwc, a Wayland compositor";
+        Documentation = "man:labwc(1)";
+      };
+      socketConfig = {
+        ListenStream = "%t/wayland-1";
+      };
+      wantedBy = ["labwc.service"];
+    };
+
+    # labwc service
+    systemd.user.services."labwc" = {
+      enable = true;
+      description = "labwc, a Wayland compositor, as a user service TEST";
+      documentation = ["man:labwc(1)"];
+      requires = ["labwc.socket"];
+      after = ["labwc.socket" "ghaf-session.service"];
+      serviceConfig = {
+        # Previously there was "notify" type, but for some reason
+        # systemd kills labwc.service because of timeout (even if it is disabled).
+        # "simple" works pretty well, so let's leave it.
+        Type = "simple";
+        #TimeoutStartSec = "60";
+        #WatchdogSec = "20";
+        # Defaults to journal
+        StandardOutput = "journal";
+        StandardError = "journal";
+        ExecStart = "${pkgs.labwc}/bin/labwc -C ${pkgs.labwc}/share/doc/labwc -s ${pkgs.labwc}/share/doc/labwc/autostart";
+        #GPU pt needs some time to start - labwc fails to restart 3 times in avg.
+        ExecStartPre = "${pkgs.coreutils}/bin/sleep 3";
+        Restart = "on-failure";
+        RestartSec = "1";
+
+        # Ivan N: adding openssh into the PATH since it is needed for waypipe to work
+        Environment = "PATH=${pkgs.openssh}/bin:$PATH WLR_RENDERER=pixman";
+      };
+      wantedBy = ["default.target"];
+    };
+  };
+}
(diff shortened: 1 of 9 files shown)


modules/graphics/labwc.nix
```diff
@@ -0,0 +1,142 @@
...
+        # systemd kills labwc.service because of timeout (even if it is disabled).
+        # "simple" works pretty well, so let's leave it.
+        Type = "simple";
+        #TimeoutStartSec = "60";
+        #WatchdogSec = "20";
+        # Defaults to journal
+        StandardOutput = "journal";
+        StandardError = "journal";
+        ExecStart = "${pkgs.labwc}/bin/labwc -C ${pkgs.labwc}/share/doc/labwc -s ${pkgs.labwc}/share/doc/labwc/autostart";
+        #GPU pt needs some time to start - labwc fails to restart 3 times in avg.
```
vilvo, line 61: I think this is ok for the first iteration to enable labwc but I'd like to ask you to elaborate "GPU pt needs some time to start" for further development.
 
What I'm after is that if "GPU pt" refers to GPU driver loading or some GPU related systemd service unit defined elsewhere, we could maybe wait for that service unit and add it to `after  = [ "gpu-pt-or-something.service"]` instead of restart-sleep-until-succeed? As said, that can be further development but please refine the comment of "GPU pt needs ..." for future readers.

overlays/custom-packages.nix
```diff
@@ -26,6 +26,25 @@
...
+             --replace swaybg ${final.swaybg}/bin/swaybg \
+             --replace kanshi ${final.kanshi}/bin/kanshi \
+             --replace waybar ${final.waybar}/bin/waybar \
+             --replace mako ${final.mako}/bin/mako \
+             --replace swayidle ${final.swayidle}/bin/swayidle
+
+             substituteInPlace ../docs/menu.xml \
+             --replace alacritty ${final.foot}/bin/foot
+
+             chmod +x ../docs/autostart
```
vilvo, line 44: Please limit chmod to `labwc` user and group only - `chmod ug+x`

vilvo: chmod ug+x actually breaks execution - tested by author

unbel13ver: As we discussed offline, unfortunately, that is how Nixpkgs suggest to keep runnable scripts.

vadika (approved):
Lets' integrate it as it is, there definitely will be imrovements later there.

modules/graphics/weston.nix
```diff
@@ -12,137 +12,63 @@ in {
...
+          # Ivan N: adding openssh into the PATH since it is needed for waypipe to work
+          Environment = "XDG_CONFIG_DIRS=$XDG_CONFIG_DIRS:/etc/xdg PATH=${pkgs.openssh}/bin:$PATH";
+        };
+        wantedBy = ["default.target"];
       };
-      wantedBy = ["multi-user.target"];
-    };
-  };
+    })
+    (lib.mkIf cfg.enable (import ./window-manager.nix {inherit lib config pkgs;}))
```
mikatammi, line 72: I think if you put all the stuff under same mkIf, you might not need the mkMerge above.

To Line 71, between } and ), you could put `// (import ./window-manager.nix {inherit lib config pkgs;})`

Actually at this point I realize, couldn't the window-manager.nix just have its own mkIf?

modules/graphics/window-manager.nix
```diff
@@ -0,0 +1,86 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}: {
```
mikatammi, line 8: You could put `{ config = lib.mkIf (config.ghaf.graphics.weston.enable || config.ghaf.graphics.labwc) { ...` here. Then you can avoid doing that lib.mkMerge in other files.

modules/graphics/labwc.nix
mikatammi: I'm actually also curious what causes this, earlier I added i915 to the initrd of the guivm to ensure it is loaded early during boot

mikatammi (changes requested):
I'm not able to merge before formatting check goes through.

Otherwise there are few improvement suggestions.

Also generally I don't like stuff being added to overlays unless there are no other options. I can let it go this time but it kind of increases techincal debt, because they need to be refactored out later

modules/graphics/window-manager.nix
mikatammi: Or simply create own option like `ghaf.graphics.window-manager.enable` and toggle it to true from weston.nix and labwc.nix modules

mikatammi: Maybe better name for the module could be like `window-manager-common`

modules/graphics/labwc.nix
unbel13ver: The issue SP-2838 is created to track this issue, all the info about that is there.

unbel13ver:
> Also generally I don't like stuff being added to overlays unless there are no other options. I can let it go this time but it kind of increases techincal debt, because they need to be refactored out later

I have though already the it would be a proper way to make a PR to nixpkgs first and bump our flake's `nixpkgs`. But it requires three (at least) consequential PRs instead of one.

Nerox9 (approved):

modules/graphics/window-manager.nix
unbel13ver: Thanks for the hint, addressed

modules/graphics/weston.nix
unbel13ver: Made it a separate module

vilvo (approved):
lgtm

vilvo:
IMG_2513

mikatammi (approved):

mikatammi:
Checked on AGX, NX, NUC and X1. Seems to work, and not break anything.

PR #345, opened 2023-11-07
devshell: for kernel development

vilvo:


flake.nix
```diff
@@ -77,6 +77,16 @@
         };
 
         formatter = pkgs.alejandra;
+
+        devShells.kernel = pkgs.mkShell {
+          packages = [pkgs.ncurses pkgs.pkg-config];
+          inputsFrom = [pkgs.linux_latest];
+          shellHook = ''
+            export src=${pkgs.linux_latest.src}
```
brianmcgillion, line 85: this is tied to the x86 mainline. Does it use the latest pinned kernel for a channel e.g. 23.05, 23.11, unstable. or does it always take the bleeding edge main?

mikatammi: I think it uses the one from the project's flake.lock file

mikatammi: So when we update the flake.lock file, it will change to the latest latest

vilvo: I think @mikatammi is right in that it uses https://github.com/tiiuae/ghaf/blob/main/flake.nix#L18 as pkgs and the exact version is based on flake.lock.

vilvo: > this is tied to the x86 mainline. 

It's actually available to all architectures - 
```
ghaf (8228000) [$?]
❯ nix develop .#devShells.
.#devShells.aarch64-linux  .#devShells.riscv64-linux  .#devShells.x86_64-linux
```
though those other than `x86_64` are currently not relevant for us in the context of `linux_latest` as we depend on the vendor BSP kernel versions for `aarch64` and `riscv64` targets.
To resolve this - we could develop target specific devShells with specific kernel source. In that case, I proposed @mikatammi that we'd refactor `devShells` to their own directory and import from there. To keep main level `flake.nix` clean.

flake.nix
```diff
@@ -77,6 +77,16 @@
...
+            unpackPhase
```
mikatammi, line 86: If we run unpackPhase I'm wondering should we also run patchPhase?

mikatammi (commented):
Generally LGTM

mikatammi (changes requested):
Tested, does not work:

```
~/ghaf]$ nix develop .#devShells.x86_64-linux.kernel
warning: Git tree '/home/mika/ghaf' is dirty
unpacking source archive /nix/store/ishbyx4xnwxkyxmqn4qpnzzl23zb5ylh-linux-6.5.5.tar.xz
unpacker appears to have produced no directories

```

I get this error message and it does not enter the devShell

vilvo:
> I get this error message and it does not enter the devShell

Addressed in 898e9a7 - also addresses https://github.com/tiiuae/ghaf/pull/345#discussion_r1385743262 (`patchPhase` added)

Tested:
```
[vilvo@carrie:~/ghaf]$ rm -rf linux-6.5.5/

[vilvo@carrie:~/ghaf]$ nix develop .#devShells.x86_64-linux.kernel
unpacking source archive /nix/store/ishbyx4xnwxkyxmqn4qpnzzl23zb5ylh-linux-6.5.5.tar.xz
source root is linux-6.5.5
setting SOURCE_DATE_EPOCH to timestamp 1695460479 of file linux-6.5.5/virt/Makefile
[ghaf-kernel-devshell:~/ghaf/linux-6.5.5]$
exit

[vilvo@carrie:~/ghaf]$ nix develop .#devShells.x86_64-linux.kernel
[ghaf-kernel-devshell:~/ghaf/linux-6.5.5]$
exit
```

flake.nix
vilvo: Addressed in 898e9a7

vilvo:
Added also `kernel-hardening-checker` in https://github.com/tiiuae/ghaf/commit/5984fd6abe11b672e561825bd024d20888dda8d6

brianmcgillion (approved):
verified

vilvo:
I rebased this with `main` and re-tested as the kernel was updated to `6.6` with 4453533  - works.

flake.nix
```diff
@@ -77,6 +77,37 @@
...
+
+            # python3+pip for kernel-hardening-checker
+            export PIP_PREFIX=$(pwd)/_build/pip_packages
+            export PYTHONPATH="$PIP_PREFIX/${pkgs.python3.sitePackages}:$PYTHONPATH"
+            export PATH="$PIP_PREFIX/bin:$PATH"
+
+            # install kernel-hardening-checker via pip under "linux-<version" for
+            # easy clean-up with directory removal - if not already installed
+            if [ ! -f "_build/pip_packages/bin/kernel-hardening-checker" ]; then
+              python3 -m pip install git+https://github.com/a13xp0p0v/kernel-hardening-checker
```
mikatammi, line 105: I think we could Nixify this Python package, but I'm fine if it happens in some follow-up commit

PR #346, opened 2023-11-07
Initial setup pulseaudio remote connection, control volume from guivm

josa41:
Initial setup for controlling pulseaudio from guivm
- Change volume toggle mute etc working
- Set default max volume level to a bit more reasonable level
- Using Pulseaudio TCP socket over ssh for now

### DIFF
--- a/modules/virtualization/microvm/guivm.nix
+++ b/modules/virtualization/microvm/guivm.nix
@@ -37,6 +37,8 @@
           };
           systemPackages = [
             pkgs.waypipe
+            pkgs.tcp-ssh
+            pkgs.pamixer
           ];
         };
 
--- a/overlays/custom-packages/default.nix
+++ b/overlays/custom-packages/default.nix
@@ -11,5 +11,6 @@
     (import ./weston)
     (import ./wifi-connector)
     (import ./qemu)
+    (import ./tcp-ssh)
   ];
 }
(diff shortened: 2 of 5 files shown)


modules/development/ssh.nix
```diff
@@ -30,6 +31,10 @@ in
     config = mkIf cfg.enable {
       services.openssh.enable = true;
       users.users.root.openssh.authorizedKeys.keys = authorizedKeys;
-      users.users.${config.ghaf.users.accounts.user}.openssh.authorizedKeys.keys = authorizedKeys;
+      users.users.${config.ghaf.users.accounts.user}.openssh.authorizedKeys = {
+        keys = authorizedKeys;
+        # Waypipe-ssh key is used here to create keys for ssh tcp tunneling for pulseaudio.
+        keyFiles = ["${pkgs.waypipe-ssh}/keys/waypipe-ssh.pub"];
```
mikatammi, line 37: The setting from this line should be in targets/lenovo-x1-carbon.nix file

mikatammi (changes requested):
At least one change is needed to prevent breaking NVIDIA builds.

Didn't test anything yet.

modules/development/ssh.nix
josa41: Thanks, fixed.
Earlier i did not find a correct place for host in lenovo-x1-carbon.nix but now i did.

mikatammi:
@josa41 Please rebase

PR #340, opened 2023-11-02
New installation approach

remimimimimi:

PR #350, opened 2023-11-08
Add Power Control buttons

baz2142:
The power control works via DBus connection from GUIVM to the host, which is established over ssh, started by systemd service in GUIVM

### DIFF
--- a/modules/development/ssh.nix
+++ b/modules/development/ssh.nix
@@ -3,24 +3,29 @@
 {
   config,
   lib,
+  pkgs,
   ...
 }: let
   cfg = config.ghaf.development.ssh.daemon;
 
-  authorizedKeys = [
-    # Add your SSH Public Keys here
-    # NOTE: adding your pub ssh key here will make accessing and "nixos-rebuild switching" development mode
-    # builds easy but still secure. Given that you protect your private keys. Do not share your keypairs across hosts.
-    #
-    # Shared authorized keys access poses a minor risk for developers in the same network (e.g. office) cross-accessing
-    # each others development devices if:
-    # - the ip addresses from dhcp change between the developers without the noticing AND
-    # - you ignore the server fingerprint checks
-    # You have been helped and you have been warned.
-    "sk-ssh-ed25519@openssh.com AAAAGnNrLXNzaC1lZDI1NTE5QG9wZW5zc2guY29tAAAAIA/pwHnzGNM+ZU4lANGROTRe2ZHbes7cnZn72Oeun/MCAAAABHNzaDo= brian@arcadia"
-    "sk-ssh-ed25519@openssh.com AAAAGnNrLXNzaC1lZDI1NTE5QG9wZW5zc2guY29tAAAAIEJ9ewKwo5FLj6zE30KnTn8+nw7aKdei9SeTwaAeRdJDAAAABHNzaDo= brian@minerva"
-    "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIDdNDuKwAsAff4iFRfujo77W4cyAbfQHjHP57h/7tJde ville.ilvonen@unikie.com"
-  ];
+  authorizedKeys =
+    [
+      # Add your SSH Public Keys here
+      # NOTE: adding your pub ssh key here will make accessing and "nixos-rebuild switching" development mode
+      # builds easy but still secure. Given that you protect your private keys. Do not share your keypairs across hosts.
+      #
+      # Shared authorized keys access poses a minor risk for developers in the same network (e.g. office) cross-accessing
+      # each others development devices if:
+      # - the ip addresses from dhcp change between the developers without the noticing AND
+      # - you ignore the server fingerprint checks
+      # You have been helped and you have been warned.
+      "sk-ssh-ed25519@openssh.com AAAAGnNrLXNzaC1lZDI1NTE5QG9wZW5zc2guY29tAAAAIA/pwHnzGNM+ZU4lANGROTRe2ZHbes7cnZn72Oeun/MCAAAABHNzaDo= brian@arcadia"
+      "sk-ssh-ed25519@openssh.com AAAAGnNrLXNzaC1lZDI1NTE5QG9wZW5zc2guY29tAAAAIEJ9ewKwo5FLj6zE30KnTn8+nw7aKdei9SeTwaAeRdJDAAAABHNzaDo= brian@minerva"
+      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIDdNDuKwAsAff4iFRfujo77W4cyAbfQHjHP57h/7tJde ville.ilvonen@unikie.com"
+
+      # So GUIVM can establish ssh-connection with the host without a password
+    ]
+    ++ lib.optional pkgs.stdenv.isx86_64 (builtins.readFile "${pkgs.waypipe-ssh}/keys/waypipe-ssh.pub");
 in
   with lib; {
     options.ghaf.development.ssh.daemon = {
@@ -29,6 +34,8 @@ in
 
     config = mkIf cfg.enable {
       services.openssh.enable = true;
+      services.openssh.extraConfig = pkgs.dbus-over-ssh.host.sshdConfigExtra;
+      services.openssh.settings.GatewayPorts = "yes";
       users.users.root.openssh.authorizedKeys.keys = authorizedKeys;
       users.users.${config.ghaf.users.accounts.user}.openssh.authorizedKeys.keys = authorizedKeys;
     };
--- a/modules/virtualization/microvm/guivm.nix
+++ b/modules/virtualization/microvm/guivm.nix
@@ -92,6 +92,23 @@
           };
           wantedBy = ["ghaf-session.target"];
         };
+
+        # Ssh service to pass DBUS from the host to GUIVM through
+        systemd.user.services.dbusOverSsh = {
+          enable = true;
+          description = "DBUS connection over ssh";
+          after = ["weston.service"];
+          serviceConfig = {
+            Environment = [];
+            ExecStart = pkgs.dbus-over-ssh.vm.establishDbusConnectionViaSshScript;
+            Restart = "always";
+            RestartSec = "5";
+            StandardError = "journal";
+            StandardOutput = "journal";
+            Type = "simple";
+          };
+          wantedBy = ["ghaf-session.target"];
+        };
       })
     ];
   };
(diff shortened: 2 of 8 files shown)


modules/virtualization/microvm/guivm.nix
vadika, line 1: Have you looked what was done for this in NetworkManager part? Should we unify this?

modules/development/ssh.nix
```diff
@@ -29,6 +34,8 @@ in
 
     config = mkIf cfg.enable {
       services.openssh.enable = true;
+      services.openssh.extraConfig = pkgs.dbus-over-ssh.host.sshdConfigExtra;
```
mikatammi, line 37: Please do not enable this for all targets, please move to targets/lenovo-x1-carbon.nix

modules/development/ssh.nix
```diff
@@ -29,6 +34,8 @@ in
...
+      services.openssh.settings.GatewayPorts = "yes";
```
mikatammi, line 38: Same for this

modules/development/ssh.nix
```diff
@@ -3,24 +3,29 @@
...
+      # - the ip addresses from dhcp change between the developers without the noticing AND
+      # - you ignore the server fingerprint checks
+      # You have been helped and you have been warned.
+      "sk-ssh-ed25519@openssh.com AAAAGnNrLXNzaC1lZDI1NTE5QG9wZW5zc2guY29tAAAAIA/pwHnzGNM+ZU4lANGROTRe2ZHbes7cnZn72Oeun/MCAAAABHNzaDo= brian@arcadia"
+      "sk-ssh-ed25519@openssh.com AAAAGnNrLXNzaC1lZDI1NTE5QG9wZW5zc2guY29tAAAAIEJ9ewKwo5FLj6zE30KnTn8+nw7aKdei9SeTwaAeRdJDAAAABHNzaDo= brian@minerva"
+      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIDdNDuKwAsAff4iFRfujo77W4cyAbfQHjHP57h/7tJde ville.ilvonen@unikie.com"
+
+      # So GUIVM can establish ssh-connection with the host without a password
+    ]
+    ++ lib.optional pkgs.stdenv.isx86_64 (builtins.readFile "${pkgs.waypipe-ssh}/keys/waypipe-ssh.pub");
```
mikatammi, line 28: And this also, read comment below

mikatammi: to he targets/lenovo-x1-carbon.nix

vilvo (commented):
I would like to question adding dbus+policies dependency to host. The option for this use case is `systemctl`:
```
systemctl [reboot|poweroff|suspend|hibernate]
```
That would leave us more options to minimize the host user space, potentially no dbus on host.  dbus between VMs is fine but potentially not necessary on host.

baz2142:
> I would like to question adding dbus+policies dependency to host. The option for this use case is `systemctl`:
> 
> ```
> systemctl [reboot|poweroff|suspend|hibernate]
> ```
> 
> That would leave us more options to minimize the host user space, potentially no dbus on host. dbus between VMs is fine but potentially not necessary on host.

It may be possible implementation, but:
- We already have NetwordManager, which works via DBus-over-ssh, -- so we just need to unify solution for both of the features
- DBus gives us flexibility: network control, power control, battery status and so on -- because everything was already done and no need for making our one bycicle. One connection to rule them all
- DBus gives us also policies, what can we allow or deny

So it's better to keep it as is. In case we have some security conserns, we can tweak our DBus-over-ssh solution

vilvo:
> > I would like to question adding dbus+policies dependency to host. The option for this use case is `systemctl`:
> > ```
> > systemctl [reboot|poweroff|suspend|hibernate]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That would leave us more options to minimize the host user space, potentially no dbus on host. dbus between VMs is fine but potentially not necessary on host.
> 
> It may be possible implementation, but:
> 
> * We already have NetwordManager, which works via DBus-over-ssh, -- so we just need to unify solution for both of the features
> * DBus gives us flexibility: network control, power control, battery status and so on -- because everything was already done and no need for making our one bycicle. One connection to rule them all

DBUS is ok between VMs. But it's not necessary on minimal host.
Network control is between VMs. Audio control is between VMs. 

Power control and battery status are the only use cases which require host dependency as of now. And for those we don't need dbus.

From security perspective, actually no. One connection will make exploitation of security vulnerabilities - https://www.cvedetails.com/vulnerability-list/vendor_id-13442/D-bus-Project.html - more easy. Limiting attack surface - especially on host - has been target and remains so.

> * DBus gives us also policies, what can we allow or deny

Limiting attack surface is even better.
 
> So it's better to keep it as is. In case we have some security conserns, we can tweak our DBus-over-ssh solution.

So we agree to disagree. We have security concerns as this is a secure platform development project. I digress.

vilvo:
I'm glad you've put thought into this. I'll try to clarify below.

 > In the solution to use ssh + systemctl commands we have one really big security problem: ssh-session to the host, which doesn't even require password -- and you can do whatever you want in the host. I guess, it's the biggest hole in the system now.
 
* Actually that's as designed - `ghaf` user and `ssh` are [enabled only for `debug`-profile](https://github.com/tiiuae/ghaf/blob/main/modules/profiles/debug.nix#L19-L26) - we have to get in to develop and debug, right? Try building the`release`-target and log into the host.
* We are in the process of enabling passwordless login with more secure PKI for the `ghaf`-user. First on `guivm`, later FIDO proxied to other VMs but not on host. Host [would have a non-root user separate from `ghaf`](https://github.com/tiiuae/ghaf/pull/359#issuecomment-1831488732
) - it should allow `sudo` 
 for `debug`-builds with password but no login otherwise.  
 
> I consider DBus with vulnerabilities and working policies less insecure, than ssh-session on the host with no password. The only problem in my current solution now -- I still have this hole as I passthrough DBus-socket via this hole-ssh-session.

For `debug`-builds we will continue to enable the ghaf internal ssh-sessions - [hopefully soon via Yubikey instead of password](https://github.com/tiiuae/ghaf/pull/359#issuecomment-1831613087).

For `release`-builds we can enable selected SSH sessions over authorized keys. Users can choose to generate secrets they trust during installation - this is currently being worked in the following PRs: #328 and #340.

> I have talked with @nesteroff, he proposed, that it might be possible to pass DBus-socket via vsock, like we have now with waypipe.

Technically yes - we even had wayland through SSH tunnel earlier but encryption proved too much overhead with wayland clients. For power controls, the overhead is acceptable (minimal) and the SSH wit keys really improves the security. In addition, we can both harden the sshd according to policies (e.g. `restricted_ssh_commands`) and extend the allowed commands more flexibly with *more minimal* dependencies when compared to vsock + DBUS. Especially when we need custom adaptations to request something from host HW. Please also remember that the minimal trusted computing base is the host target. We should be extra critical on adding dependencies and connections to host. Even when it's not yet ready. Removing something later is often more difficult than not adding something in the first place.

> 
> It's still not the perfect solution: vsock vulnerabilities + DBus vulnerabilities, but it's still should be better, than passwordless-ssh session, like we have now, + ssh own vulnerabilities: https://www.cvedetails.com/vulnerability-list/vendor_id-120/SSH.html
> 
> Or we can write our own service.
> 
> What do you think?

Thank you for questions and I really appreciate that you challenge the approach here. I hope above clarified but please let me know if I can help more.

baz2142:
> I'm glad you've put thought into this. I'll try to clarify below.
> 
> > In the solution to use ssh + systemctl commands we have one really big security problem: ssh-session to the host, which doesn't even require password -- and you can do whatever you want in the host. I guess, it's the biggest hole in the system now.
> 
> * Actually that's as designed - `ghaf` user and `ssh` are [enabled only for `debug`-profile](https://github.com/tiiuae/ghaf/blob/main/modules/profiles/debug.nix#L19-L26) - we have to get in to develop and debug, right? Try building the`release`-target and log into the host.
> * We are in the process of enabling passwordless login with more secure PKI for the `ghaf`-user. First on `guivm`, later FIDO proxied to other VMs but not on host. Host [would have a non-root user separate from `ghaf`](https://github.com/tiiuae/ghaf/pull/359#issuecomment-1831488732) - it should allow `sudo`
>   for `debug`-builds with password but no login otherwise.
> 
> > I consider DBus with vulnerabilities and working policies less insecure, than ssh-session on the host with no password. The only problem in my current solution now -- I still have this hole as I passthrough DBus-socket via this hole-ssh-session.
> 
> For `debug`-builds we will continue to enable the ghaf internal ssh-sessions - [hopefully soon via Yubikey instead of password](https://github.com/tiiuae/ghaf/pull/359#issuecomment-1831613087).
> 
> For `release`-builds we can enable selected SSH sessions over authorized keys. Users can choose to generate secrets they trust during installation - this is currently being worked in the following PRs: #328 and #340.
> 
> > I have talked with @nesteroff, he proposed, that it might be possible to pass DBus-socket via vsock, like we have now with waypipe.
> 
> Technically yes - we even had wayland through SSH tunnel earlier but encryption proved too much overhead with wayland clients. For power controls, the overhead is acceptable (minimal) and the SSH wit keys really improves the security. In addition, we can both harden the sshd according to policies (e.g. `restricted_ssh_commands`) and extend the allowed commands more flexibly with _more minimal_ dependencies when compared to vsock + DBUS. Especially when we need custom adaptations to request something from host HW. Please also remember that the minimal trusted computing base is the host target. We should be extra critical on adding dependencies and connections to host. Even when it's not yet ready. Removing something later is often more difficult than not adding something in the first place.
> 
> > It's still not the perfect solution: vsock vulnerabilities + DBus vulnerabilities, but it's still should be better, than passwordless-ssh session, like we have now, + ssh own vulnerabilities: https://www.cvedetails.com/vulnerability-list/vendor_id-120/SSH.html
> > Or we can write our own service.
> > What do you think?
> 
> Thank you for questions and I really appreciate that you challenge the approach here. I hope above clarified but please let me know if I can help more.

Thank you for so detailed clarification! I really appreciate it. I still have a few questions and some thoughts:

So, currently, if I make my solution via ssh, it will be working only in debug builds -- so we also need to enable ssh in release builds?

In general, it still feels like here is something wrong, if we look at our current approach "virtualise the critical parts": In mine and your proposals we still have direct access to the host. We have VMs for apps for isolation, VM for network, VM for GUI, but we don't have something isolated for power control, VMs control, system control (like updates, packages and so on) and maybe another devices control like disks.

So, do we have some plans to continue virtualise those parts?

baz2142:
I close this PR as was chosen another approach, but this branch still may be useful in the future. Another solution is at the new PR: https://github.com/tiiuae/ghaf/pull/406

PR #357, opened 2023-11-15
hydrajobs.nix: Drop X1 gnome target

cazfi:


brianmcgillion (approved):

mnokka-unikie:
PR357 was build ok and gnome was left out of the targets

PR #359, opened 2023-11-17
Enable Yubico Passwordless Authentication for ghaf guivm on Lenovo X1

vunnyso:
This patch enables Yubico Passwordless Authentication for debug builds on ghaf host. Users can generate U2F keys using Yubikey.

Task: SP-3243


docs/src/architecture/YubicoAuthentication.md
```diff
@@ -0,0 +1,34 @@
+<!--
+    Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+    SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Yubico Passwordless Authentication
+
+This section describes Yubico Passwordless Authentication and how to create U2F keys.
```
vilvo, line 8: Let's make this a bit more generic - "This section describes Ghaf reference implementation for passwordless authentication for fast-security online (FIDO). The reference implementation has been created with Yubico Yubikeys.
The implementation is modular and configurable - thus enabling also other implementations."

Please also add references to FIDO tech agnostic specs in the "The reader is expected to know the fundamentals ..." part.

docs/src/SUMMARY.md
```diff
@@ -17,6 +17,7 @@
     - [Platform Bus for Rust VMM](architecture/adr/platform-bus-passthrough-support.md)
   - [Hardening](architecture/hardening.md)
   - [Secure Boot](architecture/secureboot.md)
+  - [Yubico Passwordless Authentication](architecture/YubicoAuthentication.md)
```
vilvo, line 20: Please rename to `(architecture/passwordless_authentication.md)`

targets/lenovo-x1-carbon.nix
```diff
@@ -311,6 +313,11 @@
     {
       ghaf.host.secureboot.enable = false;
     }
+    # We can set it to true, if no Yubico hardware connected it works in usual way
+    ../modules/host/yubico.nix
+    {
+      ghaf.host.passwordless_authentication.enable = true;
```
vilvo, line 319: Can't make this default to `true` yet.

vilvo (changes requested):
Please address comments

targets/lenovo-x1-carbon.nix
vilvo: Above might be outdated request. I noticed the line 316 comment just know but can test this later.

modules/host/yubico.nix
```diff
@@ -0,0 +1,27 @@
...
+    environment.systemPackages = [
+      # For generating and debugging Yubikeys
+      pkgs.pam_u2f
+    ];
+
+    security.pam.services = {
+      login.u2fAuth = true;
+      sudo.u2fAuth = true;
+    };
+    security.pam.u2f.authFile = "/etc/u2f_mappings";
```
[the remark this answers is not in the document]
vilvo: Please add a combination of `assert`, `assertMsg` and `builtins.pathExists <path>` so that the user gets a proper error message if the `security.pam.u2f.authFile` has not been created by the user.

https://github.com/search?q=repo%3ANixOS%2Fnixpkgs%20assertmsg&type=code

alextserepov:
Please correct me if I'm wrong, but here is how I see the authentication process:

1. When you attempt to log in, the PAM stack processes the authentication request.
2. The pam_u2f module is invoked as per the configuration in the PAM configuration files (like /etc/pam.d/common-auth for system-wide authentication settings).
3. The module communicates with the U2F device, requesting it to sign a challenge with its private key.
4. You, as the user, activate the U2F device (usually by pressing a button on it).
5. The U2F device signs the challenge, and pam_u2f validates this signature using the stored public key.
6. If the signature is valid, authentication is successful, and you are logged in.

How do we control the physical security of the yubikey's then?

vilvo:
> Please correct me if I'm wrong, but here is how I see the authentication process:
> 
> 1. When you attempt to log in, the PAM stack processes the authentication request.
> 2. The pam_u2f module is invoked as per the configuration in the PAM configuration files (like /etc/pam.d/common-auth for system-wide authentication settings).
> 3. The module communicates with the U2F device, requesting it to sign a challenge with its private key.
> 4. You, as the user, activate the U2F device (usually by pressing a button on it).
> 5. The U2F device signs the challenge, and pam_u2f validates this signature using the stored public key.
> 6. If the signature is valid, authentication is successful, and you are logged in.

No need to correct your description of the authentication process
 
> How do we control the physical security of the yubikey's then? Or did we intend to implement 2FA instead?

 We are not at that point yet but it could be 2FA. Then again, Ghaf is a reference platform and passwordless (PKI) over password authentication would be a major improvement - even if we don't enforce 2FA from the get go.
 
In short term, I would like to see us introduce a ghaf platform capability to use passwordless authentication instead of password based in debug builds. Those should be mutually exclusive - meaning the when the passwordless authentication [is declared in ghaf config (see this PR](https://github.com/tiiuae/ghaf/pull/359/files#diff-789892fdf4e4c650d81ce42dbc84586807ced5e73a8d53c06775acf7df2d0876R319)) and authorized public keys are included in the build with proper PAM module registrations (missing from this PR - need further development), the password authentication cannot be used when the Yubikeys are not connected.
 
There are also other things that must be developed on top of this PR for this platform capability to be relevant but it's a good start.

vilvo (changes requested):
The baseline implementation was verified to work as follows:
- user runs the described process on host

I discussed with the author that we need to continue the development at least for following:
- make `yubikey.nix` a module that can be included in guest VMs - starting from GUI VM - instead of host. In that case, the yubikey must be passed through from host to GUI VM.
- `yubikey.nix` module must include the authorized public key to image and register it to PAM so that authorized hardware keys can be used to log into the system (GUI VM) without running the registration in the ghaf system. This needs further study.
- when `ghaf.host.passwordless_authentication` is declared `true`, the can be no fallback to password authentication because that would remove the benefits of having the passwordless authentication.

vilvo:
> I discussed with the author that we need to continue the development at least for following:
> 
> * make `yubikey.nix` a module that can be included in guest VMs - starting from GUI VM - instead of host. In that case, the yubikey must be passed through from host to GUI VM.

I understood from @vunnyso that with b132b8f above is supported. @leivos-unikie - can you test passwordless authentication in GUIVM shell to verify it? 

> * `yubikey.nix` module must include the authorized public key to image and register it to PAM so that authorized hardware keys can be used to log into the system (GUI VM) without running the registration in the ghaf system. This needs further study.

@vunnyso - is this already working? 

> * when `ghaf.host.passwordless_authentication` is declared `true`, there can be no fallback to password authentication because that would remove the benefits of having the passwordless authentication.

@vunnyso - is this already working?

leivos-unikie:
> > I discussed with the author that we need to continue the development at least for following:
> > 
> > * make `yubikey.nix` a module that can be included in guest VMs - starting from GUI VM - instead of host. In that case, the yubikey must be passed through from host to GUI VM.
> 
> I understood from @vunnyso that with [b132b8f](https://github.com/tiiuae/ghaf/commit/b132b8fab6a4fca8c14a22d8fde1872770449077) above is supported. @leivos-unikie - can you test passwordless authentication in GUIVM shell to verify it?

Sure, I can try tomorrow. Where I can get such Yubikey? 

Did I understand correctly that with some USB key I could ssh into gui-vm (from ghaf-host?) without having to give password? Or is this yubikey virtual in this case?

vilvo:
> > > I discussed with the author that we need to continue the development at least for following:
> > > 
> > > * make `yubikey.nix` a module that can be included in guest VMs - starting from GUI VM - instead of host. In that case, the yubikey must be passed through from host to GUI VM.
> > 
> > 
> > I understood from @vunnyso that with [b132b8f](https://github.com/tiiuae/ghaf/commit/b132b8fab6a4fca8c14a22d8fde1872770449077) above is supported. @leivos-unikie - can you test passwordless authentication in GUIVM shell to verify it?
> 
> Sure, I can try tomorrow. Where I can get such Yubikey?
Great. I asked Antti to get one for you.
 
> Did I understand correctly that with some USB key I could ssh into gui-vm (from ghaf-host?) without having to give password? Or is this yubikey virtual in this case?

Test it like it's described in the documentation. We'll collected your questions and feedback for @vunnyso to update the documentation accordingly for the next users. I'm familiar with the process from the first iteration (worked only on host) but haven't tested the latest additions to support guivm passwordless authentication.

vunnyso:
> > * `yubikey.nix` module must include the authorized public key to image and register it to PAM so that authorized hardware keys can be used to log into the system (GUI VM) without running the registration in the ghaf system. This needs further study.
> 
> @vunnyso - is this already working?
>>> Yes @vilvo I have added two public keys from my two different Yubikey in ghaf system, if we are using new Yubikey then its public key need to be added in /etc/u2f_mappings. 
> > * when `ghaf.host.passwordless_authentication` is declared `true`, there can be no fallback to password authentication because that would remove the benefits of having the passwordless authentication.
> 
> @vunnyso - is this already working?
>>> Currently user will be asked for sudo password, as user would have been previously if the Yubikey is not plugged in. Passwordless auth is only enabled when the Yubikey is plugged in.

vilvo:
> > > * `yubikey.nix` module must include the authorized public key to image and register it to PAM so that authorized hardware keys can be used to log into the system (GUI VM) without running the registration in the ghaf system. This needs further study.
> >  
> > @vunnyso - is this already working?
> Yes @vilvo I have added two public keys from my two different Yubikey in ghaf system, if we are using new Yubikey then its public key need to be added in /etc/u2f_mappings.

Ok, great. @leivos-unikie - please test this scenario. Make sure that the documentation will have enough details for users to generate the key pair on their own computer and details instructions on how to copy the public key to correct place so that it ends up in the built image. If unclear, ask @vunnyso for exact details and let's include the detailed steps to the documentation.

vilvo:
> > > * when `ghaf.host.passwordless_authentication` is declared `true`, there can be no fallback to password authentication because that would remove the benefits of having the passwordless authentication.
> > 
> > @vunnyso - is this already working?
> Currently user will be asked for sudo password, as user would have been previously if the Yubikey is not plugged in. Passwordless auth is only enabled when the Yubikey is plugged in.

Ok, this needs more development then for the reasons I mentioned earlier. 
[NixOS has mechanism to disable login using password as follows](https://search.nixos.org/options?channel=unstable&show=users.users.%3Cname%3E.initialHashedPassword&from=0&size=50&sort=relevance&type=packages&query=users):

If `users.users.<name>.initialHashedPassword` is set to `null`, "this user will not be able to log in using a password (i.e. via `login` command)"

In ghaf, we set the reference password of `ghaf`-user in debug-builds to trival - `ghaf` - here:
https://github.com/tiiuae/ghaf/blob/main/modules/users/accounts.nix#L25-L38

Please add conditional to https://github.com/tiiuae/ghaf/blob/main/modules/users/accounts.nix so that ghaf-user `initialHashedPassword = null` when [this option](https://github.com/tiiuae/ghaf/pull/359/files#diff-f293074cee2879c32954690c2bf16e0788e772035b9abef522b73a5bacdcbe08R23) is true.

Please let me know if this needs clarification.

modules/module-list.nix
```diff
@@ -18,4 +18,8 @@
   ./version
   ./virtualization/docker.nix
   ./windows-launcher
+  ./host/yubikey.nix
+  {
+    ghaf.host.passwordless_authentication.enable = true;
```
vilvo, line 23: https://github.com/tiiuae/ghaf/pull/359/files#r1408959584

vilvo:
> Please add conditional to https://github.com/tiiuae/ghaf/blob/main/modules/users/accounts.nix so that ghaf-user `initialHashedPassword = null` when [this option](https://github.com/tiiuae/ghaf/pull/359/files#diff-f293074cee2879c32954690c2bf16e0788e772035b9abef522b73a5bacdcbe08R23) is true.

Implementing this will break `ghaf`-user password authentication on host. That's ok. When you get this work, let's define a separate debug user with password authentication for host. It does not need [all the other groups](https://github.com/tiiuae/ghaf/blob/main/modules/users/accounts.nix#L40) and it's better that it will not be shared with the guest VMs when `ghaf.passwordless_authentication.enable = true;`

leivos-unikie:
> Ok, great. @leivos-unikie - please test this scenario. Make sure that the documentation will have enough details for users to generate the key pair on their own computer and detailed instructions on how to copy the public key to correct place so that it ends up in the built image. If unclear, ask @vunnyso for exact details and let's include the detailed steps to the documentation.

I created the public key for yubikey, added it to modules/host/demo-yubikeys/u2f_keys, built the image. Now running it on Lenovo with yubikey connected and I can see the key in /etc/u2f_mappings. I tested 'ssh ghaf@gui-vm.ghaf' from ghaf-host but it's still asking "Are you sure you want to continue..." and password. Yubikey blinked some time after I had logged in.

vilvo:
>  I tested 'ssh [ghaf@gui-vm.ghaf](mailto:ghaf@gui-vm.ghaf)' from ghaf-host but it's still asking "Are you sure you want to continue..." and password. Yubikey blinked some time after I had logged in.

I think it does not cover this scenario (yet) - only passwordless authentication within the guivm:
- sudo in guivm shell etc.

leivos-unikie:
> > I tested 'ssh [ghaf@gui-vm.ghaf](mailto:ghaf@gui-vm.ghaf)' from ghaf-host but it's still asking "Are you sure you want to continue..." and password. Yubikey blinked some time after I had logged in.
> 
> I think it does not cover this scenario (yet) - only passwordless authentication within the guivm:
> 
>     * sudo in guivm shell etc.

Yeah, with sudo it works!

vilvo:
@vunnyso Can you please check what’s the effort to connect yubi PAM with sshd for passwordless authentication to guivm over ssh? If it’s big effort, we can scope it out of this PR. If it’s small effort, please add the support to this PR.

docs/src/architecture/passwordless_authentication.md
```diff
@@ -0,0 +1,41 @@
...
+
+The reader is expected to know the fundamentals of [FIDO](https://fidoalliance.org/specifications/), [Yubico](https://www.yubico.com/) and [U2F](https://developers.yubico.com/U2F/).
+
+## Prerequisites for Yubikey Passwordless Authentication
+
+User must have compatible [YubiKey hardware](https://www.yubico.com/products/) to begin with.
+
+## Creating Yubikey U2F Keys
+
+Yubikey U2F keys can be created with [pamu2fcfg](https://developers.yubico.com/pam-u2f/), a module implements PAM over U2F and FIDO2. pamu2fcfg is available in Nixpkgs as pkgs.pamu2fcfg.
```
leivos-unikie, line 21: Change to "... is available in Nixpkgs as pkgs.pam_u2f.

docs/src/architecture/passwordless_authentication.md
```diff
@@ -0,0 +1,41 @@
...
+
+Use the following command to create your U2F keys for Ghaf gui-vm .
```
leivos-unikie, line 23: So is the goal to eventually be able to set up the public key by running this within gui-vm?

Currently /etc/u2f_mappings is unwritable.

docs/src/architecture/passwordless_authentication.md
```diff
@@ -0,0 +1,41 @@
...
+
+After running above command, there will be green light blinking on Yubikey and user need to touch it.
+Once executed successfully, now user authentication becomes passwordless. When prompted with password just touch Yubikey.
+More details about mapping file can found here [central-authorization-mapping](https://github.com/Yubico/pam-u2f#central-authorization-mapping)
+
+## Current Implementation
+
+Since passwordless authentication is mapped with Yubikey in use and each Yubikey will generate unique key. So its not possible to use generic key.
+Yubikey Passwordless Authentication feature is enabled in debug builds for Ghaf gui-vm only.
+We are using pre-generated public keys which are created using pam2fcfg tool on different machine and it work automatically on guivm after the public key part is included in the built image for specific Yubikey.
```
leivos-unikie, line 36: Could we add:
...guivm after the public key part is included in the built image for specific Yubikey by running
`pamu2fcfg -u ghaf -o pam://gui-vm | sudo tee ./modules/host/demo-yubikeys/u2f_mappings`
withing the ghaf repository before building.

docs/src/architecture/passwordless_authentication.md
```diff
@@ -0,0 +1,41 @@
...
+If user want to register their Yubikey hardware then public key need to generated using pamu2fcfg tool and updated in /etc/u2f_mappings file.
+Kindly note keeping plain public key in Ghaf image is not recommended way and is to be used for demo purpose only.
+
+### Yubikey Passwordless Authentication Verification
+If system is not asking you password in Ghaf host anymore then Yubikey Passwordless Authentication feature is working fine.
```
leivos-unikie, line 41: So system should not ask password in ghaf-host, or in any vm too?

Currently works only in gui-vm when authenticating sudo command.

docs/src/architecture/passwordless_authentication.md
```diff
@@ -0,0 +1,41 @@
...
+More details about mapping file can found here [central-authorization-mapping](https://github.com/Yubico/pam-u2f#central-authorization-mapping)
+
+## Current Implementation
+
+Since passwordless authentication is mapped with Yubikey in use and each Yubikey will generate unique key. So its not possible to use generic key.
+Yubikey Passwordless Authentication feature is enabled in debug builds for Ghaf gui-vm only.
+We are using pre-generated public keys which are created using pam2fcfg tool on different machine and it work automatically on guivm after the public key part is included in the built image for specific Yubikey.
+If user want to register their Yubikey hardware then public key need to generated using pamu2fcfg tool and updated in /etc/u2f_mappings file.
+Kindly note keeping plain public key in Ghaf image is not recommended way and is to be used for demo purpose only.
+
```
leivos-unikie, line 39: I checked also vendor/product id of my yubikey with
`lsusb | grep -i Yubico`

The ids happened to be the same as in targets/lenovo-x1-carbon.nix but if they were different I suppose also those need to be updated into the target nix file.

leivos-unikie (changes requested):

docs/src/architecture/passwordless_authentication.md
vunnyso: Yes @leivos-unikie currently we have tested Yubikey with vendorid=0x1050 and productid=0x0407.
In future if any new requirement comes to support different Yubikey, then we need to verify its functionality and the update target nix file accordingly.

docs/src/architecture/passwordless_authentication.md
vilvo: Now the `/etc/u2f_mappings` should be link to `/nix/store/<hash>...` - meaning that the public key you set before build gets deployed to the target image and enabled. The previous implementation was so that user had to set up the public key by running the key generation process. We might be able to have both if Yubitools support multiple authorized keys files but let's not focus on that now. I requested @vunnyso to change it like it currently is that we can authorize public keys that have been generated before on another machine. This way we can have more secure authentication from the first guivm boot.

docs/src/architecture/passwordless_authentication.md
vilvo: Something like that might be useful improvement for the initial key generation (clean Yubikey).
There's also the scenario where the user already has Yubikey (with private key in it) and wants to only add the existing public key (which matches the private key in the Yubikey) to the authorized keys (`u2f_mappings`).

docs/src/architecture/passwordless_authentication.md
vilvo: That's stale statement in documentation. Implementation was first tested on host and is to be removed from there now that the passthrough to `guivm` works.

docs/src/architecture/passwordless_authentication.md
vilvo: `lsusb` ids are vendor id and product id. The passthrough is enabled based on those. If they were different, the user is using another product and yes, the USB passthrough to `guivm` must be updated in that scenario. It might be possible to be more generic with the Yubikey reference implementation passthrough to cover some set of HW keys from Yubico but that implementation can be left for later.

docs/src/architecture/YubicoAuthentication.md
vunnyso: Updated as per review comment.

docs/src/SUMMARY.md
vunnyso: Updated as per review comment.

targets/lenovo-x1-carbon.nix
vunnyso: We can set default to false if required.

modules/module-list.nix
vunnyso: Property renamed accordingly.

docs/src/architecture/passwordless_authentication.md
vunnyso: Changed.

docs/src/architecture/passwordless_authentication.md
vunnyso: Updated documentation.

docs/src/architecture/passwordless_authentication.md
vilvo: I addressed this in [32402a8](https://github.com/tiiuae/ghaf/commit/32402a8d166f6400e47e46a3a802f927ef321088) https://github.com/tiiuae/ghaf/pull/359/files#diff-4c362f5486d800f46a0856fd12dddad0cc6d84e96e798bcfb1adb6cbf19db3c2R19

@mikatammi was working on Hardware definition here https://github.com/tiiuae/ghaf/pull/382 so I think we should leave refactoring these USB `vendorId` and `productId`  to future work https://github.com/tiiuae/ghaf/pull/359/files#diff-789892fdf4e4c650d81ce42dbc84586807ced5e73a8d53c06775acf7df2d0876R116-R184
(copy-paste of `vendorId` and `productId` are minor issue that can be addressed later - the bigger work is required if we want to enable passing through different types of `producId`s of Yubikeys)

vilvo:
The biggest blocker in current implementation is that [this](https://github.com/tiiuae/ghaf/pull/359/files#diff-303a9aa4aabd4ad66497726f14073c18a1f374d38dcc41afb91a793e60b31ee9R40-R46) works for `gui-vm` but breaks password login in debug-builds to `ghaf-host` and other guests than `gui-vm`.

I need to give this some thought but the first idea that comes to my mind is that the time is ripe for us to stop sharing the `ghaf`-user module across the system. In practice, we would introduce:
- another non-root user account for host with `wheel` group for `debug`-builds. `No login` with `release`-builds.
- another `ghaf`-user debug-account (separate module) for other guests. This allows password login for debug - at least until we can do FIDO proxying from `gui-vm` to other guests. 

Optionally we could create additional system-user account for the `systemvms` (e.g. `netvm`, etc.). In general, the common GUI-VM user role should be separated from the system management. This would then serve also `release`-builds.

vunnyso:
Future implementation of this PR will be depended on #459 and will integrate code changes with same. 
Closing this PR.

PR #363, opened 2023-11-22
Add hardened hypervisor option for lenovo X1 carbon

Kamarjam:
Set linux kernel version to 6.1.55 + apply pkvm intel X86 patch


modules/host/kernel.nix
vilvo, line 1: I think this is good solution for now and leaves room to merge host kernel hardening that combines the pkvm (virtualization hardening) and other hardening options in future. Which is likely something we want.

modules/host/kernel.nix
```diff
@@ -55,19 +67,40 @@
...
+  } ];
+  
+  kern_cfg = config.ghaf.host.kernel_hardening;
+  hyp_cfg = config.ghaf.host.hypervisor_hardening;
 in
   with lib; {
     options.ghaf.host.kernel_hardening = {
       enable = mkEnableOption "Host kernel hardening";
     };
+
```
vilvo, line 96: At the moment `ghaf.host.kernel_hardening.enable` and  `ghaf.host.hypervisor_hardening.enable` are mutually exclusive. I'm not sure if there's a nixified way to define options that are mutually exclusive - @mikatammi ? 
home-manager options seemed to document that "option x is mutually exclusive with y" and "option y is mutually exclusive with x". That's least we could add to both of these.

targets/lenovo-x1-carbon.nix
```diff
@@ -173,6 +173,8 @@
 
             ghaf = {
               host.kernel_hardening.enable = false;
+
```
vilvo, line 176: Continuing on mutually exclusive - both can be `false` both both can't be `true` at the same time. Either can be `true` separately.

vilvo (changes requested):
Let's address the mutually exclusive when we hear comments from @mikatammi - 
If there's no recommended nix way to exclude options, the least we can do is to add this to the module documentation as they get generated to here https://tiiuae.github.io/ghaf/ref_impl/modules_options.html#ghafhostkernel_hardeningenable

Also, please run `nix fmt` and force push the formatting changes to the PR branch.

vilvo:
> Also, please run `nix fmt` and force push the formatting changes to the PR branch.

Sorry that I was ambiguous - please squash the commits into one. The mainline version history stays cleaner. Thanks for understanding. 
(I meant please `nix fmt && git commit --amend -s -a && git push -f ...`)

vilvo:
> Let's address the mutually exclusive when we hear comments from @mikatammi - If there's no recommended nix way to exclude options, the least we can do is to add this to the module documentation as they get generated to here https://tiiuae.github.io/ghaf/ref_impl/modules_options.html#ghafhostkernel_hardeningenable

I tested building this and realized that the new option is nested inside the `ghaf.host.kernel_hardening` so it requires both `true`.

vilvo (approved):

mikatammi:
> > Let's address the mutually exclusive when we hear comments from @mikatammi - If there's no recommended nix way to exclude options, the least we can do is to add this to the module documentation as they get generated to here https://tiiuae.github.io/ghaf/ref_impl/modules_options.html#ghafhostkernel_hardeningenable
> 
> I tested building this and realized that the new option is nested inside the `ghaf.host.kernel_hardening` so it requires both `true`.

Asserts could be used for this

PR #364, opened 2023-11-23
Options for AGX and NX PCI Passthroughs

mikatammi:
Instead of enabling WLAN PCI Passthrough for Orin AGX or Ethernet PCI Passthrough always in the jetson-orin.nix module, refactor options for AGX and NX PCI Passthroughs, which will enable or disable PCI Passthrough related functionality and patches from the NVIDIA Jetson Orin modules.


brianmcgillion (approved):

emrahbillur:
Checked on NX and will check it on AGX also. A great refactor for adding options.

PR #365, opened 2023-11-23
Flake parts

brianmcgillion:
Start moving to flake-parts.

1st patch of many - Ultimate goal of becoming more modular.
Enabling more devshell functionality
Enables checkers for:
1. License
2. format
3. Deadcode
4. anti-patterns


juliuskoskela:
- Why do we have `lib.nix` and `lib/` (folder with default.nix so would be called as lib in Nix)? Maybe a bit confusing especially if lib.nix doesn't belong into lib/.
- What is the difference between `shell.nix` at the root and `nix/devshell.nix`?

vilvo:
> * What is the difference between `shell.nix` at the root and `nix/devshell.nix`?

`shell.nix` is compatibility shell for non-flake.
`nix/devshell.nix` has our flake devshells refactored out of top level flake - I proposed it here https://github.com/tiiuae/ghaf/pull/345#discussion_r1384704160

flake.nix
```diff
@@ -16,132 +16,123 @@
...
+      inputs = {
+        nixpkgs.follows = "nixpkgs";
+        nixpkgs-stable.follows = "nixpkgs";
+        flake-utils.follows = "flake-utils";
+        flake-compat.follows = "flake-compat";
+      };
+    };
+
+    # For preserving compatibility with non-Flake users
+    flake-compat = {
```
mikatammi, line 57: I'm not sure is adding flake-compat a good idea? I'm not completely against it but I'm also not going to use it.

Previously when we didn't have changed to declarative configurations with microvm.nix it wouldn't have even been possible to add flake-compat. But now I'm just a bit afraid that it might make it even more confusing for newcomers joining the project

brianmcgillion: the primary aim is so that folks who do nt have nix enabled in their system can just build the solution. All it does is to hide the flake goodness and give folks a standard default.nix and shell.nix experience

vilvo: I agree with Mika. In addition, it adds to complexity and future maintenance promise.

brianmcgillion:
> * Why do we have `lib.nix` and `lib/` (folder with default.nix so would be called as lib in Nix)? Maybe a bit confusing especially if lib.nix doesn't belong into lib/.

This is a tmp solution, as the comment says it is a start moving to flake-parts. There are another load of PRs that are due, but this start is just to try and reduce the big-bang releasing. In those future releases, there will be a single-unified library interface

flake.nix
brianmcgillion: I am not sure I understand the added complexity? It is used to make a flake into something that a user can run without having to enable Experimental features. It exists only in the default.nix and shell.nix entry points for those folks. For others the flake and nix3 functions are the starting point. It has no impact in the flake flow. 

Maybe there will be some breaking change that will inhibit the shell.nix from working but in terms of complexity i must have missed something on what flake compat does, other than set the hash of all the repos in a more traditional nix way.

vilvo: > I am not sure I understand the added complexity? It is used to make a flake into something that a user can run without having to enable Experimental features. It exists only in the default.nix and shell.nix entry points for those folks. For others the flake and nix3 functions are the starting point. It has no impact in the flake flow.

At high level it all boils down to this - https://github.com/NixOS/nix/issues/4024 - "Flake is not backward compatible with usual non-flake commands". Nor will it be as there is no such promise with flakes.
 
> Maybe there will be some breaking change that will inhibit the shell.nix from working but in terms of complexity i must have missed something on what flake compat does, other than set the hash of all the repos in a more traditional nix way.

Practically it will likely work for most cases as the flake-compat is written by the nix BDFL themself. For me, the added complexity comes from the question: "Is this something we can't take away?" If the answer is no, it adds to complexity. IMO - sometimes less is more. The practical side of keeping it is that if some decides to use it and finds some esoteric corner case, will anyone of us care to look at it? And when we care, can we help?

Anyway - it might be practical to have it for some people and I don't strongly oppose it. I digress but thanks for asking for clarification.

mikatammi (approved):
LGTM. Please squash commits into one before merge, as it seems the commits in between do not maintain full buildable state of the project, i.e. if you try to build things from some of the commits they are broken.

PR #369, opened 2023-11-26
Disable Gnome as a target

brianmcgillion:
This is a stepping stone to moving Gnome to its own module in another repo.


brianmcgillion:
stop the churn in with unsusd packages

PR #370, opened 2023-11-27
wifi-connector: Support nmcli interface

unbel13ver:
Now when NetworkManager is in use for NetVM, edit the script to use nmcli to avoid any conflicts.


riskuuse:
As NetworkManager is in use only in lenovo X1 target, wouldn't this brake wifi-connector for other targets?

unbel13ver:
> As NetworkManager is in use only in lenovo X1 target, wouldn't this brake wifi-connector for other targets?

Voi ei, it will!!!

riskuuse:
Looks good to me now. Tested on X1 but not on any other targets.

riskuuse (approved):
Looks good to me now. Tested on X1 but not on any other targets.

mikatammi (approved):

PR #427, opened 2023-12-21
refactor: dns+dhcp to a module, support short host queries

vilvo:
## Description of changes

* Renaming module to dns-dhcp allows us to change the internal implementation
  of DNS and DHCP services from dnsmasq to something else, like rust-based,
  later
* Current implementation stays dnsmasq-based but is only refactored into a
  module
* Disable systemd resolved on client side VMs completely, including DNS stub
  listener on 127.0.0.53 which made the name queries time out before going
  to primary nameserver 192.168.100.1 in the net-vm. Also note that it is
  required to set resolve.enable to false - even when NixOS options document
  the default as false.
* Give static IPs to guests based on their mimicked synthetic MAC
* dhcp-option '6' equals 'option:dns-server'. More descriptive this way.
* Client side queries are configured to use .ghaf domain for local name queries
* Better not set debug subnet ip addresses in dnsmasq, because sometimes we are
  debugging the dnsmasq itself.
* Add tshark network packet capturing tool for builds that include
  the ghaf debug profile

Considerations:

* ghaf-host uses resolv.conf via DNS stub resolver. It does not use
  net-vm (192.168.100.1) as default nameserver. This is considered
  more secure than using net-vm. Future development could be
  debug optional that allows ghaf-host access to local network host
  names. This has risks related to development and testing time
  assumptions on ghaf-host access to names with release builds.
* net-vm does not serve IP address for ghaf-host queries - even
  for debug builds. See rationale above.

## Checklist for things done

## Testing

```
[ghaf@gui-vm:~]$ time for host in net-vm chromium-vm gala-vm gui-vm \
> zathura-vm; do dig +short $host; done;
192.168.100.1
192.168.100.6
192.168.100.4
192.168.100.3
192.168.100.5

real	0m0.151s
user	0m0.029s
sys	0m0.031s
```

### DIFF
--- a/modules/virtualization/microvm/common/dns-dhcp.nix
+++ b/modules/virtualization/microvm/common/dns-dhcp.nix
@@ -1,0 +1,60 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+_: let
+  domain = "ghaf";
+in {
+  # Disable resolved since we are using dnsmasq
+  # This options defaults to false but has been
+  # tested not to work with dnsmasq if not set
+  # explicitly to false
+  services.resolved.enable = false;
+  # do not really use resolvconf for anything
+  networking.resolvconf.enable = false;
+
+  # dnsmasq prioritises /etc/hosts to which NixOS
+  # generates hostname entry. Here we generate
+  # additional hosts that is used to replace
+  # /etc/hosts
+  environment.etc."additional-hosts" = {
+    text = ''
+      127.0.0.1 localhost
+      192.168.100.1 net-vm
+    '';
+    mode = "0444";
+  };
+
+  # Dnsmasq is used as a DHCP/DNS server inside the NetVM
+  services.dnsmasq = {
+    enable = true;
+    settings = {
+      # keep local queries within domain by
+      # caching them within dnsmasq. query outside
+      # only if name is not available locally
+      local = ["/{domain}/192.168.100.1"];
+      server = ["8.8.8.8"];
+      dhcp-range = ["192.168.100.2,192.168.100.254"];
+      dhcp-sequential-ip = true;
+      dhcp-authoritative = true;
+      domain = "${domain}";
+      listen-address = ["127.0.0.1,192.168.100.1"];
+      dhcp-option = [
+        "option:router,192.168.100.1"
+        "option:dns-server,192.168.100.1"
+      ];
+      expand-hosts = true;
+
+      # see comment above with additional-hosts generation
+      no-hosts = true;
+      addn-hosts = "/etc/additional-hosts";
+
+      domain-needed = true;
+      bogus-priv = true;
+
+      # local host names and static IP addresses
+      dhcp-host = [
+        "02:00:00:01:01:01,192.168.100.1,net-vm"
+        "02:00:00:02:02:02,192.168.100.3,gui-vm"
+      ];
+    };
+  };
+}
--- a/modules/virtualization/microvm/common/vm-networking.nix
+++ b/modules/virtualization/microvm/common/vm-networking.nix
@@ -38,15 +38,22 @@ in {
     networks."10-${networkName}" = {
       matchConfig.MACAddress = macAddress;
       DHCP = "yes";
+      dhcpV4Config.UseDomains = "yes";
       linkConfig.RequiredForOnline = "routable";
       linkConfig.ActivationPolicy = "always-up";
     };
   };
 
-  # systemd-resolved does not support local names resolution
-  # without configuring a local domain. With the local domain,
-  # one would need also to disable DNSSEC for the clients.
-  # Disabling DNSSEC for other VM then NetVM is
-  # completely safe since they use NetVM as DNS proxy.
-  services.resolved.dnssec = "false";
+  services.resolved = {
+    # systemd-resolved does not support local names resolution
+    # without configuring a local domain. With the local domain,
+    # one would need also to disable DNSSEC for the clients.
+    # Disabling DNSSEC for other VM then NetVM is
+    # completely safe since they use NetVM as DNS proxy.
+    dnssec = "false";
+    # Disable local DNS stub listener on 127.0.0.53
+    extraConfig = ''
+      DNSStubListener=no
+    '';
+  };
 }
--- a/modules/virtualization/microvm/netvm.nix
+++ b/modules/virtualization/microvm/netvm.nix
@@ -12,6 +12,7 @@
   netvmBaseConfiguration = {
     imports = [
       (import ./common/vm-networking.nix {inherit vmName macAddress;})
+      (import ./common/dns-dhcp.nix {})
       ({lib, ...}: {
         ghaf = {
           users.accounts.enable = lib.mkDefault configHost.ghaf.users.accounts.enable;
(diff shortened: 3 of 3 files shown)


vilvo:
@riskuuse - requesting early review comments to this draft

modules/virtualization/microvm/common/dns-dhcp.nix
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{...}: let
+  domain = "ghaf";
+in {
+  # Dnsmasq is used as a DHCP/DNS server inside the NetVM
+  services.dnsmasq = {
+    enable = true;
+    resolveLocalQueries = true;
```
TanelDettenborn, line 9: A cosmetic change, but default for _[resolveLocalQueries](https://github.com/NixOS/nixpkgs/blob/nixos-23.11/nixos/modules/services/networking/dnsmasq.nix#L58)_ is true.

jpruiz84:
Today I had a routing problem in Ghaf, because my default internet router subnet is 192.168.101.0/24.

```
[ghaf@net-vm:~]$ route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         net-vm          0.0.0.0         UG    1025   0        0 wlp0s4f0
192.168.100.0   0.0.0.0         255.255.255.0   U     0      0        0 ethint0
192.168.101.0   0.0.0.0         255.255.255.0   U     0      0        0 ethint0
192.168.101.0   0.0.0.0         255.255.255.0   U     1025   0        0 wlp0s4f0
net-vm          0.0.0.0         255.255.255.255 UH    1025   0        0 wlp0s4f0
```

The problem was that the net-vm was not able to ping my laptop because the default route in that case was through ethint0 and not to wlp0s4f0. 

The solution in my case was simple, change the subnet in my Internet router to 192.168.110.0. But, many user could have a similar problem if their Internet routers subnet are 192.168.100.0 or 192.168.101.0. I think that the subnet managed by the net-vm should be one less popular, such as 192.168.170.0/

riskuuse:
> @riskuuse - requesting early review comments to this draft One key change here is also (difficult to notice in diff):
> 
> ```
> - domain-needed = true;
> + domain-needed = false;
> ```
> 
> which removes the need to use the domain `ghaf` with internal name queries (e.g. `gui-vm.ghaf`). The primary DNS as internal (`192.168.101.1`) ensures that queries do not leave outside.

My tests on X1 indicates that currently VM's don't resolve to any name, with .ghaf or without. Investigating...

modules/virtualization/microvm/common/dns-dhcp.nix
vilvo: Thanks, addressed in [301b72d](https://github.com/tiiuae/ghaf/pull/427/commits/301b72dfc704f6d22b27b177e418c0ce956589ca)

vilvo:
> > @riskuuse - requesting early review comments to this draft One key change here is also (difficult to notice in diff):
> > ```
> > - domain-needed = true;
> > + domain-needed = false;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which removes the need to use the domain `ghaf` with internal name queries (e.g. `gui-vm.ghaf`). The primary DNS as internal (`192.168.101.1`) ensures that queries do not leave outside.
> 
> My tests on X1 indicates that currently VM's don't resolve to any name, with .ghaf or without. Investigating...

Thanks, I will look into that as well. I added `tshark` to `net-vm` `-debug` builds with 428e2b8 - it might be useful.

vilvo:
> Today I had a routing problem in Ghaf, because my default internet router subnet is 192.168.101.0/24.
> 
> ```
> [ghaf@net-vm:~]$ route
> Kernel IP routing table
> Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
> default         net-vm          0.0.0.0         UG    1025   0        0 wlp0s4f0
> 192.168.100.0   0.0.0.0         255.255.255.0   U     0      0        0 ethint0
> 192.168.101.0   0.0.0.0         255.255.255.0   U     0      0        0 ethint0
> 192.168.101.0   0.0.0.0         255.255.255.0   U     1025   0        0 wlp0s4f0
> net-vm          0.0.0.0         255.255.255.255 UH    1025   0        0 wlp0s4f0
> ```
> 
> The problem was that the net-vm was not able to ping my laptop because the default route in that case was through ethint0 and not to wlp0s4f0.
> 
> The solution in my case was simple, change the subnet in my Internet router to 192.168.110.0. But, many user could have a similar problem if their Internet routers subnet are 192.168.100.0 or 192.168.101.0. I think that the subnet managed by the net-vm should be one less popular, such as 192.168.170.0/

@jpruiz84 This is good suggestion. However, I left the change of the ghaf subnet ip addresses now out of this PR because it can cause some confusion in development and breaks in testing (when they have relied on IP addresses when the internal host name queries did not work). Now that the internal host name queries work, let's give people in development and testing time to adopt the change. Then the IP address range can be changed to something less likely to conflict.

vilvo:
@henrirosten I have not idea why this check fails https://github.com/tiiuae/ghaf/actions/runs/7474453103/job/20340630773?pr=427

leivos-unikie:
Testing:

[ghaf@gui-vm:~]$ time for host in net-vm chromium-vm gala-vm gui-vm zathura-vm; do dig +short $host; done; 
192.168.100.1 
192.168.100.5 
192.168.100.4 
192.168.100.3 
192.168.100.2 

real 0m0.291s 
user 0m0.033s 
sys 0m0.048s

josa41:
Looks good to me.
As others commented ability to change the IP address space would be good.
For further development maybe pass a variable from net-vm.nix to dns-dhcp.nix for the ip address space for the dhcp.
In current implementation both the net-vm ip and dhcp range are static and "magically" happen to match.
Changing the ip address space requires changing multiple files in multiple places.

leivos-unikie:
I have recently drafted scripts for running performance tests in all VMs instead of just ghaf-host.
(https://github.com/leivos-unikie/sysbench-script/blob/main/main_x1.py)

I ran those tests also on this PR and found quite large changes in performance compared to few earlier runs. Here are the 1-thread memory read/write and cpu speed profiles of the system.

memory_1thread_wr_test

memory_1thread_read_test

cpu_1thread_test

Test runs 1-3: themisto #23 build
Test run 4: Build from PR#422 (IDS-VM)
Test run: Build from this PR #427 (refactor: dns+dhcp to a module, support short host queries)

Any idea where this change comes from?

leivos-unikie:
Looked a bit more into this and found that dnsmasq is consuming 99% of the 1 CPU in net-vm.

leivos-unikie:
IMG_2332

leivos-unikie (commented):
dnsmasq is somehow overloading net-vm CPU power (>99%) which causes significant decrease in overall performance.

In other builds net-vm CPU load is < 1% at rest.

vilvo:
> dnsmasq is somehow overloading net-vm CPU power (>99%) which causes significant decrease in overall performance.
> 
> In other builds net-vm CPU load is < 1% at rest.

Without having looked into the script implementation, this is actually excellent preliminary reference data. I had yesterday a discussion on alternative to dhcp+dns - namely static addresses in each guest `/etc/hosts` which we can generate build time. It has other limitations to dhcp+dns but those may be desired from security perspective even with the cost of dynamic guest VMs.

leivos-unikie:
https://ssrc.atlassian.net/browse/SP-3734

vadika:
I checked the PR -- when not connected to the Internet, dnsmasq continuously spoofs the 8.8.8.8 with UDP DNS requests, that causes high CPU load and performance degratation. There is no problem when the NetVM is Internet-connected.

vilvo:
> I checked the PR -- when not connected to the Internet, dnsmasq continuously spoofs the 8.8.8.8 with UDP DNS requests, that causes high CPU load and performance degratation. There is no problem when the NetVM is Internet-connected.

Thanks. That's a constructive contribution. There's either a misconfiguration or [a bug](https://bugs.launchpad.net/ubuntu/+source/dnsmasq/+bug/1980146) or both in this PR. 
The fact is that the local name queries must never leave the `net-vm` dnsmasq to upstream name servers. This must be the case whether we are online or offline.

modules/virtualization/microvm/common/dns-dhcp.nix
```diff
@@ -0,0 +1,59 @@
...
+  };
+
+  # Dnsmasq is used as a DHCP/DNS server inside the NetVM
+  services.dnsmasq = {
+    enable = true;
+    settings = {
+      # keep local queries within domain by
+      # caching them within dnsmasq. query outside
+      # only if name is not available locally
+      server = ["192.168.100.1" "8.8.8.8"];
```
riskuuse, line 33: Here we should list only upstream servers or specify that 192.168.100.1 is local server. E.g. like this
        server = ["8.8.8.8"];
        local = ["/${domain}/192.168.100.1"];
This resolves the CPU load problem. The local -part might be unnecessary.

vilvo: @riskuuse @vadika @leivos-unikie - addressed in [10c35fd](https://github.com/tiiuae/ghaf/commit/10c35fd5c342032ae4649c207e2e97a58d101b14)
Tested with "Testing"-method and no high CPU load observed with network disconnected.

leivos-unikie (approved):
Now performance test results are at normal level.

PR #430, opened 2023-12-22
Add modified element-desktop application to appvm and make start icon on gui-vm

TPiUnikie:
## Description of changes

Element-desktop from nixpkgs does not work on Ghaf. Element-desktop pathed so that application starts on guivm and appvm. Element is modified to allow GPS location sharing.

Element-web includes location sharing modifications so that GPS receiver can can be used for location sharing. An additional element-gps application is added that allows using websockets interface to receive GPS NMEA messages in Element-web implementation instead of Google location API. 
Mode detailed description of GPS implementation:
https://ssrc.atlassian.net/wiki/spaces/SP/pages/948502685/Element-desktop+-+location+sharing+using+GPS+receiver

This commit adds element desktop icon for guivm to start Element-desktop from element virtual machine (element-vm.ghaf).

## Checklist for things done

## Testing

**Guivm**
- Set Wifi ssid and password to enable network access to targets\lenovo-x1-carbon.nix
- build image and run the image on Lenovo X1
- Start the Element-desktop from desktop icon
- Login to element, and share location to someone if GPS transceiver is connected to USB port, otherwise error dialog appears


nesteroff:
Hey Timo, Ilkka asked me too look into this. When you start the element-desktop from the appvm and open the location window it loads a map using webgl which consumes quite a lot of memory. The virtual machine you've created has only 512 MB. The system immediately runs out of memory and the OOM killer kills the electron process with a message like this: `Out of memory: Killed process 748 (electron)`. It should work fine if you increase the ram size in the [vm settings](https://github.com/tiiuae/ghaf/pull/430/files#diff-789892fdf4e4c650d81ce42dbc84586807ced5e73a8d53c06775acf7df2d0876R336).

I also noticed that every time the location window is opened it consumes about 300 MB of ram and it's not being freed. If you open and close it multiple times the system will run out of memory and you'll see the same behavior again even if the vm has several gigabytes of ram. I guess you might need to force garbage collect or something like that.

TPiUnikie:
> Hey Timo, Ilkka asked me too look into this. When you start the element-desktop from the appvm and open the location window it loads a map using webgl which consumes quite a lot of memory. The virtual machine you've created has only 512 MB. The system immediately runs out of memory and the OOM killer kills the electron process with a message like this: `Out of memory: Killed process 748 (electron)`. It should work fine if you increase the ram size in the [vm settings](https://github.com/tiiuae/ghaf/pull/430/files#diff-789892fdf4e4c650d81ce42dbc84586807ced5e73a8d53c06775acf7df2d0876R336).
> 
> I also noticed that every time the location window is opened it consumes about 300 MB of ram and it's not being freed. If you open and close it multiple times the system will run out of memory and you'll see the same behavior again even if the vm has several gigabytes of ram. I guess you might need to force garbage collect or something like that.

Memory consumption goes up to 3500MB based on profiling:
image
VM RAM settings are updated.

leivos-unikie (commented):
Tested these:

1. Apps can be launched, except Gala

2. ci-test-automation suite passed

3. Didn't have USB GPS transceiver available, tested only the case when it's absent:

- Launched Element signed in and selected someone in the people list
- Tried to share location: ... Menu -> Location -> My Current Location / My live location
--> "Could not fetch location / Failed to fetch your location. Please try again later."

- When tried location sharing again the error message did not appear anymore after few trials. World map is just opened with 'Share location' button disabled.

- Closed and opened Element app again. After that error message appears at first 3 attempts when trying to share location, from then on it does not appear.

leivos-unikie:
Element-vm shows significantly poorer performance compared to other VMs. Why?

cpu_1thread_report_from_test_runs_8-14

memory_rd_1thread_report_from_test_runs_8-14

memory_wr_1thread_report_from_test_runs_8-14

leivos-unikie:
py-wrapped_taking_all

This is consuming > 99% of the cpu core assigned for element-vm
`/nix/store/qp5zys77biz7imbk6yy85q5pdv7qk84j-python3-3.11.6/bin/python3.11 /nix/store/vgh4r5z97908h4hc0l3ly0xnbv4zl53z-gpswebsock-1.0/bin/.main.py-wrapped`

That CPU load starts already at boot, not having launched the element app.

TPiUnikie:
> py-wrapped_taking_all
> 
> This is consuming > 99% of the cpu core assigned for element-vm `/nix/store/qp5zys77biz7imbk6yy85q5pdv7qk84j-python3-3.11.6/bin/python3.11 /nix/store/vgh4r5z97908h4hc0l3ly0xnbv4zl53z-gpswebsock-1.0/bin/.main.py-wrapped`
> 
> That CPU load starts already at boot, not having launched the element app.
I was able to replicate the issue. 
Issue is now fixed (main.py). 
The gpsd service does not start if the usb transceiver is not connected.  The gpspipe process will exit with error code because gpsd service is not running and that was not handled properly.

leivos-unikie:
> > py-wrapped_taking_all
> > This is consuming > 99% of the cpu core assigned for element-vm `/nix/store/qp5zys77biz7imbk6yy85q5pdv7qk84j-python3-3.11.6/bin/python3.11 /nix/store/vgh4r5z97908h4hc0l3ly0xnbv4zl53z-gpswebsock-1.0/bin/.main.py-wrapped`
> > That CPU load starts already at boot, not having launched the element app.
> 
> I was able to replicate the issue. Issue is now fixed (main.py). The gpsd service does not start if the usb transceiver is not connected. The gpspipe process will exit with error code because gpsd service is not running and that was not handled properly.

Yes, now the cpu core is not overloading when usb trasceiver is absent.

assets/icons/png/element.png
mikatammi, line 1: Should the Element icon have its own reuse specification? Eg. under what LICENSE is the icon?

TPiUnikie: I need to ask about this what is the current way. The icon is the stock icon provided by element package from nixpkgs. But resize and format change was required to make it appear on GUI vm top bar.

overlays/custom-packages/element-desktop/default.nix
```diff
@@ -0,0 +1,7 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
```
mikatammi, line 1: We are at 2024 now

packages/element-web/default.nix
```diff
@@ -0,0 +1,89 @@
...
+  stdenv,
+  fetchFromGitHub,
+  fetchYarnDeps,
+  jq,
+  yarn,
+  fixup_yarn_lock,
+  nodejs,
+  jitsi-meet,
+}: let
+  pinData = import ./pin.nix;
```
mikatammi, line 14: I wonder why do you need to have this pinData and separate pin.nix, only for the version number? As I see few lines later you remove the hashes part of it.

Maybe it could be just simpler to put `  "version" = "1.11.47";` here and remove the pin.nix

packages/element-web/default.nix
```diff
@@ -0,0 +1,89 @@
...
+  inherit (pinData.hashes) webSrcHash webYarnHash;
+  noPhoningHome = {
+    disable_guests = true; # disable automatic guest account registration at matrix.org
+  };
+in
+  stdenv.mkDerivation (finalAttrs:
+    builtins.removeAttrs pinData ["hashes"]
+    // {
```
mikatammi, line 22: Instead of all this you could just say inside the brackets: `inherit (pinData) version`

assets/icons/png/element.png
mikatammi: I'd say the right way is to put a similarly named file `element.png.LICENSE` in that directory, and put the SPDX information there, if it differs from the LICENSE specified in .reuse/dep5

mikatammi: Okay it looks like element also uses Apache License so I think we're good

targets/lenovo-x1-carbon.nix
```diff
@@ -466,6 +474,58 @@
...
                   }
+                  {
+                    name = "element";
+                    packages = [pkgs.element-desktop pkgs.gpsd pkgs.element-gps];
+                    macAddress = "02:00:00:03:08:01";
+                    ramMb = 4096;
+                    cores = 1;
+                    extraModules = [
+                      {
+                        services.gpsd = {
```
mikatammi, line 485: It feels a bit like hack to passthrough the GPS device and run gpsd in the same Virtual Machine than the element itself is running. Maybe we can let this go for now? But have you investigated the possibility of running gpsd in separate VM? What would it need for communication between the VMs? A Socket?

Maybe some TODO-line would be nice for future.

Also I generally don't like putting more stuff into this file, as we are trying to refactor all this stuff away from these target files.

mikatammi: Also these services are currently running as root, it would be nicer to create gpsd and element-gps users for these services

mikatammi: Here is a very good example for writing a service module: https://github.com/tiiuae/ghaf-caml-crush/blob/main/modules/services/caml-crush/default.nix

mikatammi:
Also as a general note, if people start writing Python-code into ghaf-project, then we should add Python formatter into the treefmt

targets/lenovo-x1-carbon.nix
```diff
@@ -466,6 +474,58 @@
                     ];
                     borderColor = "#337aff";
                   }
+                  {
+                    name = "element";
+                    packages = [pkgs.element-desktop pkgs.gpsd pkgs.element-gps];
+                    macAddress = "02:00:00:03:08:01";
+                    ramMb = 4096;
```
mikatammi, line 481: We are quite soon going to run out of RAM on the Lenovo laptop, if we have to reserve 4GB of RAM just for the sake of displaying map on the webview :D

mikatammi:
Also, why element-web is being included? I don't see it being used anywhere as a dependency. The whole element-web part is basically "dead code".

mikatammi:
Apart from the few minor nitpicks, I want to say the generally the patch look Okayish 👍 . But I'm wondering a bit about the context, why this work has been done, and why some seemingly unneeded dependencies (element-web) would need to be included in the ghaf-project.

I'm again thinking should (part of) this work go to a separate git-repository, and we could then import stuff through the flake-system.

mikatammi:
Well, I think we can take the element-web patches in, just with the reasoning that nixpkgs already has package for element-web, and this would replace it

targets/lenovo-x1-carbon.nix
mikatammi: Actually I misread, the gpsd is fine (https://github.com/NixOS/nixpkgs/blob/nixos-23.11/nixos/modules/services/misc/gpsd.nix#L113-L122), but proper service would be nice for the element-gps below.

TPiUnikie: > Does it make sense to passthrough the GPS device and run gpsd in the same Virtual Machine than the element itself is running. Maybe we can let this go for now? But have you investigated the possibility of running gpsd in separate VM? What would it need for communication between the VMs? A Socket?
> 
> Maybe some TODO-line would be nice for future.
> 
> Also I generally don't like putting more stuff into this file, as we are trying to refactor all this stuff away from these target files.

I was thinking that kind of solution, but VM for each daemon could be also quite heavy solution?

targets/lenovo-x1-carbon.nix
TPiUnikie: That was the proposed solution from weekly meeting.
Application memory consumption was profiled. The app won't even start within 1GB RAM.

TPiUnikie:
> Also, why element-web is being included? I don't see it being used anywhere as a dependency. The whole element-web part is basically "dead code".

Element-web contains the matrix-react-sdk  and is required to modified and build instead of using the default element-web. 
ref: https://github.com/element-hq/element-web?tab=readme-ov-file

targets/lenovo-x1-carbon.nix
mikatammi: Yeah I was already reading from gpsd source code that it can drop privileges if GPSD_USER and GPSD_GROUP are defined build-time: https://gitlab.com/gpsd/gpsd/-/blob/master/gpsd/gpsd.c?ref_type=heads#L2598-2666

I'm wondering does nixpkgs set these properly to "gpsd"

mikatammi: Apparently nixpkgs sets gpsdUser to "gpsd" and gpsdGroup to "dialout" by default:

https://github.com/NixOS/nixpkgs/blob/b0c816c86585ec236c15e668e3d4562d82e183b6/pkgs/servers/gpsd/default.nix#L30

mikatammi: So I guess regarding gpsd we are fine.

packages/element-web/default.nix
TPiUnikie: pin.nix defines the webSrcHash. 
Used in packages/element-web/default.nix to fetch correct version of sources from git to be patched.

overlays/custom-packages/element-desktop/default.nix
TPiUnikie: > We are at 2024 now

True, the modifications were made last year. 
I tought that there is script to update the the copy write headers, instead of manually update every year.

packages/element-gps/default.nix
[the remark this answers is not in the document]
TPiUnikie: Modified as proposed

overlays/custom-packages/element-desktop/default.nix
TPiUnikie: Modified as proposed

overlays/custom-packages/element-gps/default.nix
```diff
@@ -0,0 +1,5 @@
+# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
```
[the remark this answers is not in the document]
TPiUnikie: Modified as proposed

packages/element-web/default.nix
TPiUnikie: No changes here, hash is used from pin.nix file

packages/element-web/default.nix
TPiUnikie: No changes here, hash is used from pin.nix file

targets/lenovo-x1-carbon.nix
TPiUnikie: Ok. No changes here.

brianmcgillion:
please rebase

leivos-unikie:
I think those build checks fails should be first fixed before starting manual regression testing.

barnabakos:
ci-test-automation suite passed;
lenovo x1: do not have GPS transceiver; error dialog appears only a couple of times;
orin agx: do not have GPS transceiver; error dialog does not appear at all;

mikatammi (approved):

mikatammi:
I actually only ran test suites, but didn't check the actual functionality since I don't have the USB GPS device

PR #434, opened 2023-12-30
Orin AGX: Extract partitions while flashing

mikatammi:
## Description of changes

On NVIDIA Jetson Orin AGX, extract ESP and root partitions while running the flashing script, instead of shipping the images twice in the store path. This should make NVIDIA build targets take half the disk space the used to.

## Checklist for things done

  - [X] Tested on Jetson Orin NX or AGX `aarch64`

## Testing

Build and run NVIDIA Orin AGX flash scripts, both native and cross. Changes should not affect NX flash scripts.

### DIFF
--- a/modules/hardware/nvidia-jetson-orin/partition-template.nix
+++ b/modules/hardware/nvidia-jetson-orin/partition-template.nix
@@ -123,20 +123,23 @@ in
           mkdir -pv "$WORKDIR/bootloader"
         ''
         + lib.optionalString (!cfg.flashScriptOverrides.onlyQSPI) ''
-          echo "Decompressing ${images}/esp.img.zst into $WORKDIR/bootloader/esp.img ..."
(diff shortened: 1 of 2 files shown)


vilvo:
Would it actually make more sense to adapt these target specific flashing solutions to #340 ?
I'm not saying #340 is the final "flashing" installation implementation yet but only a "flashing" concept for x86-64 Lenovo X1. Still, it has ingredients to generalize Ghaf installation to all architectures with `nixos-anywhere (kexec)` where this NVIDIA recovery mode flashing is and will be very target specific. If we adopt, and promote, the same target specific solution for all ARM/RISC-V devices, the Ghaf installation will be mixed bag of beans. If we adopt the `nixos-anywhere`-based installation solution, we could do the whole partitioning (target specific declarations) and installation to partitions with more general solutions. That kexec approach could also be extented to support QSPI updates etc. hardware specifics with firmware.

brianmcgillion (approved):

PR #445, opened 2024-01-09
GhA: add CI step to check template integrity

remimimimimi:
Make sure that potential API breaking change will not break `x86_64-target` template.

## Description of changes

Could happen that accidentally, after refactoring some breaking changes introduced, because of lack of tests for templates they may break. This PR fixes this issue by introducing a trivial integration check for `x86_64-target` template. 

We could also expand these checks to other templates. For now, we will focus only on one template, as it is the most common one. 

## Checklist for things done

## Testing

I've used [act](https://github.com/nektos/act) to run Github Action locally with #340. And this PR planned to be ready to merge after #340.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -23,3 +23,16 @@ jobs:
         run: nix develop --command reuse lint
       - name: Check nix flake show runs successfully
         run: nix flake show
+      - name: Check that template is in tact
+        run: |
+          export GHAF_PATH=$(pwd)
+          export TMP_DIR=${{ runner.temp }}/$(tr -dc A-Za-z0-9 </dev/urandom | head -c 13)-ghaf-template-test
+
+          mkdir -p $TMP_DIR && cd $TMP_DIR
+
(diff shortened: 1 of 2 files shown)


henrirosten:
@remimimimimi : purely from github action perspective this change seems fine.
However, I would propose to have someone with more knowledge about the Ghaf templates to comment on if the added check itself also makes sense. Perhaps @vilvo  or @mikatammi ?

.github/workflows/check.yml
```diff
@@ -23,3 +23,16 @@ jobs:
         run: nix develop --command reuse lint
       - name: Check nix flake show runs successfully
         run: nix flake show
+      - name: Check that template is in tact
+        run: |
+          export GHAF_PATH=$(pwd)
+          export TMP_DIR=${{ runner.temp }}/$(tr -dc A-Za-z0-9 </dev/urandom | head -c 13)-ghaf-template-test
```
Mic92, line 29: Maybe just use $(mktemp) here?

Mic92:
Addressed in https://github.com/tiiuae/ghaf/pull/560/

PR #459, opened 2024-01-30
users: creating login and system users

vunnyso:
This patch introduces 2 different login users and
    3 different system users.
    - Login users named as 'operator' and 'ghaf'.
      * 'operator' user profile will have minimal rights.
      * 'ghaf' user profile will have specific capabilities     
          and limited access.
    - System users named as 'network'  and 'waypipe'.
      * 'network' user will manage network connections
      * 'waypipe' user will handle crossvm connections.
       * 'update' user will used for nixos-rebuild switch.
    * Tested on X1 target, System / App VM's can open,
      poweroff/reboot functionality and crossvm
      connections using ssh are working.

## Checklist for things done

## Permissions
| User  | sudo rights (Release Build) | sudo rights (Debug Build) |
| -------- | ------------- |------------- |
| ghaf  |  No | Yes |
| operator  |  No | No |

## Users
| Environment  | Users |
| ------------- | ------------- |
| ghaf-host | ghaf + operator  + update |
| gui-vm  | ghaf + operator   + update |
| net-vm  | ghaf + operator + network + update |
| chromium-vm  | ghaf + operator + waypipe + update  |
| gala-vm | ghaf + operator + waypipe + update |
| zathura-vm | ghaf + operator + waypipe + update  |

## Testing
Please note default `ghaf` user in `gui-vm` is replaced with `operator` user.
Ghaf user will there by default for backward compatibility.  

Refer to below snapshot captured from debug build.
```
$ ssh operator@x.x.x.x 
(operator@x.x.x.x) Password:
Last login: Thu Feb 15 13:52:37 2024 

$ ssh ghaf@x.x.x.x 
(ghaf@x.x.x.x) Password:
Last login: Thu Feb 15 13:56:37 2024 

[ghaf@ghaf-host:~]$ sudo su
[sudo] password for ghaf: 

[root@ghaf-host:/home/ghaf]#
```

## Change in Nixos-rebuild command

nixos-rebuild --flake .#nvidia-jetson-orin-agx-debug --target-host **update**@<ip_address_of_ghaf-host> --fast switch **--use-remote-sudo**
...
nixos-rebuild --flake .#lenovo-x1-carbon-gen11-debug --target-host **update**@<ip_address_of_ghaf-host> --fast switch **--use-remote-sudo**
...
Reference: https://tiiuae.github.io/ghaf/ref_impl/development.html


modules/graphics/window-manager.nix
```diff
@@ -58,8 +58,8 @@ in {
         ExecStart = "${pkgs.systemd}/bin/systemctl --wait --user start ghaf-session.target";
 
         # The user to run the session as. Pick one!
-        User = config.ghaf.users.accounts.user;
-        Group = config.ghaf.users.accounts.user;
+        User = "operator";
```
brianmcgillion, line 61: should still be set in the configuration not hard coded

modules/users/operator.nix
```diff
@@ -0,0 +1,13 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+_: {
+  users.users = {
+    operator = {
```
brianmcgillion, line 5: Should it not still be called Ghaf?

modules/virtualization/microvm/appvm.nix
```diff
@@ -67,6 +65,9 @@
             pkgs.waypipe
           ];
 
+          # Add root user only for debug builds
+          users.users.admin.extraGroups = lib.mkIf config.ghaf.profiles.debug.enable ["wheel"];
```
brianmcgillion, line 69: you already set this in the Groups.nix definitions

modules/virtualization/microvm/guivm.nix
```diff
@@ -30,12 +29,15 @@
           };
         };
 
+        # Add root user only for debug builds
+        users.users.admin.extraGroups = lib.mkIf config.ghaf.profiles.debug.enable ["wheel"];
```
brianmcgillion, line 33: same

modules/virtualization/microvm/netvm.nix
```diff
@@ -35,6 +34,9 @@
           firewall.allowedUDPPorts = [53];
         };
 
+        # Add root user only for debug builds
+        users.users.admin.extraGroups = lib.mkIf config.ghaf.profiles.debug.enable ["wheel"];
```
brianmcgillion, line 38: ditto

docs/src/architecture/adr/user-accounts.md
```diff
@@ -0,0 +1,120 @@
+<!--
+    Copyright 2023 TII (SSRC) and the Ghaf contributors
```
brianmcgillion, line 2: Fill this to confluence first, from there it will get placed correctly into the Ghaf documentation

brianmcgillion (changes requested):

modules/profiles/release.nix
```diff
@@ -18,6 +18,8 @@ in
       # TODO this needs to be refined when we define a policy for the
       # processes and the UID/groups that should be enabled by default
       # if not already covered by systemd
-      ghaf.users.accounts.enable = true;
+
+      users.mutableUsers = false;
+      users.users.root.hashedPassword = "!";
```
mbssrc, line 23: we probably need to keep root even in release until we have another remote update mechanism in place, suggest to comment `users.users.root.hashedPassword = "!";` out for now

modules/virtualization/microvm/appvm.nix
vunnyso: Yes right we set [here](https://github.com/tiiuae/ghaf/pull/459/files#diff-b453010599f021b67ea495c89465afc98bd4679f9e7ad5333890a1dd1ec9cc86R21), the catch is now we have additional condition along with it `lib.mkIf config.ghaf.profiles.debug.enable` due to which admin will not be part of wheel group unless explicitly [added](https://github.com/tiiuae/ghaf/pull/459/files#diff-b7c4d8b1828ec4e55c3b85b348f4c180560550100c49b79a826d4cacfd1afaedR71) in respective app and system vm's.

modules/virtualization/microvm/guivm.nix
vunnyso: same as mentioned [here](https://github.com/tiiuae/ghaf/pull/459#discussion_r1486169821)

modules/virtualization/microvm/netvm.nix
vunnyso: same as mentioned [here](https://github.com/tiiuae/ghaf/pull/459#discussion_r1486169821)

modules/development/ssh.nix
```diff
@@ -31,6 +31,5 @@ in
     config = mkIf cfg.enable {
       services.openssh.enable = true;
       users.users.root.openssh.authorizedKeys.keys = authorizedKeys;
-      users.users.${config.ghaf.users.accounts.user}.openssh.authorizedKeys.keys = authorizedKeys;
```
vilvo, line 34: I understand this is considered controversial by some but just removing it without any documented placeholder mechanism (other than developers add their keys themselves somewhere and rebuild the image) currently breaks testing, debugging and development flows (incl. `nixos-rebuild switch`) for `-debug` image.

This will have a major impact to development productivity when the step must be done manually for every PR to ghaf. And people must remember to remove their hacked keys from their PRs not to cause extra effort for reviewers.

There, I said it.

packages/nm-launcher/default.nix
```diff
@@ -13,7 +13,7 @@
       export DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/ssh_session_dbus.sock
       export DBUS_SYSTEM_BUS_ADDRESS=unix:path=/tmp/ssh_system_dbus.sock
       ${pkgs.openssh}/bin/ssh -M -S /tmp/control_socket \
-          -f -N -q ghaf@192.168.100.1 \
+          -f -N -q network@192.168.100.1 \
```
vilvo, line 16: These inter-VM integrations over SSH may not be long-lived but now that we are changing them, I think we better not hardcode the new system user account "network"  but use the account `e.g. users.users.network.account.user` default value (not yet declared) instead. That allows the username configuration without changing the hardcoded values. These usages, like line above, must use the nix variable instead.

vilvo: The same goes to other hard-coded usernames - `waypipe`

vilvo (changes requested):
Not usable for `-debug` and development needs.

```
 ssh operator@192.168.1.200
Warning: Permanently added '192.168.1.200' (ED25519) to the list of known hosts.
(operator@192.168.1.200) Password:
Last login: Wed Feb 14 13:25:58 2024 from 192.168.101.3

[operator@ghaf-host:~]$ groups
users video operator

[operator@ghaf-host:~]$ login root
login: Cannot possibly work without effective root
```

modules/development/ssh.nix
mbssrc: This needs to be changed to apply to release only, where ssh currently isn't included (and thus functionality does not work) at all. So maybe revert this and leave user keys in for now.

Perhaps it makes sense to e.g. name the "admin" to "ghaf". This will still allow older scripts to run with ghaf user, and potential sudo cmds as it is in wheel.

vilvo: Thanks @mbssrc - appreciate the common sense.

I re-requested comments on practical `-debug`-support replacement mechanism from other Ghaf contributors. So far the only proposals have been to
1. clear the `authorizedKeys` list with pre-defined public keys in the Ghaf `main`-branch and everyone rebuilds the `-debug`-images with their own keys
2. add some `--impure` hack to inject personal public key to `-debug`-builds which is essentially a variant of 1.

I think I concluded the outcome of 1. in my previous comment.

vilvo: I have to test that. I tested only `sudo su` and `login root` as `operator` from within the `-debug` in `ghaf-host`. Both failed. Even that limits the usage quite much.

vunnyso:
> Not usable for `-debug` and development needs.
> 
> ```
>  ssh operator@192.168.1.200
> Warning: Permanently added '192.168.1.200' (ED25519) to the list of known hosts.
> (operator@192.168.1.200) Password:
> Last login: Wed Feb 14 13:25:58 2024 from 192.168.101.3
> 
> [operator@ghaf-host:~]$ groups
> users video operator
> 
> [operator@ghaf-host:~]$ login root
> login: Cannot possibly work without effective root
> ```

Hi @vilvo , here you can login to `admin@ghaf-host` rather than `operator@ghaf-host` and try `sudo su` to get sudo rights then you can execute privileged commands. As `admin` user is part of wheel group under current implementation. 
```
$ ssh admin@x.x.x.x 
Last login: Thu Feb 15 13:56:37 2024 

[admin@ghaf-host:~]$ sudo su
[sudo] password for admin: 

[root@ghaf-host:/home/admin]# cat /etc/group | grep wheel
wheel:x:1:admin

```

modules/development/ssh.nix
vunnyso: Please note for `-debug` builds,  `operator` is not part of wheel group. Try with `admin` user.

vilvo: Right - I think we better document your comments already early on under the Testing section of this PR description as this will have wide impact to automated tests, development practices, etc.- even if we leave the development/debug access public keys in place also for non-root users. What do you think?

vunnyso: I have updated Testing Section. Leaving development/debug access public keys only for `-debug` builds is okay. @mbssrc are you okay with that?

vilvo:
Now that we have installer #340 merged, we could improve the installer to parametrize the `operator` (a role rather than fixed username) so that in the installer you could change the username of the operator. For example, I would like to use `vilvo` instead of `operator`. Can you please check with @remimimimimi on how we could parametrize it?

vunnyso:
> Now that we have installer #340 merged, we could improve the installer to parametrize the `operator` (a role rather than fixed username) so that in the installer you could change the username of the operator. For example, I would like to use `vilvo` instead of `operator`. Can you please check with @remimimimimi on how we could parametrize it?

I had short conversation with @remimimimimi, according to him `It could be done without any modifications to the installer codebase, as new installation approach allows users to fully configure Ghaf system, so as long there's option for configuring something in Ghaf options attribute set user can change it. If you want to make it explicit just add` [this](https://github.com/tiiuae/ghaf/blob/main/templates/targets/x86_64/generic/flake.nix#L84-L89) `line to the template, so it'll be visible for newcomers.`

modules/development/ssh.nix
vunnyso: Added keys for `root` and `operator` users.

packages/nm-launcher/default.nix
vunnyso: Fixed

vunnyso:
Hi @brianmcgillion  /@vilvo / @mbssrc 

Please help to review the patch.

modules/users/operator.nix
vunnyso: Maybe `admin` can be renamed as `ghaf`, so it won't impact much on functionality?

mbssrc:
> Hi @brianmcgillion /@vilvo / @mbssrc
> 
> Please help to review the patch.

I am fine with this for now, given that we 
- rename 'admin' to 'ghaf' for the discussed backwards compatibility
- change from operator to admin user for ssh login
- adjust the ssh-user for power/reboot functionality

modules/microvm/virtualization/microvm/appvm.nix
```diff
@@ -83,6 +85,8 @@
             pkgs.waypipe
             runWaypipe
           ];
+          # Add root user only for debug builds
```
mbssrc, line 88: Is this necessary here - user is already in wheel?

modules/microvm/virtualization/microvm/guivm.nix
```diff
@@ -40,13 +41,15 @@
             withDebug = configHost.ghaf.profiles.debug.enable;
           };
         };
+        # Add root user only for debug builds
```
mbssrc, line 44: same

modules/microvm/virtualization/microvm/netvm.nix
```diff
@@ -42,6 +44,9 @@
           firewall.allowedUDPPorts = [53];
         };
 
+        # Add root user only for debug builds
```
mbssrc, line 47: same

modules/microvm/virtualization/microvm/netvm.nix
```diff
@@ -14,7 +14,9 @@
       (import ./common/vm-networking.nix {inherit vmName macAddress;})
       ({lib, ...}: {
         ghaf = {
-          users.accounts.enable = lib.mkDefault configHost.ghaf.users.accounts.enable;
+          users.operator.account.enable = lib.mkDefault configHost.ghaf.users.operator.account.enable;
```
mbssrc, line 17: Does operator need to be enabled in netvm?

mbssrc:
Briefly tested on x1 weston/labwc. For labwc, the user in wifi-signal-strength should be changed to 'network' as well.

modules/microvm/virtualization/microvm/netvm.nix
vunnyso: Currently yes, as `netvm` have `network` user which is system user and which don't allow ssh login. In case user needs to access `netvm` through ssh that is possible using `operator` user.

vunnyso:
> Briefly tested on x1 weston/labwc. For labwc, the user in wifi-signal-strength should be changed to 'network' as well.

I have fixed wifi-signal-strength issue and pushed changes.

modules/microvm/virtualization/microvm/netvm.nix
vilvo: My 2 cents of scoping the user/ssh hardening here. Moving forward, `operator` should **not** be enabled in the netvm. `operator` must use netvm in a way that won't require ssh login nor ssh remote command but limited use-case based RPC instead. We could take at least two strategies to get there:

**a**. Disable `operator` user login and access to `netvm`. Allow remote commands over ssh for `network` system user (instead of `operator`) but do not allow `network` user login. It could be approached via sshd hardening - see for [example](https://www.digitalocean.com/community/tutorials/how-to-harden-openssh-on-ubuntu-20-04#step-3-restricting-the-shell-of-a-user).
**b**. Keep the current way as an intermediate solution and design/implement RPC between `netvm` and `guivm` to replace the ssh tunneled shell command integrations and remote network-manager gui to configure WIFI etc. It's more design and work but would focus **more** on hardening and getting rid of SSH integrations.

mbssrc: Great comment - this does little for security (yet). The question is indeed how long we keep the SSH approach before moving on. 

One of the ideas of the global "system users" (not in the sense of `isSystemUser = true;` but as in Ghaf-wide) is to have dedicated user(s) for tasks across machines, separated from the Ghaf login user (`operator`). One reason is to prevent this user to login to the VMs (and host, of course). That implies restricting access to the keys, and only allowing the operator access to a trigger (e.g., via sudo/doas), and add additional hardening as you suggested.

vilvo:
@vunnyso - can you please check if 
```
nixos-rebuild --flake .#lenovo-x1-carbon-gen11-debug --target-host root@ip_address switch
```
can be done with `ghaf` user instead of `root`?

vunnyso:
> @vunnyso - can you please check if
> 
> ```
> nixos-rebuild --flake .#lenovo-x1-carbon-gen11-debug --target-host root@ip_address switch
> ```
> 
> can be done with `ghaf` user instead of `root`?

Hi @vilvo I have tried with command `nixos-rebuild --flake .#lenovo-x1-carbon-gen11-debug --target-host`**ghaf**`@ip_address switch`

Got following error, "error: creating symlink from '/nix/var/nix/profiles/.0_system' to 'system-33-link': Permission denied"
Looks like updating in nix-store need root permissions.

mbssrc:
> > @vunnyso - can you please check if
> > ```
> > nixos-rebuild --flake .#lenovo-x1-carbon-gen11-debug --target-host root@ip_address switch
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > can be done with `ghaf` user instead of `root`?
> 
> Hi @vilvo I have tried with command `nixos-rebuild --flake .#lenovo-x1-carbon-gen11-debug --target-host`**ghaf**`@ip_address switch`
> 
> Got following error, "error: creating symlink from '/nix/var/nix/profiles/.0_system' to 'system-33-link': Permission denied" Looks like updating in nix-store need root permissions.

`nixos-rebuild` with the `--use-remote-sudo` is a bit buggy, and most of the solutions I found weren't working, but it works for me with some additional config:
`            security.sudo.extraRules = [
              { users = [ "${config.ghaf.users.ghaf.account.user}" ];
                commands = [
                  {
                    command = "/run/current-system/sw/bin/switch-to-configuration" ;
                    options = [ "NOPASSWD" ];
                  }
                  {
                    command = "/run/current-system/sw/bin/nix" ;
                    options = [ "NOPASSWD" ];
                  }
                  {
                    command = "/run/current-system/sw/bin/nix-env" ;
                    options = [ "NOPASSWD" ];
                  }
                  {
                    command = "/run/current-system/sw/bin/systemd-run" ;
                    options = [ "NOPASSWD" ];
                  }
                ];
              }
            ];
            nix.settings.trusted-users = [
              "${config.ghaf.users.ghaf.account.user}"
            ];
            security.pam.enableSSHAgentAuth = true;`

with that updated, I could run 
`nixos-rebuild --fast --flake .#lenovo-x1-carbon-gen11-debug --target-host ghaf@<host> --use-remote-sudo boot --verbose`
successfully.

So it works - we could create a non-root `update` user with off-device key for the updates. Exact config needs to be figured out, or we can write our own update script, or adjust nixos-rebuild, or use some other available tooling.

vilvo:
> 
…
> The `nixos-rebuild` with the `--use-remote-sudo` is a bit buggy, and most of the solutions I found weren't working, but it works for me with some additional config:
…
> So it works - we could create a non-root `update` user with off-device key for the updates. Exact config needs to be figured out, or we can write our own update script, or adjust nixos-rebuild, or use some other available tooling.

Thanks @mbssrc - that’s excellent.

vunnyso:
> --use-remote-sudo



> > > @vunnyso - can you please check if
> > > ```
> > > nixos-rebuild --flake .#lenovo-x1-carbon-gen11-debug --target-host root@ip_address switch
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > can be done with `ghaf` user instead of `root`?
> > 
> > 
> > Hi @vilvo I have tried with command `nixos-rebuild --flake .#lenovo-x1-carbon-gen11-debug --target-host`**ghaf**`@ip_address switch`
> > Got following error, "error: creating symlink from '/nix/var/nix/profiles/.0_system' to 'system-33-link': Permission denied" Looks like updating in nix-store need root permissions.
> 
> The `nixos-rebuild` with the `--use-remote-sudo` is a bit buggy, and most of the solutions I found weren't working, but it works for me with some additional config:
> 
> ```
> security.sudo.extraRules = [
>   { users = [ "${config.ghaf.users.ghaf.account.user}" ];
>     commands = [
>       {
>         command = "/run/current-system/sw/bin/switch-to-configuration" ;
>         options = [ "NOPASSWD" ];
>       }
>       {
>         command = "/run/current-system/sw/bin/nix" ;
>         options = [ "NOPASSWD" ];
>       }
>       {
>         command = "/run/current-system/sw/bin/nix-env" ;
>         options = [ "NOPASSWD" ];
>       }
>       {
>         command = "/run/current-system/sw/bin/systemd-run" ;
>         options = [ "NOPASSWD" ];
>       }
>     ];
>   }
> ];
> nix.settings.trusted-users = [
>   "${config.ghaf.users.ghaf.account.user}"
> ];
> security.pam.enableSSHAgentAuth = true;
> ```
> 
> With that updated, I could run `nixos-rebuild --fast --flake .#lenovo-x1-carbon-gen11-debug --target-host ghaf@<host> --use-remote-sudo boot --verbose`
> 
> So it works - we could create a non-root `update` user with off-device key for the updates. Exact config needs to be figured out, or we can write our own update script, or adjust nixos-rebuild, or use some other available tooling.

Thanks, Great finding @mbssrc, I think we can create non-root  `update` as system user it will better option rather than relying on user update script or trying to explore new tools. Are you okay with that?

mbssrc:
My recommendation for now is to put this PR on hold until we have a clearer plan and alternative mechanism to SSH in place, to avoid adding more things that ultimately may be removed.

Notes for future:
The separation of `operator` as login user and `admin` (ghaf) administrator goes in the right direction, and can be extended on. The system users separation in current state (with SSH) does not add security, and would require additional work. The `update` user should not be required anywhere except netvm and host, and should be restricted to the actions required for update. This will also be obsolete if an A/B update mechanism is put in place.

UnikieHR:
image

vunnyso:
Abandon, as changes merged into https://github.com/tiiuae/ghaf/pull/827

PR #471, opened 2024-02-08
Add an XDG handler for PDF files to the chromium-vm

nesteroff:
## Description of changes

This is a proof of concept XDG handler for PDF files. 

When a user attempts to open a PDF file from Chromium, it executes the openpdf script. The script sends the PDF file path to the GUIVM over a TCP connection. The GUIVM copies the file from chromium-vm to a temporary location in the zathura-vm and opens it there. Once the viewer is closed, the file is deleted from the zathura-vm.

Sending the file path over a TCP connection with the netcat is obviously not ideal. It should be updated once we have some inter-vm communication mechanism developed. Also, using the scp command might not be needed once we have support for permanent file storage.

## Checklist for things done

## Testing

1. Connect to the internet.
2. Open Chromium browser and download any PDF file.
3. Click on the downloaded PDF and make sure it opens in zathura-vm.
4. Try to open multiple PDF files simultaneously.

### DIFF
--- a/modules/virtualization/microvm/appvm.nix
+++ b/modules/virtualization/microvm/appvm.nix
@@ -27,7 +27,17 @@
           config,
           pkgs,
           ...
-        }: {
+        }: let
+          waypipeBorder =
+            if vm.borderColor != null
+            then "--border \"${vm.borderColor}\""
+            else "";
+          runWaypipe = with pkgs;
+            writeScriptBin "run-waypipe" ''
+              #!${runtimeShell} -e
+              ${pkgs.waypipe}/bin/waypipe --vsock -s ${toString configHost.ghaf.virtualization.microvm.guivm.waypipePort} ${waypipeBorder} server $@
+            '';
+        in {
           ghaf = {
             users.accounts.enable = lib.mkDefault configHost.ghaf.users.accounts.enable;
 
@@ -65,6 +75,7 @@
 
           environment.systemPackages = [
             pkgs.waypipe
+            runWaypipe
           ];
 
           microvm = {
@@ -159,6 +170,13 @@ in {
               '';
               type = int;
               default = 0;
+            };
+            borderColor = mkOption {
+              description = ''
+                Border color of the AppVM window
+              '';
+              type = nullOr str;
+              default = null;
             };
           };
         });
--- a/targets/lenovo-x1-carbon.nix
+++ b/targets/lenovo-x1-carbon.nix
@@ -126,9 +126,10 @@
         time.timeZone = "Asia/Dubai";
       })
     ];
-    guivmConfig = hostConfiguration.config.ghaf.virtualization.microvm.guivm;
     winConfig = hostConfiguration.config.ghaf.windows-launcher;
     networkDevice = hostConfiguration.config.ghaf.hardware.definition.network.pciDevices;
+    # TCP port used by PDF XDG handler
+    xdgPdfPort = 1200;
     guivmExtraModules = [
       {
         # Early KMS needed for GNOME to work inside GuiVM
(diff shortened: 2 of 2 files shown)


targets/lenovo-x1-carbon.nix
```diff
@@ -213,6 +230,31 @@
         ];
 
         time.timeZone = "Asia/Dubai";
+
+        # PDF XDG handler service receives a PDF file path from the chromium-vm and executes the openpdf script
```
josa41, line 234: The open TCP port allows anyone/anything on the system an access to any file in chromium VM.
Anyone can send any chromium-vm "filename" to the guivm tcp port and that file is copied from chromium-vm to zathura-vm and (possibly) opened (not sure how zathura is able to open random files).

targets/lenovo-x1-carbon.nix
```diff
@@ -149,29 +150,45 @@
           "pa,id=pa1,server=unix:/run/pulse/native"
         ];
       }
-      ({pkgs, ...}: {
+      ({pkgs, ...}: let
```
josa41, line 153: With the TCP port open anyone can inject "any" data to the TCP port.
With any data injection of random commands becomes possible.
Basename does not do much sanitation for the file, it only removes the path from file and this might make it posible to inject arbitrary commands at least to the last two "openssh ... <path>" statements
I did not test this is practice but if the send to TCP port 1200 is `"filename\"; echo hello>file2"` there might be "file2 with content "hello" in zathura vm?

josa41 (commented):
Its nice clean code overall, well commented.
I'm a bit worried about the possibility of arbitrary command injections.

modules/virtualization/microvm/appvm.nix
```diff
@@ -27,7 +27,17 @@
           config,
           pkgs,
           ...
-        }: {
+        }: let
```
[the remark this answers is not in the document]
nesteroff: Thanks for reviewing it. You are right, it's always best to have a single logical change per commit. In this particular case I'm not really adding support for the border color. In fact it has been there for quite a while. The problem is that the value was hardcoded in the launchers and I moved it to the appvm settings to avoid hardcoding it in the PDF script as well. That's why it's a part of this PR.

targets/lenovo-x1-carbon.nix
nesteroff: That's correct. As I mentioned, this simple tcp based rpc is just a placeholder until we have a proper inter-vm communication mechanism developed. Also, once we have support for file storage there should be no need to copy files between vms. I guess here I can improve it a little bit by restricting access by remote address in the script and/or in a firewall rule. Any other ideas are welcome.

targets/lenovo-x1-carbon.nix
nesteroff: Well, all parameters are wrapped with quotes in the script so I guess when it gets a malformed file path it should fail to copy the file and stop the execution but I think it's a good idea to test different scenarios here. It's actually quite easy to do: ssh to chromium-vm and execute "openpdf some_file".

josa41: Isn't this openPdf script below at guivm and executed when the url from chromium is sent to port tcp:1200. So easier method to test this part is to send some nicely formatted data to the tcp port from anywhere.
With some malformed content the scp may fail but i think the execution continues to the two ssh remote executors at the end.
I need to try this just out of curiosity,,

nesteroff: Sure, it would be great if you could find time to test it. There are two openpdf scripts. One is in the guivm and the other one is in the chromium-vm which just sends a string over tcp using netcat. It's used by the xdg handler but it can be executed manually. The script should stop because of the '-e' option which means “if any pipeline ever ends with a non-zero ('error') exit status, terminate the script immediately” but I'm not the biggest bash expert to be honest.

leivos-unikie (approved):
- Went through the steps of the **Testing** section. OK. Created also a Zephyr test case of those.
- All apps launch
- ci-test-automation pass

targets/lenovo-x1-carbon.nix
nesteroff: I've updated the script. In this version it checks the source address that is passed from systemd and uses it in the scp command. Only chromium-vm is allowed to open pdf files for now.

josa41 (approved):

brianmcgillion (approved):

PR #473, opened 2024-02-13
Wayland displaying using shared memory between virtual machines

jkuro-tii:
## Description of changes
Add memory sharing mechanism between virtual machines and use it for wayland/waypipe displaying.

### Host

- Added the shared memory manager within QEMU.
- Direct Memory Mapping via SysBus:
  - Modified QEMU's code to map shared memory directly into the VM's memory address space using the SysBus interface.
- Option for Defining Shared Memory Manager:
  - Added an option to QEMU for defining a shared memory manager when running VMs.
- BigPages Area Allocation:
  - Allocated shared memory within the BigPages area for optimized memory usage.
 
### GUI and app VMs

1. **Linux Driver for Shared Memory and Interrupts:**
   - Added a Linux driver to manage shared memory operations and interrupts and creating the `/dev/ivshmem` device.

2. **Memsocket Application for Inter-VM Communication:**
   - Introduced the `memsocket` application to facilitate socket data forwarding between VMs.

3. **Integration with Waypipe:**
   - Modified Waypipe service options to interface with the `memsocket` application, enabling seamless communication between VMs.

## Checklist for things done

- [x] PR linked to architecture documentation and requirement(s) [ticket id](https://ssrc.atlassian.net/browse/SP-3805)

## Testing
Run Chromium and other wayland enabled application. Verify that they execute and display properly.
Run performance tests described in [documentation](https://ssrc.atlassian.net/wiki/spaces/~62552e6ffdb60b006927ad98/pages/825720835/Wayland+displaying+with+shared+memory)

## Jira ticket:
https://ssrc.atlassian.net/browse/SP-3805

## Documentation:
https://ssrc.atlassian.net/wiki/spaces/~62552e6ffdb60b006927ad98/blog/2022/09/29/612958326/Memory+sharing+between+virtual+machines
https://ssrc.atlassian.net/wiki/spaces/~62552e6ffdb60b006927ad98/pages/825720835/Wayland+displaying+with+shared+memory

## Summary of performance
Playing YouTube is smooth, up to a resolution of 2160p. 
CPU consumption by the memsocket application is below 5%, usually 3-6%.
iperf results are presented in [documentation](https://ssrc.atlassian.net/wiki/spaces/~62552e6ffdb60b006927ad98/pages/825720835/Wayland+displaying+with+shared+memory)

### DIFF
--- a/modules/hardware/common/shared-mem.nix
+++ b/modules/hardware/common/shared-mem.nix
@@ -1,0 +1,259 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Module for Shared Memory Definitions
+#
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.ghaf.shm;
+  inherit (lib)
+    foldl'
+    lists
+    mkMerge
+    mkIf
+    mkOption
+    mdDoc
+    types
+    ;
+in
+{
+  options.ghaf.shm = {
+    enable = mkOption {
+      type = types.bool;
+      default = false;
+      description = mdDoc ''
+        Enables shared memory communication between virtual machines (VMs)
+      '';
+    };
+    memSize = mkOption {
+      type = types.int;
+      default = 16;
+      description = mdDoc ''
+        Specifies the size of the shared memory region, measured in 
+        megabytes (MB)
+      '';
+    };
+    hugePageSz = mkOption {
+      type = types.str;
+      default = "2M";
+      description = mdDoc ''
+        Specifies the size of the large memory page area. Supported kernel 
+        values are 2 MB and 1 GB
+      '';
+      apply =
+        value:
+        if value != "2M" && value != "1G" then
+          builtins.throw "Invalid huge memory area page size"
+        else
+          value;
+    };
+    hostSocketPath = mkOption {
+      type = types.path;
+      default = "/tmp/ivshmem_socket"; # The value is hardcoded in the application
+      description = mdDoc ''
+        Specifies the path to the shared memory socket, used by QEMU 
+        instances for inter-VM memory sharing and interrupt signaling
+      '';
+    };
+    flataddr = mkOption {
+      type = types.str;
+      default = "0x920000000";
+      description = mdDoc ''
+        Maps the shared memory to a physical address if set to a non-zero value.
+        The address must be platform-specific and arbitrarily chosen to avoid 
+        conflicts with other memory areas, such as PCI regions.
+      '';
+    };
+    vms_enabled = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      description = mdDoc ''
+        List of vms having access to shared memory
+      '';
+    };
+    enable_host = mkOption {
+      type = types.bool;
+      default = false;
+      description = mdDoc ''
+        Enables the memsocket functionality on the host system
+      '';
+    };
+    instancesCount = mkOption {
+      type = types.int;
+      default =
+        if cfg.enable_host then (builtins.length cfg.vms_enabled) + 1 else builtins.length cfg.vms_enabled;
+      description = mdDoc ''
+        Number of memory slots allocated in the shared memory region
+      '';
+    };
+    serverSocketPath = mkOption {
+      type = types.path;
+      default = "/run/user/${builtins.toString config.ghaf.users.accounts.uid}/memsocket-server.sock";
+      description = mdDoc ''
+        Specifies the path of the listening socket, which is used by Waypipe 
+        or other server applications as the output socket in server mode for 
+        data transmission
+      '';
+    };
+    clientSocketPath = mkOption {
+      type = types.path;
+      default = "/run/user/${builtins.toString config.ghaf.users.accounts.uid}/memsocket-client.sock";
+      description = mdDoc ''
+        Specifies the location of the output socket, which will connected to 
+        in order to receive data from AppVMs. This socket must be created by 
+        another application, such as Waypipe, when operating in client mode
+      '';
+    };
+    display = mkOption {
+      type = types.bool;
+      default = false;
+      description = mdDoc ''
+        Enables the use of shared memory with Waypipe for Wayland-enabled 
+        applications running on virtual machines (VMs), facilitating 
+        efficient inter-VM communication
+      '';
+    };
+  };
+  config =
+    let
+      user = "microvm";
+      group = "kvm";
+    in
+    mkIf cfg.enable (mkMerge [
+      {
+        boot.kernelParams =
+          let
+            hugepages = if cfg.hugePageSz == "2M" then cfg.memSize / 2 else cfg.memSize / 1024;
+          in
+          [
+            "hugepagesz=${cfg.hugePageSz}"
+            "hugepages=${toString hugepages}"
+          ];
+      }
+      {
+        systemd.tmpfiles.rules = [
+          "d /dev/hugepages 0755 ${user} ${group} - -"
+        ];
+      }
+      (mkIf cfg.enable_host {
+        environment.systemPackages = [
+          (pkgs.callPackage ../../../packages/memsocket { vms = cfg.instancesCount; })
+        ];
+      })
+      {
+        systemd.services.ivshmemsrv =
+          let
+            pidFilePath = "/tmp/ivshmem-server.pid";
+            ivShMemSrv =
+              let
+                vectors = toString (2 * cfg.instancesCount);
+              in
+              pkgs.writeShellScriptBin "ivshmemsrv" ''
+                if [ -S ${cfg.hostSocketPath} ]; then
+                  echo Erasing ${cfg.hostSocketPath} ${pidFilePath}
+                  rm -f ${cfg.hostSocketPath}
+                fi
+                ${pkgs.qemu_kvm}/bin/ivshmem-server -p ${pidFilePath} -n ${vectors} -m /dev/hugepages/ -l ${(toString cfg.memSize) + "M"}
+              '';
+          in
+          {
+            enable = true;
+            description = "Start qemu ivshmem memory server";
+            path = [ ivShMemSrv ];
+            wantedBy = [ "multi-user.target" ];
+            serviceConfig = {
+              Type = "oneshot";
+              RemainAfterExit = true;
+              StandardOutput = "journal";
+              StandardError = "journal";
+              ExecStart = "${ivShMemSrv}/bin/ivshmemsrv";
+              User = user;
+              Group = group;
+            };
+          };
+      }
+      {
+        microvm.vms =
+          let
+            memsocket = pkgs.callPackage ../../../packages/memsocket { vms = cfg.instancesCount; };
+            vectors = toString (2 * cfg.instancesCount);
+            makeAssignment = vmName: {
+              ${vmName} = {
+                config = {
+                  config = {
+                    microvm = {
+                      qemu = {
+                        extraArgs = [
+                          "-device"
+                          "ivshmem-doorbell,vectors=${vectors},chardev=ivs_socket,flataddr=${cfg.flataddr}"
+                          "-chardev"
+                          "socket,path=${cfg.hostSocketPath},id=ivs_socket"
+                        ];
+                      };
+                      kernelParams = [ "kvm_ivshmem.flataddr=${cfg.flataddr}" ];
+                    };
+                    boot.extraModulePackages = [
+                      (pkgs.linuxPackages.callPackage ../../../packages/memsocket/module.nix {
+                        inherit (config.microvm.vms.${vmName}.config.config.boot.kernelPackages) kernel;
+                        vmCount = cfg.instancesCount;
+                      })
+                    ];
+                    services = {
+                      udev = {
+                        extraRules = ''
+                          SUBSYSTEM=="misc",KERNEL=="ivshmem",GROUP="kvm",MODE="0666"
+                        '';
+                      };
+                    };
+                    environment.systemPackages = [
+                      memsocket
+                    ];
+                    systemd.user.services.memsocket =
+                      if vmName == "gui-vm" then
+                        {
+                          enable = true;
+                          description = "memsocket";
+                          after = [ "labwc.service" ];
+                          serviceConfig = {
+                            Type = "simple";
+                            ExecStart = "${memsocket}/bin/memsocket -c ${cfg.clientSocketPath}";
+                            Restart = "always";
+                            RestartSec = "1";
+                          };
+                          wantedBy = [ "ghaf-session.target" ];
+                        }
+                      else
+                        # machines connecting to gui-vm
+                        let
+                          vmIndex = lists.findFirstIndex (vm: vm == vmName) null cfg.vms_enabled;
+                        in
+                        {
+                          enable = true;
+                          description = "memsocket";
+                          serviceConfig = {
+                            Type = "simple";
+                            ExecStart = "${memsocket}/bin/memsocket -s ${cfg.serverSocketPath} ${builtins.toString vmIndex}";
+                            Restart = "always";
+                            RestartSec = "1";
+                          };
+                          wantedBy = [ "default.target" ];
+                        };
+                  };
+                };
+              };
+            };
+          in
+          foldl' lib.attrsets.recursiveUpdate { } (map makeAssignment cfg.vms_enabled);
+      }
+      {
+        microvm.vms.gui-vm.config.config.boot.kernelParams = [
+          "kvm_ivshmem.flataddr=${cfg.flataddr}"
+        ];
+      }
+    ]);
+}
--- a/modules/microvm/virtualization/microvm/guivm.nix
+++ b/modules/microvm/virtualization/microvm/guivm.nix
@@ -117,7 +117,7 @@ let
                   # Switch off display, if wayland is running
                   if ${pkgs.procps}/bin/pgrep -fl "wayland" > /dev/null; then
                     wl_running=1
-                    WAYLAND_DISPLAY=/run/user/1000/wayland-0 ${pkgs.wlopm}/bin/wlopm --off '*'
+                    WAYLAND_DISPLAY=/run/user/${builtins.toString config.ghaf.users.accounts.uid}/wayland-0 ${pkgs.wlopm}/bin/wlopm --off '*'
                   else
                     wl_running=0
                   fi
@@ -127,7 +127,7 @@ let
 
                   # Enable display
                   if [ "$wl_running" -eq 1 ]; then
-                    WAYLAND_DISPLAY=/run/user/1000/wayland-0 ${pkgs.wlopm}/bin/wlopm --on '*'
+                    WAYLAND_DISPLAY=/run/user/${builtins.toString config.ghaf.users.accounts.uid}/wayland-0 ${pkgs.wlopm}/bin/wlopm --on '*'
                   fi
                   ;;
                 "button/lid LID open")
--- a/overlays/custom-packages/qemu/0001-ivshmem-flat-memory-support.patch
+++ b/overlays/custom-packages/qemu/0001-ivshmem-flat-memory-support.patch
@@ -1,0 +1,258 @@
+From 13229d9e11eaf15eb6679be036364b9d0256f1c1 Mon Sep 17 00:00:00 2001
+From: Jaroslaw Kurowski <jaroslaw.kurowski@tii.ae>
+Date: Wed, 10 Jul 2024 15:21:07 +0400
+Subject: ivshmem flat memory support
+
+---
+ contrib/ivshmem-server/ivshmem-server.c |  6 +-
+ hw/i386/pc_q35.c                        |  2 +
+ hw/misc/ivshmem.c                       | 74 ++++++++++++++++++++++---
+ include/hw/misc/ivshmem.h               |  1 +
+ 4 files changed, 71 insertions(+), 12 deletions(-)
+
+diff --git a/contrib/ivshmem-server/ivshmem-server.c b/contrib/ivshmem-server/ivshmem-server.c
+index 2f3c732..906c5d0 100644
+--- a/contrib/ivshmem-server/ivshmem-server.c
++++ b/contrib/ivshmem-server/ivshmem-server.c
+@@ -11,6 +11,7 @@
+ 
+ #include <sys/socket.h>
+ #include <sys/un.h>
++#include <sys/mman.h>
+ 
+ #include "ivshmem-server.h"
+ 
+@@ -297,11 +298,10 @@ ivshmem_server_start(IvshmemServer *server)
+                              server->shm_path);
+         shm_fd = shm_open(server->shm_path, O_CREAT | O_RDWR, S_IRWXU);
+     } else {
+-        gchar *filename = g_strdup_printf("%s/ivshmem.XXXXXX", server->shm_path);
++        gchar *filename = g_strdup_printf("%s/ivshmem", server->shm_path);
+         IVSHMEM_SERVER_DEBUG(server, "Using file-backed shared memory: %s\n",
+                              server->shm_path);
+-        shm_fd = mkstemp(filename);
+-        unlink(filename);
++        shm_fd = open(filename, O_RDWR | O_CREAT, 0666);
+         g_free(filename);
+     }
+ 
+diff --git a/hw/i386/pc_q35.c b/hw/i386/pc_q35.c
+index c7bc8a2..d76939e 100644
+--- a/hw/i386/pc_q35.c
++++ b/hw/i386/pc_q35.c
+@@ -28,6 +28,7 @@
+  * THE SOFTWARE.
+  */
+ 
++#include "hw/misc/ivshmem.h"
+ #include "qemu/osdep.h"
+ #include "qemu/units.h"
+ #include "hw/acpi/acpi.h"
+@@ -361,6 +362,7 @@ static void pc_q35_machine_options(MachineClass *m)
+     machine_class_allow_dynamic_sysbus_dev(m, TYPE_INTEL_IOMMU_DEVICE);
+     machine_class_allow_dynamic_sysbus_dev(m, TYPE_RAMFB_DEVICE);
+     machine_class_allow_dynamic_sysbus_dev(m, TYPE_VMBUS_BRIDGE);
++    machine_class_allow_dynamic_sysbus_dev(m, TYPE_IVSHMEM_FLAT);
+     compat_props_add(m->compat_props,
+                      pc_q35_compat_defaults, pc_q35_compat_defaults_len);
+ }
+diff --git a/hw/misc/ivshmem.c b/hw/misc/ivshmem.c
+index de49d1b..b1761b3 100644
+--- a/hw/misc/ivshmem.c
++++ b/hw/misc/ivshmem.c
+@@ -36,6 +36,7 @@
+ #include "chardev/char-fe.h"
+ #include "sysemu/hostmem.h"
+ #include "qapi/visitor.h"
++#include "hw/sysbus.h"
+ 
+ #include "hw/misc/ivshmem.h"
+ #include "qom/object.h"
+@@ -59,6 +60,7 @@
+ 
+ #define TYPE_IVSHMEM_COMMON "ivshmem-common"
+ typedef struct IVShmemState IVShmemState;
++typedef struct IvshmemFTState IvshmemFTState;
+ DECLARE_INSTANCE_CHECKER(IVShmemState, IVSHMEM_COMMON,
+                          TYPE_IVSHMEM_COMMON)
+ 
+@@ -74,6 +76,9 @@ DECLARE_INSTANCE_CHECKER(IVShmemState, IVSHMEM_DOORBELL,
+ DECLARE_INSTANCE_CHECKER(IVShmemState, IVSHMEM,
+                          TYPE_IVSHMEM)
+ 
++#define TYPE_IVSHMEM_FLAT "ivshmem-flat"
++DECLARE_INSTANCE_CHECKER(IvshmemFTState, IVSHMEM_FLAT, TYPE_IVSHMEM_FLAT)
++
+ typedef struct Peer {
+     int nb_eventfds;
+     EventNotifier *eventfds;
+@@ -117,6 +122,15 @@ struct IVShmemState {
+     /* migration stuff */
+     OnOffAuto master;
+     Error *migration_blocker;
++
++    /* flat memory stuff */
++    uint64_t flataddr;
++    DeviceState *flat_dev;
++    MemoryRegion flat_mem;
++};
++
++struct IvshmemFTState {
++    SysBusDevice parent_obj;
+ };
+ 
+ /* registers for the Inter-VM shared memory device */
+@@ -476,8 +490,11 @@ static void setup_interrupt(IVShmemState *s, int vector, Error **errp)
+ 
+ static void process_msg_shmem(IVShmemState *s, int fd, Error **errp)
+ {
++    Error *local_err = NULL;
+     struct stat buf;
+     size_t size;
++    void *ptr;
++    SysBusDevice *sbd;
+ 
+     if (s->ivshmem_bar2) {
+         error_setg(errp, "server sent unexpected shared memory message");
+@@ -494,13 +511,26 @@ static void process_msg_shmem(IVShmemState *s, int fd, Error **errp)
+ 
+     size = buf.st_size;
+ 
++    if (s->flataddr) {
++
++        ptr = mmap(0, size, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_HUGETLB|MAP_LOCKED,
++                fd, 0);
++        s->flat_dev = sysbus_create_simple(TYPE_IVSHMEM_FLAT, -1, 0);
++
++        memory_region_init_ram_ptr(&s->flat_mem, OBJECT(IVSHMEM_FLAT(s->flat_dev)),
++                                   "ivshmem.flat", size, ptr);
++        sbd = SYS_BUS_DEVICE(s->flat_dev);
++        sysbus_init_mmio(sbd, &s->flat_mem);
++        sysbus_mmio_map(sbd, 0, s->flataddr);
++    }
++
+     /* mmap the region and map into the BAR2 */
+-    if (!memory_region_init_ram_from_fd(&s->server_bar2, OBJECT(s),
+-                                        "ivshmem.bar2", size, RAM_SHARED,
+-                                        fd, 0, errp)) {
++    memory_region_init_ram_from_fd(&s->server_bar2, OBJECT(s), "ivshmem.bar2",
++                                   size, RAM_SHARED, fd, 0, &local_err);
++    if (local_err) {
++        error_propagate(errp, local_err);
+         return;
+     }
+-
+     s->ivshmem_bar2 = &s->server_bar2;
+ }
+ 
+@@ -832,7 +862,6 @@ static void ivshmem_write_config(PCIDevice *pdev, uint32_t address,
+ 
+ static void ivshmem_common_realize(PCIDevice *dev, Error **errp)
+ {
+-    ERRP_GUARD();
+     IVShmemState *s = IVSHMEM_COMMON(dev);
+     Error *err = NULL;
+     uint8_t *pci_conf;
+@@ -902,7 +931,8 @@ static void ivshmem_common_realize(PCIDevice *dev, Error **errp)
+     if (!ivshmem_is_master(s)) {
+         error_setg(&s->migration_blocker,
+                    "Migration is disabled when using feature 'peer mode' in device 'ivshmem'");
+-        if (migrate_add_blocker(&s->migration_blocker, errp) < 0) {
++        if (migrate_add_blocker(s->migration_blocker, errp) < 0) {
++            error_free(s->migration_blocker);
+             return;
+         }
+     }
+@@ -920,7 +950,10 @@ static void ivshmem_exit(PCIDevice *dev)
+     IVShmemState *s = IVSHMEM_COMMON(dev);
+     int i;
+ 
+-    migrate_del_blocker(&s->migration_blocker);
++    if (s->migration_blocker) {
++        migrate_del_blocker(s->migration_blocker);
++        error_free(s->migration_blocker);
++    }
+ 
+     if (memory_region_is_mapped(s->ivshmem_bar2)) {
+         if (!s->hostmem) {
+@@ -1014,7 +1047,7 @@ static const VMStateDescription ivshmem_plain_vmsd = {
+     .minimum_version_id = 0,
+     .pre_load = ivshmem_pre_load,
+     .post_load = ivshmem_post_load,
+-    .fields = (const VMStateField[]) {
++    .fields = (VMStateField[]) {
+         VMSTATE_PCI_DEVICE(parent_obj, IVShmemState),
+         VMSTATE_UINT32(intrstatus, IVShmemState),
+         VMSTATE_UINT32(intrmask, IVShmemState),
+@@ -1068,7 +1101,7 @@ static const VMStateDescription ivshmem_doorbell_vmsd = {
+     .minimum_version_id = 0,
+     .pre_load = ivshmem_pre_load,
+     .post_load = ivshmem_post_load,
+-    .fields = (const VMStateField[]) {
++    .fields = (VMStateField[]) {
+         VMSTATE_PCI_DEVICE(parent_obj, IVShmemState),
+         VMSTATE_MSIX(parent_obj, IVShmemState),
+         VMSTATE_UINT32(intrstatus, IVShmemState),
+@@ -1083,6 +1116,7 @@ static Property ivshmem_doorbell_properties[] = {
+     DEFINE_PROP_BIT("ioeventfd", IVShmemState, features, IVSHMEM_IOEVENTFD,
+                     true),
+     DEFINE_PROP_ON_OFF_AUTO("master", IVShmemState, master, ON_OFF_AUTO_OFF),
++    DEFINE_PROP_UINT64("flataddr", IVShmemState, flataddr, 0),
+     DEFINE_PROP_END_OF_LIST(),
+ };
+ 
+@@ -1115,6 +1149,20 @@ static void ivshmem_doorbell_class_init(ObjectClass *klass, void *data)
+     dc->vmsd = &ivshmem_doorbell_vmsd;
+ }
+ 
++static Property ivshmem_flat_props[] = {
++    DEFINE_PROP_END_OF_LIST(),
++};
++
++static void ivshmem_flat_class_init(ObjectClass *klass, void *data)
++{
++    DeviceClass *dc = DEVICE_CLASS(klass);
++
++    dc->hotpluggable = true;
++    set_bit(DEVICE_CATEGORY_MISC, dc->categories);
++    device_class_set_props(dc, ivshmem_flat_props);
++    dc->user_creatable = false;
++}
++
+ static const TypeInfo ivshmem_doorbell_info = {
+     .name          = TYPE_IVSHMEM_DOORBELL,
+     .parent        = TYPE_IVSHMEM_COMMON,
+@@ -1123,11 +1171,19 @@ static const TypeInfo ivshmem_doorbell_info = {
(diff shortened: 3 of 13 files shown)


jenninikko:
Are you planning to add documentation to the repository?

vilvo (changes requested):
1. Please make PR description according to the template - https://github.com/tiiuae/ghaf/blob/main/.github/pull_request_template.md 
2. Please fill the PR template checklist, describe testing and remove "Draft" status.
3. Re-request review

jkuro-tii:
The purpose of this PR is to share the code and try together to solve the mentioned problem of high kernel overhead.

jkuro-tii:
Getting pert data. perf is a standard Linux tool (https://perf.wiki.kernel.org/index.php/Main_Page). In this PR the ``perf`` component is included in the host, GuiVM and AppVM VMs. The first stage of using ``perf`` is collecting execution data (e.g. ``sudo perf record -a -g -p PID``) for certain period of time. The data is written into the ``perf.data`` file. Then the data can be visualized using the ``sudo perf report`` subcommand. ``perf report`` uses ``/proc/kallsyms`` for obtaining information about kernel symbols. If you want to process the data offline, copy ``/proc/kallsyms`` and refer to it with the ``perf record`` ``--kallsyms=<file>`` option. The files that I attached contain both ``perf.data`` and kernel symbols files.
 
To check the memsocket application performace on GuiVM:

1. Find the PID of the memsocket service: ``systemctl status --user memsocket.service | grep "Main PID"``
2. Collect ``perf`` data: ``sudo perf record -a -g -p PID``
3. Process it with ``perf report``.

To check GuiVM's qemu performance on the host:
1. Start terminal
2. Login to host: ``ssh 192.168.101.2``
3. Find PID of qemu instance running GuiVM: ``systemctl status microvm@gui-vm.service | grep "Main PID"``
4. collect perf data: ``sudo perf record -a -g -p PID``
5. Process it with ``perf report``.

All files kept in the ``/home/ghaf`` directory on the host are stored on the Ghaf boot disk. It may be used for accessing them on another computer. E.g. files from the GuiVM machine can be copied to host using ``scp _file_ ghaf@192.168.101.2:`` and accessed offline.
Sample reading about ``perf``:
[1](https://linux-kernel-labs.github.io/refs/heads/master/labs/kernel_profiling.html), [2](https://perf.wiki.kernel.org/index.php/Tutorial)

vilvo:
> To check the memsocket application performace on GuiVM

Thanks. I can reproduce your `perf` measurement and `perf report` analysis in the Youtube playing scenario. In addition, the user experience is terrible.

[Behind the link of the description](https://ssrc.atlassian.net/wiki/spaces/~62552e6ffdb60b006927ad98/pages/825720835/Wayland+displaying+with+shared+memory), there an interesting mention on 
```
Result
Throughput socket to socket: 
2.8 GB/s = 22.4 Gb/s
using ssh: 140 MB/s = 1.1 Gb/s

Lenovo Thinkpad P1 Intel i7 @2.5 GHz

Sending 1 MiB buffer 8 K times (total 80 GiB sent)
``` 

Can you please share reproducible `iperf` test description? I'd like to reproduce and establish Lenovo X1 Carbon Gen11 baseline throughput. Another thing we could then check is to parametrize the buffer size and find where we hit the ceiling.

jkuro-tii:
I did iperf measuring, with the unsock tool - in order to tunnel TCP traffic over socket and shared memory. 

Test configuration:
_chromium-vm_: `iperf -c` with `unsock` -> [socket] -> `memsocket`-> 
[shared memory]
 _guivm_: `memsocket` -> [socket] -> `iperf -s` with `unsock`

Performance achieved:
Lenovo X1 with Ghaf: ~90 Mbits/sec
Lenovo P1 with Fedora Linux and two Ghaf vm-debug machines (my development environment): ~32 **Gbit**/s


[result.client_x1.log](https://github.com/tiiuae/ghaf/files/14308081/result.client_x1.log)
[result.server_x1.log](https://github.com/tiiuae/ghaf/files/14308082/result.server_x1.log)
[result.server.log](https://github.com/tiiuae/ghaf/files/14308083/result.server.log)
[result.client.log](https://github.com/tiiuae/ghaf/files/14308084/result.client.log)

vilvo:
> I did iperf measuring, with the unsock tool - in order to tunnel TCP traffic over socket and shared memory.
> 
> Test configuration: _chromium-vm_: `iperf -c` with `unsock` -> [socket] -> `memsocket`-> [shared memory] _guivm_: `memsocket` -> [socket] -> `iperf -s` with `unsock`
> 
> Performance achieved: Lenovo X1 with Ghaf: ~90 Mbits/sec Lenovo P1 with Fedora Linux and two Ghaf vm-debug machines (my development environment): ~32 **Gbit**/s
> 
> [result.client_x1.log](https://github.com/tiiuae/ghaf/files/14308081/result.client_x1.log) [result.server_x1.log](https://github.com/tiiuae/ghaf/files/14308082/result.server_x1.log) [result.server.log]

>(https://github.com/tiiuae/ghaf/files/14308083/result.server.log) [result.client.log](https://github.com/tiiuae/ghaf/files/14308084/result.client.log)

```
[ghaf@ghaf-host:~/shmsockproxy/app]$ UNSOCK_ADDR=127.0.0.1 UNSOCK_DIR=`pwd` LD_P
RELOAD=~/unsock/libunsock.so.1.1.0  iperf -c 127.0.0.1 -p 1234
```
This indicates you have compiled some dependency in the current working directory (`pwd`?) of shmsockproxy app to use some version(which git commit hash?) of `libunsock` with iperf. From above instructions it's unclear on how to replicate the test setup as I rather not do guessing.

Optionally the test tools you have used could be nix packaged and included in the PR `-debug` build.

Would it make sense to add some simple unit throughput test through memsocket instead? What do you think?

vilvo:
If I try a bit further, to compile the `memsocket` on `ghaf-host` I get a lot of gcc warnings on `-Wformat-extra-args` and `-Wformat-overflow=` (just a small snippet below).

```
[ghaf@ghaf-host:~]$ nix-shell -p git stdenv

[nix-shell:~]$ cd shmsockproxy/app/

[nix-shell:~/shmsockproxy/app]$ gcc memsocket.c
memsocket.c: In function ‘server_init’:
memsocket.c:175:8: warning: too many arguments for format [-Wformat-extra-args]
  175 |   INFO("server initialized", "");
...
memsocket.c: In function ‘wayland_connect’:
memsocket.c:50:19: warning: ‘%s’ directive writing up to 255 bytes into a region of size 235 [-Wformat-overflow=]
...
```

vilvo:
Also, I think this is relevant to performance comparison between X1 and desktop - https://github.com/tiiuae/ghaf/pull/340#pullrequestreview-1884426193
So we can now use Ghaf internal NVMe instead of external SSD over USB3.

jkuro-tii:
chromium -> [socket] -> waypipe  -> [socket] -> memsocket -> 

[shared memory]

-> memsocket send() -> [socket] -> waypipe -> wayland

> If I try a bit further, to compile the `memsocket` on `ghaf-host` I get a lot of gcc warnings on `-Wformat-extra-args` and `-Wformat-overflow=` (just a small snippet below).
> 
> ```
> [ghaf@ghaf-host:~]$ nix-shell -p git stdenv
> 
> [nix-shell:~]$ cd shmsockproxy/app/
> 
> [nix-shell:~/shmsockproxy/app]$ gcc memsocket.c
> memsocket.c: In function ‘server_init’:
> memsocket.c:175:8: warning: too many arguments for format [-Wformat-extra-args]
>   175 |   INFO("server initialized", "");
> ...
> memsocket.c: In function ‘wayland_connect’:
> memsocket.c:50:19: warning: ‘%s’ directive writing up to 255 bytes into a region of size 235 [-Wformat-overflow=]
> ...
> ```
> 
> Edit. `git` was used to clone:
> 
> ```
> [nix-shell:~]$ git clone https://github.com/tiiuae/shmsockproxy.git
> [nix-shell:~]$ cd shmsockproxy/
> [nix-shell:~/shmsockproxy]$ git log --oneline -1
> 8b1bf60 (HEAD -> main, origin/main, origin/HEAD) Removed forking, as systemd sends SIGTERM
> ```

I added binary library and a few commands by hand into chromium and gui VMs. Is it OK it I attach them here, or you want to include them in the Ghaf build?

vilvo:
> I added binary library and a few commands by hand into chromium and gui VMs. Is it OK it I attach them here, or you want to include them in the Ghaf build?

This command:
```
[ghaf@ghaf-host:~/shmsockproxy/app]$ UNSOCK_ADDR=127.0.0.1 UNSOCK_DIR=`pwd` LD_P
RELOAD=~/unsock/libunsock.so.1.1.0  iperf -c 127.0.0.1 -p 1234
```
from [here](https://github.com/tiiuae/ghaf/files/14308084/result.client.log) indicates that some commands are also required on `ghaf-host` instead of only `chromium-vm` and `gui-vm` _if one wants to reproduce the iperf throughput measurements._ Now I'm quite not sure if that's the case or not. So I still have some trouble following the testing procedure and if we are talking about the integrated YouTube-streaming case or the iperf measurement case. So it's probably more clear if we add reproducible test setup in the Ghaf build for `-debug`-target and clarify instructions.

jkuro-tii:
It doesn't seem we are facing data loss. Connection between remote and waypipe instances is multichannel, data is sent both directions. After portion of data sent from AppVM, GuiVM sends a short package of data, probably it's confirmation. If its not received by AppVM, there is no further transmission. Also if we faced data loss, GUI stuff - windows, menu, scroll bars would be damaged. We don't face it.

I've pushed my branch rebased with main.

vilvo:
> I've pushed my branch rebased with main.

Thanks, I installed your rebased draft branch build to X1 NVMe with:
```
nix build .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug-installer
dd if=result/iso/nixos-23.11.20231218.3a9928d-x86_64-linux.iso ...
<boot and install to nvme0n1>
```

The problem with `dd` method is the numbers tell very little. Writing stops with
```
[ghaf@gui-vm:~]$ time dd if=/dev/zero of=/dev/ivshmem
dd: writing to '/dev/ivshmem': No space left on device
32769+0 records in
32768+0 records out
16777216 bytes (17 MB, 16 MiB) copied, 0.688331 s, 24.4 MB/s

real	0m0.692s
user	0m0.000s
sys	0m0.690s
```

vilvo:
During the Youtube streaming scenario there's very heavy load in QEMU processes as seen from `htop` on `ghaf-host`.
The message signaled interrupts on `chromium-vm` and `gui-vm` are roughly ~50-60/s as seen with `watch -n1 "cat /proc/interrupts"`
qemu-load-with-youtube-streaming
When you stop the streaming, there's no  QEMU load.

vilvo:
We're not using `aio=native` with QEMU. Please check
https://hex.ro/wp/blog/kvm-qemu-high-cpu-load-but-low-cpu-usage/
I checked the `strace` during streaming and we've got the same - littered with `events=POLLIN` and a ton of `anon_inode:[eventfd]`s under the QEMU `/proc/<pid>/fd`

vilvo:
And this presentation in the blog post link of the previous comment is extremely relevant for optimizing our case - https://vmsplice.net/~stefan/stefanha-kvm-forum-2017.pdf

jkuro-tii:
The ``aio`` parameter is not accepted for device that is used for shared memory and interrupts:
``                          "-device"
                          "ivshmem-doorbell,vectors=${vectors},chardev=ivs_socket,aio=native,cache=none"
                          "-chardev"
                          "socket,path=${config.ghaf.profiles.applications.ivShMemServer.hostSocketPath},id=ivs_socket"
``
:
``qemu-kvm: -device ivshmem-doorbell,vectors=10,chardev=id,aio=native: Property 'ivshmem-doorbell.aio' not found``

It seems that it's designated for disks/block devices. I'll dig deeper.

vilvo:
> The `aio` parameter is not accepted for device that is used for shared memory and interrupts: ` "-device" "ivshmem-doorbell,vectors=${vectors},chardev=ivs_socket,aio=native,cache=none" "-chardev" "socket,path=${config.ghaf.profiles.applications.ivShMemServer.hostSocketPath},id=ivs_socket"` : `qemu-kvm: -device ivshmem-doorbell,vectors=10,chardev=id,aio=native: Property 'ivshmem-doorbell.aio' not found`
> 
> It seems that it's designated for drives/block devices. I'll dig deeper.

aio=native was fairly new to QEMU and I think must be implemented in the device. Our ivmshmem device is based on fairly old implementation https://github.com/tiiuae/shmsockproxy/blob/main/drivers/char/ivshmem/kvm_ivshmem.c#L8
Is this something to consider?

jkuro-tii:
The kernel driver uses regular PCI interface. I think the matter is the counterpart implementation in [qemu](https://github.com/qemu/qemu/blob/master/hw/misc/ivshmem.c). And we still don't know why in development environment it runs without overhead.

vilvo:
> The kernel driver uses regular PCI interface. I think the matter is the counterpart implementation in [qemu]

Could be only qemu. Or could be both. Regarding the actual driver the differences between X1 and your desktop can be the MMU and how MSIs are handled. Please correct me if I'm wrong but my understanding is that the driver is only using logical virtio-PCI-device model, not actual PCI bus but shared memory instead. On X1 from `chromium-vm` shared memory PCI driver perspective it looks like:
```
00:07.0 RAM memory: Red Hat, Inc. Inter-VM shared memory (rev 01)
	Subsystem: Red Hat, Inc. QEMU Virtual Machine
	Control: I/O+ Mem+ BusMaster+ SpecCycle- MemWINV- VGASnoop- ParErr- Stepping- SERR+ FastB2B- DisINTx+
	Status: Cap+ 66MHz- UDF- FastB2B- ParErr- DEVSEL=fast >TAbort- <TAbort- <MAbort- >SERR- <PERR- INTx-
	Latency: 0
	Region 0: Memory at febfc000 (32-bit, non-prefetchable) [size=256]
	Region 1: Memory at febfd000 (32-bit, non-prefetchable) [size=4K]
	Region 2: Memory at fd000000 (64-bit, prefetchable) [size=16M]
	Capabilities: [40] MSI-X: Enable+ Count=12 Masked-
		Vector table: BAR=1 offset=00000000
		PBA: BAR=1 offset=00000800
	Kernel driver in use: kvm-shmem
	Kernel modules: virtio_pci
```

> And we still don't know why in Chromium-VM/Zathura-VM/dev environment there is no overhead.

We don't know the measured overhead in your dev environment. If your desktop CPU running the guest is not throttling, it may have observable impact but still have overhead. We also don't know the QEMU version. Please share.

jkuro-tii:
I designed an utility for testing connection.
On one side (e.g. chromium-vm) it sends a file to a socket, on the other side (e.g. gui-vm) it receives it, without any writing. The transmission is protected by CRC-8.
Both sockets are connected with the memsocket app.
It's possible to specify how many data to send, and to force send broken CRC (by adding anything as 4th parameter), e.g.:

``memtest socket_path /dev/random 10M``
``memtest socket_path /dev/random 10M anything # for sending broken CRC``

First start it on the receiving site (any of the VMs):
``memsocket -c ./test.socket & # run it once``
``memtest ./test.socket``

Now file(s) can be send from another VM:
``memsocket -s ./test.socket 2 & # run it once``

``memtest ./test.socket /dev/random 10M``
``memtest ./test.socket /dev/random 10M wrong_crc``

It's prints some statistics.
**Before using it it's necessary to stop the memsocket service: ``systemctl stop memsocket.service`` on the gui-vm and the other peer VM (e.g. chromium)!**

Sample sessions:

[ghaf@chromium-vm:~]$ memtest
Usage: memtest socket_path for receiving.
memtest socket_path input_file [size CRC_error] for sending.
Before using stop the memsocket service: 'systemctl stop --user memsocket.service'.
For receiving run e.g.: 'memsocket -c ./test.sock &; memtest ./test.sock'.
To send a file run on other VM: 'memsocket -s ./test.sock 3 &; memtest ./test.sock /dev/random 10M.'
To force wrong CRC on sending: 'memtest ./test.sock /dev/random 10M xxx
'
[ghaf@chromium-vm:~]$ systemctl stop --user memsocket.service

[ghaf@chromium-vm:~]$ memsocket -s ./test.sock 3 &
[1] 776

[ghaf@chromium-vm:~]$ memtest ./test.sock /dev/random 10M
Sent/received 10485761 bytes 2 Mbytes/sec
real 5.030s
user 0.010s
sys  0.030s

[ghaf@chromium-vm:~]$ memtest ./test.sock /dev/random 100M
Sent/received 104857601 bytes 4 Mbytes/sec
real 27.640s
user 0.250s
sys  0.320s
[ghaf@chromium-vm:~]$ memtest ./test.sock /dev/random 100M
Sent/received 104857601 bytes 244 Mbytes/sec
real 0.410s
user 0.200s
sys  0.200s

[ghaf@chromium-vm:~]$ memtest ./test.sock /dev/random 100M ui **# to brake the CRC**
Sent/received 104857601 bytes 263 Mbytes/sec
real 0.380s
user 0.170s
sys  0.200s

[ghaf@chromium-vm:~]$ memtest ./test.sock /dev/random 100M 
Sent/received 104857601 bytes 263 Mbytes/sec
real 0.380s
user 0.190s
sys  0.180s




--------- 
Usage: memtest socket_path for receiving.
memtest socket_path input_file [size CRC_error] for sending.
Before using stop the memsocket service: 'systemctl stop --user memsocket.service'.
For receiving run e.g.: 'memsocket -c ./test.sock &; memtest ./test.sock'.
To send a file run on other VM: 'memsocket -s ./test.sock 3 &; memtest ./test.sock /dev/random 10M.'
To force wrong CRC on sending: 'memtest ./test.sock /dev/random 10M xxx

[ghaf@zathura-vm:~]$ systemctl stop --user memsocket.service

[ghaf@zathura-vm:~]$ memsocket -c ./test.sock &
[1] 670

[ghaf@zathura-vm:~]$ memtest ./test.sock
Waiting for a connection.
Connected...
Sent/received 104857602 bytes 159 Mbytes/sec
real 0.630s
user 0.270s
sys  0.030s
Read 104857601 bytes. CRC OK
Connected...
Sent/received 104857602 bytes 244 Mbytes/sec
real 0.410s
user 0.200s
sys  0.010s
Read 104857601 bytes. CRC OK
Connected...
Sent/received 104857602 bytes 263 Mbytes/sec
real 0.380s
user 0.180s
sys  0.020s
Read 104857601 bytes. TRAMISSION ERROR!
Connected...
Sent/received 104857602 bytes 263 Mbytes/sec
real 0.380s
user 0.190s
sys  0.010s
Read 104857601 bytes. CRC OK
``

jkuro-tii:
I noticed that sending data between gui-vm and any other machine (zathura-vm, chromium-vm) is significantly slower then sending between app VMs (e.g. zathura-vm and chromium-vm).

On my desktop I use qemu version 8.0.5. The full run command line (it's a Ghaf vm-debug build):
``exec /nix/store/ppfkk3h94b3s3vhifp55q80kr5y8bzvc-qemu-host-cpu-only-8.0.5/bin/qemu-kvm -cpu max \
    -name ghaf-host \
    -m 1024 \
    -smp 1 \
    -device virtio-rng-pci \
    -net nic,netdev=user.0,model=virtio -netdev user,id=user.0,"$QEMU_NET_OPTS" \
    -virtfs local,path=/nix/store,security_model=none,mount_tag=nix-store \
    -virtfs local,path="${SHARED_DIR:-$TMPDIR/xchg}",security_model=none,mount_tag=shared \
    -virtfs local,path="$TMPDIR"/xchg,security_model=none,mount_tag=xchg \
    -drive cache=writeback,file="$NIX_DISK_IMAGE",id=drive1,if=none,index=1,werror=report -device virtio-blk-pci,bootindex=1,drive=drive1 \
    -device virtio-keyboard \
    -usb \
    -device usb-tablet,bus=usb-bus.0 \
    -kernel ${NIXPKGS_QEMU_KERNEL_ghaf_host:-/nix/store/sgv6vmzrbkghz8a0hr05garms07cqz2m-nixos-system-ghaf-host-23.05.20230927.5cfafa1/kernel} \
    -initrd /nix/store/sgv6vmzrbkghz8a0hr05garms07cqz2m-nixos-system-ghaf-host-23.05.20230927.5cfafa1/initrd \
    -append "$(cat /nix/store/sgv6vmzrbkghz8a0hr05garms07cqz2m-nixos-system-ghaf-host-23.05.20230927.5cfafa1/kernel-params) init=/nix/store/sgv6vmzrbkghz8a0hr05garms07cqz2m-nixos-system-ghaf-host-23.05.20230927.5cfafa1/init regInfo=/nix/store/w7b65bipqayzsi6w2ksjvd1yi7m8b0x3-closure-info/registration console=ttyS0,115200n8 console=tty0 $QEMU_KERNEL_PARAMS" \
    $QEMU_OPTS \
    "$@"
``

modules/development/debug-tools.nix
```diff
@@ -37,6 +37,9 @@ in
           # Performance testing
           speedtest-cli
           iperf
+          perf-tools
+          linuxPackages.perf
```
Mic92, line 41: Off-topic: I find bcc's profile script a bit easier to use ` sudo profile -f -F 99 10 > /tmp/folded-graph` and the result can be looked at in https://www.speedscope.app/

jkuro-tii:
I ran my tool in several configuration. Results:
- there is no data lost or data corruption in any scenario
- transfers between chromium-vm and zathura-vm are fast - ~250 MBytes/s
- transfers between any of the above VMs and gui-vm are very slow. They hardly reach 12 MB/s

Shutting down weston gave no result. But doing that and eliminating passing-through the GPU solved the problem. After that transfer to/from gui-vm have the same high speed as inter app-vm's.

I'll continue with different virtual PCI devices setup, and try to use qemu own monitor in order to find the bottleneck.

jkuro-tii:
The driver already contains interrupts counters. I verified number of interrupts sent by one VM against numbers listed in /proc/interrupts/. There are in line, i.e. number of interrupts raised by one VM is equal to number /proc/interrupts in the other VM as well as counted by the device driver. It means that there is no interrupt loop, missed interrupts.

Delays and huge CPU load happen when GPU is pass-through. Without it CPU load is less then 5%.

jkuro-tii:
Added some improvements:
- allocating shared memory area using huge pages (2MB size)
- adding to the qemu ivshmem driver ability to map the above memory directly in VM address space (currently it's mapped into PCI area). The physical memory address is set in kernel command line.

vilvo (changes requested):
Main observations:
- Previous PCI device model-based shared memory has been removed. I hid my outdated review comments from February.

The main issues:
- PR description field is still not following the PR template. In particular, the Testing section is both outdated (not updated) and it is still missing the exact steps to reproduce the performance testing results. The exact steps must be specific enough that people who are not familiar with the work can reproduce the measurement results.
This comment from February is still valid https://github.com/tiiuae/ghaf/pull/473#pullrequestreview-1877609249 
Please keep the PR description up-to-date.

- The PR branch does not build:
```
[vilvo@carrie:~/ghaf]$ git checkout jarek/jkuro-add-wayland-shm-lenovo_rebase
...
HEAD is now at a68cf7b Added memory sharing between VMs

[vilvo@carrie:~/ghaf]$ nix build .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug
error: hash mismatch in fixed-output derivation '/nix/store/n49a0yrl19lagg6yi2vjkcxs7mmyy55i-source.drv':
         specified: sha256-y1o9Md9apgn+mjBZ9wScWUNQ3ZgxbT0TKxl8bIMZxGs=
            got:    sha256-qNeIjJj3VJ1ngvWTa914zacQMwzpM+9KLysJyJTcnGk=
```

vilvo:
Structurally, these parts are copies
https://github.com/tiiuae/ghaf/pull/473/files#diff-092df22f199bcdee3f75c01eb9e440429aabe51e8fd706c44b479975f70fee72R132-R139
https://github.com/tiiuae/ghaf/pull/473/files#diff-4cd0792a1806df6e4de8a4de75cf451ec4536470b07fa5b4a3f10ad041ed989eR129-R136
so I would recommend removing the duplication.

vilvo:
Thanks, I tested the PR and compared 1080p streaming with main as follows:

## https://github.com/tiiuae/ghaf/pull/473

ghaf on HEAD (139d614) [$] via ❄️  impure (Ghaf-devshell-env)
❯ nix build .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug

## reference - origin/main

ghaf on HEAD (2db4cee) [$] via ❄️  impure (Ghaf-devshell-env)
❯ nix build .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug

## Test protocol - load on Youtube Full HD streaming

* Connect wifi network
* Open chromium-vm with https://www.youtube.com/watch?v=RzVvThhjAKw
* Quality > 1080p
* Play full screen
* Right click > Stats for nerds
* Open terminal1	> htop (gui-vm)
* Open terminal2 > ssh 192.168.101.2 > htop (ghaf-host)
* Open terminal3 > ssh chromium-vm > htop (chromium-vm)
* Place terminal windows on top of full screen video (see images below)
* Take an image with camera

## Results

### https://github.com/tiiuae/ghaf/pull/473 @ 139d614
shmem

### main @ 2db4cee

non_shmem

## Conclusions

Snapshot of subsystem load numbers (incl. average) are inconclusive as they are ~ in the same ballpark.

modules/desktop/profiles/applications.nix
```diff
@@ -4,21 +4,112 @@
...
+          description = "Enable memory sharing between virtual machines";
+        };
+        memSize = mkOption {
+          type = lib.types.int;
+          default = 16;
+          description = mdDoc ''
+            Defines shared memory size
+          '';
+        };
+        vmCount = mkOption {
```
vilvo, line 30: I have some concerns with `vmCount` used as a duplicate limit for already declared number of application VMs.
If/when we must(?) set this as [a build-time parameter to memsocket](https://github.com/tiiuae/ghaf/pull/473/files#diff-4cd0792a1806df6e4de8a4de75cf451ec4536470b07fa5b4a3f10ad041ed989eR31), could we instead [reuse the length of `enabled-app-vms` list](https://github.com/tiiuae/ghaf/blob/2db4ceeff045a8bc4c858bb1ff33481f70ed7c26/modules/reference/appvms/default.nix#L22) to follow DRY-principle?

As a practical question - what happens if the the `vmCount` is declared smaller than the amount of system enabled application VMs?

modules/desktop/profiles/applications.nix
```diff
@@ -4,21 +4,112 @@
...
+          type = lib.types.int;
+          default = 6;
+          description = mdDoc ''
+            Defines maximum number of application VMs
+          '';
+        };
+        serverSocketPath = mkOption {
+          type = lib.types.str;
+          default = "/run/user/1000/memsocket-server.sock";
```
vilvo, line 39: Binding the path to `ghaf` `UID` can be error-prone. It might be an acceptable risk but the user-model is due to get changed to operator and admin model where we can't assume the operator user to have the same rights as it currently has. Depending on how nix creates the declared operator and admin users, the UID bound path can break things.

modules/desktop/profiles/applications.nix
```diff
@@ -4,21 +4,112 @@
...
+          description = mdDoc ''
+            Defines location of the listening socket.
+            It's used by waypipe as an output socket when running in server mode
+          '';
+        };
+        clientSocketPath = mkOption {
+          type = lib.types.str;
+          default = "/run/user/1000/memsocket-client.sock";
```
vilvo, line 47: https://github.com/tiiuae/ghaf/pull/473/files#r1698205321

modules/desktop/profiles/applications.nix
```diff
@@ -4,21 +4,112 @@
...
+            Defines location of the shared memory socket. It's used by qemu
+            instances for memory sharing and sending interrupts.
+          '';
+        };
+        flataddr = mkOption {
+          type = lib.types.str;
+          default = "0x920000000";
+          description = mdDoc ''
+            If set to a non-zero value, it maps the shared memory
+            into this physical address.
```
vilvo, line 67: Would you please add to `description` on how the default value was/is chosen?

modules/hardware/common/kernel.nix
```diff
@@ -50,11 +50,21 @@ in {
           ++ config.ghaf.hardware.definition.gpu.pciDevices
           ++ config.ghaf.hardware.definition.audio.pciDevices
         ));
+        hugepagesz = 2; # MBytes
```
vilvo, line 53: if you use "optional scale suffix [kKmMgG]" here, the comment gets redundant and you can also change the line https://github.com/tiiuae/ghaf/pull/473/files#diff-7e202e79cee502470d29582cd4b072d36760f5cdc8c1eca7e87026577f04ae03R58 where you already use the scale suffix. Changing this also removes a source of error - if one wants to test different page sizes, they can't define the suffix on line 53 because it would break on line 58 with two suffices. Defining it only on line 53 solves the problem.

modules/microvm/virtualization/microvm/appvm.nix
```diff
@@ -25,6 +25,11 @@
       if vm.cid > 0
       then vm.cid
       else cfg.vsockBaseCID + index;
+    shmConfig = configHost.ghaf.profiles.applications.ivShMemServer;
+    memsocket = pkgs.callPackage ../../../../packages/memsocket {
```
vilvo, line 29: memsocket should be thoroughly reviewed and tested before integrating this PR

modules/microvm/virtualization/microvm/guivm.nix
```diff
@@ -227,20 +267,52 @@ in {
...
+          vectors = toString (2 * shmConfig.vmCount);
+        in
+          pkgs.writeShellScriptBin "ivshmemsrv" ''
+            chown microvm /dev/hugepages
+            chgrp kvm /dev/hugepages
+            if [ -S ${shmConfig.hostSocketPath} ]; then
+              echo Erasing ${shmConfig.hostSocketPath} ${pidFilePath}
+              rm -f ${shmConfig.hostSocketPath}
+            fi
+            ${pkgs.sudo}/sbin/sudo -u microvm -g kvm ${pkgs.qemu_kvm}/bin/ivshmem-server -p ${pidFilePath} -n ${vectors} -m /dev/hugepages/ -l ${(toString config.ghaf.profiles.applications.ivShMemServer.memSize) + "M"}
```
vilvo, line 299: Please avoid redefining the suffix `"M"` here.

overlays/custom-packages/qemu/0001-ivshmem-flat-memory-support.patch
vilvo, line 1: Have you asked QEMU upstream for comments on this patch?

vilvo (changes requested):
In general, this is a great step forward on functional shared memory support on an area where it can matter a lot. My testing was not conclusive on performance improvement https://github.com/tiiuae/ghaf/pull/473#issuecomment-2259930145 but I recommend involving testing team on through regression testing and further performance testing.

The main change request is to get proper peer review for `memsocket` before integrating this PR.
Smaller change requests are added in detailed comments. 
Thanks for addressing previous review comments and improving the PR description. All in all, nice work. A lot of potential for perf improvements, optimization and security (potentially with pKVM).

packages/memsocket/default.nix
```diff
@@ -0,0 +1,49 @@
...
+  CFLAGS =
+    "-O2 -DVM_COUNT="
+    + (toString vms)
+    + (
+      if debug
+      then " -DDEBUG_ON"
+      else ""
+    );
+  prePatch = ''
+    cd app
```
avnik, line 34: If app not in repo root, but in `app/` subdirectory -- use `sourceRoot="app/"` statement, not chdir during prePatch phase

packages/memsocket/memtest.nix
```diff
@@ -0,0 +1,39 @@
...
+  name = "memtest";
+
+  src = fetchFromGitHub {
+    owner = "tiiuae";
+    repo = "shmsockproxy";
+    rev = "4eef1a7e1f6993a9ba0ef1aa066145c410e2eec6";
+    sha256 = "sha256-uaKBtb+bD42RK/Z96dhMrGxNXg+ZV/6iW52ouNQ+QSY=";
+  };
+
+  nativeBuildInputs = [gcc gnumake];
```
avnik, line 21: GCC and gnumake is in stdenv, no need to set them explicitly. Also I suggest to join memsocket and memtest to single package (just add there root makefile to make/make install all at once.

packages/memsocket/default.nix
jkuro-tii: Done. Surprisingly, the `sourceRoot` must be headed with `source`, i.e.: `sourceRoot="source/app/"`.

packages/memsocket/memtest.nix
jkuro-tii: I decided to remove the memtest component. It's used only for checking the speed and reliability of the socket-to-socket connection.

modules/desktop/profiles/applications.nix
jkuro-tii: I added the `uid` parameter to `options.ghaf.users.accounts`, as the same issue exists for nm-launcher.

overlays/custom-packages/qemu/0001-ivshmem-flat-memory-support.patch
jkuro-tii: Not yet. I plan to do it after finishing this review.

modules/desktop/profiles/applications.nix
jkuro-tii: Done.

vilvo:
> The main change request is to get proper peer review for `memsocket` before integrating this PR.

I propose (at least) static analyser protocol and fixes to `memsocket` (a.k.a `shmsockproxy`):

```
$ git clone https://github.com/tiiuae/shmsockproxy.git
$ cd shmsockproxy/
$ nix-shell -p flawfinder
$ flawfinder .
...
ANALYSIS SUMMARY:

Hits = 43
Lines analyzed = 4110 in approximately 0.02 seconds (190030 lines/second)
Physical Source Lines of Code (SLOC) = 3207
Hits@level = [0]  17 [1]  10 [2]  18 [3]   0 [4]  15 [5]   0
Hits@level+ = [0+]  60 [1+]  43 [2+]  33 [3+]  15 [4+]  15 [5+]   0
Hits/KSLOC@level+ = [0+] 18.7091 [1+] 13.4082 [2+] 10.29 [3+] 4.67727 [4+] 4.67727 [5+]   0
Dot directories skipped = 1 (--followdotdir overrides)
Minimum risk level = 1

Not every hit is necessarily a security vulnerability.
You can inhibit a report by adding a comment in this form:
// flawfinder: ignore
Make *sure* it's a false positive!
You can use the option --neverignore to show these.
$ flawfinder . > results.txt
```
Please find the results attached.
[results.txt](https://github.com/user-attachments/files/16452262/results.txt)

modules/hardware/common/kernel.nix
jkuro-tii: Fixed

modules/desktop/profiles/applications.nix
jkuro-tii: > As a practical question - what happens if the the `vmCount` is declared smaller than the amount of system enabled application VMs?
1. Interrupts from excessive VMs will not interrupt the peer as it allocates only the `vmCount` pairs of interrupts.
2. The `memsocket` application will fail as there is insufficient shared memory for excessive VMs.

modules/microvm/virtualization/microvm/guivm.nix
jkuro-tii: It's not a redefinition. I added a comment that defined shared memory size unit are MBytes:
https://github.com/jkuro-tii/ghaf/blob/4717c4e9df348dbde0e336a273365fa050daa440/modules/desktop/profiles/applications.nix#L27

modules/desktop/profiles/applications.nix
jkuro-tii: > [reuse the length of enabled-app-vms list](https://github.com/tiiuae/ghaf/blob/2db4ceeff045a8bc4c858bb1ff33481f70ed7c26/modules/reference/appvms/default.nix#L22) to follow DRY-principle?

Done.

modules/microvm/virtualization/microvm/appvm.nix
```diff
@@ -42,10 +47,18 @@
...
             then "--border \"${vm.borderColor}\""
             else "";
-          runWaypipe = pkgs.writeScriptBin "run-waypipe" ''
-            #!${pkgs.runtimeShell} -e
-            ${pkgs.waypipe}/bin/waypipe --vsock -s ${toString configHost.ghaf.virtualization.microvm.guivm.waypipePort} ${waypipeBorder} server "$@"
-          '';
+          runWaypipe =
+            if shmConfig.display
+            then
+              pkgs.writeScriptBin "run-waypipe" ''
```
vunnyso, line 53: Is it possible to have `if, else` in script rather creating than creating different scripts?

jkuro-tii:
Updates:

- Refactored Nix code by consolidating most host and VM configuration changes into a single file, improving maintainability.

- Added support for shared memory on the host system, enabling any application utilizing Unix Sockets (e.g. gRPC) to seamlessly run on both the host and within a virtual machine (VM).

- Implemented the shared memory driver in the NixOS build as a kernel module instead of a kernel patch, simplifying maintenance and eliminating the need for kernel rebuilding.

- Replaced the hardcoded ghaf user ID (1000) with a configurable parameter, increasing flexibility.

UnikieHR:
image

modules/hardware/common/shared-mem.nix
```diff
@@ -0,0 +1,232 @@
...
+#
+# Module for Shared Memory Definitions
+#
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+with lib;
```
mbssrc, line 12: Common pattern used atm is 
``` 
: let
  inherit (lib) types; # + all the functions you need
in {
```

modules/hardware/common/shared-mem.nix
```diff
@@ -0,0 +1,232 @@
...
+{
+  options.ghaf.shm = {
+    enable = lib.mkOption {
```
mbssrc, line 15: and no need for .lib

modules/hardware/common/shared-mem.nix
```diff
@@ -0,0 +1,232 @@
...
+      apply =
+        enable:
+        if enable && !config.ghaf.shm.enable then
+          builtins.throw "Enable shared memory in order to use shm display"
+        else
+          enable;
+    };
+  };
+
+  config.boot.kernelParams =
```
mbssrc, line 109: For better visibility in the module, you can write 
```
config = mkIf cfg.enable {
  # all configurations here, e.g.,
  boot.kernelParams = ...
}
```

modules/hardware/common/shared-mem.nix
```diff
@@ -0,0 +1,232 @@
...
+      default = false;
+      description = "Display VMs using shared memory";
+      apply =
+        enable:
+        if enable && !config.ghaf.shm.enable then
+          builtins.throw "Enable shared memory in order to use shm display"
+        else
+          enable;
+    };
+  };
```
mbssrc, line 107: Do you need all of these options to be configurable, or are they referenced outside of the module? If not, you can define them in the `let ... in` part. Also, you can define assertions in the config if needed

modules/hardware/common/shared-mem.nix
```diff
@@ -0,0 +1,232 @@
...
+
+  config.boot.kernelParams =
+    let
+      hugepagesz = "2M"; # valid values: "2M" and "1G", as kernel supports these huge pages' size
+      hugepages =
+        if hugepagesz == "2M" then config.ghaf.shm.memSize / 2 else config.ghaf.shm.memSize / 1024;
```
mbssrc, line 113: Should `hugepagez = "2M";` be an option? Otherwise the `hugepages` calculation is always the same

modules/hardware/common/shared-mem.nix
```diff
@@ -0,0 +1,232 @@
...
+          vectors = toString (2 * config.ghaf.shm.instancesCount);
+        in
+        pkgs.writeShellScriptBin "ivshmemsrv" ''
+          chown microvm /dev/hugepages
+          chgrp kvm /dev/hugepages
+          if [ -S ${config.ghaf.shm.hostSocketPath} ]; then
+            echo Erasing ${config.ghaf.shm.hostSocketPath} ${pidFilePath}
+            rm -f ${config.ghaf.shm.hostSocketPath}
+          fi
+          ${pkgs.sudo}/sbin/sudo -u microvm -g kvm ${pkgs.qemu_kvm}/bin/ivshmem-server -p ${pidFilePath} -n ${vectors} -m /dev/hugepages/ -l ${
```
mbssrc, line 137: Could you run the systemd service as `User = "microvm"; Group = "kvm";` instead of using sudo?

modules/hardware/common/shared-mem.nix
```diff
@@ -0,0 +1,232 @@
...
+        };
+      };
+    in
+    mkIf config.ghaf.shm.enable (
+      foldl' lib.attrsets.recursiveUpdate { } (map makeAssignment config.ghaf.shm.vms_enabled)
+    );
+
+  config.ghaf.hardware.definition.gpu.kernelConfig.kernelParams = optionals config.ghaf.shm.enable [
+    "kvm_ivshmem.flataddr=${config.ghaf.shm.flataddr}"
+  ];
```
mbssrc, line 231: Could you use `config.microvm.vms.gui-vm.config.config.boot.kernelParams` directly instead of the hardware configuration?

modules/hardware/common/shared-mem.nix
jkuro-tii: Running the service under root is needed for setting ownership of /dev/hugepages - it must be done under `root`. Elevating privileges in a user-owned service using /run/wrappers/bin/sudo is not feasible since it prompts for a password. Additionally, using ${pkgs.sudo}/sbin/sudo is also not an option in this case.

mbssrc (approved):

brianmcgillion:
@jkuro-tii can you squash those commits to allow easier bisect?

leivos-unikie:
I tried to run the tests described in 
https://ssrc.atlassian.net/wiki/spaces/~62552e6ffdb60b006927ad98/pages/825720835/Wayland+displaying+with+shared+memory
but I think I have missed something.

```
[ghaf@gui-vm:~]$ systemctl start --user memsocket.service
Failed to start memsocket.service: Unit memsocket.service not found.
```
```
[ghaf@chromium-vm:~]$ systemctl stop --user memsocket.service
Failed to stop memsocket.service: Unit memsocket.service not loaded.
```
```
[ghaf@ghaf-host:~]$ vsockproxy 5201 100 5201
vsockproxy: command not found
```

leivos-unikie:
And also I am wondering if those Confluence pages under Documentation have been migrated to new tii hosted Confluence.

leivos-unikie:
tests in ci-test-automation passed

leivos-unikie:
Since the shared memory feature is disabled by default (for now) I ran the iperf tests described in documentation with separate testing branch
https://github.com/jkuro-tii/ghaf/tree/jkuro-add-wayland-shm-lenovo_rebase_test

**In gui-vm:**
`systemctl stop --user memsocket.service`
`memsocket -s 1234.sock 3 &`
`UNSOCK_ADDR=127.0.0.1 UNSOCK_DIR=``pwd`` LD_PRELOAD=/nix/store/dlavg00s1ckmkdkcaz1p9igqnmzgnhf6-unsock/lib/libunsock.so.1.1.0 iperf -c 127.0.0.1 -p 1234`

**In chromium:**
`systemctl stop --user memsocket.service`
`UNSOCK_BLOCK_INET6=1 UNSOCK_PORT=55555 UNSOCK_ADDR=127.0.0.1/0 UNSOCK_DIR="/home/ghaf/" LD_PRELOAD=/nix/store/dlavg00s1ckmkdkcaz1p9igqnmzgnhf6-unsock/lib/libunsock.so.1.1.0 iperf -s -p 55555 &`

-> Speed 20.1Gbits/sec


For comparison without shared memory

**In gui-vm:**
`iperf -s -1 -p 5201`

**In chromium-vm:**
`iperf3 -c 192.168.100.3 -p 5201`

-> Speed 15Gbits/sec

leivos-unikie:
Should there be some documentation about this feature in ghaf/docs?
(Jenni had already asked this but didn't receive any answer.)

At least in my opinion it is not enough to have the iperf test instructions in the old Confluence.

vadika:
> Should there be some documentation about this feature in ghaf/docs? (Jenni had already asked this but didn't receive any answer.)
> 
> At least in my opinion it is not enough to have the iperf test instructions in the old Confluence.

Documentation is definitely needs to be in place in the Documentation repository

PR #478, opened 2024-02-16
Introduce new disk partitioning scheme

unbel13ver:
New disk configuration provides grounds for upcoming features, such as AB software updates and Storage VM and many more.

The Lenovo X1 config has two LVM pools, first one is fixed-size 250G "system" partition (which is going to be encrypted) and the rest of the disk is dedicated to the StorageVM.

## Checklist for things done

## Testing
Current upstream partition scheme is as follows:
```
[ghaf@ghaf-host:~]$ lsblk
NAME          MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
sda             8:0    0 465.8G  0 disk 
├─sda1          8:1    0   236M  0 part 
└─sda2          8:2    0   8.6G  0 part 
nvme0n1       259:0    0 953.9G  0 disk 
├─nvme0n1p1   259:1    0     1M  0 part 
├─nvme0n1p2   259:2    0   500M  0 part /boot
└─nvme0n1p3   259:3    0 953.4G  0 part 
  └─pool-root 254:0    0 953.4G  0 lvm  /nix/store
                                        /
```
To test new partitioning scheme, run `lenovo-x1-carbon-gen11-debug-installer` image and install Ghaf into the internal SSD storage of Lenovo-X1.
After insallation is completed and the laptop is booted into the Ghaf system, check partitions with `lsblk` command.
It should be as follows:
```
[ghaf@ghaf-host:~]$ lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
nvme0n1     259:0    0 953.9G  0 disk 
├─nvme0n1p1 259:1    0     1M  0 part 
├─nvme0n1p2 259:2    0   500M  0 part /boot
├─nvme0n1p3 259:3    0   500M  0 part 
└─nvme0n1p4 259:4    0    14G  0 part 
```
Check zpool is online:
```
[ghaf@ghaf-host:~]$ zpool list
NAME      SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
zroot_1    14G  5.00G  9.00G        -         -     0%    35%  1.00x    ONLINE  -
```
Check datasets are mounted successfully:
```
[ghaf@ghaf-host:~]$ mount | grep zroot_1
zroot_1/root_a on / type zfs (rw,relatime,xattr,posixacl,casesensitive,x-initrd.mount)
zroot_1/root_a on /nix/store type zfs (ro,relatime,xattr,posixacl,casesensitive)
zroot_1/storagevm on /storagevm type zfs (rw,relatime,xattr,posixacl,casesensitive)
zroot_1/vm_storage_a on /vm_storage type zfs (rw,relatime,xattr,posixacl,casesensitive)
zroot_1/gp_storage on /gp_storage type zfs (rw,relatime,xattr,posixacl,casesensitive)
```
Optionally - check there are no failed `systemd` units:
```
[ghaf@ghaf-host:~]$ systemctl list-units --all --state=failed
  UNIT LOAD ACTIVE SUB DESCRIPTION

0 loaded units listed.
To show all installed unit files use 'systemctl list-unit-files'.
```

### DIFF
--- a/modules/disko/disko-ab-partitions.nix
+++ b/modules/disko/disko-ab-partitions.nix
@@ -1,0 +1,161 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# This partition scheme contains three common partitions and ZFS pool.
+# Some partitions are duplicated for the future AB SWupdate implementation.
+#
+# First three partitions are related to the boot process:
+# - boot : Bootloader partition
+# - ESP-A : (500M) Kernel and initrd
+# - ESP-B : (500M)
+#
+# ZFS datasets do not necessary need to have specified size and can be
+# allocated dynamically. Quotas only restrict the maximum size of
+# datasets, but do not reserve the space in the pool.
+# The ZFS pool contains next datasets:
+# - root-A : (30G) Root FS
+# - root-B : (30G)
+# - vm-storage-A : (30G) Possible standalone pre-built VM images are stored here
+# - vm-storage-B : (30G)
+# - reserved-A : (10G) Reserved dataset, no use
+# - reserved-B : (10G)
+# - gp-storage : (50G) General purpose storage for some common insecure cases
+# - recovery : (no quota) Recovery factory image is stored here
+# - storagevm: (no quota) Dataset is meant to be used for StorageVM
+{pkgs, ...}: {
+  #TODO Probably the 'networking.hostId' should be set
+  # somewhere else instead.
+  networking.hostId = "8425e349";
+  disko = {
+    memSize = 4096;
+    extraPostVM = ''
+      ${pkgs.zstd}/bin/zstd --compress $out/*raw
+      rm $out/*raw
+    '';
+    extraRootModules = ["zfs"];
+    devices = {
+      disk.disk1 = {
+        type = "disk";
+        imageSize = "15G";
+        content = {
+          type = "gpt";
+          partitions = {
+            boot = {
+              name = "boot";
+              size = "1M";
+              type = "EF02";
+              priority = 1; # Needs to be first partition
+            };
+            esp_a = {
+              name = "ESP_A";
+              size = "500M";
+              type = "EF00";
+              content = {
+                type = "filesystem";
+                format = "vfat";
+                mountpoint = "/boot";
+                mountOptions = [
+                  "umask=0077"
+                  "nofail"
+                ];
+              };
+            };
+            esp_b = {
+              name = "ESP_B";
+              size = "500M";
+              type = "EF00";
+              content = {
+                type = "filesystem";
+                format = "vfat";
+                mountOptions = [
+                  "umask=0077"
+                  "nofail"
+                ];
+              };
+            };
+            zfs_1 = {
+              size = "100%";
+              content = {
+                type = "zfs";
+                pool = "zfspool";
+              };
+            };
+          };
+        };
+      };
+      zpool = {
+        zfspool = {
+          type = "zpool";
+          rootFsOptions = {
+            mountpoint = "none";
+            acltype = "posixacl";
+          };
+          datasets = {
+            "root_a" = {
+              type = "zfs_fs";
+              mountpoint = "/";
+              options = {
+                mountpoint = "/";
+                quota = "30G";
+              };
+            };
+            "vm_storage_a" = {
+              type = "zfs_fs";
+              options = {
+                mountpoint = "/vm_storage";
+                quota = "30G";
+              };
+            };
+            "reserved_a" = {
+              type = "zfs_fs";
+              options = {
+                mountpoint = "none";
+                quota = "10G";
+              };
+            };
+            "root_b" = {
+              type = "zfs_fs";
+              options = {
+                mountpoint = "none";
+                quota = "30G";
+              };
+            };
+            "vm_storage_b" = {
+              type = "zfs_fs";
+              options = {
+                mountpoint = "none";
+                quota = "30G";
+              };
+            };
+            "reserved_b" = {
+              type = "zfs_fs";
+              options = {
+                mountpoint = "none";
+                quota = "10G";
+              };
+            };
+            "gp_storage" = {
+              type = "zfs_fs";
+              options = {
+                mountpoint = "/gp_storage";
+                quota = "50G";
+              };
+            };
+            "recovery" = {
+              type = "zfs_fs";
+              options = {
+                mountpoint = "none";
+              };
+            };
+            "storagevm" = {
+              type = "zfs_fs";
+              options = {
+                mountpoint = "/storagevm";
+              };
+            };
+          };
+        };
+      };
+    };
+  };
+}
--- a/modules/common/systemd/hardened-configs/common/systemd-remount-fs.nix
+++ b/modules/common/systemd/hardened-configs/common/systemd-remount-fs.nix
@@ -43,7 +43,7 @@
   # Devices #
   ###########
 
-  PrivateDevices = true;
+  # PrivateDevices = true;
   # DeviceAllow=/dev/null
 
   ##########
--- a/modules/disko/disko-zfs-postboot.nix
+++ b/modules/disko/disko-zfs-postboot.nix
@@ -1,0 +1,36 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{pkgs, ...}: let
+  postBootCmds = ''
+    set -xeuo pipefail
+
+    # Check which physical disk is used by ZFS
+    ZFS_POOLNAME=$(${pkgs.zfs}/bin/zpool list | ${pkgs.gnugrep}/bin/grep -v NAME |  ${pkgs.gawk}/bin/awk '{print $1}')
+    ZFS_LOCATION=$(${pkgs.zfs}/bin/zpool status -P | ${pkgs.gnugrep}/bin/grep dev | ${pkgs.gawk}/bin/awk '{print $1}')
+
+    # Get the actual device path
+    P_DEVPATH=$(readlink -f "$ZFS_LOCATION")
+
+    # Extract the partition number using regex
+    if [[ "$P_DEVPATH" =~ [0-9]+$ ]]; then
+      PARTNUM=$(echo "$P_DEVPATH" | ${pkgs.gnugrep}/bin/grep -o '[0-9]*$')
+      PARENT_DISK=$(echo "$P_DEVPATH" | ${pkgs.gnused}/bin/sed 's/[0-9]*$//')
+    else
+      echo "No partition number found in device path: $P_DEVPATH"
+    fi
+
+    # Fix GPT first
+    ${pkgs.gptfdisk}/bin/sgdisk "$PARENT_DISK" -e
+
+    # Call partprobe to update kernel's partitions
+    ${pkgs.parted}/bin/partprobe
+
+    # Extend the partition to use unallocated space
+    ${pkgs.parted}/bin/parted -s -a opt "$PARENT_DISK" "resizepart $PARTNUM 100%"
+
+    # Extend ZFS pool to use newly allocated space
+    ${pkgs.zfs}/bin/zpool online -e "$ZFS_POOLNAME" "$ZFS_LOCATION"
+  '';
+in {
+  boot.postBootCommands = postBootCmds;
+}
--- a/modules/disko/flake-module.nix
+++ b/modules/disko/flake-module.nix
@@ -7,5 +7,11 @@
       ./disko-basic-partition-v1.nix
       ./disko-basic-postboot.nix
     ];
+
+    disko-ab-partitions-v1.imports = [
+      inputs.disko.nixosModules.disko
+      ./disko-ab-partitions.nix
+      ./disko-zfs-postboot.nix
+    ];
   };
 }
--- a/modules/hardware/x86_64-generic/x86_64-linux.nix
+++ b/modules/hardware/x86_64-generic/x86_64-linux.nix
@@ -32,11 +32,14 @@ in {
       initrd.availableKernelModules = [
         "nvme"
         "uas"
+        "zfs"
       ];
       loader = {
         efi.canTouchEfiVariables = true;
         systemd-boot.enable = true;
       };
+      supportedFilesystems = ["zfs"];
+      kernelPackages = config.boot.zfs.package.latestCompatibleLinuxPackages;
     };
   };
 }
--- a/targets/laptop/flake-module.nix
+++ b/targets/laptop/flake-module.nix
@@ -14,7 +14,7 @@
   targets = [
     # Laptop Debug configurations
     (laptop-configuration "lenovo-x1-carbon-gen10" "debug" [
-      self.nixosModules.disko-basic-partition-v1
+      self.nixosModules.disko-ab-partitions-v1
       {
         ghaf = {
           hardware.definition.configFile = "/lenovo-x1/definitions/x1-gen10.nix";
@@ -23,7 +23,7 @@
       }
     ])
     (laptop-configuration "lenovo-x1-carbon-gen11" "debug" [
-      self.nixosModules.disko-basic-partition-v1
+      self.nixosModules.disko-ab-partitions-v1
       {
         ghaf = {
           hardware.definition.configFile = "/lenovo-x1/definitions/x1-gen11.nix";
@@ -52,7 +52,7 @@
 
     # Laptop Release configurations
     (laptop-configuration "lenovo-x1-carbon-gen10" "release" [
-      self.nixosModules.disko-basic-partition-v1
+      self.nixosModules.disko-ab-partitions-v1
       {
         ghaf = {
           hardware.definition.configFile = "/lenovo-x1/definitions/x1-gen10.nix";
@@ -61,7 +61,7 @@
       }
     ])
     (laptop-configuration "lenovo-x1-carbon-gen11" "release" [
-      self.nixosModules.disko-basic-partition-v1
+      self.nixosModules.disko-ab-partitions-v1
       {
         ghaf = {
           hardware.definition.configFile = "/lenovo-x1/definitions/x1-gen11.nix";


templates/targets/x86_64/generic/disk-config.nix
```diff
@@ -14,19 +15,38 @@ _: {
...
-            name = "ESP";
+          esp_a = {
+            name = "ESP_A";
             size = "500M";
             type = "EF00";
             content = {
               type = "filesystem";
               format = "vfat";
               mountpoint = "/boot";
-              mountOptions = ["umask=0077"];
```
vilvo, line 25: There was no testing description so was this PR tested with this change. We had to add `umask=0077` because of warning:
```
⚠️ Mount point '/boot' which backs the random seed file is world accessible, which is a security hole! ⚠️
⚠️ Random seed file '/boot/loader/.#bootctlrandom-seed5c51b0dd8aff4d60' is world accessible, which is a security hole! ⚠️
```
More info: https://discourse.nixos.org/t/security-warning-when-installing-nixos-23-11/37636
and https://github.com/tiiuae/ghaf/compare/45dedba37c862186c92e417ca36b227492df6319..0a59206396b2997ad757a0ee508b3aab3caa0c12#diff-8a1fbff8658406e95003d28116ba21ced73c81b4fb8d1a5edad5fe5aeef1c427R25

templates/targets/x86_64/generic/disk-config.nix
```diff
@@ -51,6 +71,95 @@ _: {
               ];
             };
           };
+          vm_storage_a = {
+            size = "30G";
+            content = {
+              type = "filesystem";
+              format = "ext4";
+              mountpoint = "/vmstor";
```
vilvo, line 79: Is the missing `e` in the end of the mountpoint intentional or accidental? https://github.com/tiiuae/ghaf/pull/478/files#diff-8a1fbff8658406e95003d28116ba21ced73c81b4fb8d1a5edad5fe5aeef1c427R44

templates/targets/x86_64/generic/disk-config.nix
```diff
@@ -51,6 +71,95 @@ _: {
...
+                "defaults"
+              ];
+            };
+          };
+          recovery = {
+            size = "20G";
+            content = {
+              type = "filesystem";
+              format = "ext4";
+              #mountpoint = "/vmstor";
```
vilvo, line 141: This can be removed.

templates/targets/x86_64/generic/disk-config.nix
```diff
@@ -1,6 +1,7 @@
-# Copyright 2022-2023 TII (SSRC) and the Ghaf contributors
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
-# Example to create a bios compatible gpt partition
+#
+# The partitioning scheme is hardcoded for 500G disk.
```
vilvo, line 4: With this comment we could actually make this specific to Lenovo X1 Carbon Gen11 SKU (stock keeping unit) with 500G of internal NVMe. I'm afraid we must build target specific installers as we have not yet described disk space extension in case the internal NVMe is larger. Optionally, the installer could resolve the size of the disk with `lsblk` and parametrize the size to generate disk config for `generic` use. I think all the other fixed sizes could remain the same and only the remaining space here https://github.com/tiiuae/ghaf/pull/478/files#diff-8a1fbff8658406e95003d28116ba21ced73c81b4fb8d1a5edad5fe5aeef1c427R154 would have to be calculated. Please check with  @remimimimimi if this is feasible. The next iterations of installer must support other installer parameters - like user specific secrets - as well.

vilvo (changes requested):
Please see the detailed comments.

leivos-unikie:
@unbel13ver can you provide test instructions for this PR? What is the expected outcome of new partitioning scheme?

templates/targets/x86_64/generic/disk-config.nix
unbel13ver: Thanks, addressed.

templates/targets/x86_64/generic/disk-config.nix
unbel13ver: Thanks for noticing! Addressed.

unbel13ver:
> @unbel13ver can you provide test instructions for this PR? What is the expected outcome of new partitioning scheme?

Testing instructions added.

templates/targets/x86_64/generic/disk-config.nix
unbel13ver: > Optionally, the installer could resolve the size of the disk with lsblk and parametrize the size to generate disk config for `generic` use.

This exact approach assumes that the installer program changes `disk-config.nix` file before passing it into `disko` utility, which I am pretty sure we want to avoid.

> I think all the other fixed sizes could remain the same and only the remaining space <here> would have to be calculated.

Unfortunately, `disko` utility for some reason does not support `size = "100%"` parameter in that exact place. It fails with error while partitioning the disk. Probably I should have mentioned this in the comment inside the file itself.

> The next iterations of installer must support other installer parameters - like user specific secrets - as well.

This is out of scope of this PR.

leivos-unikie (approved):
- lsblk gives expected output
- ci-test-automation passed
- performance ok
- apps launch

templates/targets/x86_64/generic/disk-config.nix
remimimimimi: > Unfortunately, disko utility for some reason does not support size = "100%" parameter in that exact place. It fails with error while partitioning the disk. Probably I should have mentioned this in the comment inside the file itself.

It seems that [this should work](https://github.com/nix-community/disko/commit/e1d5b132d1ac75b7f87aa407024fdb24766e695f) @unbel13ver .

Mic92: Regarding secrets one can use the `diskoImagesScript` to generate full systems that also contain secrets. The script is run outside of the NixOS sandbox.

unbel13ver: > My proposal for placement is <ghaf_repo_root>/targets/lenovo-x1-carbon-disk-config.nix because this is very specific to that device.

That is reasonable, implemented.

Also I managed to take the rest of the disk in use for StorageVM (should have added "100%FREE" instead of "100%").
@leivos-unikie could you retest this please? I've updated the expected partitioning in the header of this PR.

leivos-unikie: I spotted these deviations in log messages

- at booting installer
failed_creating_link

- in the beginning of installation process
dont_know_msg

- after installation was finished and pressed Enter for reboot
failed_unmounting


Otherwise no issues found in testing.

unbel13ver: Thanks for testing! Those are not related to the matter of this PR (those are not something worth to worry about, anyway).

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
```diff
@@ -0,0 +1,166 @@
...
+            content = {
+              type = "filesystem";
+              format = "ext4";
+              mountpoint = "/";
+              mountOptions = [
+                "defaults"
+              ];
+            };
+          };
+          vm_storage_a = {
```
Mic92, line 74: How is the vm_storage_a and vm_storage_b used? For system updates it's clear to me that one would apply changes to alternating partitions, but how does this work for vm state?

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
```diff
@@ -0,0 +1,166 @@
...
+      device = "DRIVE_PATH";
+      type = "disk";
+      content = {
+        type = "gpt";
+        partitions = {
+          boot = {
+            name = "boot";
+            size = "1M";
+            type = "EF02";
+          };
```
Mic92, line 16: The EFI standard allows only one boot loader to be registered, how would the second partition used?

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
```diff
@@ -0,0 +1,166 @@
...
+            name = "ESP_B";
+            size = "500M";
+            type = "EF00";
+            content = {
+              type = "filesystem";
+              format = "vfat";
+              mountOptions = ["umask=0077"];
+            };
+          };
+          # LVM pool that is going to be encrypted
```
Mic92, line 38: How is the encryption done here? Would it be not cryptsetup -> lvm?

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
```diff
@@ -0,0 +1,166 @@
...
+          other_1 = {
+            name = "lvm_pv_1";
+            size = "250G";
+            content = {
+              type = "lvm_pv";
+              vg = "pool";
+            };
+          };
+          # LVM pool that is going to be passed to the Storage VM
+          other_2 = {
```
Mic92, line 48: Are VMs not encrypted?

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
```diff
@@ -0,0 +1,166 @@
...
+            size = "50G";
+            content = {
+              type = "filesystem";
+              format = "ext4";
+              mountOptions = [
+                "defaults"
+              ];
+            };
+          };
+          recovery = {
```
Mic92, line 136: Is the recover also some NixOS? How is it booted?

Mic92:
Brian asked me to compare using LVM vs ZFS for this setup. Here a comparison:

- with LVM one has to plan in advance how much space needs to be used for what purpose. The A/B partitioning will take 50% of the space.
- with zfs one would create a single pool and datasets can dynamically allocate space from that. Instead of A/B partitioning, one would use snapshots -> potentially also create boot loader entries for rollback.
- Ext4 probably will provide best performance for most desktop workloads.
- Zfs makes better use of storage. 500GB is not much and compression can help here. Checksums are also nice to be more confident about correct data.
- LVM is accepted upstream so any Linux kernel will work. With zfs there is some delay usually until the latest kernel can be used. NixOS by default uses LTS version of the kernel, which will always support zfs. Apart from that there is `zfs.latestCompatibleLinuxPackages` that will give the latest linux kernel supported
-  I'd recommend having a swap for zfs because the page eviction in ARC is not as quick as the normal page cache - to normal users it appears that zfs uses way more memory but this is mostly because ARC is not accounted as page cache
- Encryption: There some open issues with zfs's native encryption. Cryptsetup is recommended if one wants to use zfs send at some point.
- Impermanence: It's possible to implement automatic state reset with both LVM/Zfs. For LVM one would probably recreate/reformat a volatile state partitions. With zfs it would be a of a dataset to some blank snapshot (exceptionally a bit easier)

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
```diff
@@ -0,0 +1,166 @@
...
+          };
+          esp_a = {
+            name = "ESP_A";
+            size = "500M";
+            type = "EF00";
+            content = {
+              type = "filesystem";
+              format = "vfat";
+              mountpoint = "/boot";
+              mountOptions = ["umask=0077"];
```
Mic92, line 25: I like adding `"nofail"` for non-root mounts. It makes recovering easier in case there is something wrong with mountpoints. Otherwise systemd will go into "emergency" mode. This can even happen on nixos-rebuild.

```suggestion
              mountOptions = ["umask=0077" "nofail"];
```

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
```diff
@@ -0,0 +1,166 @@
...
+            content = {
+              type = "filesystem";
+              format = "ext4";
+              mountpoint = "/vmstore";
+              mountOptions = [
+                "defaults"
+              ];
+            };
+          };
+          reserved_a = {
```
Mic92, line 85: Maybe some comment what this is for would be helpful.
If this is just a spaceholder do we need a filesystem on it?

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
unbel13ver: There were some ideas that some VMs can be supplied as pre-built images or so, and the end user will not be able to change the contents of that images. Those VMs will be updated together with the main system and must correspond the specific version of the base system.

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
unbel13ver: Currently there is only one bootloader partition. The duplication in this case require some external tooling to switch between A and B, which we decided to avoid.

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
unbel13ver: There is some discussion going about that in the mail thread. I do not know at this moment how it is going to be implemented, and I decided to remove this comment to avoid the confusion.

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
unbel13ver: StorageVM partition is not going to be encrypted, since it is supposed that it stores encrypted data. In this case it allows us to backup and recover the stored data safely if needed.

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
```diff
@@ -0,0 +1,166 @@
...
+        type = "lvm_vg";
+        lvs = {
+          root_a = {
+            size = "50G";
+            content = {
+              type = "filesystem";
+              format = "ext4";
+              mountpoint = "/";
+              mountOptions = [
+                "defaults"
```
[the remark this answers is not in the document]
unbel13ver: Thanks for the hint, done

targets/disc-configs/lenovo-x1-carbon-disk-config.nix
unbel13ver: I have removed the filesystem from the reserved partitions and the recovery partition. I have also added a comment section in the beginning of the file to explain some of its parts.

vilvo:
> Brian asked me to compare using LVM vs ZFS for this setup. Here a comparison:
...
> * LVM is accepted upstream so any Linux kernel will work. With zfs there is some delay usually until the latest kernel can be used. NixOS by default uses LTS version of the kernel, which will always support zfs. Apart from that there is `zfs.latestCompatibleLinuxPackages` that will give the latest linux kernel supported

[ghaf hardened](https://github.com/tiiuae/ghaf/blob/364d64a1cc7dec021997f58da4d522cd41506f46/packages/kernel/default.nix#L14) requires a ZFS kernel config fragment. It's not an issue but it's a dependency to keep in mind. 

> I currently do not have the full picture of all how this a/b partitioning will be implemented in the end. Especially around VM storage.

Yeah, the A/B partitioning was specified and documented before the latest arch documentation process change. It would probably help to collect the prev documented designs and LVM/ZFS comparisons together according to the new process. With such many contributors I'd like to call it a collaborative documentation process to ensure a better full picture for contributors.

vunnyso:
Is it possible to use smaller partitioning size and resizing mechanism later with current implementation? As final image generated will be 260GB which will take more time to generate and flash. 

If not whether following postboot [commands](https://github.com/tiiuae/ghaf/blob/main/modules/disko/disko-basic-postboot.nix#L45-L58)  will be valid to execute still?

unbel13ver:
I have added a second commit with ZFS configuration so it is easier to compare what has changed.

Known issues: For some reason, `disko` now fails with OOM during the nix store copying process if I give it's VM less than 16G of RAM.

brianmcgillion:
> Known issues: For some reason, `disko` now fails with OOM during the nix store copying process if I give it's VM less than 16G of RAM.

that should be fixed if you rebase against main. that was fixed in disko and the main has been bumped to use the newer version.

modules/disko/disko-ab-partitions.nix
```diff
@@ -0,0 +1,181 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
```
lheckemann, line 1: Should we actually have this file if we're not referencing it anywhere?

modules/disko/disko-ab-partitions-zfs.nix
```diff
@@ -0,0 +1,161 @@
...
+          };
+          "recovery" = {
+            type = "zfs_fs";
+            options = {
+              mountpoint = "none";
+            };
+          };
+        };
+      };
+      zroot_2 = {
```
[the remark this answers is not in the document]
unbel13ver: This is what I am looking for - "extract" that partition from the host.

unbel13ver (commented):
@lheckemann thank you for your input. I'll rebase my branch and refresh the PR.

modules/disko/disko-ab-partitions-zfs.nix
```diff
@@ -0,0 +1,161 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# !!! To utilize this partition scheme, the disk size must be >252G !!!
```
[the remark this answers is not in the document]
lheckemann: You can make the pool itself much smaller -- as long as the actual data being copied in (and dataset reservations, if applicable) still have enough space.

modules/disko/disko-ab-partitions-zfs.nix
```diff
@@ -0,0 +1,161 @@
...
+# - boot : Bootloader partition
+# - ESP-A : (500M) Kernel and initrd
+# - ESP-B : (500M)
+#
+# First ZFS pool contains next partitions:
+# - root-A : (50G) Root FS
+# - root-B : (50G)
+# - vm-storage-A : (30G) Possible standalone pre-built VM images are stored here
+# - vm-storage-B : (30G)
+# - reserved-A : (10G) Reserved partition, no use
```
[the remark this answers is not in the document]
lheckemann: That's a general problem with disko -- it doesn't provide the ability to change the layout later on without a complete wipe. Reserved partitions also won't help with that.

modules/hardware/lenovo-x1/default.nix
```diff
@@ -24,19 +24,24 @@ in {
...
+      disk1.device = "/dev/nvme0n1";
+      disk1.imageSize = "250G";
+    };
 
     # Hardware specific kernel parameters
     boot = {
       inherit (hwDefinition.generic) kernelParams;
+      initrd.availableKernelModules = ["nvme" "zfs"];
+      supportedFilesystems = ["zfs"];
+      kernelPackages = config.boot.zfs.package.latestCompatibleLinuxPackages;
```
vilvo, line 44: @vunnyso FYI - this requires follow-up work with the ghaf kernel hardening but is fine in the context of this PR.
In the follow-up work, it might be the easiest to add the ZFS required kernel config fragments to `host_hardened_kernel` https://github.com/tiiuae/ghaf/blob/ae44f5928131415ab578cbe5051baa9c7f78c612/modules/hardware/x86_64-generic/kernel/host/default.nix#L60

vilvo:
@unbel13ver - can you please re-request review when you are done with the changes? Just press the icon with "recycling"-arrows icon next to people you want to re-review/test. Thanks.
Screenshot 2024-06-13 at 15 05 35

modules/disko/disko-ab-partitions.nix
```diff
@@ -0,0 +1,181 @@
...
+      disk.disk1 = {
+        type = "disk";
+        imageSize = "15G";
+        content = {
+          type = "gpt";
+          partitions = {
+            boot = {
+              name = "boot";
+              size = "1M";
+              type = "EF02";
```
Mic92, line 46:
```suggestion
              type = "EF02";
              priority = 1; # Needs to be first partition
```

Mic92: I got reports this is important for some EFI/BIOS implementations.

unbel13ver: Applied, thanks!

unbel13ver:
Testing instructions updated.

johannarautanen:
**Tested in Lenovo X1 with commit:** bfcb361c5262bafd9f14fbcc330f354b5cbe44cd

Screenshot from 2024-07-01 11-01-58

- ci-test-automation cases ok
- apps launched and manual regression cases ok

modules/disko/disko-ab-partitions.nix
```diff
@@ -0,0 +1,160 @@
...
+# - vm-storage-A : (30G) Possible standalone pre-built VM images are stored here
+# - vm-storage-B : (30G)
+# - reserved-A : (10G) Reserved partition, no use
+# - reserved-B : (10G)
+# - gp-storage : (50G) General purpose storage for some common insecure cases
+# - recovery : (rest of the ZFS pool) Recovery factory image is stored here
+#
+# Second ZFS pool is dedicated for Storage VM completely.
+{pkgs, ...}: {
+  networking.hostId = "8425e349";
```
vilvo, line 27: Future structural refactoring proposition - Should this be tagged with `TODO` as `networking.hostId` sounds like something that could be refactored to #668 with @tervis-unikie work? 
Especially when it's not used anywhere?
```
[vilvo@carrie:~/ghaf]$ grep -r hostId *
modules/disko/disko-ab-partitions.nix:  networking.hostId = "8425e349";
grep: result/disk1.raw: binary file matches
[vilvo@carrie:~/ghaf]$
```

modules/disko/disko-ab-partitions.nix
```diff
@@ -0,0 +1,160 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# !!! To utilize this partition scheme, the disk size must be >252G !!!
+#
+# This partition scheme contains three common partitions and two ZFS pools.
+# First LVM pool occupies 250G and second one occupies the rest of the disk space.
+# Some paritions are duplicated for the future AB SWupdate implementation.
```
vilvo, line 8: minor: `s/paritions/partitions/`

vilvo (approved):
Lgtm. Did not retest as @johannarautanen already confirmed verification.

modules/disko/disko-ab-partitions.nix
unbel13ver: Fixed, thanks for noticing

modules/disko/disko-ab-partitions.nix
unbel13ver: Added TODO comment in the code to not to forget about this.

modules/disko/disko-ab-partitions.nix
```diff
@@ -0,0 +1,161 @@
...
+        type = "disk";
+        imageSize = "15G";
+        content = {
+          type = "gpt";
+          partitions = {
+            boot = {
+              name = "boot";
+              size = "1M";
+              type = "EF02";
+              priority = 1; # Needs to be first partition
```
Mic92, line 47: I made this a default now. But it also doesn't hurt to leave it: https://github.com/nix-community/disko/pull/688/files#diff-a91aff802e945d3aaba098fc7593c18276073bfc7877604a41c463498c50b0c2R53

modules/hardware/laptop.nix
```diff
@@ -35,6 +35,13 @@ in {
     # Disk configuration
     disko.devices.disk = hwDefinition.disks;
 
+    # Hardware specific kernel parameters
+    boot = {
+      initrd.availableKernelModules = ["nvme" "zfs"];
+      supportedFilesystems = ["zfs"];
+      kernelPackages = config.boot.zfs.package.latestCompatibleLinuxPackages;
```
brianmcgillion, line 42: this section should be included in https://github.com/tiiuae/ghaf/blob/main/modules/hardware/x86_64-generic/x86_64-linux.nix#L27

brianmcgillion (changes requested):

modules/hardware/laptop.nix
unbel13ver: Done

brianmcgillion (approved):

clayhill66:
With the latest commit https://github.com/tiiuae/ghaf/pull/478/commits/9d7ab74cfd9defee141651ccf0b4e5810751fd39:
-automated cases are ok
-all apps launch - ok
-nixos rebuild does NOT work

johannarautanen:
Screenshot from 2024-07-10 06-10-01

vunnyso:
Reason for limited size is probably because of [disko-basic-postboot](https://github.com/tiiuae/ghaf/blob/main/modules/disko/disko-basic-postboot.nix) commands didn't work as expected due to change in partitioning scheme.

leivos-unikie:
Any idea if this PR can address this problem?
Current mainline ghaf boots partially from USB SSD and nvme if ghaf is installed (with ghaf-installer) to nvme and trying to boot ghaf from USB SSD.

IMG_2866

unbel13ver:
> With the latest commit [9d7ab74](https://github.com/tiiuae/ghaf/commit/9d7ab74cfd9defee141651ccf0b4e5810751fd39): -automated cases are ok -all apps launch - ok -nixos rebuild does NOT work

`nixos-rebuild` does not work for updating partition scheme, in this case the system needs to be re-installed.

> Reason for limited size is probably because of [disko-basic-postboot](https://github.com/tiiuae/ghaf/blob/main/modules/disko/disko-basic-postboot.nix) commands didn't work as expected due to change in partitioning scheme.

Yes, thanks for noticing this. The `disco-basic-postboot.nix` does not consider ZFS as a possible filesystem. I need to take a look into that.

unbel13ver:
> Any idea if this PR can address this problem? Current mainline ghaf boots partially from USB SSD and nvme if ghaf is installed (with ghaf-installer) to nvme and trying to boot ghaf from USB SSD.
> 
> IMG_2866

I am not sure the issue is related to this PR. NixOS picks the device to boot by its label, and in this case there are two devices with the same label (`nixos`). We got the same issue with Orin devices previously. The solution either to remove external drive or change the label of the internal partition.

leivos-unikie:
> > Any idea if this PR can address this problem? Current mainline ghaf boots partially from USB SSD and nvme if ghaf is installed (with ghaf-installer) to nvme and trying to boot ghaf from USB SSD.
> > IMG_2866
> 
> I am not sure the issue is related to this PR. NixOS picks the device to boot by its label, and in this case there are two devices with the same label (`nixos`). We got the same issue with Orin devices previously. The solution either to remove external drive or change the label of the internal partition.

Actually nowadays 'LABEL' is empty for every partition, they have 'PARTLABEL' though.

I made a test, changed the partlabels of partitions on nvme with gdisk (nix-shell -p gptfdisk). Then tried to boot ghaf from USB SSD. Still it booted partially from USB and nvme.

image

So it seems that the boot is not by LABEL nor by PARTLABEL. 

After installing ghaf to nvme the only way for booting ghaf from USB SSD is to boot first non-ghaf OS from USB SSD and wipe nvme. 

Is it ok to discuss this here, or should we move elsewhere...

leivos-unikie:
> So it seems that the boot is not by LABEL nor by PARTLABEL.

Well, not quite so. If there is no `disk-disk1-ESP` partlabel available on any device the boot fails with timeout (not finding /dev/disk/by-partlabel/disk-disk1-ESP). Other partlabels seem to be insignificant, for example:

[ghaf@ghaf-host:~]$ lsblk -o name,mountpoints,label,partlabel
NAME          MOUNTPOINTS
                         LABEL PARTLABEL
sda                            
├─sda1                         s1
├─sda2        /boot            disk-disk1-ESP
└─sda3                         s3
nvme0n1                        
├─nvme0n1p1                    p1
├─nvme0n1p2                    p2
└─nvme0n1p3                    p3
  └─pool-root /nix/store       
              /

unbel13ver:
ZFS partition is now extended during the first boot.
This example is when the system is booted from the external SSD (root is on the `/dev/sda`):

```
[ghaf@ghaf-host:~]$ lsblk
NAME    MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
sda       8:0    0 465.8G  0 disk 
├─sda1    8:1    0     1M  0 part 
├─sda2    8:2    0   500M  0 part /boot
├─sda3    8:3    0   500M  0 part 
└─sda4    8:4    0 464.8G  0 part 
nvme0n1 259:0    0 953.9G  0 disk 

[ghaf@ghaf-host:~]$ zpool list
NAME      SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
zfspool   464G  5.04G   459G        -         -     0%     1%  1.00x    ONLINE  -
```

milva-unikie:
### **Tested on Lenovo-X1**

Everything seems to be good with both installer and debug-image! (I did not test nixos-rebuild though, please let me know if that needs to be still checked)

- Installer installed to internal SSD
	- Installation works
	- Test-automation passes
	- The iso-image is a bit bigger than it was previously (8,5 GB now vs 7,3 GB previously)
	- Same testing results as in Testing-section except that `zroot_1` has been renamed to `zfspool`

- Debug-image installed to external SSD
	- Test-automation passes
	- `disk1.raw.zst` is a bit bigger than it was previously (4,5 GB now vs 3,8 GB previously), but the extracted image is the same size (16,1 GB)

(As a note for our testing team: The free space in host is reduced with this PR and there is no longer an excessive amount of free space to run fileio-tests. We need to decide how we want to proceed with that.)

PR #439, opened 2024-01-08
hardening: kernel config modularization

vunnyso:
* Created separate Kconfig file to enable virtualization on top of hardened baseline kernel.
* Using kernel merge_config.sh to merge kconfig fragment

## Checklist for things done

- [x] Summary of the proposed changes in the PR description
- [x] More detailed description in the commit message(s)
- [x] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [x] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] PR linked to architecture documentation and requirement(s) (ticket id)
- [x] Test procedure described (or includes tests). Select one or more:
  - [x] Tested on Lenovo X1 `x86_64`
  - [ ] Tested on Jetson Orin NX or AGX `aarch64`
  - [ ] Tested on Polarfire `riscv64`
- [x] Author has run `nix flake check --accept-flake-config` and it passes
- [x] All automatic Github Action checks pass - see [actions](https://github.com/tiiuae/ghaf/actions)
- [x] Author has added reviewers and removed PR draft status

## Testing
**This feature is only enabled for Lenovo X1.**

 *Tested following to work*
  nix build .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug
  nix build .#checks.x86_64-linux.module-test-hardened-generic-host-kernel
  nix build .#checks.x86_64-linux.module-test-hardened-lenovo-x1-guest-guivm-kernel
  nix build .#checks.x86_64-linux.module-test-hardened-pkvm-kernel
  nix develop .#devShells.x86_64-linux.kernel-x86
  [ghaf-kernel-x86-devshell:~/ghaf/linux-6.6.7]$ eval $checkPhase

**How to test**
By default kernel config modularization is disabled, its can be enabled using below configuration:
```
diff --git a/targets/lenovo-x1-carbon.nix b/targets/lenovo-x1-carbon.nix
index 35c35bd1c8..0850dea156 100644
--- a/targets/lenovo-x1-carbon.nix
+++ b/targets/lenovo-x1-carbon.nix
@@ -322,13 +322,13 @@
             ghaf = {
               hardware.definition = hwDefinition;
               # To enable guest hardening enable host hardening first
-              host.kernel.hardening.enable = false;
-              host.kernel.hardening.virtualization.enable = false;
[...]

PR #479, opened 2024-02-18
kernel hardening checker update

avnik:
## Description of changes

KHC tool is fast evolving, and bumping hash manually each time is inconvenient.
I propose using flake.nix/lock for managing kernel-hardening-checker version pinning.

## Checklist for things done

## Testing

`kernel-hardening-checker --version` show newer version, also checker works on our baseline config

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -141,6 +141,11 @@
         flake-compat.follows = "flake-compat";
       };
     };
+
+    kernel-hardening-checker = {
+      url = "github:a13xp0p0v/kernel-hardening-checker";
+      flake = false;
+    };
   };
 
   outputs = inputs @ {flake-parts, ...}: let
--- a/flake.lock
+++ b/flake.lock
@@ -221,6 +221,22 @@
         "type": "github"
       }
     },
+    "kernel-hardening-checker": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1705445702,
+        "narHash": "sha256-xpVazB9G0cdc0GglGpna80EWHZXfTd5mc5mTvvvoPfE=",
+        "owner": "a13xp0p0v",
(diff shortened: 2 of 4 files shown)


vilvo:
In general, this is a good idea. I propose we look at this again after #439 is merged and extend [the hardened kernel module tests here](https://github.com/tiiuae/ghaf/pull/439/files#diff-49f7a3ba809188883d2a60aef5bcd93529587f91ebb40004b7fc3f412feae79cR19-R22) to cover also the `kernel-hardening-checker` testing. Even if just run for the merged, hardened kernel configuration. That way we will find possible regression in the `kernel-hardening-checker`. If we only update the tool, we won't notice the regression.

flake.nix
```diff
@@ -141,6 +141,11 @@
         flake-compat.follows = "flake-compat";
       };
     };
+
+    kernel-hardening-checker = {
```
Mic92, line 145: Could be also integrated into the CI like this:

{
  runCommand "kernel-check" {} ''
    kernel-hardening-checker -c ${pkgs.hardened-kernel.configfile}
  '':
}

Mic92: Where the hardened-kernel is the attribute of this PR: https://github.com/tiiuae/ghaf/pull/439

avnik: We have plan for it -- make json report from KHC, then use custom script to assert checks (because KHC asserts on too much things, we need to ignore some checks like CONFIG_MODULE)

mikatammi:
Couldn't you simply make some script/tool to update the hash in the package file? I know there are even ready-made ones

Mic92:
That would also works. Shameless self-plug: https://github.com/Mic92/nix-update/ has support for flakes...

UnikieHR:
image

PR #390, opened 2023-12-03
Apps refactor

brianmcgillion:
Refactoring the apps to be more modular

WIP

## Checklist for things done

- [ ] Summary of the proposed changes in the PR description
- [ ] More detailed description in the commit message(s)
- [ ] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [ ] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] PR linked to architecture documentation and requirement(s) (ticket id)
- [ ] Test procedure described (or includes tests). Select one or more:
  - [ ] Tested on Lenovo X1 `x86_64`
  - [ ] Tested on Jetson Orin NX or AGX `aarch64`
  - [ ] Tested on Polarfire `riscv64`
- [ ] Author has run `nix flake check --accept-flake-config` and it passes
- [ ] All automatic Github Action checks pass - see [actions](https://github.com/tiiuae/ghaf/actions)
- [ ] Author has added reviewers and removed PR draft status

## Testing

<How this was tested by the author? How is this supposed to be tested by people doing system testing?>

PR #490, opened 2024-03-01
Refactor bit by bit

brianmcgillion:
## Description of changes

This change is the first in a series of changes to refactor the structure to be more modular and reusable. Currently the target files are used to define large parts of the system and as such they are not inheritable by any other downstream project.

This change starts on the Lenovo X1 and attempts to define the reusable parts before they are refactored into the correct place in the tree.

There is no new functionality to be tested. It just should not cause any regressions

## Checklist for things done


targets/lenovo-x1/everything.nix
```diff
@@ -271,19 +187,21 @@
...
-          ../modules/virtualization/microvm/netvm.nix
-          ../modules/virtualization/microvm/guivm.nix
-          ../modules/virtualization/microvm/appvm.nix
+          (import ../../modules/partitioning/lenovo-x1-disko-basic.nix {device = "/dev/nvme0n1";}) #TODO define device in hw def file
+          ../../modules/partitioning/disko-basic-postboot.nix
+          ../../modules/host
+          ../../modules/virtualization/microvm/microvm-host.nix
+          ../../modules/virtualization/microvm/netvm.nix
+          ../../modules/virtualization/microvm/guivm.nix
+          ../../modules/virtualization/microvm/appvm.nix
```
juliuskoskela, line 196: I wonder if we could refactor out these kinds of imports that import from one branch to another. In my own NixOS configs I import my modules at flake root level and then pass them to other branches as arguments. The advantages of this would be that we don't get so wedded into our folder structure and arguably it would enhance readability as well. We could have `ghafModules` that we pass along.

brianmcgillion: we can and we will. this patch set is called bit by bit for a reason :)
  
it will eventually become this
https://github.com/tiiuae/ghaf/pull/390/files#diff-24cbae53da3e4c7b899685295737b955daa048345f53ef85ab51b07270d7b31e

targets/lenovo-x1/releaseModules.nix
```diff
@@ -0,0 +1,5 @@
+[
```
mikatammi, line 1: This file is missing the SPDX stuff on

targets/lenovo-x1/debugModules.nix
```diff
@@ -0,0 +1,16 @@
+[
```
mikatammi, line 1: This file is missing the SPDX stuff

mikatammi (approved):
Tests pass and otherwise look fine expect for few missing SPDX headers

PR #497, opened 2024-03-06
Micro vm target refactor

brianmcgillion:
## Description of changes

This change does not modify any functionality, creats no new build. it only factors out some common components from the x1 target so that they could be used in other targets as required.

## Checklist for things done

  - [] Tested on Lenovo X1 `x86_64`

### DIFF
--- a/targets/lenovo-x1/everything.nix
+++ b/targets/lenovo-x1/everything.nix
@@ -20,29 +20,6 @@
   ## To here
 
   lenovo-x1 = variant: extraModules: let
-    netvmExtraModules = [
-      ({
-          pkgs,
-          config,
-          ...
-        }:
-        #TODO convert these to modules when the lenovox1 module is created and imports these
-        # also requires the hostConfiguration to be a module that can be imported
-          import ./netvmExtraModules.nix {inherit lib config pkgs microvm hostConfiguration;})
-    ];
-    guivmExtraModules = [
-      ({
-          pkgs,
-          config,
-          ...
-        }:
-        #TODO convert these to modules when the lenovox1 module is created and imports these
-        # also requires the hostConfiguration to be a module that can be imported
-          import ./guivmExtraModules.nix {inherit lib config pkgs microvm hostConfiguration;})
-    ];
-
-    xdgPdfPort = 1200;
-
     hostConfiguration = lib.nixosSystem {
       inherit system;
       specialArgs = {inherit lib;};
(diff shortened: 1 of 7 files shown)


targets/lenovo-x1/guivmExtraModules.nix
```diff
@@ -12,122 +13,150 @@
...
-    sockets."pdf" = {
-      unitConfig = {
-        Description = "PDF socket";
-      };
-      socketConfig = {
-        ListenStream = "${toString xdgPdfPort}";
-        Accept = "yes";
+
+  #TODO how to replace this flake-parts
+  configH = config;
```
mikatammi, line 22: This looks wrong. configH is supposed to be the config from Host. From this diff it is hard to see was it originally the way it was supposed to be.

mikatammi: I'll actually check with git in terminal to see if the levels on nesting are right

mikatammi: Okay looks like this is a real issue

targets/lenovo-x1/netvmExtraModules.nix
```diff
@@ -8,62 +8,75 @@
...
-  microvm = {
-    shares = [
-      {
-        tag = config.ghaf.security.sshKeys.waypipeSshPublicKeyName;
-        source = config.ghaf.security.sshKeys.waypipeSshPublicKeyDir;
-        mountPoint = config.ghaf.security.sshKeys.waypipeSshPublicKeyDir;
-      }
-    ];
+}: let
+  configH = config;
```
mikatammi, line 12: Either repleace this line with `configH = hostConfiguration.config` or just remove the `configH` altogether and replace later occurences of `configH` with `hostConfiguration.config`

targets/lenovo-x1/guivmExtraModules.nix
mikatammi, line 22: Either replace this line with `configH = hostConfiguration.config` or just remove the `configH` altogether and replace later occurences of `configH` with `hostConfiguration.config`

targets/lenovo-x1/everything.nix
```diff
@@ -124,140 +101,18 @@
...
-                        inherit (d) path;
-                      })
-                      configH.ghaf.hardware.definition.network.pciDevices
-                    );
-                  };
-                in
-                  [netvmPCIPassthroughModule]
-                  ++ netvmExtraModules;
+                extraModules =
+                  import ./netvmExtraModules.nix {inherit lib config pkgs microvm hostConfiguration;};
```
mikatammi, line 105: Actually I'm not sure, it looks like you pass in the host configuration twice, first as a config, then as a hostConfiguration.config

mikatammi: Maybe we could have a single parameter `configH` or `configHost` instead of `config` and `hostConfiguration`.

And then here we say inside the `import ./netvmExtraModules.nix { inherit lib pkgs microvm; configHost = config };`, to make it very explicit

mikatammi (approved):

PR #506, opened 2024-03-08
Initial minimal systemd for Ghaf

mbssrc:
## Description of changes

Initial PR for minimal systemd for Ghaf (x1 only at the moment). Based on NixOS systemdMinimal, it aims to provide a baseline systemd configuration with options to extend or limit the functionality per VM and host to the necessary minimal.

The implementation provides 2 things:
1. Systemd package configuration
2. Removal of unwanted units (in some cases necessary to allow built)

Hereby, most exposed options follow the respective NixOS systemd config flags. Some options are removing units only, or group flags together. Both package config and unit removal was done in anticipation of future removals and configuration, and will be subject to change.

While the original idea was also to reduce image size; this is not the case at the moment with a shared nix store compared to individual VM images. 

Going forward, this change may require some work on the developer side to ensure that appropriate systemd functionality is available.

## Checklist for things done

## Testing

Tested on x1:
- Startup and applications work with the exception of sound in chromium (pulseaudio blames snd_hda_intel), but this doesn't seem to work in current version at all.
- Note that the disablement of timesyncd in the VMs bypasses an issue with networkd that causes internal DNS failure 

Tested on AGX:
- Works, proper usage of host applications requires this change in microvm-host.nix:
```
        withNss = config.ghaf.hardware.nvidia.orin.enable;
        withResolved = config.ghaf.hardware.nvidia.orin.enable;
``` 
(and setting net-vm as default gw)

### DIFF
--- a/modules/common/systemd/base.nix
+++ b/modules/common/systemd/base.nix
@@ -1,0 +1,295 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  # Ghaf systemd config
+  cfg = config.ghaf.systemd;
+
+  # Override minimal systemd package configuration
+  package =
+    pkgs.systemdMinimal.override {
+      pname = cfg.withName;
+      withAcl = true;
+      withAnalyze = cfg.withDebug;
+      inherit (cfg) withApparmor;
+      inherit (cfg) withAudit;
+      withCompression = true;
+      withCoredump = cfg.withDebug || cfg.withMachines;
+      inherit (cfg) withCryptsetup;
+      inherit (cfg) withEfi;
+      withBootloader = cfg.withEfi; # systemd-boot fails if not explicity set
+      inherit (cfg) withFido2;
+      inherit (cfg) withHostnamed;
+      withImportd = cfg.withMachines;
+      withKexectools = cfg.withDebug;
+      withKmod = true;
+      withLibBPF = true;
+      withLibseccomp = true;
+      inherit (cfg) withLocaled;
+      inherit (cfg) withLogind;
+      withMachined = cfg.withMachines;
+      inherit (cfg) withNetworkd;
+      inherit (cfg) withNss;
+      withOomd = true;
+      withPam = true;
+      inherit (cfg) withPolkit;
+      inherit (cfg) withResolved;
+      withShellCompletions = cfg.withDebug;
+      withTimedated = true;
+      inherit (cfg) withTimesyncd;
+      inherit (cfg) withTpm2Tss;
+      withUtmp = cfg.withJournal || cfg.withAudit;
+    } # To be removed, current systemd version 254.6 < 255
+    // lib.optionalAttrs (lib.hasAttr "withVmspawn" (lib.functionArgs pkgs.systemd.override)) {
+      withVmspawn = cfg.withMachines;
+    };
+
+  # Definition of suppressed system units in systemd configuration. This removes the units and has priority.
+  # Required to avoid build failures compared to only disabling units for some options. Note that errors will be silently ignored.
+  suppressedSystemUnits =
+    [
+      ## Default disabled units
+      "remote-cryptsetup.service"
+      "remote-cryptsetup.target"
+      "remote-fs-pre.service"
+      "remote-fs-pre.target"
+      "remote-fs.service"
+      "remote-fs.target"
+      "rpcbind.service"
+      "rpcbind.target"
+      "systemd-update-done.service"
+      "system-update.target"
+      "system-update-pre.target"
+      "system-update-cleanup.service"
+    ]
+    ++ (lib.optionals (!cfg.withMachines) [
+      "container-getty.service"
+      "container-getty@.service"
+      "container@.service"
+      "systemd-nspawn@.service"
+    ])
+    ++ (lib.optionals ((!cfg.withDebug) && (!cfg.withSerial)) [
+      "getty.service"
+      "getty@.service"
+      "getty.target"
+      "getty-pre.target"
+      "serial-getty.service"
+      "serial-getty@.service"
+      "serial-getty.target"
+      "serial-getty@.target"
+    ])
+    ++ (lib.optionals ((!cfg.withDebug) && (!cfg.withJournal)) [
+      "systemd-journald-audit.socket"
+      "systemd-journal-catalog-update.service"
+      "systemd-journal-flush.service"
+      "systemd-journald.service"
+      "systemd-journald@.service"
+      "systemd-journal-gatewayd.socket"
+      "systemd-journald-audit.socket"
+      "systemd-journald-dev-log.socket"
+      "systemd-journald-varlink@.socket"
+      "systemd-journald.socket"
+      "systemd-journald@.socket"
+      "systemd-update-utmp.service"
+    ])
+    ++ (lib.optionals (!cfg.withAudit) [
+      "audit.service"
+      "auditd.service"
+      "systemd-journald-audit.socket"
+    ])
+    ++ (lib.optionals ((!cfg.withDebug) && (!cfg.withMachines)) [
+      "systemd-coredump.socket"
+    ])
+    ++ (lib.optionals (!cfg.withLogind) [
+      "systemd-logind.service"
+      "dbus-org.freedesktop.login1.service"
+    ])
+    ++ (lib.optionals (!cfg.withNss) [
+      "nscd.service"
+      "nss.service"
+      "nss.target"
+      "nss-lookup.target"
+      "nss-user-lookup.target"
+      "nss-lookup.target.requires"
+      "nss-user-lookup.target.requires"
+    ])
+    ++ (lib.optionals (!cfg.withTimesyncd) [
+      "systemd-timesyncd.service"
+    ])
+    ++ (lib.optionals (!cfg.withResolved) [
+      "systemd-resolved.service"
+    ])
+    ++ (lib.optionals (!cfg.withNetworkd) [
+      "network.target"
+      "network-pre.target"
+      "network-online.target"
+      "network-interfaces.target"
+      "network-setup.service"
+      "network-local-commands.service"
+      "systemd-timesyncd.service"
+      "systemd-networkd-wait-online.service"
+      "systemd-networkd.service"
+      "systemd-networkd.socket"
+    ])
+    ++ (lib.optionals (!cfg.withDebug) [
+      ## Units kept with debug
+      "kbrequest.target"
+      "rescue.service"
+      "rescue.target"
+      "emergency.service"
+      "emergency.target"
+      "systemd-vconsole-setup.service"
+      "reload-systemd-vconsole-setup.service"
+      "console-getty.service"
+      "sys-kernel-debug.mount"
+      "sys-fs-fuse-connections.mount"
+      "systemd-pstore.service"
+      "mount-pstore.service"
+      "systemd-ask-password-console.path"
+      "systemd-ask-password-console.service"
+      "systemd-ask-password-wall.path"
+      "systemd-ask-password-wall.service"
+      "systemd-kexec.service"
+      "kexec.service"
+      "kexec.target"
+      "kexec-tools.service"
+      "kexec-tools.target"
+      "prepare-kexec.service"
+      "prepare-kexec.target"
+    ]);
+in
+  with lib; {
+    options.ghaf.systemd = {
+      enable = mkEnableOption "Enable minimal systemd configuration.";
+
+      withName = mkOption {
+        description = "Set systemd name.";
+        type = types.str;
+        default = "base-systemd";
+      };
+
+      withLogind = mkOption {
+        description = "Enable systemd login daemon.";
+        type = types.bool;
+        default = true;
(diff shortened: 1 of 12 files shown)


modules/common/systemd/base.nix
```diff
@@ -0,0 +1,295 @@
...
+      inherit (cfg) withAudit;
+      withCompression = true;
+      withCoredump = cfg.withDebug || cfg.withMachines;
+      inherit (cfg) withCryptsetup;
+      inherit (cfg) withEfi;
+      withBootloader = cfg.withEfi; # systemd-boot fails if not explicity set
+      inherit (cfg) withFido2;
+      inherit (cfg) withHostnamed;
+      withImportd = cfg.withMachines;
+      withKexectools = false;
```
vilvo, line 28: This is kind of potential minor opportunity to smooth development in `-debug` builds because `kexec` enables `nixos-anywhere` and could potentially allow re-running the ghaf installer directly without rebooting to installer. Take it with "I'm not sure if this kind of functionality could be added to the ghaf installer. If yes, this could be `withKexectools = cfg.withDebug`

modules/common/systemd/base.nix
```diff
@@ -0,0 +1,295 @@
...
+      inherit (cfg) withTimesyncd;
+      inherit (cfg) withTpm2Tss;
+      withUtmp = cfg.withJournal || cfg.withAudit;
+    } # To be removed, current systemd version 254.6 < 255
+    // lib.optionalAttrs (lib.hasAttr "withVmspawn" (lib.functionArgs pkgs.systemd.override)) {
+      withVmspawn = cfg.withMachines;
+    };
+
+  # Definition of suppressed system units in systemd configuration. This removes the units and has priority.
+  # Required to avoid build failures compared to only disabling units for some options. Note that errors will be silently ignored.
```
vilvo, line 52: Could there be a way to log the errors with `-debug` builds?

modules/common/systemd/base.nix
```diff
@@ -0,0 +1,295 @@
...
+      "console-getty.service"
+      "sys-kernel-debug.mount"
+      "sys-fs-fuse-connections.mount"
+      "systemd-pstore.service"
+      "mount-pstore.service"
+      "systemd-ask-password-console.path"
+      "systemd-ask-password-console.service"
+      "systemd-ask-password-wall.path"
+      "systemd-ask-password-wall.service"
+    ]);
```
vilvo, line 163: Regarding lines https://github.com/tiiuae/ghaf/pull/506/files#diff-8799260fcdef43f44d9f5a097ac22bbc9122cada55afae08610edf81ccf480fbR53-R163
I feel the `optionals` lists of units to be removed could be refactored into files that are then imported for unit removal from each VM. This is kind of opinionated comment as I find (multiple) exclusive conditions a tad hard to read because one would have to go to the VM declarations to check which config options apply to which VM. If the systemd unit exclusions were imported and declared in the VMs, it could be clearer. We did conceptually something similar with hardened kernel configuration fragments (separate files) based on config option declarations.

So, I guess I'm writing this down as something to consider in further iterations.

vilvo (changes requested):
Builds but the image fails to boot from SSD.
```
git checkout manuel/systemd-minimal
git log --oneline -1
7534792 (HEAD, manuel/systemd-minimal) Initial minimal systemd for Ghaf
nix build .#lenovo-x1-carbon-gen11-debug
sudo dd if=result/disk1.raw of=/dev/sdc bs=1M status=progress
sudo sync
<detach and boot X1 from the SSD>
```
screencapture
I double checked.

mbssrc:
> Builds but the image fails to boot from SSD.
> 
> ```
> git checkout manuel/systemd-minimal
> git log --oneline -1
> 7534792 (HEAD, manuel/systemd-minimal) Initial minimal systemd for Ghaf
> nix build .#lenovo-x1-carbon-gen11-debug
> sudo dd if=result/disk1.raw of=/dev/sdc bs=1M status=progress
> sudo sync
> <detach and boot X1 from the SSD>
> ```
> 
> screencapture I double checked.

I encountered the same issue when using the SSD image, so I checked with a vanilla Ghaf version and had the same issue, so my guess is that it is unrelated. Could you check and verify that generic build .#lenovo-x1-carbon-gen11-debug works for you without this PR?

vilvo:
...
> I encountered the same issue when using the SSD image, so I checked with a vanilla Ghaf version and had the same issue, so my guess is that it is unrelated. Could you check and verify that generic build .#lenovo-x1-carbon-gen11-debug works for you without this PR?

You are right - `a38ac1c (HEAD, origin/main, origin/HEAD) labwc: fix clock in waybar` fails the same way.

vilvo:
> Installer and remote build worked fine.

Installer build of 7534792 fails with:
```
nix build .#lenovo-x1-carbon-gen11-debug-installer
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'ghaf-23.11.20231218.3a9928d-x86_64-linux.iso'
         whose name attribute is located at /nix/store/0082jh2mcly1y2smsv5r8bk4al3px0zb-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'sources' of derivation 'ghaf-23.11.20231218.3a9928d-x86_64-linux.iso'

         at /nix/store/0082jh2mcly1y2smsv5r8bk4al3px0zb-source/nixos/lib/make-iso9660-image.nix:57:3:

           56|
           57|   sources = map (x: x.source) contents;
             |   ^
           58|   targets = map (x: x.target) contents;

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: getting status of '/nix/store/s6p66ggmkv94i7h41vj41w6w0klax5zw-source/targets/packages': No such file or directory
```

mbssrc:
> > Installer and remote build worked fine.
> 
> Installer build of [7534792](https://github.com/tiiuae/ghaf/commit/75347927db9e59885f3df6cfab2dfdc048bb299c) fails with:
> 
> ```
> nix build .#lenovo-x1-carbon-gen11-debug-installer
> error:
>        … while calling the 'derivationStrict' builtin
> 
>          at /builtin/derivation.nix:9:12: (source not available)
> 
>        … while evaluating derivation 'ghaf-23.11.20231218.3a9928d-x86_64-linux.iso'
>          whose name attribute is located at /nix/store/0082jh2mcly1y2smsv5r8bk4al3px0zb-source/pkgs/stdenv/generic/make-derivation.nix:348:7
> 
>        … while evaluating attribute 'sources' of derivation 'ghaf-23.11.20231218.3a9928d-x86_64-linux.iso'
> 
>          at /nix/store/0082jh2mcly1y2smsv5r8bk4al3px0zb-source/nixos/lib/make-iso9660-image.nix:57:3:
> 
>            56|
>            57|   sources = map (x: x.source) contents;
>              |   ^
>            58|   targets = map (x: x.target) contents;
> 
>        (stack trace truncated; use '--show-trace' to show the full trace)
> 
>        error: getting status of '/nix/store/s6p66ggmkv94i7h41vj41w6w0klax5zw-source/targets/packages': No such file or directory
> ```

Think this is the `installScript = pkgs.callPackage ../packages/installer {` that should be `installScript = pkgs.callPackage ../../packages/installer {` - forgot to include that here

vilvo:
> Think this is the `installScript = pkgs.callPackage ../packages/installer {` that should be `installScript = pkgs.callPackage ../../packages/installer {` - forgot to include that here

Thanks. Got it working with
```
ghaf (7534792) [$!?] via ❄️  impure (Ghaf-devshell-env)
❯ git diff
diff --git a/targets/lenovo-x1-installer/flake-module.nix b/targets/lenovo-x1-installer/flake-module.nix
index 8125da7..e0de741 100644
--- a/targets/lenovo-x1-installer/flake-module.nix
+++ b/targets/lenovo-x1-installer/flake-module.nix
@@ -20,7 +20,7 @@
           modulesPath,
           ...
         }: let
-          installScript = pkgs.callPackage ../packages/installer {
+          installScript = pkgs.callPackage ../../packages/installer {
             inherit imagePath;
           };
         in {
```
Builds the installer and installs the PR build. 

PR `-debug` build `ghaf-host` `systemctl` shows `137 loaded units listed`
main `-debug` build `ghaf-host` `systemctl` shows `145 loaded units listed`

The basics on desktop seem to work.

modules/common/systemd/base.nix
vilvo: I had a discussion with @TanelDettenborn over lunch on possibilities to quickly build and test kernel changes without modifying them by building to nix-store and updating boot entries everytime. `kexec` was the only thing that came to my mind so I'd keep that `cfg.withDebug` to support development.

mbssrc: Ok will add that to debug

modules/common/systemd/base.nix
```diff
@@ -0,0 +1,295 @@
...
+      withFido2 = mkOption {
+        description = "Enable systemd Fido2 token functionality.";
+        type = types.bool;
+        default = false;
+      };
+
+      withTpm2Tss = mkOption {
+        description = "Enable systemd TPM functionality.";
+        type = types.bool;
+        default = false;
```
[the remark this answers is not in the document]
mbssrc: Can you elaborate if/how systemd's tpm2tss is required for the current TPM module? It seems like abrmd is standalone and not dependent on systemd (correct me if I am wrong). Of course once we use any systemd functionality requiring TPM this needs to be enabled (e.g., measure, creds, cryptsetup/enroll).

vilvo (approved):
LGTM. 

(Squashing of small commits from yesterday into one would be nice.)

modules/common/systemd/base.nix
mbssrc: No I don't think so - the units are filtered with the list of suppressedUnits, so any string therein will work without errors.
`enabledUnits = filterAttrs (n: v: ! elem n cfg.suppressedSystemUnits) cfg.units;`

mbssrc: No I don't think so - the units are filtered with the list of suppressedUnits, so any string therein will work without errors.
`enabledUnits = filterAttrs (n: v: ! elem n cfg.suppressedSystemUnits) cfg.units;`

Mic92 (approved):

milva-unikie (commented):
I was asked to test this with Lenovo-X1 on Labwc

- All of the apps open from icons
- Sound is working on Chromium
- A few test cases from ci-test-automation failed
	- Time does not get synchronized with internet connection
	- (Automated tests for opening apps failed. That is to be expected because the tests have been written for Weston.)
- Performance tests are ok

mbssrc:
> I was asked to test this with Lenovo-X1 on Labwc
> 
>     * All of the apps open from icons
> 
>     * Sound is working on Chromium
> 
>     * A few test cases from ci-test-automation failed
>       
>       * Time does not get synchronized with internet connection
>       * (Automated tests for opening apps failed. That is to be expected because the tests have been written for Weston.)
> 
>     * Performance tests are ok

Thanks for testing. Time doesn't get synced because timesyncd is currently disabled everywhere but the host. Time sync in apps caused  some issues with name resolution/networkd in my setup. What exactly is the time that is tested - in the guivm? Should be fine to re-enable that until we have a better system time solution.

barnabakos:
automated tests look to be fine on orin AGX

milva-unikie:
> Thanks for testing. Time doesn't get synced because timesyncd is currently disabled everywhere but the host. Time sync in apps caused some issues with name resolution/networkd in my setup. What exactly is the time that is tested - in the guivm? Should be fine to re-enable that until we have a better system time solution.

The automated test are actually only testing time in host. 

The time in host used to get synchronized with ethernet connection and time in vms with wifi connection (`timedatectl -a` should show that system clock was synchronized). The automated tests are run with ethernet.

The time shown on desktop is connected to gui-vm. Without the time updating also there, user might see wrong time. For example my X1 is currently running 20 minutes behind and with this pr will not get updated to the correct time.

jrautane:
Run the manual regression tests with Orin NX. No new findings during testing.

milva-unikie:
Polarfire is also ok. Automated tests passed.

milva-unikie:
> Thanks for the clarification, the reason it doesn't work is (presumably) that the host here cannot resolve the ntp dns name. Instead of adding that in again, I removed timesycnd from the x1 host as our goal is to remove host (and especially network) functionality. So the test should be removed for the x1 target as well. For the other targets, it is still in the host after the change.
> 
> Good point with the clock, I added timesyncd into the guivm so the clock is getting updated. Works for me on Weston, would appreciate if you can double check on labwc.

I confirmed with Labwc that time is now syncing correctly in gui-vm.

Also I let the automated testing team to know that there will be changes needed to the time tests.

mikatammi:
I think we should leave the `hostNetworking` enabled for now, at least in debug builds. Here's PR to do that: https://github.com/tiiuae/ghaf/pull/520

This does not only break timesyncd, but big parts of the host networking on X1 Carbon, because the DNS related systemd-resolved is also disabled :(

mbssrc:
> I think we should leave the `hostNetworking` enabled for now, at least in debug builds. Here's PR to do that: #520
> 
> This does not only break timesyncd, but big parts of the host networking on X1 Carbon, because the DNS related systemd-resolved is also disabled :(

Yes, the point was to remove host network services. PR #520 enables it in both debug and release. I'll keep it in mind to be removed again when we focus on getting release working...

PR #513, opened 2024-03-13
Add gen10 target and refactor hardware definition

mbssrc:
## Description of changes

PR for discussion. Changes:

- Added x1-gen10 target
- Refactored hwDefinition 
- Parametrized udev rule 
- Generalized USB camera 'passthrough'
- Simplified udev rules to include all possible mouse/touchpad hw for a generation

## Checklist for things done


targets/lenovo-x1/appvms/chromium.nix
```diff
@@ -53,6 +48,11 @@ in {
         "intel-hda"
         "-device"
         "hda-duplex,audiodev=pa1"
+        # Lenovo X1 integrated usb webcam
+        "-device"
+        "qemu-xhci"
+        "-device"
+        "usb-host,hostbus=3,hostport=8"
```
vilvo, line 55: Could this inherit these values from `modules/common/hardware/lenovo-x1/definitions/x1-<gen>.nix` ?
The bus and port values per hardware device are not there yet but could be. That way we could continue decoupling the chromium appvm from the hardware though there's some mode work to do with audio devices but I think the future development of audio architecture would allow that.

vilvo (approved):
Looks good to me and definitely improves device SKU declarations. The comment on USB host bus and port is not a blocker imo.

modules/common/hardware/lenovo-x1/definitions/default.nix
```diff
@@ -0,0 +1,50 @@
...
+      '');
+  in ''
+    # Laptop keyboard
+    SUBSYSTEM=="input", ATTRS{name}=="AT Translated Set 2 keyboard", GROUP="kvm"
+    # Laptop TrackPoint
+    SUBSYSTEM=="input", ATTRS{name}=="TPPS/2 Elan TrackPoint", GROUP="kvm"
+    # Lenovo X1 integrated webcam
+    KERNEL=="3-8", SUBSYSTEM=="usb", ATTR{busnum}=="3", ATTR{devnum}=="3", GROUP="kvm"
+    # Mouse and Touchpad
+    ${lib.strings.concatStrings (mapMouseRules hwDefinition.mouse)}
```
Mic92, line 38: We potentially also want to include Human interface devices in generally here?

```
Bus 001 Device 043: ID 046d:c548 Logitech, Inc. Logi Bolt Receiver
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0
  bMaxPacketSize0        64
  idVendor           0x046d Logitech, Inc.
  idProduct          0xc548 Logi Bolt Receiver
  bcdDevice            5.01
  iManufacturer           1 Logitech
  iProduct                2 USB Receiver
  iSerial                 0
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0054
    bNumInterfaces          3
    bConfigurationValue     1
    iConfiguration          4
    bmAttributes         0xa0
      (Bus Powered)
      Remote Wakeup
    MaxPower               98mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device <-----
```

Mic92: There is ID_TYPE=hid, but not sure if it's too broad. It would also include my jabra headphone dongle.

Mic92: `ID_INPUT_KEYBOARD=1` and `ID_INPUT_MOUSE=1` might work.

Mic92: But instead of adding the KVM group here, could the VM that is responsible for input not be in the input group? This one already exists.

modules/common/hardware/lenovo-x1/definitions/default.nix
```diff
@@ -0,0 +1,50 @@
...
+  # Notes:
+  #   1. This assembles udev rules for different hw configurations (i.e., different mice/touchpads) by adding
+  #      all of them to the configuration. This was chosen for simplicity to not have to provide hw identifier at build,
+  #      but is not ideal and should be changed.
+  #   2. USB camera "passthrough" is handled by qemu and thus available on host. If peripheral VM is implemented,
+  #      the entire host controller should be passthrough'd using the PCI bus (14.0). In x1, bluetooth and fingerprint
+  #      reader are on this bus.
+  udevRules = let
+    mapMouseRules =
+      builtins.map (d: ''        SUBSYSTEM=="input", ATTRS{name}=="${d}", KERNEL=="event*", GROUP="kvm", SYMLINK+="mouse"
```
mikatammi, line 25: I know this is a refactor commit, but interestingly now as code gets more clear to read, some things become more visible.

I was thinking could there be something better than the `name` attribute to match to, like combination of `idVendor` and `idProduct`? I'm just thinking the name probably gets read from some database, which might change in the future, but at least the USB vendor id and product id should stay same.

mikatammi: but this can also be fixed in some future commit

mikatammi (approved):
Briefly tested and seems to work

targets/lenovo-x1/appvms/chromium.nix
mbssrc: Yes that could be an option. I think it really depends on how we deal with different hardware definitions in the future - will we support different generations and/or HW SKUs as targets? See also the comment below.

modules/common/hardware/lenovo-x1/definitions/default.nix
mbssrc: I think at the moment all the VMs have the same user microvm and require the input, hence the adding to KVM group for microvm user, but adding microvm to input may also be an option. ID_INPUT_* may actually a good solution here - haven't really looked into that here. 
If we move to a dedicated peripheral VM, the host passthroughs+udev rules become trivial for usb devices if we simply pass the controllers.

brianmcgillion (approved):

PR #517, opened 2024-03-15
disko: Enable Luks disk encryption

vunnyso:
This patch creates luks filesystem which is encrypted and can be decrypted using Yubikey or system password.

## Checklist for things done

## Testing
Please follow the below steps:
1) After fresh installation on first boot system will ask for Yubikey PIN:
image
2) After which need to tap Yubikey 2 times (Repeat steps 1 and 2 depending on how many devices need to encrypted)
3) System will automatically continue to boot up, please check in ghaf-host terminal you will see lvm block devices encrypted.
 
image

4) On Next boot user need to enter Yubikey PIN once + tap on Yubikey device (depending on devices to be decrypted).
image

## Improvements
-  For encryption depending on number of devices need to encrypted user need to enter Yubikey PIN each time + double tap the Yubikey.
- For decryption user need to enter Yubikey PIN once + tap once Yubikey for each device which need to be decrypted.
- More block devices can be added later.

### DIFF
--- a/modules/common/hardware/definition.nix
+++ b/modules/common/hardware/definition.nix
@@ -84,11 +84,18 @@
             Path to the disk
           '';
         };
+        options.imageSize = mkOption {
+          type = types.str;
+          description = ''
+            Size of the image
+          '';
+        };
       });
       default = {};
       example = literalExpression ''
         {
           disk1.device = "/dev/nvme0n1";
+          disk1.imageSize = "100G"
         }
       '';
     };
--- a/modules/common/hardware/lenovo-x1/definitions/default.nix
+++ b/modules/common/hardware/lenovo-x1/definitions/default.nix
@@ -9,9 +9,16 @@
 in {
   inherit (hwDefinition) mouse;
   inherit (hwDefinition) touchpad;
-  inherit (hwDefinition) disks;
   inherit (hwDefinition) network;
   inherit (hwDefinition) gpu;
+
+  disks = {
+    disk1.device = "/dev/nvme0n1";
+    # 250G is the size of the 1st LVM pool, 10G is reserved for the second LVM pool
+    # Second LVM pool can be extended safely to the end of the disk, and its restricted size
+    # makes flashing quicker.
+    disk1.imageSize = "260G";
+  };
 
   # Notes:
   #   1. This assembles udev rules for different hw configurations (i.e., different mice/touchpads) by adding
--- a/modules/common/hardware/lenovo-x1/definitions/x1-gen10.nix
+++ b/modules/common/hardware/lenovo-x1/definitions/x1-gen10.nix
@@ -7,10 +7,6 @@
   mouse = ["ELAN067B:00 04F3:31F8 Mouse" "SYNA8016:00 06CB:CEB3 Mouse"];
   touchpad = ["ELAN067B:00 04F3:31F8 Touchpad" "SYNA8016:00 06CB:CEB3 Touchpad"];
 
-  disks = {
-    disk1.device = "/dev/nvme0n1";
-  };
-
   network.pciDevices = [
     {
       # Passthrough Intel WiFi card
--- a/modules/common/hardware/lenovo-x1/definitions/x1-gen11.nix
+++ b/modules/common/hardware/lenovo-x1/definitions/x1-gen11.nix
@@ -16,10 +16,6 @@
     "ELAN067B:00 04F3:31F8 Touchpad"
   ];
 
-  disks = {
-    disk1.device = "/dev/nvme0n1";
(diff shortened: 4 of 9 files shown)


modules/disko/lenovo-x1-disko-luks.nix
```diff
@@ -0,0 +1,9 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{pkgs, ...}: {
+  environment.systemPackages = [pkgs.fido2luks];
+  boot.initrd.luks.fido2Support = true;
+  boot.initrd.luks.devices.encrypted.fido2.credentials = ["b610d7d28dc9f3a87a643e2be6f35df7c7faaa09c2439bbc3679ffd07b726c46898e49f7ac9a1feeed872cea91d830b8"];
```
Mic92, line 6: Is this option needed for fido to figure out the right device to use?

vilvo: I understand `credentials` allow multiple [fidos per luks](https://search.nixos.org/options?channel=23.11&show=boot.initrd.luks.devices.%3Cname%3E.fido2.credentials&from=0&size=50&sort=relevance&type=packages&query=fido2).
This raises good questions on:
- if we could include step 1. to the ghaf installer so that some kind of guided wizard would instruct installer to enroll the fido device(s) without manually copying the credential
- how we want to define the installer to guide fido device(s) enrollment? I think it could make sense to guide (or at least provide an option) the user to enroll two fido devices. If one is lost... Like the saying goes - two is one and one is none.

vunnyso: > Is this option needed for fido to figure out the right device to use?

Yes as per my understanding credentials option needed to open luks encrypted drive and key should be same what we have used with `fido2luks add-key ...` command. So during NixOS stage 1 there will be decryption/device opening happening [here](https://github.com/NixOS/nixpkgs/blob/nixos-23.11/nixos/modules/system/boot/luksroot.nix#L470-L471)

vunnyso: > I understand `credentials` allow multiple [fidos per luks](https://search.nixos.org/options?channel=23.11&show=boot.initrd.luks.devices.%3Cname%3E.fido2.credentials&from=0&size=50&sort=relevance&type=packages&query=fido2). This raises good questions on:
> 
> * if we could include step 1. to the ghaf installer so that some kind of guided wizard would instruct installer to enroll the fido device(s) without manually copying the credential
> * how we want to define the installer to guide fido device(s) enrollment? I think it could make sense to guide (or at least provide an option) the user to enroll two fido devices. If one is lost... Like the saying goes - two is one and one is none.

Correct @vilvo if GUI/Wizard installer way will be more user friendly. Or maybe we can have some app in `gui-vm` to view/modify fido2 credentials at later stages of use in terms of adding, deleting as per needs of user.

brianmcgillion:
I believe this has to be done in an installation phase? Either choose passwd or yubikey. Because rebuilding an image per installation is not really viable.

If full luks can't be done on the fly in the installer, can we prompt for a key rotation (i.e. force a password change on first boot, and give option for either passwd or yubikey)

Secondly, shouldn't the passwd,/yubikey be mixed with tpm on the device? Or does fido2luks do this behind the scene?

vunnyso:
> I believe this has to be done in an installation phase? Either choose passwd or yubikey. Because rebuilding an image per installation is not really viable.
> 
> If full luks can't be done on the fly in the installer, can we prompt for a key rotation (i.e. force a password change on first boot, and give option for either passwd or yubikey)
> 
> Secondly, shouldn't the passwd,/yubikey be mixed with tpm on the device? Or does fido2luks do this behind the scene?

Hi @remimimimimi can we do Luks encryption along with yubikey enrollment during installation phase? 
Hi @TanelDettenborn can you please comment about tpm question?

vunnyso:
Hi @Mic92 please help to review my commit https://github.com/tiiuae/ghaf/pull/517/commits/b11ed1ca92558cff00d57379f4a65c4a52341db8.

vunnyso:
Currently put on hold, as I was planning to change ZFS datasets declared [here](https://github.com/tiiuae/ghaf/blob/main/modules/disko/disko-ab-partitions.nix#L109-L115) from `zfs_fs` to `zfs_volume` as `zfs_fs` cannot be encrypted runtime using LUKS.  Need discussion later with @unbel13ver

UnikieHR:
image

vunnyso:
Not valid, since we moved to ZFS partitioning scheme.

PR #523, opened 2024-03-19
Add perf and unixbench to debug tools

leivos-unikie:
## Description of changes

Add perf and unixbench to debug-tools.nix. Need to exclude unixbench from RISCV because runtimeShell is not enabled for that target.

## Checklist for things done

- [x] Author has run `nix flake check --accept-flake-config` and it passes

## Testing

These tools are meant for performance monitoring, mainly for Q&A team. 

Perf was added especially for RISCV Icicle-kit target but it was tested also on Lenovo-X1. On Icicle-kit custom perf script will be used (to get execution time decreased):
`{ perf bench sched messaging; perf bench sched pipe -l 50000; perf bench syscall basic; perf bench mem memcpy; perf bench mem memset; perf bench mem find_bit -i 5 -j 1000; perf bench numa mem -p 1 -t 1 -P 1024 -C 0 -M 0 -s 5 -zZq --thp 1 --no-data_rand_walk; perf bench futex all; perf bench epoll wait; perf bench epoll ctl; perf bench internals synthesize -i 1000; perf bench internals kallsyms-parse -i 10; } | tee -a perf_results`

The plan is to run unixbench (default) index test once per month on ghaf-host and on net-vm for Lenovo-X1 and Orin AGX/NX. Unixbench is not included to cross-compiled targets since they lack runtimeShell. Run the index test set with command:
`ubench`

Single unixbench index run takes around 0,5h. In multi-core environment it takes 1h (1 parallel process / multiple parallel processes).

### DIFF
--- a/modules/common/development/debug-tools.nix
+++ b/modules/common/development/debug-tools.nix
@@ -37,16 +37,13 @@ in
           # Performance testing
           speedtest-cli
           iperf
-
-          # Let's have this fixed version according to kernel.
-          # It would be possible to select also latest producing currently (12-3-2024) perf version 6.6.7
-          # linuxPackages_latest.perf
-          linuxKernel.packages.linux_6_1.perf
+          # Match perf version with kernel.
+          config.boot.kernelPackages.perf
         ]
         # TODO Can this be changed to platformPkgs to filter ?
         # LuaJIT (which is sysbench dependency) not available on RISC-V
-        ++ lib.lists.optionals (config.nixpkgs.hostPlatform.system != "riscv64-linux") [sysbench]
+        ++ lib.optional (config.nixpkgs.hostPlatform.system != "riscv64-linux") sysbench
         # runtimeShell (unixbench dependency) not available on RISC-V nor on cross-compiled Orin AGX/NX
-        ++ lib.lists.optionals (stdenv.hostPlatform == stdenv.buildPlatform) [unixbench];
+        ++ lib.optional (stdenv.hostPlatform == stdenv.buildPlatform) unixbench;
     };
   }
--- a/modules/common/development/debug-tools.nix
+++ b/modules/common/development/debug-tools.nix
@@ -37,10 +37,16 @@ in
           # Performance testing
           speedtest-cli
           iperf
+
+          # Let's have this fixed version according to kernel.
+          # It would be possible to select also latest producing currently (12-3-2024) perf version 6.6.7
+          # linuxPackages_latest.perf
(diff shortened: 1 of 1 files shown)


leivos-unikie:
Cross-compiled Orin builds are failing, the reason seems to be
`error: attribute 'runtimeShell' missing`

@mikatammi, any idea how to fix this or should I try to exclude unixbench from those targets with lib.optional?

modules/common/development/debug-tools.nix
```diff
@@ -37,10 +37,16 @@ in
           # Performance testing
           speedtest-cli
           iperf
+
+          # Let's have this fixed version according to kernel.
+          # It would be possible to select also latest producing currently (12-3-2024) perf version 6.6.7
+          # linuxPackages_latest.perf
+          linuxKernel.packages.linux_6_1.perf
```
flokli, line 44:
```suggestion
          config.boot.kernelPackages.perf
```

This automatically pulls in perf for the specific kernel the system is built with.

modules/common/development/debug-tools.nix
```diff
@@ -37,10 +37,16 @@ in
...
         ]
-        ++
-        # LuaJIT (which is sysbench dependency) not available on RISC-V
         # TODO Can this be changed to platformPkgs to filter ?
-        lib.optional (config.nixpkgs.hostPlatform.system != "riscv64-linux") sysbench;
+        # LuaJIT (which is sysbench dependency) not available on RISC-V
+        ++ lib.lists.optionals (config.nixpkgs.hostPlatform.system != "riscv64-linux") [sysbench]
```
vunnyso, line 48: If you are adding just one package, then please use
`lib.optional (config.nixpkgs.hostPlatform.system != "riscv64-linux") sysbench;`

Same goes to below package.

modules/common/development/debug-tools.nix
leivos-unikie: Thanks. Changed to that.

modules/common/development/debug-tools.nix
leivos-unikie: Changed back to lib.optional. I had that lib.lists.optionals because first I was adding two packages with single conditional.

milva-unikie (approved):
Checked that `ubench` is working for Lenovo-X1 and Orin-NX. Automated tests also passed for these targets.

Cross-compiled NX boots and there `ubench` is not working as was expected.

Perf is working on Icicle-kit with the command given in Testing-section.

brianmcgillion (approved):

PR #541, opened 2024-04-03
Hardware definition refactoring

mbssrc:
## Description of changes

This PR introduces

1. Refactoring of hardware definition:
	- Moving hardware from common into a dedicated flake-module. This implies that hardware definition is no longer part of the VM namespace by default. 
	- Re-factoring `generation` parameter, which is now an option
2. Moving TPM related code to generic-x86_64 as hardware dependent module
3. Enabling fingerprint reader in the `guivm`

## Checklist for things done

## Testing

Tested by author:

1. Build + run `.#lenovo-x1-carbon-gen11-debug`
2. Builld + run `.#vm-debug`
3. Build + run `.#generic-x86_64-debug` (on x1)
4. Verified TPM stack running on x1 host
5. Fingerprint: (more info in commit message)
	- tested enrollment (+multiple fingers)
	- tested verification: sudo, systemd, and swaylock

Test instructions:
1. Normal build + run on all x86-based targets with functionality test
2. TPM: Verify that abrmd service runs on host
3. Fingerprint:
    - In terminal, use `fprintd-enroll` to enroll fingerprint. Run a `sudo` command and `systemctl restart systemd-timesyncd.service`, authenticate with fingerprint. In-depth testing not required at this point, as features are missing (see commit message). Expected outcome: fingerprint enrollment works, sudo and systemctl (system) commands can be authenticated with fingerprint
    - To test swaylock, enable `config.ghaf.graphics.labwc.lock.enable` and build, enroll fingerprint, then run `swaylock` (e.g., run in terminal), press Enter and use fingerprint reader to authenticate. Expected outcome: fingerprint authentication unlocks swaylock


modules/hardware/lenovo-x1/modules/fprint.nix
```diff
@@ -0,0 +1,94 @@
...
+                // Allow user to enroll fingerprints
+                polkit.addRule(function(action, subject) {
+                if (action.id == "net.reactivated.fprint.device.enroll" &&
+                    subject.user == "ghaf") {
+                  return polkit.Result.YES;
+                  }
+                });
+              '';
+            };
+            # PAM rules for swaylock fingerprint reader
```
Mic92, line 71: I assume that the default pam configuration for fprintd in NixOS did not quite work?

targets/lenovo-x1/guivmExtraModules.nix
```diff
@@ -136,24 +136,28 @@
...
+        "-device"
+        "battery"
+        # Lenovo X1 AC adapter
+        "-device"
+        "acad"
+        # Connect sound device to hosts pulseaudio socket
+        "-audiodev"
+        "pa,id=pa1,server=unix:/run/pulse/native"
+      ]
+      ++ lib.optionals configH.ghaf.hardware.fprint.enable configH.ghaf.hardware.fprint.qemuExtraArgs;
```
Mic92, line 154: More of a general comment and not scope of this pull request, but it might make sense to use [toGNUCommandLine](https://github.com/NixOS/nixpkgs/blob/0d28066770464d19d637f6e8e42e8688420b6ac6/lib/cli.nix#L31) for qemu extraArgs at some point given in how many places it is set. The thinking is that than one could make use of the module system and do overrides in one module for where default was set in a different one.

modules/desktop/graphics/hardware.nix
```diff
@@ -0,0 +1,14 @@
+# Copyright 2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{lib, ...}:
+with lib; {
+  options.ghaf.graphics.hardware = {
+    networkDevice = mkOption {
```
Mic92, line 6: To me it looks like this is plural:

```suggestion
    networkDevices = mkOption {
```

modules/desktop/graphics/hardware.nix
```diff
@@ -0,0 +1,14 @@
...
+      type = types.anything;
+      default = {};
+      description = ''
+        Network device interface for use with graphics stack.
```
Mic92, line 10: Maybe add and `example = {}` value for an interface. From the description it's not clear what needs to be set here. Without reading the description, I would expect to set the string of the network interface name.

modules/hardware/lenovo-x1/definitions/default.nix
```diff
@@ -0,0 +1,69 @@
...
+      # Laptop keyboard
+      SUBSYSTEM=="input", ATTRS{name}=="AT Translated Set 2 keyboard", GROUP="kvm"
+      # Laptop TrackPoint
+      SUBSYSTEM=="input", ATTRS{name}=="TPPS/2 Elan TrackPoint", GROUP="kvm"
+      # Lenovo X1 integrated webcam
+      KERNEL=="3-8", SUBSYSTEM=="usb", ATTR{busnum}=="3", ATTR{devnum}=="3", GROUP="kvm"
+      # Lenovo X1 integrated fingerprint reader
+      KERNEL=="3-6", SUBSYSTEM=="usb", ATTR{busnum}=="3", ATTR{devnum}=="2", GROUP="kvm"
+      # Mouse and Touchpad
+      ${lib.strings.concatStrings (mapMouseRules hwDefinition.mouse)}
```
Mic92, line 65: Any particular reason to not use `lib.strings.concatMapStringsSep` here?

mikatammi:
This is now in conflict, please rebase

modules/hardware/lenovo-x1/modules/fprint.nix
mbssrc: These are swaylock specific PAM rules - other fprintd PAM rules (sudo, systemctl, etc.) are fine.

targets/lenovo-x1/everything.nix
```diff
@@ -62,25 +58,26 @@
...
 
-            disko.devices.disk = config.ghaf.hardware.definition.disks;
+            # Hardware configurations
+            security.polkit.extraConfig = powerControl.polkitExtraConfig;
 
             ghaf = {
-              hardware.definition = hwDefinition;
-              # To enable guest hardening enable host hardening first
+              # Hardware definitions
+              hardware.x86_64.common.enable = true;
```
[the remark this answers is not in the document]
mbssrc: Currently, we are setting `nixpkgs.hostPlatform.system = "x86_64-linux";` here so this won't work?!? Maybe we can handle the nixpkgs options differently ...

modules/desktop/graphics/hardware.nix
mbssrc: It is the same definition as in hardware namespace - copied that over including example so that it is clear. Ideally, we can remove this from here in the future.

modules/hardware/lenovo-x1/definitions/default.nix
mbssrc: Not really.

barnabakos:
ci-test-automation passed;
apps launch;
missing fingerprint testing instructions, though it works;
missing tpm testing instructions

brianmcgillion (approved):

mikatammi (approved):
Test sets also run nicely :+1:

PR #547, opened 2024-04-08
Wayland java - DRAFT

emrahbillur:
This work is for trying to achieve a pure Wayland support version of OpenJDK in Ghaf. The work includes two steps. First is to replace the stock JDK with Wayland version of it. The second stage is a little more problematic where the developers of the Pure Wayland OpenJDK has created a libwakefield library inside their source but not compiling it with the OpenJDK builds. The libwakefield should be seperately build from its subdirectory and manually added into OpenJDK prerequisites. The work is aimed for Nvidia Orin but it will also work on Lenovo and other platforms with wayland display.

Some very bad news here as the libwakefield requires the weston9 where our weston version is 12 the weston development interface has changed far beyond repair for porting this. Now my build errors are completely weston.h specific and fixing one by one is not going to fix it anymore as every fix introduces multiple new problems. We may not even have a chance for a proper working version.

## Checklist for things done

### DIFF
--- a/modules/common/development/default.nix
+++ b/modules/common/development/default.nix
@@ -4,6 +4,7 @@
   imports = [
     ./debug-tools.nix
     ./usb-serial.nix
+    ./java.nix
     ./nix.nix
     ./ssh.nix
   ];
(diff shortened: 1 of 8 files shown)


brianmcgillion (changes requested):
To be clear this will not live in Ghaf repo. Fine if you want to keep this patch here to polish it but overall this belongs upstream in nixpkgs.

PR #549, opened 2024-04-08
waybar: lock the wifi-signal-strength script

Nerox9:
## Description of changes

Multiple call from multiple instance of waybar causes race condition. flock and trap are used to eliminate. 

## Checklist for things done

### DIFF
--- a/packages/wifi-signal-strength/default.nix
+++ b/packages/wifi-signal-strength/default.nix
@@ -13,8 +13,20 @@
     pkgs.writeShellScript
     "wifi-signal-strength"
     ''
+      NETWORK_STATUS_FILE=/tmp/network-status
+
       export DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/ssh_session_dbus.sock
       export DBUS_SYSTEM_BUS_ADDRESS=unix:path=/tmp/ssh_system_dbus.sock
+
+      # Lock the script to reuse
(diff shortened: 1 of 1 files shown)


packages/wifi-signal-strength/default.nix
```diff
@@ -13,8 +13,12 @@
     pkgs.writeShellScript
     "wifi-signal-strength"
     ''
+      trap "${pkgs.openssh}/bin/ssh -q -S /tmp/nmcli_socket -O exit ghaf@${netvm_address}" EXIT
```
brianmcgillion, line 16: https://github.com/tiiuae/ghaf/pull/549/files#diff-0dd9147b12941844c66d8519a708fb64e934748abb257ff28c09de8bec9dc667R47

Do you not already do that here? 

If I am not mistaken EXIT only triggers on normal exit (0), which only occurs after the line listed above. So could we replace line 47 with an exit 0 to trigger the trap?

brianmcgillion (commented):
Looking good, just the one question on the trap

packages/wifi-signal-strength/default.nix
Nerox9: Yes, you are right. The script will be clearer with it, thanks!

brianmcgillion (approved):

milva-unikie:
Tested with X1 that this does fix the problem and everything else is working normally.

PR #552, opened 2024-04-10
GPIO proxy drivers ready for passthrough testing in VM using systemd

KimGSandstrom:
Kernel update for GPIO passthrough.
Starts gpio-vm VM for testing

<!--

- Kernel patches for tegra-5.10 kernel (used by optee)
- update of TEGRA_GPIO_HOST_PROXY and TEGRA_GPIO_GUEST_PROXY kernel modules
- starts gpio-vm virtual machine for testing GPIO passthrough
- VM has a systemd service called gpiotest that will use some of the GPIO pins as outputs

- code is work in progress and not tested yet.
- further documentation is yet to follow.

## Testing

- not tested yet / work in progress
- expected behaviour is that Ghaf starts up with a VM called gpio-vm, said VM has an 
  autostarting systemd service called gpiotest that will run a testing script. Script is a 
  bare sanity test testing only output on a few pins.
- verification of test result is done with a logic analyser on pins on the Jetson 40-pin 
  header for GPIO08, GPIO09, GPIO27, GPIO35. Which are pins 16, 32, 15, 18, respectively 
  on the 40-pin header


KimGSandstrom:
After merging with newest tiiuae ghaf the virtual machine and its service (microvm@gpio-vm.service) do not start up anymore. No VM seem to be running.

this is missing:
./targets/nvidia-jetson-orin/default.nix:              virtualization.microvm.gpiovm.enable = true;

KimGSandstrom:
The guest's device tree is not set. That's why the guest kernel fails at startup of microvm@gpio-vm

UnikieHR:
image

Issue #551, opened 2024-04-10
Evaluation against nixos-unstable fails


henrirosten:
Ghaf flake evaluation against `nixos-unstable` fails after commit https://github.com/tiiuae/ghaf/commit/c7cc82b1e2360cd94cdf55a23ff1b3a896f7dff0 which was merged in PR https://github.com/tiiuae/ghaf/pull/550.

**Repro steps**
- Pin Ghaf flake to `nixos-unstable`:

```
❯ git diff flake.nix
diff --git a/flake.nix b/flake.nix
index 1d1d736..eee65e2 100644
--- a/flake.nix
+++ b/flake.nix
@@ -25,7 +25,7 @@
   };
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
 
     #
     # Flake and repo structuring configurations
```
- Update flake lock
```
❯ nix flake lock --update-input nixpkgs
```
- Observe the evaluation fails with `error: infinite recursion encountered`:
```
❯ nix eval --raw .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug -v
         ...
         at /nix/store/450afzqlzzgw6wnyc3dwysf3i5yxyqkr-source/lib/customisation.nix:80:32:

           79|       newDrv = derivation (drv.drvAttrs // (f drv));
           80|     in flip (extendDerivation (seq drv.drvPath true)) newDrv (
             |                                ^
           81|       { meta = drv.meta or {};

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: infinite recursion encountered
      ...
```

- Evaluation passes if c7cc82b1e2360cd94cdf55a23ff1b3a896f7dff0 is removed:
```
❯ git revert c7cc82b1e2360cd94cdf55a23ff1b3a896f7dff0
[main 554965f] Revert "Add USB network card kernel patch"
 2 files changed, 26 deletions(-)
 delete mode 100644 modules/common/hardware/usb-network-kernel-patch.nix

❯ nix eval --raw .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug
...
/nix/store/w0yanw2x2m1nh62awapzzlnf35pjv6ir-ghaf-host-disko-images
```

mikatammi:
This should now at least fix the evaluation issues: https://github.com/tiiuae/ghaf/pull/553

[...]

PR #553, opened 2024-04-10
Refactor USB network card patch a bit

mikatammi:
## Description of changes

* Rename the module to ax88179_178a, which is the kernel module name.
* Enable manually based on option instead of condition based on config.boot.kernelPackages.kernel.version, which started to cause infinite recursion when evaluating against nixos-unstable.
* Enable manually only for platforms knwon having affected 6.1.x kernel.

## Checklist for things done

## Testing

Just regular builds. This still does not fix builds against nixos-unstable which have other issues, but at least makes them evaluate. Did not cause rebuiilds so I'm guessing this is fine.


mikatammi:
This should fix #551

PR #558, opened 2024-04-11
Create a virtual machine with a virtio-gpu device

nesteroff:
## Description of changes

Currently Ghaf uses Waypipe to run graphical application in virtual machines. It works quite well over VSOCK but it might be not very optimal in terms of performance as it serializes all data including pixel buffers over sockets. Also, Waypipe has some security implications, see [this section of the man page](https://gitlab.freedesktop.org/mstoeckl/waypipe/-/blob/master/waypipe.scd?ref_type=heads#L330).

As an alternative solution designed specifically for virtual machine we can evaluate using crosvm virtual machines with a [virtio-gpu device](https://crosvm.dev/book/devices/wayland.html). It requires a proxy Wayland compositor running in the guest, which can be either Sommelier or Wayland-proxy-virtwl.

This PR adds a virtiogpu-vm virtual machine which can be enabled for generic-x86_64 target. It has both Sommelier and Wayland-proxy-virtwl installed as well as Waypipe. The idea is that it can be used as a playground to test virtio-gpu device implemented in crosvm. It's possible to run applications using all available methods with the following scripts: run-sommelier, run-wayland-proxy, run-waypipe.

Known issues:
- The virtiogpu-vm can only be used in generic-x86_64 target at the moment. The use case with GPU passthrough in the GUIVM is not supported.
- The virtiogpu-vm doesn’t autostart because it requires a wayland socket which is only available after the compositor starts. There is also a permissions problem because microvm virtual machines are system services while the compositor is a user service.

My current observations show that Sommelier is a bit unstable. Sometimes it crashes when moving windows, etc. but Wayland-proxy-virtwl seems to be pretty stable. Comparing to Waypipe the performance looks better and the system load is lower but more testing is required.

## Checklist for things done

## Testing

- Enable virtgpu-vm for the generic-x86_64 target by uncommenting `virtualization.microvm.virtgpuvm.enable = true;`
- Build the image, boot Lenovo X1 and manually start the VM with the following command: `sudo -E /var/lib/microvms/virtgpu-vm/current/bin/microvm-run`
- Run Chromium using different methods, test performance, stability, system load, etc:
```
ghaf@virtiogpu-vm$ run-sommelier chromium --ozone-platform=wayland --disable-gpu
ghaf@virtiogpu-vm$ run-wayland-proxy chromium --ozone-platform=wayland --disable-gpu
ghaf@virtiogpu-vm$ run-waypipe chromium --ozone-platform=wayland --disable-gpu
```

### DIFF
--- a/modules/microvm/default.nix
+++ b/modules/microvm/default.nix
@@ -10,5 +10,6 @@
     ./virtualization/microvm/appvm.nix
     ./virtualization/microvm/guivm.nix
     ./networking.nix
+    ./virtualization/microvm/virtgpuvm.nix
   ];
 }
--- a/modules/microvm/virtualization/microvm/virtgpuvm.nix
+++ b/modules/microvm/virtualization/microvm/virtgpuvm.nix
@@ -1,0 +1,156 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  configHost = config;
+  vmName = "virtgpu-vm";
+  virtgpuvmBaseConfiguration = {
+    imports = [
+      (import ./common/vm-networking.nix {
+        inherit vmName;
+        macAddress = "02:00:00:03:05:01";
+      })
+      ({
+        lib,
+        pkgs,
+        ...
+      }: {
+        ghaf = {
+          users.accounts.enable = lib.mkDefault configHost.ghaf.users.accounts.enable;
+
+          development = {
+            ssh.daemon.enable = lib.mkDefault configHost.ghaf.development.ssh.daemon.enable;
+            debug.tools.enable = lib.mkDefault configHost.ghaf.development.debug.tools.enable;
+            nix-setup.enable = lib.mkDefault configHost.ghaf.development.nix-setup.enable;
+          };
+        };
+
+        # SSH is very picky about the file permissions and ownership and will
+        # accept neither direct path inside /nix/store or symlink that points
+        # there. Therefore we copy the file to /etc/ssh/get-auth-keys (by
+        # setting mode), instead of symlinking it.
+        environment.etc."ssh/get-auth-keys" = {
+          source = let
+            script = pkgs.writeShellScriptBin "get-auth-keys" ''
+              [[ "$1" != "ghaf" ]] && exit 0
+              ${pkgs.coreutils}/bin/cat /run/waypipe-ssh-public-key/id_ed25519.pub
+            '';
+          in "${script}/bin/get-auth-keys";
+          mode = "0555";
+        };
+        services.openssh = {
+          authorizedKeysCommand = "/etc/ssh/get-auth-keys";
+          authorizedKeysCommandUser = "nobody";
+        };
+
+        system.stateVersion = lib.trivial.release;
+
+        nixpkgs.buildPlatform.system = configHost.nixpkgs.buildPlatform.system;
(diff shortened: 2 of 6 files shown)


Mic92:
I also played around with wayland-proxy-virtwl and also did add some features needed for Gaming (https://github.com/talex5/wayland-proxy-virtwl/pull/75 https://github.com/talex5/ocaml-wayland/pull/40). Things that I noticed: Chromium worked sort of but firefox crashed on startup. That didn't happen with waypipe. I think the whole thing is not yet compatible with any GPU stuff. I coulnd't actually notice any performance benefit when I was testing it with qemu rutagaba vs waypipe, also this was just the perceived performance not doing any measurements.

Mic92:
I think crosvm might be more stable than the stuff in qemu. Than there is also this cloud-hypervisor fork: https://spectrum-os.org/software/cloud-hypervisor/

nesteroff:
> I also played around with wayland-proxy-virtwl and also did add some features needed for Gaming ([talex5/wayland-proxy-virtwl#75](https://github.com/talex5/wayland-proxy-virtwl/pull/75) [talex5/ocaml-wayland#40](https://github.com/talex5/ocaml-wayland/pull/40)). Things that I noticed: Chromium worked sort of but firefox crashed on startup. That didn't happen with waypipe. I think the whole thing is not yet compatible with any GPU stuff. I couldn't actually notice any performance benefit when I was testing it with qemu rutagaba vs waypipe, also this was just the perceived performance not doing any measurements.

That's cool! To me it's even hard to read OCaml to be honest. I just tested Firefox and it seems to work fine in this configuration. At least it starts and it's possible to open tabs, etc. It's hard to measure performance but I noticed that when watching a 4K video full screen the system load is not as high as when using waypipe.

Performance problems become really noticeable in the lenovo-x1 configuration when the compositor is running in the GUIVM. In this scenario the data goes from one VM to the host and then from the host to another VM through vsockproxy. There is a lot of extra IO and 4K video becomes more like a slide show. I wonder if there is a way to make wayland-proxy-virtwl support the use case when the compositor is in its own VM?

nesteroff:
> I think crosvm might be more stable than the stuff in qemu. I did back than not test crosvm at all because I don't think it has audio support on Linux. Than there is also this cloud-hypervisor fork: https://spectrum-os.org/software/cloud-hypervisor/

Yes, I believe Alyssa developed the initial version when she worked with us. It's a really good patch but it uses the same GPU backend implemented in crosvm so I assume it should work more or less the same.

nesteroff:
Updated to support autostart of the VM with virtio-gpu

Mic92:
I wonder if there is a way to make wayland-proxy-virtwl support the use case when the compositor is in its own VM?

I think something related nested compositing was added only very recently to the wayland protocol or was related to having x11 window manager in a wayland session? I cannot remember exactly.

Mic92:
> That's cool! To me it's even hard to read OCaml to be honest. 

Yeah. It's also not my favorite language and the person maintaining the project also writes advanced abstractions at least for my level.

nesteroff:
> I think something related nested compositing was added only very recently to the wayland protocol or was related to having x11 window manager in a wayland session? I cannot remember exactly.

Nested compositing seems to work fine on Wayland nowadays but I'm not sure if it's something that we could use for the Lenovo X1 target. It has a single compositor but it's running in its own VM with GPU passthrough instead of running on the host. I guess we can try to figure out how to share blobs and pipes over virtio-gpu in the guest-to-guest scenario instead of guest-to-host if it's even feasible.

nesteroff:
Closing since this feature is not needed at the moment and the PR is outdated.

PR #561, opened 2024-04-11
Update jetpack-nixos in flake.lock

mikatammi:
## Checklist for things done

  - [X] Tested on Jetson Orin NX *and* AGX `aarch64`

## Testing

Basic test sets on all platforms.

Still would need to test by hand how does this affect BPMP passthrough and UART passthrough, etc. NVIDIA kernel related patches that are part of ghaf and not tested yet by our test suites

### DIFF
--- a/modules/common/systemd/systemd-boot-double-dtb-buffer-size.patch
+++ b/modules/common/systemd/systemd-boot-double-dtb-buffer-size.patch
@@ -1,0 +1,46 @@
+From f3b523c80624a164889928bc255c6d16da7b20da Mon Sep 17 00:00:00 2001
+From: Mika Tammi <mika.tammi@unikie.com>
+Date: Thu, 11 Apr 2024 20:23:32 +0300
+Subject: Double the device tree buffer size
+
+Signed-off-by: Mika Tammi <mika.tammi@unikie.com>
+---
+ src/boot/efi/devicetree.c | 9 ++++++++-
+ 1 file changed, 8 insertions(+), 1 deletion(-)
+
+diff --git a/src/boot/efi/devicetree.c b/src/boot/efi/devicetree.c
+index caafef24d3..8395fd2c79 100644
+--- a/src/boot/efi/devicetree.c
++++ b/src/boot/efi/devicetree.c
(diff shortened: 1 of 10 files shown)


modules/common/systemd/systemd-boot-double-dtb-buffer-size.patch
```diff
@@ -0,0 +1,46 @@
...
+ #define FDT_V1_SIZE (7*4)
+ 
+ static EFI_STATUS devicetree_allocate(struct devicetree_state *state, size_t size) {
+@@ -87,7 +89,8 @@ EFI_STATUS devicetree_install(struct devicetree_state *state, EFI_FILE *root_dir
+                 /* 32MB device tree blob doesn't seem right */
+                 return EFI_INVALID_PARAMETER;
+ 
+-        len = info->FileSize;
++        /* Double the length to be allocated so there's space for modifications */
++        len = info->FileSize * 2;
```
nesteroff, line 30: Just one thing here. Initially when I developed this patch I used FileSize * 2 because I only needed a little bit of space to make sure it updates the MAC address. Do you think now we should use FileSize * 4 to be consistent with the L4TLauncher? Here is how they do it: https://github.com/NVIDIA/edk2-nvidia/blob/main/Silicon/NVIDIA/Application/L4TLauncher/L4TLauncher.c#L1491C61-L1491C74

mikatammi: Maybe we could do 4x although it feels a bit overkill :D

nesteroff: Who knows, maybe there is a reason why they do 4x in their launcher :) Obviously it is a nasty hack anyway. I wish they used the device tree fixup protocol or something like that.

TanelDettenborn: If we would not like to patch C-code then another option would be patching device tree build. `-S`-parameter to dtc would do a similar hack.

brianmcgillion (approved):

PR #539, opened 2024-04-03
labwc redesign, Gnome removal

humaidtii:
This aligns the system design closer to the prototypes in Figma, improving UI/UX state of Ghaf. Also as Gnome is not functional, it is removed.

### labwc redesign
- Add a custom theme configuration for labwc, matching our design language.
- Revamp waybar to modern layout (bottom dock matching our design).
- Add new assets (SVG) for waybar.
- Add resize function for PNG.
- Add conversion function for SVG to PNG.
- Add configuration for mako (notification daemon) to match our design.
- Use Inter font for system by default.
- Window border style changes
  - Use darker colors.
  - Use frameColoring alongside waypipe colouring (nicer for demos, need a more robust & secure solution later).
  - Make waypipe border smaller (5px->3px) as not to overlay apps too much.
- Add PrintScreen keyboard shortcut (saves PNG to `~`).
- Use natural scrolling by default.
- Add new wallpaper.

### Gnome removal

Removed as Gnome is not officially supported or working. If it is later decided Gnome should be supported, it should be brought back as a separate PR with proper support.

## Checklist for things done

- [x] Summary of the proposed changes in the PR description
- [x] More detailed description in the commit message(s)
- [x] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [x] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] PR linked to architecture documentation and requirement(s) (ticket id)
- [x] Test procedure described (or includes tests). Select one or more:
  - [x] Tested on Lenovo X1 `x86_64`
  - [ ] Tested on Jetson Orin NX or AGX `aarch64`
  - [ ] Tested on Polarfire `riscv64`
- [ ] Author has run `nix flake check --accept-flake-config` and it passes
- [x] All automatic Github Action checks pass - see [actions](https://github.com/tiiuae/ghaf/actions)
- [x] Author has added reviewers and removed PR draft status

## Testing

1. Waybar buttons work. They render icons correctly.
2. Drawer opens no issues. All launchers in drawers have icons.

PR #582, opened 2024-04-29
Full migration to labwc

humaidtii:
Full migration to labwc, removing Weston compositor support and adding hi-res icons for apps.

This also fixes labwc performance issues and other bugs in Jetson Orin.

~Depends on: https://github.com/tiiuae/ghaf/pull/539~

## Description of changes

Weston was deprecated in Ghaf 24.03, now removal for Ghaf 24.06. With Weston removed, we do not need to provide backwards compatibility.

- Completely remove weston implementation & configuration.
- Use [Papirus icon theme](https://github.com/PapirusDevelopmentTeam/papirus-icon-theme) for all launcher entries.
  - Remove icon assets as they are replaced by the theme.
  - Created subset of Papirus icon theme package to only include icons we use.
  - Remove power icons custom package.
- Fix letter casing for launcher entries.
- Fix labwc performance issues on Jetson Orin by using gles2.
  - Renderer can be defined as an option.
- Fix duplicate icons in drawer.
- Replace i3lock with gtklock, make it match system theme.
- Added window icons from Figma.
  - Had to create custom package `ghaf-openbox-theme`.
- Include default fonts in the system.

## Known issues

- Cannot interact with gtklock using mouse, only keyboard.

## Checklist for things done

## Testing

- labwc available on all platforms, no major performance issues.
- New icon theme available for all launcher icons.
- No more duplicate icons in the drawer on Orin.
- Lock screen should work as expected. When autolock is enabled, there is dimming before locking.
- Enrol with fingerprint, test fingerprint login with gtklock (there are pre-existing quirks with this feature).


brianmcgillion (approved):

johannarautanen:
**Test results on Orin-AGX**

Manual testing: 

- Cannot open Element-application 
- In Lock-view the "Unlock"-button or the "image" in the password line does nothing.
(You can unlock with entering the password and then hitting keyboard's Enter-button)


Other manual regression tests were ok.

 ci-test automation:

- bat tests passed

leivos-unikie:
I ran ci-test-automation for native build on Orin-NX. All tests passed. Performance seems normal except there is some drop in sysbench FileIO write test. It is at 42.75MiB/s with this build. Automated test history of FileIO write results on Orin NX shows somewhat similar dip at build #127 (56.16MiB/s) but then it recovered in the following builds, normally the level being consistently above 75MiB/s.

I did two FileIO test runs in to confirm.

milva-unikie:
On X1 automated tests passed, performance is ok and all apps launch.

The were some bugs that I noticed, but most of them were happening only sometimes so it was a bit of a challenge to describe them clearly. Happy to give more information if something is unclear!

- When an external display is connected by HDMI, sometimes an extra task bar is created. I couldn't find any clear logic as to when this will happen. At least if the HDMI is connected at startup and after startup you disconnect and reconnect, this will happen most of the time.
- On the lock screen, the Unlock button does not work. Also, the password cannot be seen with the button to the right of the password field. Pressing Enter will work to confirm the password.
- If you have registered a fingerprint but are using a password to log in, it will take about 30 seconds for the laptop to open.
- When entering the fingerprint on lock screen, the screen sometimes goes black (mouse can be seen). This doesn't happen every time, but it happened to me three times.

leivos-unikie:
> I ran ci-test-automation for native build on Orin-NX. All tests passed. Performance seems normal except there is some drop in sysbench FileIO write test. It is at 42.75MiB/s with this build. Automated test history of FileIO write results on Orin NX shows somewhat similar dip at build #127 (56.16MiB/s) but then it recovered in the following builds, normally the level being consistently around 75MiB/s.
> 
> I did two FileIO test runs to confirm.

I think my fileio write issue is not related to this PR. I made some more testing. Booted from another USB SSD and with that I got fileio results above 70MiB/s (multiple runs). Then booted again from the original USB SSD, and it didn't even boot to ghaf. It booted to ubuntu which is installed on nvme (plugged under the carried board). Then changed the orientation of the SSD USB cable and tried booting again, and it booted to ghaf! So it seems my USB SSD is worn out. I had already strange errors during other tests few days ago, probably caused by that SSD.

milva-unikie:
> Fingerprint quirks probably existed previously, due to PAM configuration. But maybe more obvious with gtklock?

We have not done much testing with the fingerprint before, so this is very possible. 


> Task bar should show on all screens no matter when the display is plugged in, maybe try moving the cursor where the taskbar should be, it might be a rendering glitch?

This is what I meant with extra taskbar. The screen on the left is the laptop screen and screen on the right is an external screen.

20240520_09h06m11s_grim

johannarautanen:
**Orin-AGX (commit 1c55128d9a07ce873292b538d020e07110d09b38):**

- Now the element -application is launched (it took the "normal" 46 seconds to launch)
- mouse click inputs are still ignored in lock screen. 
- When you have opened more then one browser-views and you have the "gmail-browser" first in the browser-list and there is for example "Aamulehti"-browser visible in the taskbar. 


When you want to select the "Aamulehti-browser" by click the name on the taskbar, the "gmail-browser" opened, cause it was the "first" in the list.

leivos-unikie:
Checked that labwc works also in vm-debug target.

Mic92:
Attempt to fix cross compilation: https://github.com/NixOS/nixpkgs/pull/315119

leivos-unikie:
Tested both native and cross-compiled build on Orin NX
- ci-test-automation passed and performance ok
- apps launch
- no issues found except some lag in mouse movement on desktop

milva-unikie:
Tested on Lenovo-X1:
- Automated tests for apps fail. The test fetches app paths from `/run/current-system/sw/share/applications/${app_name}.desktop` in gui-vm. In this pr those files don't exist for apps like chromium and element. This might already have been an issue with the previous version I tested. 
- Other automated tests pass.
- Performance ok.
- All apps launch.

These old notes are still valid:
- On the lock screen, the Unlock-button does not work. Also, the password cannot be seen with the button to the right of the password field. Pressing enter works to confirm the password.
- If you have registered a fingerprint but are using a password to log in, it will take about 30 seconds for the laptop to open.
- When entering the fingerprint on lock screen, the screen sometimes goes black (mouse can be seen). This doesn't happen every time, but it is happening regularly. The only way to recover I found is to reboot the laptop.

milva-unikie:
**Tested on Lenovo-X1** (used commit [1b4c539](https://github.com/tiiuae/ghaf/commit/1b4c5392885fbfe27abd188222c0ced7b9a39146))
- Test-automation passes (except app tests that need to be updated. I created a PR for that so the change is ready to go into test-automation when this PR is merged).
- Performance ok.
- All apps launch.

These old notes are still valid:
- On the lock screen, the Unlock-button does not work. Also, the password cannot be seen with the button to the right of the password field. Pressing enter works to confirm the password.
- If you have registered a fingerprint but are using a password to log in, it will take about 30 seconds for the laptop to open.
- When entering the fingerprint on lock screen, the screen sometimes goes black (mouse can be seen). This doesn't happen every time, but it is happening regularly. The only way to recover I found is to reboot the laptop.

brianmcgillion (approved):

overlays/cross-compilation/papirus-icon-theme/default.nix
```diff
@@ -0,0 +1,11 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# papirus-icon-theme cross-compilation fixes (removing qt dependency)
```
Mic92, line 4: Was this done because qt doesn't cross-compile just now or to make the build process faster?

Mic92: cc @humaidq-tii

PR #584, opened 2024-04-30
networking: module for static names

vilvo:
## Description of changes

* based on decisions to use static names over internal DNS https://github.com/tiiuae/ghaf/pull/427
* removes dnsmasq and dhcp from net-vm to other guests
* `/etc/hosts` with static hosts to IP mapping generated and made available to host and guests
* resolves https://github.com/tiiuae/ghaf/pull/600

## Checklist for things done

- [x] PR linked to architecture documentation and requirement(s) (ticket id)

### Testing the debug network access with names - PASS

Connect to `ghaf-host` over debug ethernet. Check the IP in your network.
```
❯ ssh ghaf@nn.nn.nn.nn
Last login: Tue Jun  4 10:09:53 2024 from yy.yy.yy.yy

[ghaf@ghaf-host:~]$ less /etc/hosts

[ghaf@ghaf-host:~]$ ssh gui-vm-debug
(ghaf@gui-vm-debug) Password:
[ghaf@gui-vm:~]$
logout
Connection to gui-vm-debug closed.

[ghaf@ghaf-host:~]$ ssh net-vm-debug
(ghaf@net-vm-debug) Password:

[ghaf@net-vm:~]$ ssh gui-vm-debug
(ghaf@gui-vm-debug) Password:
Last login: Tue Jun  4 10:11:35 2024 from 192.168.101.2

[ghaf@gui-vm:~]$
logout
Connection to gui-vm-debug closed.

[ghaf@net-vm:~]$ ifconfig ethint0
ethint0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 192.168.100.1  netmask 255.255.255.0  broadcast 192.168.100.255
        inet6 fe80::ff:fe01:101  prefixlen 64  scopeid 0x20<link>
        ether 02:00:00:01:01:01  txqueuelen 1000  (Ethernet)
        RX packets 940  bytes 171017 (167.0 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 891  bytes 543302 (530.5 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```
### Testing the access to internet over Wifi - PASS
1. Connect to WIFI using nm-applet in the Ghaf desktop - PASS
2. Start chromium - PASS
3. Test browsing with chromium - PASS

### DIFF
--- a/modules/common/networking/hosts.nix
+++ b/modules/common/networking/hosts.nix
@@ -1,0 +1,74 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  ...
+}: let
+  # please note that .100. network is not
+  # reachable from ghaf-host. It's only reachable
+  # guest-to-guest.
+  # Use to .101. (debug) to access guests from host.
+  # debug network hosts are post-fixed: <hostname>-debug
+  ipBase = "192.168.100";
+  debugBase = "192.168.101";
+  hostsEntries = [
+    {
+      ip = 1;
+      name = "net-vm";
+    }
+    {
+      ip = 2;
+      name = "ghaf-host";
+    }
+    {
+      ip = 3;
+      name = "gui-vm";
+    }
+    {
+      ip = 4;
+      name = "ids-vm";
+    }
+    {
+      ip = 5;
+      name = "audio-vm";
+    }
+    {
+      ip = 10;
+      name = "admin-vm";
+    }
+    {
+      ip = 100;
+      name = "chromium-vm";
+    }
+    {
+      ip = 101;
+      name = "gala-vm";
+    }
+    {
+      ip = 102;
+      name = "zathura-vm";
+    }
+    {
+      ip = 103;
+      name = "element-vm";
+    }
+    {
+      ip = 104;
+      name = "appflowy-vm";
+    }
+  ];
+  mkHostEntry = {
+    ip,
+    name,
+  }:
+    "${ipBase}.${toString ip}\t${name}\n"
+    + lib.optionalString config.ghaf.profiles.debug.enable
+    "${debugBase}.${toString ip}\t${name}-debug\n";
+  entries = map mkHostEntry hostsEntries;
+in {
+  environment.etc.hosts = lib.mkForce {
+    text = lib.foldl' (acc: x: acc + x) "127.0.0.1 localhost\n" entries;
+    mode = "0444";
+  };
+}
--- a/modules/microvm/virtualization/microvm/adminvm.nix
+++ b/modules/microvm/virtualization/microvm/adminvm.nix
@@ -11,10 +11,14 @@
 
   adminvmBaseConfiguration = {
     imports = [
-      (import ./common/vm-networking.nix {inherit vmName macAddress;})
+      (import ./common/vm-networking.nix {
+        inherit config lib vmName macAddress;
+        internalIP = 10;
+      })
       ({lib, ...}: {
         ghaf = {
           users.accounts.enable = lib.mkDefault configHost.ghaf.users.accounts.enable;
+          profiles.debug.enable = lib.mkDefault configHost.ghaf.profiles.debug.enable;
           development = {
             # NOTE: SSH port also becomes accessible on the network interface
             #       that has been passed through to VM
@@ -44,15 +48,6 @@
           enable = true;
           networks."10-ethint0" = {
             matchConfig.MACAddress = macAddress;
-            addresses = [
-              {
-                addressConfig.Address = "192.168.100.10/24";
-              }
-              {
-                # IP-address for debugging subnet
-                addressConfig.Address = "192.168.101.10/24";
-              }
-            ];
             linkConfig.ActivationPolicy = "always-up";
           };
         };
--- a/packages/wifi-signal-strength/default.nix
+++ b/packages/wifi-signal-strength/default.nix
@@ -6,8 +6,6 @@
   wifiDevice,
   ...
 }: let
-  # Replace the IP address with "net-vm.ghaf" after DNS/DHCP module merge
-  netvm_address = "192.168.100.1";
   wifiSignalStrength =
     pkgs.writeShellScript
     "wifi-signal-strength"
@@ -23,11 +21,11 @@
       ${pkgs.util-linux}/bin/flock -w 60 -x 99 || exit 1
 
       # Return the result as json format for waybar and use the control socket to close the ssh tunnel.
-      trap "${pkgs.openssh}/bin/ssh -q -S /tmp/nmcli_socket -O exit ghaf@${netvm_address} && ${pkgs.coreutils-full}/bin/cat $NETWORK_STATUS_FILE" EXIT
+      trap "${pkgs.openssh}/bin/ssh -q -S /tmp/nmcli_socket -O exit ghaf@net-vm && ${pkgs.coreutils-full}/bin/cat $NETWORK_STATUS_FILE" EXIT
 
       # Connect to netvm
       ${pkgs.openssh}/bin/ssh -M -S /tmp/nmcli_socket \
-          -f -N -q ghaf@${netvm_address} \
+          -f -N -q ghaf@net-vm \
           -i /run/waypipe-ssh/id_ed25519 \
           -o StrictHostKeyChecking=no \
           -o StreamLocalBindUnlink=yes \
--- a/targets/lenovo-x1/netvmExtraModules.nix
+++ b/targets/lenovo-x1/netvmExtraModules.nix
@@ -25,12 +25,15 @@
 
     internalNic = let
       vmNetworking = import ../../modules/microvm/virtualization/microvm/common/vm-networking.nix {
+        config = configH;
+        inherit lib;
         vmName = microvm.name;
         inherit (microvm) macAddress;
+        internalIP = 1;
       };
     in "${lib.head vmNetworking.networking.nat.internalInterfaces}";
 
-    elemen-vmIp = "192.168.100.253";
+    element-vmIp = "192.168.100.103";
   in {
     # For WLAN firmwares
     hardware = {
@@ -65,7 +68,7 @@
           ];
 
         # DNS host record has been added for element-vm static ip
-        host-record = "element-vm,element-vm.ghaf,${elemen-vmIp}";
+        host-record = "element-vm,element-vm,${element-vmIp}";
       };
 
       openssh = configH.ghaf.security.sshKeys.sshAuthorizedKeysCommand;
@@ -107,7 +110,7 @@
       firewallConfig = true;
       externalNic = "${externalNic}";
       internalNic = "${internalNic}";
-      serverIpAddr = "${elemen-vmIp}";
+      serverIpAddr = "${element-vmIp}";
     };
   };
 in [netvmPCIPassthroughModule netvmAdditionalConfig]
--- a/modules/common/default.nix
+++ b/modules/common/default.nix
@@ -17,5 +17,6 @@
     ./systemd
     ./services
     ./programs
+    ./networking
   ];
 }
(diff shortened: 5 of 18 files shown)


modules/common/networking/hosts.nix
```diff
@@ -0,0 +1,21 @@
...
+    # guests from host (no names)
+    text = ''
+      127.0.0.1 localhost
+      192.168.100.1 net-vm
+      192.168.100.2 log-vm
+      192.168.100.3 gala-vm
+      192.168.100.4 chromium-vm
+      192.168.100.5 zathura-vm
+      192.168.100.6 element-vm
+      192.168.100.7 gui-vm
```
vunnyso, line 17: I would suggest we can keep `gui-vm`  IP to older value `192.168.100.3`
and what about `ghaf-host` IP assignment?

tervis-unikie: I thought it would have been preferable too, but I did not change the address in hosts.

vilvo: > I would suggest we can keep `gui-vm` IP to older value `192.168.100.3` and what about `ghaf-host` IP assignment?

I'm fine with this. In the end, the ip addresses won't matter. When all of this works, we can use the names only for inter-vm and it will allow us even to change the network if needed/preferred.

modules/common/networking/hosts.nix
```diff
@@ -0,0 +1,74 @@
...
+      ip = 3;
+      name = "gui-vm";
+    }
+    {
+      ip = 4;
+      name = "ids-vm";
+    }
+    {
+      ip = 5;
+      name = "audio-vm";
```
josa41, line 34: I think earlier the name address of each vms ended with .ghaf (for example: audio-vm.ghaf). In some vm config that use audio (chromium and element) there is pulseaudio config that refer to audio-vm.ghaf:4713. If the .ghaf ending addresses don't work anymore those configs should be updated to reflect the new name.

modules/common/networking/hosts.nix
```diff
@@ -0,0 +1,74 @@
...
+      ip = 101;
+      name = "gala-vm";
+    }
+    {
+      ip = 102;
+      name = "zathura-vm";
+    }
+    {
+      ip = 103;
+      name = "element-vm";
```
josa41, line 54: In element-vm config element.nix:79 there is a refrence to a static ip address (of element-vm) 192.168.100.253 
The same ip is also set as internal config variable in netvmExtraModules.nix:33 (element-vmIp = "192.168.100.253";)
To me it seems that this is to get dendrite routing done with iptables on dendrite-pinecone.nix.
For now I think it should work if the ip is set correctly in all places but this (dendrite networking setup) would need a bit more love. :)

targets/lenovo-x1/netvmExtraModules.nix
```diff
@@ -25,8 +25,11 @@
 
     internalNic = let
       vmNetworking = import ../../modules/microvm/virtualization/microvm/common/vm-networking.nix {
+        config = configH;
+        inherit lib;
         vmName = microvm.name;
         inherit (microvm) macAddress;
+        internalIP = 1;
       };
     in "${lib.head vmNetworking.networking.nat.internalInterfaces}";
```
josa41, line 35: Element-vmIp static address different in the new hosts file.

josa41 (changes requested):

modules/common/networking/hosts.nix
```diff
@@ -0,0 +1,74 @@
...
+  # please note that .100. network is not
+  # reachable from ghaf-host. It's only reachable
+  # guest-to-guest.
+  # Use to .101. (debug) to access guests from host.
+  # debug network hosts are post-fixed: <hostname>-debug
+  ipBase = "192.168.100";
+  debugBase = "192.168.101";
+  hostsEntries = [
+    {
+      ip = 1;
```
vunnyso, line 17: One generic suggestion, instead of using hardcoded numbers here and in other files [loc1](https://github.com/tiiuae/ghaf/pull/584/files#diff-477e661df628667d7a61761a39c61bfb59fa57fdba1e3b7e3e58468d2f5b396dR24), [loc2](https://github.com/tiiuae/ghaf/pull/584/files#diff-ed0b8ba6a36a522c74a303a9dc0f1c41cfc3dc6a07ef06450aabf6df1318db8aR32)   maybe we can associate those values with nix options, example instead of ip = 1 we can use ip = `config.ghaf.ip.index.netvm`  so it will be easy to access/read across different files and easy to maintain :)
Not necessary to be in scope of this PR but can be taken up later.

modules/common/networking/hosts.nix
tervis-unikie: .ghaf network references should be fixed now.

modules/common/networking/hosts.nix
tervis-unikie: This should be fixed now.

targets/lenovo-x1/netvmExtraModules.nix
tervis-unikie: Should be fixed now.

modules/common/networking/hosts.nix
tervis-unikie: Original idea was to try to use the list of appvms to generate hosts file, but hosts file gets generated while the vms are created, so there was a bit of an chicken and egg problem.
Right away, even with the help of some nix experts, we could not figure out a reasonably dynamic way to handle this, so at least for now we decided to have some redundancy (several lists of vms) so we get this thing working. A more elegant solution can be implemented later.

tervis-unikie:
.ghaf network references should be fixed now.

modules/common/networking/hosts.nix
josa41: I think this is good for now.
Maybe consider updating the ipBase to a separate config variable. The issue is that there is are a few "magic" values with the ipbase / gateway address in various .nix files (at least in hosts.nix and vm-networking.nix).
I agree with the difficulty of handling this dynamically (that is what dhcp is for). In our use Dhcp just causes its own issues that would be really complex to solve.

tervis-unikie: What I was referring to by "dynamic" was "build-time-dynamic" solution, so that we'd have one list of vms in one place, and both the actual vms and hosts entries would get generated from that. Not referring to dynamic host configuration protocol here.

vilvo: Yeah, we discussed to get the scattered/distributed networking configuration refactoring in the next phase. The coming centralised networking configuration nix module could look like [target hw definitions](https://github.com/tiiuae/ghaf/blob/main/modules/hardware/lenovo-x1/definitions/x1-gen11.nix) but with MACs, IPs, network bases etc. - collected to a single file and referenced elsewhere via config. Anyone configuring the ghaf target network would see all the values in one file. That structural change is less prone to regression than this and can bring some build-time checks via value references.

This PR mainly addresses the transition from internal DNS/DHCP with dnsmasq to static internal network with names.

josa41 (approved):
Looks good to me

modules/microvm/virtualization/microvm/adminvm.nix
```diff
@@ -11,10 +11,14 @@
 
   adminvmBaseConfiguration = {
     imports = [
-      (import ./common/vm-networking.nix {inherit vmName macAddress;})
+      (import ./common/vm-networking.nix {
+        inherit config lib vmName macAddress;
+        internalIP = 10;
```
brianmcgillion, line 16: For all of the places where the internalIP is set should this not come from the modules/common/networking/hosts.nix? like this we are doing the accounting twice. or is this coming in a followup patch?

packages/wifi-signal-strength/default.nix
```diff
@@ -6,8 +6,7 @@
   wifiDevice,
   ...
 }: let
-  # Replace the IP address with "net-vm.ghaf" after DNS/DHCP module merge
-  netvm_address = "192.168.100.1";
+  netvm_address = "net-vm";
```
brianmcgillion, line 9: any reason to define a variable here? in e.g. [nm-launcher](https://github.com/tiiuae/ghaf/pull/584/files#diff-8e5d414928ef2d58c4b50f71cb7e0ca65585d16ce036993905e32f5b44a0fd4aR30) we just use the name directly?

targets/lenovo-x1/netvmExtraModules.nix
```diff
@@ -25,12 +25,15 @@
...
 
-    elemen-vmIp = "192.168.100.253";
+    element-vmIp = "192.168.100.104";
```
brianmcgillion, line 36: again, shouldnt this be taken from the [host.nix](https://github.com/tiiuae/ghaf/pull/584/files#diff-8180f7d1e010f8a2734247b6e1878f4e1df2e4b72459fc4d7af8e0215d3c7488R57).

As you can see there is already appearing to be a copy error as AppFlowy is .104 and element is .103

brianmcgillion (changes requested):

modules/microvm/virtualization/microvm/adminvm.nix
vilvo: Yes, removal of IP duplication is to come in a follow-up patch. We actually tried that first but with a wrong approach. Merging of VMs declared IPs with built-time generated /etc/hosts that go into the VMs turned out as a build-time chicken-egg problem with nix - even with nix expert support.

I'd rather not debate that further in this PR but [refactor that in the follow-up PR like this](https://github.com/tiiuae/ghaf/pull/584#discussion_r1628929192).

packages/wifi-signal-strength/default.nix
tervis-unikie: Variable was already there, I suppose that was the only reason to have it.
Variable removed.

targets/lenovo-x1/netvmExtraModules.nix
tervis-unikie: Good catch. In this case IP is required, hostname won't work.
Getting it from unified config is in the future plans.
IP changed to correct one.

barnabakos:
tested with latest commit: 2ab7c5ae5ccbff67ac371d23b9d3f74d6b99a9e0
- merge conflict; should be rebased

platform: Lenovo X1 Carbon
- ci-test-automation suite passed
- all apps start
- static names are working to access the vms
- no new findings

brianmcgillion (approved):

tervis-unikie:
Rebased on latest main.

johannarautanen:
tested with the commit: https://github.com/tiiuae/ghaf/commit/2ab7c5ae5ccbff67ac371d23b9d3f74d6b99a9e0

**Native Orin-AGX:**
Only different thing comparing to the yesterday's regression test results:
Not able to ping "google.com" via net-vm with wi-fi. 
Via ethernet in ghaf-host ping "google.com" was successfully.


There is double icons available, but that is known issue for yesterday's regression tests: https://ssrc.atlassian.net/browse/SP-4707 and the nm-applet thing: https://ssrc.atlassian.net/browse/SP-4715

- In Orin-AGX there is only net-vm and ghaf-host available, "ssh net-vm-debug" and "ssh ghaf-host-debug" works ok.
- other manual regressions tests were ok.
- all apps launched
- automation cases ok

johannarautanen:
tested with the commit: https://github.com/tiiuae/ghaf/commit/2ab7c5ae5ccbff67ac371d23b9d3f74d6b99a9e0

Native Orin-NX:
Only different thing comparing to the yesterday's regression test results:
Not able to ping "google.com" via net-vm via ethernet connection, the "ping 8.8.8.8" gives response
Via ethernet in ghaf-host ping "google.com" was successfully.

Same double icon and nm-applet thing like in Orin-AGX. 

- In Orin-NX there is only net-vm and ghaf-host available, "ssh net-vm-debug" and "ssh ghaf-host-debug" works ok.
- other manual regressions tests were ok.
- all apps launched
- automation cases ok

tervis-unikie:
net-vm name resolving should work now.

brianmcgillion:
verified `ping google.com` on the x1 and it is a pass

vunnyso:
Verified on Orin AGX ping "google.com"  and ping 8.8.8.8 are working fine with PR changes 
image

brianmcgillion (approved):

mbssrc (approved):

PR #585, opened 2024-04-30
Change internal installation approach

remimimimimi:
## Description of changes

Change approach which is used on installer. Instead of flashing pre-build image - partition disk and assemble system during installation process. Like it's done during normal NixOS installation process using `nixos-install` command.

## Checklist for things done

## Testing

To run introduced test execute `nix build .#checks.x86_64-linux.installer --log-format bar-with-logs`, otherwise testing process same as in previous installer-related PRs.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -155,6 +155,7 @@
         ./packages/flake-module.nix
         ./targets/flake-module.nix
         ./hydrajobs/flake-module.nix
+        ./tests/flake-module.nix
         ./templates/flake-module.nix
       ];
 
--- a/packages/installer/default.nix
+++ b/packages/installer/default.nix
@@ -1,23 +1,30 @@
 # Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
 {
-  bash,
-  imagePath,
-  substituteAll,
-}:
-substituteAll {
-  dir = "bin";
-  isExecutable = true;
-
-  pname = "ghaf-installer";
-  src = ./ghaf-installer.sh;
-  inherit imagePath;
-
-  buildInputs = [
-    bash
-  ];
-
-  postInstall = ''
-    patchShebangs $out/bin/ghaf-installer.sh
-  '';
-}
+  lib,
+  stdenvNoCC,
+  makeWrapper,
+  diskoInstall,
+  targetName,
+  ghafSource,
+  diskName,
+}: let
+  name = "ghaf-installer";
+in
+  stdenvNoCC.mkDerivation {
+    inherit name;
+    src = ./.;
+    nativeBuildInputs = [
+      makeWrapper
+    ];
+    installPhase = ''
+      mkdir -p $out/bin
+      cp ${name}.sh $out/bin/${name}.sh
+      chmod 755 $out/bin/${name}.sh
+      wrapProgram $out/bin/${name}.sh \
+        --set GHAF_SOURCE "${ghafSource}" \
+        --set TARGET_NAME "${targetName}" \
+        --set DISKO_DISK_NAME "${diskName}" \
+        --prefix PATH : ${lib.makeBinPath [diskoInstall]}
+    '';
+  }
--- a/packages/installer/ghaf-installer.sh
+++ b/packages/installer/ghaf-installer.sh
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 # Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
+
+# FIXME: Otherwise test fails
+if [ -z ${DEVICE_PATH+x} ]; then
+    set -euo pipefail
+fi
+
 clear
 cat <<"EOF"
   ,----..     ,---,
@@ -20,15 +26,17 @@ EOF
 
 echo "Welcome to Ghaf installer!"
 
-echo "To install image choose path to the device on which image will be installed."
+if [ -z ${DEVICE_PATH+x} ]; then
+    echo "To install image choose path to the device on which image will be installed."
 
-lsblk
-read -r -p "Device name [e.g. /dev/nvme0n1]: " DEVICE_NAME
+    lsblk
+    read -r -p "Device path [e.g. /dev/nvme0n1]: " DEVICE_PATH
 
-read -r -p 'WARNING: Next command will destroy all previous data from your device, press Enter to proceed. '
+    read -r -p 'WARNING: Next command will destroy all previous data from your device, press Enter to proceed. '
+fi
 
 echo "Installing..."
-dd if=@imagePath@ of="${DEVICE_NAME}" bs=32M status=progress
+disko-install --write-efi-boot-entries --debug --flake "$GHAF_SOURCE#$TARGET_NAME" --disk "$DISKO_DISK_NAME" "$DEVICE_PATH" --option substitute false --option binary-caches ""
 
 echo ""
 echo "Installation done. Please remove the installation media and reboot"
--- a/targets/lenovo-x1-installer/flake-module.nix
+++ b/targets/lenovo-x1-installer/flake-module.nix
@@ -5,23 +5,41 @@
 {
   lib,
   self,
+  inputs,
   ...
 }: let
   name = "lenovo-x1-carbon";
   system = "x86_64-linux";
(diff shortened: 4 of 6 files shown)


leivos-unikie (commented):
Tested this
nix build github:remimimimimi/ghaf/fixed-installer#checks.x86_64-linux.installer --log-format bar-with-logs

The check seemed to pass. At first run there was a lot of output lines and some cleanup in the end. When I tried running it second time there was no output in the terminal. It created a result empty directory linked to nix store.

Is there be some explanation/documentation how that check works? Is it basically a build check and an installer test withing a VM or does it check something more?


**Tests on Lenovo-X1**
- Built the installer with
nix build --verbose github:remimimimimi/ghaf/fixed-installer#lenovo-x1-carbon-gen11-debug-installer
- ci-test-automation tests ok
- All apps launch and poweroff/reboot buttons work

remimimimimi:
> Tested this nix build github:remimimimimi/ghaf/fixed-installer#checks.x86_64-linux.installer --log-format bar-with-logs
> 
> The check seemed to pass. At first run there was a lot of output lines and some cleanup in the end. When I tried running it second time there was no output in the terminal. It created an empty result directory linked to nix store.
> 
> Is there be some explanation/documentation how that check works? Is it basically a build check and an installer test withing a VM or does it check something more?

There's [documentation on wiki](https://nixos.wiki/wiki/NixOS_Testing_library). But tl;dr is that it builds derivation, and during the build process runs tests. So because you haven't changed derivation, you saw no output.

leivos-unikie:
When testing this ghaf installer yesterday at the first time (built with nix build --verbose github:remimimimimi/ghaf/fixed-installer#lenovo-x1-carbon-gen11-debug-installer) I got 2 NixOS generations available at boot menu:
IMG_2592

The older generation had still old labwc version.

Today I tried the installer from ghaf mainline and from this PR again but could not reproduce the situation above. I am wondering how did that happen? I know that I have had some older ghaf version in the nvme but I had "disabled" that by booting from ghaf-installer and running something like this: `sudo dd if=/dev/zero of=/dev/nvme0n1 count=100 bs=4M`. Could it be that if there were some leftovers from the previous installation that the installer did some nixos-rebuild on top of that? I would expect the installer to wipe everything and install fresh ghaf.

leivos-unikie:
And this morning I also noticed that the installer from this PR creates always a new 'Linux Boot Manager' entry to the Boot Priority Order list in BIOS. To be able to boot from USB SSD again it requires manually booting to UEFI BIOS and lowering or discarding the 'Linux Boot Manager' entry. This can be a problem from the automated testing point of view. We have been preparing to run also installer tests in automated setup but 'Linux Boot Manager' will prevent that.

One observation more: This installer takes around 7min to run, whereas the old made it in 1min. But I understand this is the cost we have to pay for using nixos-install instead of just flashing.

remimimimimi:
> When testing this ghaf installer yesterday at the first time (built with nix build --verbose github:remimimimimi/ghaf/fixed-installer#lenovo-x1-carbon-gen11-debug-installer) I got 2 NixOS generations available at boot menu: IMG_2592
> 
> The older generation had still old labwc version.
> 
> Today I tried the installer from ghaf mainline and from this PR again but could not reproduce the situation above. I am wondering how did that happen? I know that I have had some older ghaf version in the nvme but I had "disabled" that by booting from ghaf-installer and running something like this: `sudo dd if=/dev/zero of=/dev/nvme0n1 count=100 bs=4M`. Could it be that if there were some leftovers from the previous installation that the installer did some nixos-rebuild on top of that? I would expect the installer to wipe everything and install fresh ghaf.

That's probably because you overwrote with zeros only 400M. Also, full system wipe is done (optionally) to my knowledge only by Debian installer, others simply write on top.

> And this morning I also noticed that the installer from this PR creates always a new 'Linux Boot Manager' entry to the Boot Priority Order list in BIOS. To be able to boot from USB SSD again it requires manually booting to UEFI BIOS and lowering or discarding the 'Linux Boot Manager' entry. This can be a problem from the automated testing point of view. We have been preparing to run also installer tests in automated setup but 'Linux Boot Manager' will prevent that.
> 
> One observation more: This installer takes around 7min to run, whereas the old made it in 1min. But I understand this is the cost we have to pay for using nixos-install instead of just flashing.

It creates new boot entry on every installation because otherwise the system may be unbootable. This behavior is because of `disko-install` `--write-efi-boot-entries` flag, and it’s also required to build a system without internet.

Otherwise, because of internal structure it attempts to rebuild packages, which require internet connection. [That was fixed upstream only recently](https://github.com/NixOS/nixpkgs/pull/306339).

leivos-unikie:
> It creates new boot entry on every installation because otherwise the system may be unbootable. This behavior is because of `disko-install` `--write-efi-boot-entries` flag, and it’s also required to build a system without internet.

ghaf-installer in the current ghaf mainline does not produce any 'Linux Boot Manager' entry to BIOS. I think in that case it's enough that NVMe0 is in the Boot Order list.

leivos-unikie:
> It creates new boot entry on every installation because otherwise the system may be unbootable. This behavior is because of `disko-install` `--write-efi-boot-entries` flag, and it’s also required to build a system without internet.
> 
> Otherwise, because of internal structure it attempts to rebuild packages, which require internet connection. [That was fixed upstream only recently](https://github.com/NixOS/nixpkgs/pull/306339).

I tested without that `--write-efi-boot-entries` flag. Installation fails both with and without internet connection.

brianmcgillion:
PXL_20240506_135341824 MP

Mic92:
If we switch from the iso here to pure disko than these hash mismatch issues should be fixed by: https://github.com/nix-community/disko/pull/625

Mic92:
Noticed that in the latest version I switched to use `closureInfo`, which is now also needed for offline installation: https://github.com/nix-community/disko/pull/625/files?short_path=cb7823f#diff-cb7823f0751d86126c961181be2eea6ecf59d368c3019ddd0ce86dab10ec92a3

leivos-unikie:
Related to test automation, @remimimimimi thanks for mentioning about efibootmgr. I drafted a script to delete all Linux Boot Manager entries, so that we are able to automate it and boot again from USB SSD after running test with ghaf-installer.
```
for f in $(efibootmgr | grep Linux | awk 'NR > 0 {print $1}' | cut -c 5-8)
do
    sudo efibootmgr -q -b ${f} -B
done
```

There is just the potential risk that in case ghaf-installer installs a broken ghaf which cannot boot (and the active Linux Boot Manager entry) we can't run the script above and cannot handle the situation anymore automatically (need to boot manually to UEFI BIOS and edit the boot order). Maybe the best way to reduce this risk is to run the tests first on ghaf image (booted from USB SSD) and if the plain ghaf image is bootable only then proceed to run tests for ghaf-installer -> ghaf.

leivos-unikie:
```
[samuli@nixos:~/builds/installer]$ nix build github:remimimimimi/ghaf/fixed-installer#lenovo-x1-carbon-gen11-debug-installer
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'nixos-24.05.20240515.a6c1060-x86_64-linux.iso'
         whose name attribute is located at /nix/store/vgh4cg6w837n512lfwkqhw23xbvs7gh4-source/pkgs/stdenv/generic/make-derivation.nix:331:7

       … while evaluating attribute 'sources' of derivation 'nixos-24.05.20240515.a6c1060-x86_64-linux.iso'

         at /nix/store/vgh4cg6w837n512lfwkqhw23xbvs7gh4-source/nixos/lib/make-iso9660-image.nix:76:3:

           75|
           76|   sources = map (x: x.source) contents;
             |   ^
           77|   targets = map (x: x.target) contents;

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: The option `disko.devices.disk.disk1.device' is defined multiple times while it's expected to be unique.

       Definition values:
       - In `/nix/store/95xn80gk76xjpfddhylp9y98b3irank5-source/modules/hardware/lenovo-x1/definitions': "/dev/nvme0n1"
       - In `/nix/store/vgh4cg6w837n512lfwkqhw23xbvs7gh4-source/flake.nix': "/dev/nvme0n1"
       Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```

UnikieHR:
image

PR #595, opened 2024-05-08
windows-launcher: Replace zenity with yad

mikatammi:
## Description of changes

[windows-launcher: Replace zenity with yad](https://github.com/tiiuae/ghaf/pull/595/commits/3649a0d2eaf11eb7fa78605b59d4eafcfe79a053)

yad is a fork of zenity and it has much less heavy dependencies, so it
works as a nice drop-in replacement for zenity for now.

## Checklist for things done

## Testing

Haven't built or tested this yet at all. For testing windows-launcher qemu issue also needs to be fixed first


packages/windows-launcher/default.nix
```diff
@@ -143,14 +143,14 @@
         fi
 
         if ! ${windowsLauncher} $FILE; then
-          ${gnome.zenity}/bin/zenity --error --text="Failed to run Windows VM: $?"
+          ${yad}/bin/yad --error --text="Failed to run Windows VM: $?"
```
nesteroff, line 146: Thanks, that is a nice tool. I didn't know about it. I can't test it at the moment but it looks like there is no --error parameter. I guess it should be replaced with --image=gtk-dialog-error.

mikatammi:
The problem with zenity is its massive dependencies, I think it needs both Gtk3 and Gtk4 and Qt. Yad only needs Gtk3 I think. It's not the optimal or the best solution to have this shellscript at all, but I understand that it was quite quickly created for demo purposes, and this would just make our dependency load much lighter. The problem with these GNOME applications is that they will always bring half of the GNOME as a dependency with them.

brianmcgillion:
cherry-picked so closing

PR #624, opened 2024-05-21
Add helper scripts in debug mode

leivos-unikie:
## Description of changes

Added few scripts which can make testing faster and easier:
- 2 sysbench scripts which are used in automated testing but currently copied into every VM at each test round
- perf test script for icicle kit target, no need to copy that either
- script for removing Linux Boot Manager entries from UEFI Boot order list

## Checklist for things done

## Testing

**Levono-X1**
ghaf-host:
`rm-linux-bootmgrs`

**Orin AGX/NX**
ghaf-host:
`nvpmodel-check 3` -> Power mode check ok
`nvpmodel-check 2` -> Unexpected power mode detected

**Icicle-kit**
`perf-test-icicle-kit`

**All targets except Icicle-kit**
ghaf-host:
`sysbench-fileio-test`

ghaf-host or any VM:
`sysbench-test`

### DIFF
--- a/modules/common/development/scripts/perf_test_icicle_kit.nix
+++ b/modules/common/development/scripts/perf_test_icicle_kit.nix
@@ -1,0 +1,29 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  writeShellApplication,
+  lib,
+  ...
+}:
+writeShellApplication {
+  name = "perf-test-icicle-kit";
+  text = ''
+    time {
+    perf bench sched messaging;
+    perf bench sched pipe -l 50000;
+    perf bench syscall basic;
+    perf bench mem memcpy;
+    perf bench mem memset;
+    perf bench mem find_bit -i 5 -j 1000;
+    perf bench numa mem -p 1 -t 1 -P 1024 -C 0 -M 0 -s 5 -zZq --thp 1 --no-data_rand_walk;
+    perf bench futex all;
+    perf bench epoll wait;
+    perf bench epoll ctl;
+    perf bench internals synthesize -i 1000;
+    perf bench internals kallsyms-parse -i 10;
+    } | tee -a perf_results_YYYY-MM-DD_BUILDER-BuildID_SDorEMMC
+  '';
+  meta = with lib; {
+    description = "Perf test script customized for measuring ghaf performance on Microchip Icicle Kit target";
+  };
+}
--- a/modules/common/development/scripts/sysbench_fileio_test.nix
+++ b/modules/common/development/scripts/sysbench_fileio_test.nix
@@ -1,0 +1,53 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  writeShellApplication,
+  lib,
+  ...
+}:
+writeShellApplication {
+  name = "sysbench-fileio-test";
+  text = ''
+    # Test set to be run with sysbench
+
+    # These variable needs to be given on the command line. For example: sysbench-fileio-test 20
+    THREADS="''$1"
+
+    # Create a directory for the results
+    RESULT_DIR="sysbench_results"
+    echo -e "\nCreating directory for test results:\n./''$RESULT_DIR"
+    mkdir -p ''$RESULT_DIR
+
+    # Create test_info file with system information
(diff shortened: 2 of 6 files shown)


brianmcgillion (commented):
Could you use writeShellApplication instead of writeScriptBin. It adds some sanity checking (https://nixos.asia/en/writeShellApplication)  just requires adding [text = " .... "] Around what you have.

modules/common/development/scripts/sysbench_fileio_test.nix
```diff
@@ -0,0 +1,61 @@
...
+    # Create test_info file with system information
+    echo -e "\nSaving information about test environment to ./''$RESULT_DIR/test_info\n"
+    echo -e "''$(lscpu)" "\n\n" "''$(free)" "\n\n" "''$(df)" "\n\n" >> ./''$RESULT_DIR/test_info
+    echo -e "\nHost: ''$(hostname)\n" | tee -a ./''$RESULT_DIR/test_info
+
+    # Check cpu architecture for specific actions
+    if lscpu | grep -qw "aarch64"; then
+        echo -e "\nSystem architecture is aarch64." | tee -a ./''$RESULT_DIR/test_info
+        if hostname | grep -qw "ghaf-host"; then
+            echo -e "''$(nvpmodel -q)" "\n" >> ./''$RESULT_DIR/test_info
```
mikatammi, line 29: I want to point out nvpmodel is NVIDIA specific, and not aarch64 specific

modules/common/development/scripts/perf_test_icicle_kit.nix
```diff
@@ -0,0 +1,20 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{pkgs, ...}:
```
mikatammi, line 3: Don't put pkgs here if you are using callPackage mechanism.

The callPackage mechansim works so, you can put every package name here into this argument type attribute set, which seems to be just `{writeScriptBin}` at this point.

I think generally in all these files it would be better idea to use writeShellApplication instead of writeScriptBin

modules/common/development/scripts/perf_test_icicle_kit.nix
```diff
@@ -0,0 +1,20 @@
...
+with pkgs;
```
mikatammi, line 4: Don't use `with pkgs;`

mikatammi: This also applies to all other files, I just don't want to add 4 comments with same text

mikatammi:
Examples of how I fixed things here: https://github.com/tiiuae/ghaf/pull/626

mikatammi:
Example of using writeShellApplication here: https://github.com/tiiuae/ghaf/blob/main/packages/wifi-connector/default.nix
or here: https://github.com/tiiuae/ghaf/blob/main/packages/windows-launcher/default.nix

One benefit of using writeShellApplication is also that it uses shellcheck to check the script during build time

modules/common/development/scripts/sysbench_fileio_test.nix
mikatammi: More compatible way would be probably to just enable the cores, and set the cores to use performance governor through sysfs. (That is actually partly what nvpmodel internally does.) Maybe we could have both?

leivos-unikie: > I want to point out nvpmodel is NVIDIA specific, and not aarch64 specific

In our context we will check nvpmodel only on Orin AGX/NX targets. I cleaned and separated the check script to own file.

leivos-unikie:
> Example of using writeShellApplication here: https://github.com/tiiuae/ghaf/blob/main/packages/wifi-connector/default.nix or here: https://github.com/tiiuae/ghaf/blob/main/packages/windows-launcher/default.nix
> 
> One benefit of using writeShellApplication is also that it uses `shellcheck` to check the script during build time
> 
> EDIT: Didn't notice Brian already commented the same thing :)

Thanks. Changed to that writeShellApplication. Shellcheck at least forced to add a lot of double quotes "to prevent globbing and word splitting".

modules/common/development/scripts/sysbench_fileio_test.nix
mikatammi: This is good and more future proof. Now if we start testing the iMX8 targets, it won't cause problems

brianmcgillion (approved):

PR #625, opened 2024-05-21
[X1]: logging: To centralize logs using grafana

vunnyso:
## Checklist for things done

## Testing
This PR can be tested with multiple `Lenovo X1` devices and each system will upload system logs with unique `mac address`  to remote cloud server.

**Prerequisite:** 
Its recommended to  connect with `wifi` to get unique Mac Address and to upload logs to remote server.

By default system will start collecting logs once its booted.

WAL(Write Ahead Log) will be stored in `/var/lib/private/alloy/data-alloy/loki.write.remote/wal` path which can be accessed through `admin-vm` as well `ghaf-host` (WAL is not in human readable format)

**To View logs remotely on grafana-server**

Open browser and type following [address](https://ghaflogs.vedenemo.dev/explore), it will prompt with `credentials`.

Please request for `Credentials`.  Once you managed to login, choose appropriate labels as in example below.

**Note**:

- If there are are logs from multiple devices, it will be go under `systemdJournalLogs` label.
- To find `mac address` with which logs are pushed run below command in `admin-vm`
 [ghaf@admin-vm:~]$ sudo cat /var/lib/private/alloy/MACAddress

image

### DIFF
--- a/modules/common/default.nix
+++ b/modules/common/default.nix
@@ -18,5 +18,6 @@
     ./services
     ./networking
     ../hardware/definition.nix
+    ./logging
   ];
 }
--- a/modules/common/logging/client.nix
+++ b/modules/common/logging/client.nix
@@ -45,5 +45,9 @@ in {
     };
 
     services.alloy.enable = true;
+    # Once alloy.service in admin-vm stopped this service will
+    # still keep on retrying to send logs batch, so we need to
+    # stop it forcefully.
+    systemd.services.alloy.serviceConfig.TimeoutStopSec = 4;
   };
 }
--- a/modules/common/logging/client.nix
+++ b/modules/common/logging/client.nix
@@ -1,0 +1,49 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  ...
+}: let
+  cfg = config.ghaf.logging.client;
+  endpointUrl = config.ghaf.logging.client.endpoint;
+in {
+  options.ghaf.logging.client.endpoint = lib.mkOption {
+    description = ''
+      Assign endpoint url value to the alloy.service running in
+      different log producers. This endpoint URL will include
+      protocol, upstream, address along with port value.
+    '';
+    type = lib.types.str;
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.etc."alloy/client.alloy" = {
+      text = ''
+        discovery.relabel "journal" {
+          targets = []
+          rule {
+            source_labels = ["__journal__hostname"]
+            target_label  = "nodename"
+          }
+        }
+
+        loki.source.journal "journal" {
+          path          = "/var/log/journal"
+          relabel_rules = discovery.relabel.journal.rules
+          forward_to    = [loki.write.adminvm.receiver]
+        }
+
+        loki.write "adminvm" {
+          endpoint {
+            url = "${endpointUrl}"
+          }
+        }
+      '';
+      # The UNIX file mode bits
+      mode = "0644";
+    };
+
+    services.alloy.enable = true;
+  };
+}
--- a/modules/common/logging/default.nix
+++ b/modules/common/logging/default.nix
@@ -1,0 +1,38 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{lib, ...}: let
+  inherit (lib) mkOption types;
+in {
+  # Creating logging configuration options needed across the host and vms
+  options.ghaf.logging = {
+    client.enable = mkOption {
+      description = ''
+        Enable logging client service. Currently we have grafana alloy
+        running as client which will upload system journal logs to
+        grafana alloy running in admin-vm.
+      '';
+      type = types.bool;
+      default = false;
+    };
+
+    listener.address = mkOption {
+      description = ''
+        Listener address will be used where log producers will
+        push logs and where admin-vm alloy.service will be
+        keep on listening or receiving logs.
+      '';
+      type = types.str;
+    };
+
+    listener.port = mkOption {
+      description = ''
+        Listener port for the logproto endpoint which will be
+        used to receive logs from different log producers.
+        Also this port value will be used to open the port in
+        the admin-vm firewall.
+      '';
+      type = types.port;
+      default = 9999;
+    };
+  };
+}
--- a/modules/common/logging/hw-mac-retrieve.nix
+++ b/modules/common/logging/hw-mac-retrieve.nix
@@ -1,0 +1,42 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  # TODO: replace sshCommand and MacCommand with givc rpc to retrieve Mac Address
+  sshCommand = "${pkgs.sshpass}/bin/sshpass -p ghaf ${pkgs.openssh}/bin/ssh -o StrictHostKeyChecking=no ghaf@net-vm";
+  macCommand = "cat /sys/class/net/wlp0s5f0/address";
+  macAddressPath = config.ghaf.logging.identifierFilePath;
+in {
+  options.ghaf.logging.identifierFilePath = lib.mkOption {
+    description = ''
+      This configuration option used to specify the identifier file path.
+      The identifier file will be text file which have unique identification
+      value per machine so that when logs will be uploaded to cloud
+      we can identify its origin.
+    '';
+    type = lib.types.path;
+    example = "/tmp/MACAddress";
+  };
+
+  config = lib.mkIf config.ghaf.logging.client.enable {
+    # TODO: Remove hw-mac.service and replace with givc rpc later
+    systemd.services."hw-mac" = {
+      description = "Retrieve MAC address from net-vm";
+      wantedBy = ["alloy.service"];
+      requires = ["network-online.target"];
+      serviceConfig = {
+        Type = "oneshot";
+        RemainAfterExit = "yes";
+        # Make sure we can ssh before we retrieve mac address
+        ExecStartPre = "${sshCommand} ls";
+        ExecStart = ''
+          ${pkgs.bash}/bin/bash -c "echo -n $(${sshCommand} ${macCommand}) > ${macAddressPath}"
+        '';
+      };
+    };
+  };
+}
--- a/modules/microvm/virtualization/microvm/adminvm.nix
+++ b/modules/microvm/virtualization/microvm/adminvm.nix
@@ -83,8 +83,8 @@
                 # Creating a persistent log-store which is mapped on ghaf-host
                 # This is only to preserve logs state across adminvm reboots
                 tag = "log-store";
-                source = "/var/lib/loki";
-                mountPoint = "/var/lib/loki";
+                source = "/var/lib/private/alloy";
+                mountPoint = "/var/lib/private/alloy";
                 proto = "virtiofs";
               }
             ];
--- a/modules/microvm/virtualization/microvm/adminvm.nix
+++ b/modules/microvm/virtualization/microvm/adminvm.nix
@@ -8,6 +8,7 @@
   configHost = config;
   vmName = "admin-vm";
   macAddress = "02:00:00:AD:01:01";
+  isLoggingEnabled = config.ghaf.logging.client.enable;
 
   adminvmBaseConfiguration = {
     imports = [
@@ -15,6 +16,9 @@
         inherit config lib vmName macAddress;
         internalIP = 10;
       })
+      # We need to retrieve mac address and start log aggregator
+      ../../../common/logging/hw-mac-retrieve.nix
+      ../../../common/logging/logs-aggregator.nix
       ({lib, ...}: {
         ghaf = {
           users.accounts.enable = lib.mkDefault configHost.ghaf.users.accounts.enable;
@@ -29,10 +33,19 @@
           systemd = {
             enable = true;
             withName = "adminvm-systemd";
+            withNss = true;
+            withResolved = true;
             withPolkit = true;
             withTimesyncd = true;
             withDebug = configHost.ghaf.profiles.debug.enable;
           };
+
+          # Log aggregation configuration
+          logging.client.enable = isLoggingEnabled;
+          logging.listener.address = configHost.ghaf.logging.listener.address;
+          logging.listener.port = configHost.ghaf.logging.listener.port;
+          logging.identifierFilePath = "/var/lib/private/alloy/MACAddress";
+          logging.server.endpoint = "http://ghaflogs.vedenemo.dev:3100/loki/api/v1/push";
         };
 
         system.stateVersion = lib.trivial.release;
@@ -41,7 +54,7 @@
         nixpkgs.hostPlatform.system = configHost.nixpkgs.hostPlatform.system;
 
         networking = {
-          firewall.allowedTCPPorts = [];
+          firewall.allowedTCPPorts = lib.mkIf isLoggingEnabled [config.ghaf.logging.listener.port];
           firewall.allowedUDPPorts = [];
         };
 
@@ -56,14 +69,26 @@
         microvm = {
           optimize.enable = true;
           hypervisor = "cloud-hypervisor";
-          shares = [
-            {
-              tag = "ro-store";
-              source = "/nix/store";
-              mountPoint = "/nix/.ro-store";
-              proto = "virtiofs";
-            }
-          ];
+          shares =
+            [
+              {
+                tag = "ro-store";
+                source = "/nix/store";
+                mountPoint = "/nix/.ro-store";
+                proto = "virtiofs";
+              }
+            ]
+            ++ lib.optionals isLoggingEnabled [
+              {
+                # Creating a persistent log-store which is mapped on ghaf-host
+                # This is only to preserve logs state across adminvm reboots
+                tag = "log-store";
+                source = "/var/lib/loki";
+                mountPoint = "/var/lib/loki";
+                proto = "virtiofs";
+              }
+            ];
+
           writableStoreOverlay = lib.mkIf config.ghaf.development.debug.tools.enable "/nix/.rw-store";
         };
         imports = [../../../common];
--- a/modules/microvm/virtualization/microvm/appvm.nix
+++ b/modules/microvm/virtualization/microvm/appvm.nix
@@ -32,6 +32,8 @@
           inherit (vm) macAddress;
           internalIP = index + 100;
         })
+        # To push logs to central location
+        ../../../common/logging/client.nix
         ({
           lib,
           config,
@@ -66,6 +68,9 @@
               withDebug = configHost.ghaf.profiles.debug.enable;
               withHardenedConfigs = true;
             };
+            # Logging client configuration
+            logging.client.enable = configHost.ghaf.logging.client.enable;
+            logging.client.endpoint = configHost.ghaf.logging.client.endpoint;
           };
 
           # SSH is very picky about the file permissions and ownership and will
--- a/modules/profiles/laptop-x86.nix
+++ b/modules/profiles/laptop-x86.nix
@@ -8,6 +8,8 @@
 }: let
   powerControl = pkgs.callPackage ../../packages/powercontrol {};
   cfg = config.ghaf.profiles.laptop-x86;
+  listenerAddress = config.ghaf.logging.listener.address;
+  listenerPort = toString config.ghaf.logging.listener.port;
 in {
   imports = [
     ../desktop/graphics
@@ -112,6 +114,12 @@ in {
       profiles = {
         applications.enable = false;
       };
+
+      # Logging configuration
+      logging.client.enable = true;
+      logging.client.endpoint = "http://${listenerAddress}:${listenerPort}/loki/api/v1/push";
+      logging.listener.address = "admin-vm-debug";
+      logging.listener.port = 9999;
     };
   };
 }
--- a/modules/common/logging/logs-aggregator.nix
+++ b/modules/common/logging/logs-aggregator.nix
@@ -20,6 +20,7 @@ in {
   };
 
   config = lib.mkIf config.ghaf.logging.client.enable {
+    environment.etc."loki/pass" = {text = "ghaf";};
     environment.etc."alloy/logs-aggregator.alloy" = {
       text = ''
         local.file "macAddress" {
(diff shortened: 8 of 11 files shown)


modules/microvm/virtualization/microvm/loki-local-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+# SPDX-License-Identifier: Apache-2.0
+auth_enabled: false
+
+server:
+  http_listen_port: 3100
+  grpc_listen_port: 9096
+
+common:
+  instance_addr: 127.0.0.1
+  path_prefix: /var/lib
```
vunnyso, line 11: Please change path_prefix: `/var/lib` to path_prefix: `/var/lib/loki`

modules/common/log/grafana-alloy.nix
```diff
@@ -0,0 +1,51 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  pkgs,
+  config,
+  lib,
+  hostName,
+  ...
+}: let
+  admin-ip-port = "192.168.101.10:3100";
```
vilvo, line 10: Please use the `admin-vm` hostname from `/etc/hosts` instead of duplicating the IP here. Port is fine.

modules/common/log/loki.nix
```diff
@@ -0,0 +1,51 @@
...
+  systemd.tmpfiles.rules = [
+    "d /var/lib/loki 0777 loki loki - -"
+  ];
+
+  services.loki = {
+    inherit (config.ghaf.logging.client) enable;
+    configuration = {
+      auth_enabled = false;
+      server = {
+        http_listen_port = 3100;
```
vilvo, line 16: Please add an option for the port in [modules/common/log/default.nix](https://github.com/tiiuae/ghaf/pull/625/files#diff-251d008edd24f623160e847eb1ced62c07dea3ebf0b710f10554b8e7fa524ff0)

modules/common/log/loki.nix
```diff
@@ -0,0 +1,51 @@
...
+
+      ruler = {
+        alertmanager_url = "http://localhost:9093";
+      };
+
+      query_range.cache_results = true;
+    };
+  };
+
+  networking.firewall.allowedTCPPorts = [3100];
```
vilvo, line 50: Please reuse the configured logging port here instead of hard-coded literal. Please mention in the option description that a firewall allowed TCP port policy will be created.

modules/common/log/grafana-alloy.nix
vunnyso: Done.

modules/common/log/loki.nix
vunnyso: Done.

modules/common/log/loki.nix
vunnyso: Done.

modules/common/log/default.nix
```diff
@@ -0,0 +1,56 @@
...
+    default = 9999;
+  };
+
+  options.ghaf.logging.server.address = lib.mkOption {
+    description = ''
+      This is the IP of grafana cloud server. If server logging is enabled
+      we need to specify the IP where device logs will be uploaded.
+    '';
+    type = lib.types.str;
+    default = "x.x.x.x";
```
vilvo, line 54: I would remove invalid default and leave the specific declaration to targets.

modules/common/log/logs-process.nix
```diff
@@ -0,0 +1,77 @@
...
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  cfg = config.ghaf.logging.client;
+  listenerAddress = "admin-vm-debug";
+  listenerPort = config.ghaf.logging.listener.port;
+  sshCommand = "${pkgs.sshpass}/bin/sshpass -p ghaf ${pkgs.openssh}/bin/ssh -o StrictHostKeyChecking=no ghaf@net-vm";
```
vilvo, line 12: Please add a `# TODO replace sshCommand MacCommand with givc rpc` to indicate that the SSH-based integration is not intended as long term

modules/common/log/logs-source.nix
```diff
@@ -0,0 +1,48 @@
...
+  config,
+  lib,
+  hostName,
+  ...
+}: let
+  admin-ip-port = "admin-vm-debug:${toString config.ghaf.logging.listener.port}";
+  cfg = config.ghaf.logging.client;
+in {
+  config = lib.mkIf cfg.enable {
+    # environment.systemPackages = [pkgs.grafana-alloy];
```
vilvo, line 13: Minor: please remove commented out declaration

modules/common/log/loki.nix
```diff
@@ -0,0 +1,60 @@
...
+    systemd.tmpfiles.rules = [
+      "d /var/lib/loki 0777 loki loki - -"
+    ];
+
+    services.loki = {
+      enable = true;
+      configuration = {
+        auth_enabled = false;
+        server = {
+          http_listen_port = 3100;
```
vilvo, line 22: Please use https://github.com/tiiuae/ghaf/pull/625/files#diff-251d008edd24f623160e847eb1ced62c07dea3ebf0b710f10554b8e7fa524ff0R33 instead

modules/profiles/laptop-x86.nix
```diff
@@ -112,6 +112,10 @@ in {
       profiles = {
         applications.enable = false;
       };
+
+      # Logging services
+      logging.client.enable = true;
+      logging.server.enable = true;
     };
```
vilvo, line 119: The cloud grafana logging server ip declaration could go here

modules/common/log/default.nix
vunnyso: Removed default value.

modules/common/log/logs-process.nix
vunnyso: Added TODO comment.

modules/profiles/laptop-x86.nix
vunnyso: Added server address declaration.

modules/common/log/loki.nix
vunnyso: Fixed.

modules/common/log/logs-source.nix
vunnyso: Removed.

modules/common/default.nix
```diff
@@ -18,5 +18,6 @@
     ./services
     ./networking
     ../hardware/definition.nix
+    ./log
```
flokli, line 21:
```suggestion
    ./logging
```

modules/common/log/logging-client.nix
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
```
flokli, line 2: This is not a module, and only calls logs-source.nix, passing along the parameters it's called with. Can we remove that trampoline?

modules/common/log/logs-process.nix
```diff
@@ -0,0 +1,78 @@
...
+  remoteLoggingAddress = config.ghaf.logging.server.address;
+  loggingPort = config.ghaf.logging.port;
+  lokiForward =
+    if config.ghaf.logging.server.enable
+    then "loki.write.local.receiver, loki.write.remote.receiver"
+    else "loki.write.local.receiver";
+in {
+  config = lib.mkIf cfg.enable {
+    # TODO: Remove hw-mac.service and replace with givc rpc later
+    systemd.services."hw-mac" = {
```
flokli, line 24: Can this be moved to another module, even with the hacky workaround we have currently? All that alloy has to do with it is read from /var/lib/private/alloy/MACAddress, but it doesn't need to be coupled to actually also contain the implementation of that.

modules/common/log/logs-process.nix
```diff
@@ -0,0 +1,78 @@
...
+        loki.write "local" {
+          endpoint {
+            url = "http://localhost:${toString loggingPort}/loki/api/v1/push"
+          }
+          external_labels = {systemdJournalLogs = local.file.macAddress.content }
+        }
+
+        loki.write "remote" {
+          endpoint {
+            url = "http://${remoteLoggingAddress}:${toString loggingPort}/loki/api/v1/push"
```
flokli, line 55: I'd just ask for the full endpoint in the module system (including path) here. We might want to change the scheme, or path coming after that. Just asking for an URL is gonna be less work later.

modules/common/log/loki.nix
```diff
@@ -0,0 +1,61 @@
...
+  lib,
+  ...
+}: let
+  loki_data_dir = "/var/lib/loki";
+  cfg = config.ghaf.logging.client;
+  loggingPort = config.ghaf.logging.port;
+in {
+  config = lib.mkIf cfg.enable {
+    systemd.tmpfiles.rules = [
+      "d /var/lib/loki 0777 loki loki - -"
```
flokli, line 15: StateDirectory=loki on the loki systemd service should already do that, no?

modules/microvm/virtualization/microvm/adminvm.nix
```diff
@@ -55,14 +63,25 @@
...
+              {
+                tag = "ro-store";
+                source = "/nix/store";
+                mountPoint = "/nix/.ro-store";
+                proto = "virtiofs";
+              }
+            ]
+            ++ lib.optionals isLoggingEnabled [
+              {
+                # Creating a persistent log-store which is mapped on ghaf-host
```
flokli, line 77: This is only to preserve state across adminvm reboots, correct? Can the comment be extended to mention that?

modules/common/log/default.nix
```diff
@@ -0,0 +1,55 @@
...
+      are created
+    '';
+    type = lib.types.int;
+    default = 3100;
+  };
+
+  options.ghaf.logging.listener.port = lib.mkOption {
+    description = ''
+      Assign port value to grafana alloy source service running in
+      different clients. Using this port value firewall allowed TCP port
```
flokli, line 39: The grammar is a bit broken here. Maybe just "opens that port in the firewall"?

Though technically I disagree with doing this, I'd feel more comfortable if this would be done somewhere closer in the machine configs, and only for the network interfaces we need to accept traffic to.

modules/common/log/logs-process.nix
vunnyso: Done.

modules/common/log/logs-process.nix
vunnyso: Good point. Decoupled and moved different module.

modules/common/default.nix
vunnyso: Changed.

modules/common/log/default.nix
vunnyso: Can you please suggest where it can done?

modules/common/logging/logs-source.nix
```diff
@@ -0,0 +1,46 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  hostName,
+  ...
+}: let
+  admin-ip-port = "admin-vm-debug:${toString config.ghaf.logging.listener.port}";
```
flokli, line 9: Same here, this should just ask for the url to push to as a config option.

modules/common/log/default.nix
flokli: Instead of opening the firewall in this module, I'd expect the part setting `ghaf.logging.listener.port` in the machine config to also contain a statement opening the firewall.

modules/microvm/virtualization/microvm/adminvm.nix
flokli: Thanks!

modules/common/log/loki.nix
flokli: Not a "bug" necessarily, but having a StateDirectory = loki in that unit file in case cfg.dataDir is /var/lib/loki sounds like a good thing to upstream. We can unconditionally extend the Loki systemd unit file config downstream in the meantime.

flokli: The nice part is that systemd automatically ensures everything is owned by the user when the service is started.

modules/common/log/logs-process.nix
flokli: No, this is still not a config option, but some math in a let binding. See the other comments.

modules/common/log/logging-client.nix
flokli: We can import code that sets config options in a more generic NixOS module, but instantiating like this feels wrong.

modules/common/logging/hw-mac-retrieve.nix
```diff
@@ -0,0 +1,31 @@
...
+  config = lib.mkIf cfg.enable {
+    # TODO: Remove hw-mac.service and replace with givc rpc later
+    systemd.services."hw-mac" = {
+      description = "Retrieve MAC address from net-vm";
+      wantedBy = ["alloy.service"];
+      requires = ["network-online.target"];
+      serviceConfig = {
+        Type = "oneshot";
+        RemainAfterExit = "yes";
+        # Make sure we can ssh before we retrieve mac address
```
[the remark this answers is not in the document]
flokli: Ah, the unconditional redirection to the output file is the problem.

We could use pkgs.writers.writeBash, buffer the output and only write on success, but that's probably ok to do in a followup too.

modules/common/logging/logs-source.nix
```diff
@@ -0,0 +1,46 @@
...
+in {
+  config = lib.mkIf cfg.enable {
+    environment.etc."alloy/config.alloy" = {
+      text = ''
+        discovery.relabel "journal" {
+          targets = []
+          rule {
+            source_labels = ["__journal__systemd_unit"]
+            target_label  = "unit"
+          }
```
flokli, line 20:
```suggestion
      rule {
        source_labels = ["__journal__systemd_unit"]
        target_label = "systemd_unit"
      }
      rule {
        source_labels = ["__journal__hostname"]
        target_label = "nodename"
      }
      rule {
        source_labels = ["__journal_syslog_identifier"]
        target_label = "syslog_identifier"
      }

```

(maybe different indent ofc)

modules/profiles/laptop-x86.nix
```diff
@@ -112,6 +112,11 @@ in {
       profiles = {
         applications.enable = false;
       };
+
+      # Logging services
+      logging.client.enable = true;
+      logging.server.enable = true;
+      logging.server.address = "ghaflogs.vedenemo.dev";
```
flokli, line 119: This is a bit confusing. Is this running in all VMs, just the host or both?

The log collection bits should be configured in a profile that gets imported in all VMs and the host, and the log processing thing in a profile that only gets imported on the admin-vm.

modules/common/logging/hw-mac-retrieve.nix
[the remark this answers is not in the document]
vunnyso: Sure, This `hw-mac.service` is temporary code, I have added TODO as well, It will be soon replaced with `givc` implementation. I prefer not to spend much efforts here.

barnabakos:
1 commit: eb93878834978b24001db7a1a84e72b00a9dd5cf
rebased to current tip of Ghaf main without conflicts: 4269cce084bf048ff8d536d0e1049623051f9d11

platform: **Lenovo X1 Carbon**
- starting apps: ok except issues with **AppFlowy**
- regression / ad-hoc testing: ok

#### works:
- logs are available both locally and remotely on grafana server

#### issues:
- starting **AppFlowy**:
     seems to crash often; sometimes cannot even select `Start with an anonymous session`
- **shutdown** takes a lot longer than mainline build; using the power icons from the app drawer
  - when **not connected** to Wifi network
    - shutdown for Reboot
      mainline: 10-11s
      PR:       100-105s
    - shutdown for Shutdown
      mainline: 4-5s
      PR:       90-100s
  - **connected** to Wifi network
    - shutdown for Reboot
      mainline: 11s
      PR:       14s
    - shutdown for Shutdown
      mainline: 4s
      PR:       8s
- **startup** time takes a few seconds longer
  mainline: 42-44s
  PR:       45-46s

johannarautanen:
**Tested with Lenovo X1 Carbon**, using commit: eb93878834978b24001db7a1a84e72b00a9dd5cf

**Issues:**

- Occasionally hard to open the Chromium manually (automated chromium-case was ok)

- Tried the reboot and shutdown -issues that Barna noticed and the issues reproduced: long reboot and shutdown times when there is no wifi-connection.


**Working:**

- `/var/lib/loki ` was visible in admin-vm and ghaf-host
- commands: `logcli query '{job="systemd-journal"}'` , `logcli query '{hostname="ghaf-host"}'` and `logcli query '{systemdJournalLogs="aa:bb:cc:dd:ee:ff"}' ` working ok.

(the wifi mac address that be found from Grafana can be checked via ` logcli query '{job="systemd-journal"}'`: 
Screenshot from 2024-07-09 07-59-57

- possible to view and save logs remotely on grafana-server:
Screenshot from 2024-07-08 16-10-21

modules/common/log/logging-client.nix
vunnyso: Now importing NixOS module into `admin-vm` instead of putting in another file.

modules/common/log/logs-process.nix
vunnyso: Fixed.

modules/common/log/loki.nix
vunnyso: We don't need `loki.service` anymore on device.

modules/common/log/default.nix
vunnyso: Done.

modules/common/logging/default.nix
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{lib, ...}:
+with lib; {
```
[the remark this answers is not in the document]
vunnyso: Fixed.

modules/common/logging/hw-mac-retrieve.nix
```diff
@@ -0,0 +1,41 @@
...
+}: let
+  cfg = config.ghaf.logging.client;
+  # TODO: replace sshCommand and MacCommand with givc rpc to retrieve Mac Address
+  sshCommand = "${pkgs.sshpass}/bin/sshpass -p ghaf ${pkgs.openssh}/bin/ssh -o StrictHostKeyChecking=no ghaf@net-vm";
+  macCommand = "cat /sys/class/net/wlp0s5f0/address";
+  macAddressPath = config.ghaf.logging.identifierFilePath;
+in {
+  options.ghaf.logging.identifierFilePath = lib.mkOption {
+    description = ''
+      Listener port for the logproto endpoint which will be
```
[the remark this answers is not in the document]
vunnyso: Sorry wrong description. Updated now.

modules/common/logging/hw-mac-retrieve.nix
```diff
@@ -0,0 +1,41 @@
...
+      used to receive logs from different log producers.
+    '';
+    type = lib.types.path;
+    example = "/tmp/MACAddress";
+  };
+
+  config = lib.mkIf cfg.enable {
```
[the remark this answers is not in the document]
vunnyso: Understood. Fixed.

modules/microvm/virtualization/microvm/appvm.nix
```diff
@@ -32,6 +32,9 @@
           inherit (vm) macAddress;
           internalIP = index + 100;
         })
+        (import ../../../common/logging/logs-source.nix {
```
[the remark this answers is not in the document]
vunnyso: Now imported as NixOS module.

vunnyso:
Hi @flokli shall I mark above comments from you as "Resolve conversation" if you okay with current implementation?

flokli:
> Hi @flokli shall I mark above comments from you as "Resolve conversation" if you okay with current implementation?

Ok

vunnyso:
With latest changes, long delay in reboot and shutdown services is fixed. However there can be still 3-4 sec extra delay when not connected to `Wifi` network.

brianmcgillion:
1 nit pick. then please rebase and we can do the final review to get this merged

brianmcgillion (approved):

modules/common/logging/client.nix
```diff
@@ -0,0 +1,53 @@
...
+      '';
+      # The UNIX file mode bits
+      mode = "0644";
+    };
+
+    services.alloy.enable = true;
+    # Once alloy.service in admin-vm stopped this service will
+    # still keep on retrying to send logs batch, so we need to
+    # stop it forcefully.
+    systemd.services.alloy.serviceConfig.TimeoutStopSec = 0;
```
flokli, line 51: Do we want to give it a second or two to actually have a chance to flush things out when applicable?

flokli: I propose setting this to 4 seconds like in the other file.

vunnyso: Thanks @flokli this is good point. I have kept **0**  because in case of shutdown/reboot I noticed in `admin-vm` alloy.service is getting stopped first and I will wait for 4 sec [here](https://github.com/tiiuae/ghaf/pull/625/files#diff-87d86a1e265f336e25adef2e8f43d3e185ddb94bbf7c68e8e7a59d0befa1c9fcR70) to flush things and log senders to stop, assuming once alloy.service in log-aggregator stopped, does it make sense for clients to wait more ? Since there will be no listener/receiver running it will be only retry happening to send batch during that time and it will never reach target. 
I could be wrong here in terms timing of services are stopped in case of shutdown/reboot.

flokli: if you shut down both adminvm and a regular "client vm", flushing logs from the "client vm" to the adminvm might not succeed, sure. However, both shutdowns happen at the same time, so using the same timeout as on the admin shouldn't cause an increase in total shutdown time.

If I interpret the manpage right, setting this to 0 does immediately send a `SIGKILL` right after the `SIGTERM`, which is quite brutal. We should give it some time to flush things to the adminvm if possible, clean up open files etc. This code path is also exercised if the "client vm" is simply rebooted.

barnabakos:
2 commits up to: 2e168da39a60d6c7838f47bafac824e8e38d31b2
PR is on top of latest tip of Ghaf main: 9d1a8faa23b11d9cc2ef15da01d9e6d2938fcb07

platform: **Lenovo X1 Carbon**
- starting apps: ok
- regression / ad-hoc testing: ok

#### works:
- logs are available both locally (WAL) and remotely on grafana server

#### fixed issues:
- Shutdown times are improved/fixed

#### notes:
- shutdown can take 5-10 seconds longer than mainline build
- startup takes a few seconds longer than mainline build

#### issues:
- N/A

modules/common/logging/client.nix
vunnyso: Sure, increasing timeout in client.nix didn't impact on total system shutdown time. So modified code accordingly.

PR #638, opened 2024-05-28
hardening: Adding hardening profile for lenovo X1 target

vunnyso:
## Description of changes
Booting Issue: [SP-4701](https://ssrc.atlassian.net/browse/SP-4701)

## Checklist for things done

## Testing

1) If we test only `host.secureboot` by disabling `host.kernel.hardening` it works fine on Lenovo X1 Carbon.
2) If we enable `host.kernel.hardening`  without `secureboot` system cannot boot and it will be stuck at `EFI stub: Measured initrd data into PCR9` stage as we are using Linux latest and is currently its pointing to `6.8.9` which is not tested before.
Booting Issue: [SP-4701](https://ssrc.atlassian.net/browse/SP-4701)
3) Last time hardened kernel was tested with `6.6.7` version and it was working fine.

### DIFF
--- a/modules/common/profiles/default.nix
+++ b/modules/common/profiles/default.nix
@@ -4,5 +4,6 @@
   imports = [
     ./debug.nix
     ./release.nix
+    ./hardening.nix
   ];
 }
--- a/modules/common/profiles/hardening.nix
+++ b/modules/common/profiles/hardening.nix
@@ -7,36 +7,47 @@
   ...
 }: let
   cfg = config.ghaf.profiles.hardening;
+  has_host = builtins.hasAttr "host" config.ghaf;
+  has_secureBoot = builtins.hasAttr "secureboot" config.ghaf.host;
+  has_guest = builtins.hasAttr "guest" config.ghaf;
 in {
   options.ghaf.profiles.hardening = {
     enable = lib.mkEnableOption "hardened profile";
   };
+  imports = [../../hardware/x86_64-generic/kernel/hardening.nix];
 
   config = lib.mkIf cfg.enable {
-    ghaf = {
-      host = {
-        # Enable some security features in the host configuration
-        secureboot.enable = true;
-
-        # Kernel hardening
-        kernel.hardening = {
-          enable = true;
-          usb.enable = true;
-          debug.enable = true;
-          virtualization.enable = true;
-          networking.enable = true;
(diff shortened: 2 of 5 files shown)


modules/common/profiles/host_hardening.nix
```diff
@@ -0,0 +1,47 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
```
brianmcgillion, line 3: should this file should be called hardening.nix? and not host-hardening? Because it includes settings for both guest and host

vunnyso: Changed to `hardening.nix`

vilvo (commented):
On Testing 2. - Does the `6.8.9` hardened kernel boot without `secureboot`? Or `6.8.9` hardened kernel just does not boot at all?

The structural change looks good to me. I think merging should depend on clarifications to above questions. When we have the hardening profile, we should maintain it even if it's not enabled by default. Otherwise users may get wrong impression on ghaf hardening API. The debugging and maintenance could be scoped outside this PR but my personal preference would be to include the maintenance with this. Maybe as a distinct commit?

In any case, make sure to create an issue on hardened `6.8.9` not booting and link to this issue.

vunnyso:
> On Testing 2. - Does the `6.8.9` hardened kernel boot without `secureboot`? Or `6.8.9` hardened kernel just does not boot at all?
> 
> The structural change looks good to me. I think merging should depend on clarifications to above questions. When we have the hardening profile, we should maintain it even if it's not enabled by default. Otherwise users may get wrong impression on ghaf hardening API. The debugging and maintenance could be scoped outside this PR but my personal preference would be to include the maintenance with this. Maybe as a distinct commit?
> 
> In any case, make sure to create an issue on hardened `6.8.9` not booting and link to this issue.

On Testing 2. Yes `6.8.9` hardened kernel boot is without `secureboot` option enabled.
System will stuck during EFI stub as in pic below, I think during initrd stage. 
image

@brianmcgillion  and @mbssrc any thoughts on maintenance?

UnikieHR:
image

vunnyso:
Closing as host-hardening profile integrated in https://github.com/tiiuae/ghaf/pull/758

PR #656, opened 2024-06-11
Host-to-guest storage module

mbssrc:
## Description of changes

- System-wide service to configure shares for guest vms using virtiofs
- Persistent fingerprint storage configured for gui-vm
- Preparation for other services using this feature

## Checklist for things done

### DIFF
--- a/modules/common/services/storage.nix
+++ b/modules/common/services/storage.nix
@@ -1,0 +1,114 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{lib, ...}: let
+  inherit (lib) mkEnableOption mkOption types literalExpression;
+
+  storageSubmodule = types.submodule {
+    options = {
+      tag = mkOption {
+        type = types.nullOr types.str;
+        description = ''
+          Storage tag used as label for the microvm share.
+        '';
+      };
+      host-path = mkOption {
+        type = types.str;
+        description = ''
+          Storage directory in the host. If it does not exist, it will be generated
+          with the default owner 'microvm' and group 'kvm'.
+        '';
+      };
+      tmp-path = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = ''
+          (Optional) Temporary directory in the VM to mount the host directory. If this is set,
+          the folder from host-path will be recursively copied from 'tmp-path' to 'vm-path' and owner, group,
+          and permissions applied.
+          If this parameter is null, the host-path is mounted to vm-path directly.
+        '';
+      };
+      vm-path = mkOption {
+        type = types.str;
+        description = ''
+          Storage directory in the guest where the share is mounted to.
+        '';
+      };
+      target-vm = mkOption {
+        type = types.str;
+        description = ''
+          Name of the VM the share is mounted into. The name must match the microvm
+          name in 'config.microvm.vms': e.g., "gui-vm" (note the '-').
+        '';
+      };
+      target-owner = mkOption {
+        type = types.nullOr types.str;
+        description = ''
+          Ownership to be applied after copying from 'tmp-path' or passing the files to the vm-path.
+          Note that the owner, group, and permissions are applied to the actual host files. This is
+          reset on next boot by the host storage service to allow microvm to access the files.
+        '';
+      };
+      target-group = mkOption {
+        type = types.nullOr types.str;
+        description = ''
+          Group to be applied after copying from 'tmp-path' or passing the files to the vm-path.
+          Note that the owner, group, and permissions are applied to the actual host files. This is
+          reset on next boot by the host storage service to allow microvm to access the files.
+        '';
+      };
+      target-permissions = mkOption {
+        type = types.nullOr types.str;
+        description = ''
+          Permissions to be applied after copying from 'tmp-path' or passing the files to the vm-path.
+          Format is an integer string (e.g., "755") as input for chmod.
+          Note that the owner, group, and permissions are applied to the actual host files. This is
+          reset on next boot by the host storage service to allow microvm to access the files.
+        '';
+      };
+      target-service = mkOption {
+        type = types.str;
+        default = "default.target";
+        description = ''
+          Systemd unit that requires the share. If set, the share is processed in the VM before the service is started.
+          Defaults to 'default.target'.
+        '';
+      };
+    };
+  };
+in {
+  options.ghaf.services.storage = {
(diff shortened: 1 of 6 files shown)


modules/microvm/virtualization/microvm/host/storage.nix
```diff
@@ -0,0 +1,115 @@
...
+    forEach (filterSharesByName vm) (vmShare: {
+      environment.systemPackages = optionals (vmShare.tmp-path != null) [pkgs.umount];
+      systemd.services."storage-prep-${vmShare.tag}" = let
+        # Service to set owner, permissions, and (optionally) copy files
+        prepStorage = pkgs.writeShellScriptBin "storage_prep" ''
+          set -xeuo pipefail
+          if [ ! -d "${vmShare.vm-path}" ]; then
+            mkdir -p ${vmShare.vm-path}
+          fi
+          ${optionalString (vmShare.tmp-path != null) "cp -r ${vmShare.tmp-path}/* ${vmShare.vm-path}"}
```
Mic92, line 48: The way I understand the semantic, I would expect the previous content to be removed.
Maybe using `rsync` with `--delete` would be than a better choice?

modules/common/services/storage.nix
```diff
@@ -0,0 +1,114 @@
...
+          Storage directory in the host. If it does not exist, it will be generated
+          with the default owner 'microvm' and group 'kvm'.
+        '';
+      };
+      tmp-path = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = ''
+          (Optional) Temporary directory in the VM to mount the host directory. If this is set,
+          the folder from host-path will be recursively copied from 'tmp-path' to 'vm-path' and owner, group,
```
Mic92, line 26:
```suggestion
          the folder from host-path will be recursively copied from 'tmp-mountpoint' to 'vm-path' and owner, group,
```

modules/microvm/virtualization/microvm/host/storage.nix
Mic92: Rsync, also would have an option to directly apply the right owner/group without an additional chown.

modules/common/services/storage.nix
```diff
@@ -0,0 +1,114 @@
...
+          and permissions applied.
+          If this parameter is null, the host-path is mounted to vm-path directly.
+        '';
+      };
+      vm-path = mkOption {
+        type = types.str;
+        description = ''
+          Storage directory in the guest where the share is mounted to.
```
Mic92, line 34:
```suggestion
          Storage directory in the guest where the share is mounted or where the data is copied to if `tmp-mountpoint` is set.
```

modules/common/services/storage.nix
```diff
@@ -0,0 +1,114 @@
...
+        '';
+      };
+      target-vm = mkOption {
+        type = types.str;
+        description = ''
+          Name of the VM the share is mounted into. The name must match the microvm
+          name in 'config.microvm.vms': e.g., "gui-vm" (note the '-').
```
Mic92, line 41: You can use `assertions = [{}];` nixos option and `config.microvm.vms ? ${target-vm}` condition here to test if the microvm actually exists to check for misconfiguration.

modules/microvm/virtualization/microvm/host/storage.nix
```diff
@@ -0,0 +1,115 @@
...
+          chown -R ${vmShare.target-owner}:${vmShare.target-group} ${vmShare.vm-path}
+          chmod -R ${vmShare.target-permissions} ${vmShare.vm-path}
+          ${optionalString (vmShare.tmp-path != null) "${pkgs.umount}/bin/umount ${vmShare.tmp-path}"}
```
Mic92, line 51: Is this not something that would need to run inside the VM? As far as I understand the tmp-path is not mounted on the host but it's virtiofsd exporting host path as a filesystem to the VM.

modules/microvm/virtualization/microvm/host/storage.nix
```diff
@@ -0,0 +1,115 @@
...
+          path = [prepStorage];
+          after = ["local-fs.target"];
+          requiredBy = ["${vmShare.target-service}"];
+          serviceConfig = {
+            Type = "oneshot";
+            RemainAfterExit = true;
+            Restart = "no";
+            StandardOutput = "journal";
+            StandardError = "journal";
+            ExecStart = "${prepStorage}/bin/storage_prep";
```
Mic92, line 66: I believe this will be re-runs whenever a new NixOS is deployed. Might be a bit disruptive if a VM gets part of the state wiped without when it's already running.
But I also don't know if life-updates is a requirement here.

modules/common/services/storage.nix
mbssrc: Yes will add assertions later if this is going to be used

modules/microvm/virtualization/microvm/host/storage.nix
mbssrc: The idea with this was that the host data remains present. This 'temporary' mountpoint option is intended for the use case where one wants to copy data from the host on VM boot, and apply specific permissions without changing the host permissions or rely on some UID mapping. Deletion could be added as an option as well.

mbssrc: I was also looking into using the impermanence module instead of this script. At the moment, I am waiting for the storage vm implementation to see how to finally implement this

modules/microvm/virtualization/microvm/host/storage.nix
mbssrc: Yes the service configs are passed to the VMs.

modules/microvm/virtualization/microvm/host/storage.nix
mbssrc: Good point - my assumption was that the VM service is restarted, but as we see with netvm now this may not necessary be the case. Adding `restartIfChanged = false;` should do the trick?

Mic92: Yes. restartIfChanged = false; would solve this.

modules/microvm/virtualization/microvm/host/storage.nix
Mic92: ok. I now wrapped my hand around it.

modules/microvm/virtualization/microvm/host/storage.nix
Mic92: As far as I know, impermanence doesn't provide copy semantics.

UnikieHR:
image

PR #658, opened 2024-06-11
lenovo-x1: compress image using zstd

humaidtii:
This avoids having images on the builders and cache take too much storage.

## Description of changes

The goal of this change is to compress all build artefacts to a reasonable size, in a reasonable amount of time. This is so that:

- Disko image files takes less space in the binary cache and builders.
- Downloading disko images files would be much faster from the binary cache, reducing bandwidth and time required to build a system.
- Building the ISO image would be much faster.

These are results as tested on my Lenovo P1 Gen3 laptop. This table is similar to Fedora's [comparison of compression levels and algorithms](https://fedoraproject.org/wiki/Changes/Switch_RPMs_to_zstd_compression#Comparison_of_compression_algorithms_and_levels), so we can compare easily.

| Target | Compression | Level | Size | Comp. time | Decomp. time | Comments |
| --- | --- | --- | --- | --- | --- | --- |
| `lenovo-x1-carbon-gen11-debug` (disko image) | None | - | 16G | - | - | Default before this patch |
| `lenovo-x1-carbon-gen11-debug` (disko image) | zstd | 3 | 3.8G |  36s | 12s | This patch |
| `lenovo-x1-carbon-gen11-debug` (disko image) | zstd | 8 | 3.32G |  2min | 15s |  |
| `lenovo-x1-carbon-gen11-debug-installer` (iso) | disko: none <br> squashfs: xz <br> iso: none | [See default](https://github.com/NixOS/nixpkgs/blob/nixos-24.05/nixos/modules/installer/cd-dvd/iso-image.nix#L509) | 6.3G | 25min | - | Default before this patch |
| `lenovo-x1-carbon-gen11-debug-installer` (iso) | disko: none <br> squashfs: zstd <br> iso: none | 3 | 7.6G | 1min | - | |
| `lenovo-x1-carbon-gen11-debug-installer` (iso) | disko: zstd <br> squashfs: zstd <br> iso: none | 3 | 7.5G | 50s | - | This patch |

> [!IMPORTANT]  
> With the installer target, compression time is the build time for the last item in dependency chart (iso build), not including disko image build time/compression. The decompression time for installer is only for the iso, as squashfs is decompressed during boot.

Using zstandard as a compression algorithm makes sense for our use case, especially with the speed improvement. This improves development speed, and we can still use this compression for production builds.
Distributions such as Fedora, Ubuntu, and Arch have switched to using zstd for compressing packages.

## Checklist for things done

## Testing

- Build disko image
  - Flash using: `zstdcat result/disk1.raw.zst | sudo dd of=/dev/<DEVICE> bs=32M status=progress`
- Create an installer, flash on USB. Test with `sudo ghaf-installer.sh`

For disko image, expect the build to take slightly longer, and to a less extent also decompression. But image should take significantly less disk space.

For iso installer image, expect the build to be significantly faster, but the result would be larger by about 1.3GB.

### DIFF
--- a/modules/disko/lenovo-x1-disko-basic.nix
+++ b/modules/disko/lenovo-x1-disko-basic.nix
@@ -3,7 +3,7 @@
 # Example to create a bios compatible gpt partition
 # To use this example, you will need to specify a device i.e.
 #   { disko.devices.disk1.device = "/dev/sda"; }
-{
+{pkgs, ...}: {
   disko.devices = {
     disk.disk1 = {
       type = "disk";
@@ -59,5 +59,11 @@
       };
     };
   };
-  disko.memSize = 2048;
+  disko = {
+    memSize = 2048;
+    extraPostVM = ''
+      ${pkgs.zstd}/bin/zstd --compress $out/*raw
+      rm $out/*raw
+    '';
+  };
 }
--- a/packages/installer/ghaf-installer.sh
+++ b/packages/installer/ghaf-installer.sh
@@ -28,7 +28,7 @@ read -r -p "Device name [e.g. /dev/nvme0n1]: " DEVICE_NAME
 read -r -p 'WARNING: Next command will destroy all previous data from your device, press Enter to proceed. '
 
 echo "Installing..."
-dd if=@imagePath@ of="${DEVICE_NAME}" bs=32M status=progress
+zstdcat @imagePath@ | dd of="${DEVICE_NAME}" bs=32M status=progress
 
 echo ""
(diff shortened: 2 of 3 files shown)


leivos-unikie (approved):
Tested disko image and ghaf installed by installer on Lenovo-X1
- ci-test-automation passed, performance ok
- All apps launch

Compressed disk image took only 3.5G

Installer iso file is about 1G larger than from mainline (6G) but built faster.

brianmcgillion (approved):

PR #665, opened 2024-06-12
Use cloud-hypervisor for net-vm on x86

slakkala:
## Description of changes

Replace QEMU with cloud-hypervisor for net-vm guest on x86 hardware.

cloud-hypervisor makes a differently connected virtual device, so the permanent name of the interface changes.

Move smcroute startup to NetworkManager dispatcher script to allow the guest to reach multi-user.target before host timeout reboots the VM.

## Checklist for things done

## Testing

Verify that network connectivity works after configuring it.

### DIFF
--- a/modules/microvm/virtualization/microvm/netvm.nix
+++ b/modules/microvm/virtualization/microvm/netvm.nix
@@ -62,13 +62,19 @@
 
         microvm = {
           optimize.enable = true;
-          hypervisor = "qemu";
+          hypervisor = {
+            # Use the same machine type as the host
+            x86_64-linux = "cloud-hypervisor";
+            aarch64-linux = "qemu";
+          }
+          .${configHost.nixpkgs.hostPlatform.system};
           shares =
             [
               {
                 tag = "ro-store";
                 source = "/nix/store";
                 mountPoint = "/nix/.ro-store";
+                proto = "virtiofs";
               }
             ]
             ++ lib.optionals isGuiVmEnabled [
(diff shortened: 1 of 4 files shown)


modules/microvm/virtualization/microvm/netvm.nix
```diff
@@ -62,13 +62,19 @@
 
         microvm = {
           optimize.enable = true;
-          hypervisor = "qemu";
+          hypervisor = {
+            # Use the same machine type as the host
+            x86_64-linux = "cloud-hypervisor";
+            aarch64-linux = "qemu";
```
brianmcgillion, line 68: why having a different vmm per architecture?

modules/microvm/virtualization/microvm/netvm.nix
```diff
@@ -62,13 +62,19 @@
...
+          }
+          .${configHost.nixpkgs.hostPlatform.system};
           shares =
             [
               {
                 tag = "ro-store";
                 source = "/nix/store";
                 mountPoint = "/nix/.ro-store";
+                proto = "virtiofs";
```
brianmcgillion, line 77: can you check the extra time delay in booting and size that is incurred by using virtiofs?

modules/microvm/virtualization/microvm/netvm.nix
slakkala: The cloud-hypervisor PCI support on AArch64 is limited in functionality, and does not work with our wifi adaptor. Further, the PCI passing on AArch64 would only work in UEFI mode, which microvm does not support; and would probably be slower.

modules/microvm/virtualization/microvm/netvm.nix
slakkala: In my measurements the delay is negligible, less than 1s difference from boot to VM up.

The storage usage grows by 36 kB on AArch64.

UnikieHR:
image

PR #678, opened 2024-06-28
Refactor the laptop out

brianmcgillion:
Refactor the x1 target to extract out the common functionality for laptops and hardware definitions.

define a series of profiles to better define the use cases for each x86 target
prepare to be able to support further HW

## Checklist for things done

## Testing

Needs the usual round of tests on the Lenovo X1 target.
no new functionality jsut refactoring effort


targets/lenovo-x1/flake-module.nix
```diff
@@ -3,15 +3,71 @@
...
   lib,
   self,
   ...
 }: let
-  inherit (inputs) microvm lanzaboote disko;
-  name = "lenovo-x1-carbon";
   system = "x86_64-linux";
-  targets = import ./everything.nix {inherit self lib microvm lanzaboote disko name system;};
+
+  mvp-definition = machineType: variant: extraModules: let
```
mbssrc, line 12: This file seems generic enough now to be a `targets/laptop`? And `mvp-definition` could be `laptop-definition` as mvp settings are now a profile.

targets/lenovo-x1/flake-module.nix
```diff
@@ -3,15 +3,71 @@
...
+    };
+  in {
+    inherit hostConfiguration;
+    name = "${machineType}-${variant}";
+    package = hostConfiguration.config.system.build.diskoImages;
+  };
+
+  targets = [
+    (mvp-definition "lenovo-x1-carbon-gen10" "debug" [
+      self.nixosModules.disko-basic-partition-v1
```
mbssrc, line 55: Do we expect the disk partitioning to be configured? I wouldn't expect much differences for a particular disk in a downstream project, but rather a stronger dependency of the disk layout to the system setup.

brianmcgillion: Yes, that is coming in the storage VM work.

targets/lenovo-x1/flake-module.nix
brianmcgillion: Glad you agree 😉

targets/lenovo-x1/flake-module.nix
```diff
@@ -3,15 +3,71 @@
...
+      {ghaf.hardware.definition.configFile = "/lenovo-x1/definitions/x1-gen10.nix";}
```
mbssrc, line 56: If we keep defining targets depending on the hardware, we could require that the hardware definition for each `machineType` lives in `target`, and automatically import the right configuration according to the value of `machineType`?  So we have a clean separation between hw definition and the Ghaf hw usage. All that needs to be done downstream is to generate a hw definition file and put it in target, naming it accordingly. But perhaps how we handle the hardwares may be a topic for another time.

modules/hardware/laptop.nix
```diff
@@ -6,17 +6,18 @@
...
+    ./definition.nix
   ];
 
-  options.ghaf.hardware.generation = mkOption {
-    description = "Generation of the hardware configuration";
-    type = types.nullOr types.str;
-    default = null;
+  options.ghaf.hardware.definition.configFile = mkOption {
+    description = "Path to the hardware configuration file.";
+    type = types.str;
```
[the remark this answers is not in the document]
brianmcgillion: null, no because it is required to be set to a valid file. As for being a path rather than a string. the handling of paths is somewhat different and will cause the file referenced as a path to be copied to the nix store. secondly the path is being created for th hwDefinition as it is relative to that place and not relative to where it is being called e.g. in the targets/laptop directory. so it behaves better as a string.

targets/laptop/laptop-configuration-builder.nix
```diff
@@ -0,0 +1,44 @@
...
+{
+  lib,
+  self,
+  ...
+}: let
+  system = "x86_64-linux";
+
+  #TODO move this to a standalone function
+  #should it live in the library or just as a function file
+  laptop-configuration = machineType: variant: extraModules: let
```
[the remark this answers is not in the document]
brianmcgillion: Good call. More expressive. Amended patch coming

modules/hardware/laptop.nix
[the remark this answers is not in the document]
brianmcgillion: it is the current way it is built. that is another refactoring to happen later, when we decide on config file vs maybe a module. yes the entire source is there, but path will add it again as a standalone path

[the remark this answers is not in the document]
brianmcgillion: https://github.com/tiiuae/ghaf/pull/678/files#diff-f62a48ac04e1d3df38117cf98c13bcabfd4714ec99e6609701e1bd7bd6ec6a10R10

barnabakos (approved):
11 commits up to: 188b88621d79917096d35cf77a59c325fa86fbae
PR is on top of latest tip of Ghaf main: a7167f494be88a3eb54209525d45e6e7b31ce787

platform: **Lenovo X1 Carbon**
- starting apps: ok
- ad-hoc tesing: no new findings

mbssrc (approved):

PR #686, opened 2024-07-11
Add support for USB hot plugging

nesteroff:
## Description of changes

This adds a service that runs on the host and listens for device add and remove events using libudev. When a new USB device is attached, it is automatically assigned to the designated virtual machine using the official qemu.qmp library."

## Checklist for things done

## Testing

Input Devices:
- Connect a USB keyboard or mouse to the Lenovo X1 and verify that it is working in the GUI VM.

Audio Devices:
- Connect a USB headset to the Lenovo X1 and ssh into the audio-vm.
- Run `aplay -l` and find the headset in the list. It should be listed as either card 0 or card 1.
- To test the sound, execute `speaker-test -D hw:1,0 -c 2 -t sine`. If the headset appears as card 0, change hw:1,0 to hw:0,0.

If a bug is found or if something is not working as expected, please collect the vhotplug service logs from journalctl on the host.

### DIFF
--- a/modules/microvm/virtualization/microvm/audiovm.nix
+++ b/modules/microvm/virtualization/microvm/audiovm.nix
@@ -59,7 +59,7 @@ let
               pkgs.pulseaudio
               pkgs.pamixer
               pkgs.pipewire
-            ];
+            ] ++ lib.optional config.ghaf.development.debug.tools.enable pkgs.alsa-utils;
           };
 
           time.timeZone = config.time.timeZone;
@@ -73,7 +73,8 @@ let
           services.openssh = config.ghaf.security.sshKeys.sshAuthorizedKeysCommand;
 
           microvm = {
-            optimize.enable = true;
+            # Optimize is disabled because when it is enabled, qemu is built without libusb
+            optimize.enable = false;
             vcpu = 1;
             mem = 256;
             hypervisor = "qemu";
@@ -102,6 +103,10 @@ let
                   aarch64-linux = "virt";
                 }
                 .${configHost.nixpkgs.hostPlatform.system};
+              extraArgs = [
+                "-device"
+                "qemu-xhci"
+              ];
             };
           };
 
--- a/packages/qemuqmp/default.nix
+++ b/packages/qemuqmp/default.nix
@@ -1,0 +1,26 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  python3Packages,
+  fetchPypi,
+  lib,
+}:
+python3Packages.buildPythonPackage rec {
+  pname = "qemu.qmp";
+  version = "0.0.3";
+
+  src = fetchPypi {
+    inherit pname version;
+    sha256 = "sha256-y8iPvMEV7pQ9hER9FyxkLaEgIgRRQWwvYhrPM98eEBA=";
+  };
+
+  pyproject = true;
+
+  nativeBuildInputs = [ python3Packages.setuptools-scm ];
+
+  meta = {
+    homepage = "https://www.qemu.org/";
+    description = "QEMU Monitor Protocol library";
+    license = lib.licenses.lgpl2Plus;
+  };
+}
--- a/modules/hardware/common/default.nix
+++ b/modules/hardware/common/default.nix
@@ -4,6 +4,7 @@
   imports = [
     ./usb/internal.nix
     ./usb/external.nix
+    ./usb/vhotplug.nix
     ./devices.nix
     ./kernel.nix
     ./qemu.nix
--- a/modules/hardware/common/qemu.nix
+++ b/modules/hardware/common/qemu.nix
@@ -35,7 +35,11 @@ in
           "acad"
         ]
         ++ optionals (hasAttr "yubikey" config.ghaf.hardware.usb.external.qemuExtraArgs) config.ghaf.hardware.usb.external.qemuExtraArgs.yubikey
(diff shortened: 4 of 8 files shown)


packages/qemuqmp/default.nix
```diff
@@ -0,0 +1,29 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  python3Packages,
+  fetchPypi,
+  lib,
+}:
+with python3Packages;
```
brianmcgillion, line 8: no need for the with as you can be explicit in the line below. similar to the vhotplug/default.nix below

packages/qemuqmp/default.nix
```diff
@@ -0,0 +1,29 @@
...
+    version = "0.0.3";
+
+    src = fetchPypi {
+      inherit pname version;
+      sha256 = "sha256-y8iPvMEV7pQ9hER9FyxkLaEgIgRRQWwvYhrPM98eEBA=";
+    };
+
+    pyproject = true;
+
+    nativeBuildInputs = with python3Packages; [
```
brianmcgillion, line 20: avoid with and just use it explicitly for the one package below?

brianmcgillion (changes requested):

packages/qemuqmp/default.nix
nesteroff: Right. Fixed.

packages/qemuqmp/default.nix
nesteroff: Sure, fixed.

brianmcgillion (approved):

mbssrc:
This is great. Tested this briefly and worked fine. Let me know what your test results with the other usb devices are!
Generally (maybe not as part of this PR) we could use a static configuration file as general policy based on device type and VM, and later add-on user defined passthrough for switcheroo between VMs.

nesteroff:
> This is great. Tested this briefly and worked fine. Let me know what your test results with the other usb devices are! If you could add `ID_INPUT_TOUCHSCREEN` etc. (https://github.com/tiiuae/ghaf/blob/main/packages/hardware-scan/hardware-scan.sh#L242) as well, it would also support touchscreens.

Thanks. I added touchscreens to the list as well.

> Generally (maybe not as part of this PR) we could use a static configuration file as general policy based on device type and VM, and later add-on user defined passthrough for switcheroo between VMs.

Sure. I tried to keep this app simple but in future versions we will need to improve the configuration to implement more complicated rules, add exclusions, support for the cloud hypervisor and so on.

I’ve tested this with input devices, audio headsets, removable disks, and ethernet adapters. It seems to work fine but I left network devices disabled by default because attaching them to the net-vm breaks remote nixos-rebuild with target-host. Also hot-plugging of disk devices is disabled until we have a storage vm or something like that. Do you know if there are any other device types that we need to support?

mbssrc:
Perhaps the yubikey and gps devices, so we can get rid of the static external device definition?

nesteroff:
> Perhaps the yubikey and gps devices, so we can get rid of the static external device definition?

Sure, I'll look into it.

nesteroff:
Recent updates:
- Introduced a configuration file for hot-plugging settings.
- Updated the passthrough algorithm to use USB interface information (class, subclass, protocol). This allows us to passthrough USB devices to VMs without requiring drivers on the host. It should also work when the host is configured not to authorize USB devices by default.
- When defining a rule for a USB interface class, it is possible to add certain devices to the ignore list so that they remain connected to the host.
- Unfortunately, some devices do not provide correct USB interface information. For example, our GPS receiver returns 0xFF (Vendor Defined). In such cases we can passthrough the device by VID/PID using either `ghaf.hardware.usb.external` or `ghaf.hardware.usb.vhotplug`.

modules/profiles/laptop-x86.nix
```diff
@@ -121,6 +121,68 @@ in {
       logging.client.endpoint = "http://${listenerAddress}:${listenerPort}/loki/api/v1/push";
       logging.listener.address = "admin-vm-debug";
       logging.listener.port = 9999;
+
+      # USB hot plugging configuration
+      hardware.usb.vhotplug = {
+        enable = true;
+        rules = [
```
mbssrc, line 128: Perhaps we can move the rule configuration as this file should be a high level laptop config. Perhaps we can put this into the hardware/usb/vhotplug as default configuration?

modules/microvm/virtualization/microvm/guivm.nix
```diff
@@ -118,10 +118,15 @@
           writableStoreOverlay = lib.mkIf config.ghaf.development.debug.tools.enable "/nix/.rw-store";
 
           qemu = {
-            extraArgs = [
-              "-device"
-              "vhost-vsock-pci,guest-cid=${toString cfg.vsockCID}"
-            ];
+            extraArgs =
```
mbssrc, line 121: You could use the qemu.nix for the qemu `extraArgs`, conditionally on the vhotplug being enabled?

modules/microvm/virtualization/microvm/audiovm.nix
```diff
@@ -57,6 +57,7 @@
             pkgs.pulseaudio
             pkgs.pamixer
             pkgs.pipewire
+            pkgs.alsa-utils
```
mbssrc, line 60: Are the `alsa-utils` required?

nesteroff: It's very convenient to use aplay and speaker-test from alsa-utils to test USB headsets until we have the audio control application merged. I've updated it to enable this conditionally when debug is on.

mbssrc (approved):

nesteroff:
Updated to make hot-plugged devices survive suspend and resume.

brianmcgillion (approved):

PR #692, opened 2024-07-18
Hardware detection scanner

mbssrc:
## Description of changes

Adding a (temporary) hardware detection script to generate hardware definition files for different laptop hardware.
The script can generate a Ghaf laptop target compatible hardware definition file (details below).
 
Main changes:
- Add `hardware-scan` shell script to generate a hardware definition file
- Add a laptop hardware scan target, minimal image without having to generate an image to flash
- Change the udev rule generation for input devices. This is necessary to allow more input devices and rule auto-generation 
- Add a hardware definition for the Dell Latitude 7230 laptop

## Checklist for things done

### Generate the hardware-scan package
```
nix build .#packages.x86_64-linux.hardware-scan
./result/bin/hardware-scan
```
or run directly with
`nix run github:tiiuae/ghaf/<commit>#hardware-scan`
### Generate and run hardware-scan on a laptop
1. Add your ssh key
2. Build target 
`nix build .#laptop-hw-scan`
3. Flash 
`dd if=/result/iso/ghaf-... of=/dev/sdX bs=32M status=progress; sync`
4. Boot, then type `hardware-scan`
5. The script should ask you if you want to select various devices for passthrough. If selected, the respective entries are generated. Note that there may be errors caused by IOMMU group dependencies.
6. Move the generated files to your development machine (e.g., `scp -r nixos@<your-ip>:~/* local-dir/`)
7. The script should generate a hardware definition profile, as well as a `hwinfo/` folder with dumps from udevadm, lspci, etc.
8. Use the generated `.nix` as basis for your hardware definition

Few words of caution:
- Don't fully rely on the generated output file, walk through all options that are being set 
- Currently, the host kernel module blacklist is commented out by default. Select the suggested modules carefully, and/or try without blacklisting
- Uncomment required `misc` input devices, e.g., for `acpi` 

## Testing
- Test builds of all lenovo-x1-carbon targets 
- Verify that input devices are working as before (although errors should result in not being able see GUI)
- Perhaps run `udevadm verify`

(Optional) Feel free to run the `hardware-scan` on various targets and report any findings
(Optional) Report any SKU identifiers to be added to the x1 definitions

### DIFF
--- a/modules/hardware/definitions/dell-latitude/dell-latitude-7230.nix
+++ b/modules/hardware/definitions/dell-latitude/dell-latitude-7230.nix
@@ -1,0 +1,155 @@
+# Copyright 2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  # System name
+  name = "Dell Latitude 7230 Rugged";
+
+  # List of system SKUs covered by this configuration
+  skus = [
+    "0BB7 Latitude 7230 Rugged Extreme Tablet"
+  ];
+
+  # Host configuration
+  host = {
+    kernelConfig.kernelParams = [
+      "intel_iommu=on,sm_on"
+      "iommu=pt"
+      "acpi_backlight=vendor"
+      "acpi_osi=linux"
+      "module_blacklist=i915,iwlwifi,snd_hda_intel,snd_sof_pci_intel_tgl"
+    ];
+  };
+
+  # Input devices
+  input = {
+    keyboard = {
+      name = ["AT Translated Set 2 keyboard"];
+      evdev = ["/dev/keyboard0"];
+    };
+
+    mouse = {
+      name = ["PS/2 Generic Mouse" "SYNAPTICS Synaptics HIDUSB TouchPad V1.05 Mouse"];
+      evdev = ["/dev/mouse0" "/dev/mouse1"];
+    };
+
+    touchpad = {
+      name = ["SYNAPTICS Synaptics HIDUSB TouchPad V1.05 Touchpad" "EETI8082:00 0EEF:C004" "EETI8082:00 0EEF:C004 Stylus"];
+      evdev = ["/dev/touchpad0" "/dev/touchpad1" "/dev/touchpad2"];
+    };
+
+    misc = {
+      name = [
+        # "Intel HID events" "Dell WMI hotkeys" "Video Bus" "HDA Intel PCH Headphone Mic" "HDA Intel PCH HDMI/DP,pcm=3" "HDA Intel PCH HDMI/DP,pcm=7" "HDA Intel PCH HDMI/DP,pcm=8" "HDA Intel PCH HDMI/DP,pcm=9" "Intel HID 5 button array" "Lid Switch" "Power Button" "Sleep Button"
+      ];
+      evdev = [
+        # /dev/input/by-path/platform-INTC1070:00-event /dev/input/by-path/platform-PNP0C14:02-event
+      ];
+    };
+  };
+
+  # Main disk device
+  disks = {
+    disk1.device = "/dev/nvme0n1";
+  };
+
+  # Network devices for passthrough to netvm
+  network = {
+    pciDevices = [
+      {
+        # Network controller: Intel Corporation Alder Lake-P PCH CNVi WiFi (rev 01)
+        name = "wlp0s5f0";
+        path = "0000:00:14.3";
+        vendorId = "8086";
+        productId = "51f0";
+        # Detected kernel driver: iwlwifi
+        # Detected kernel modules: iwlwifi
+      }
+    ];
+    kernelConfig = {
+      stage1.kernelModules = [];
+      stage2.kernelModules = ["iwlwifi"];
+      kernelParams = [];
+    };
+  };
+
+  # GPU devices for passthrough to guivm
+  gpu = {
+    pciDevices = [
+      {
+        # VGA compatible controller: Intel Corporation Alder Lake-UP4 GT2 [Iris Xe Graphics] (rev 0c)
+        name = "gpu0";
+        path = "0000:00:02.0";
+        vendorId = "8086";
+        productId = "46aa";
+        # Detected kernel driver: i915
+        # Detected kernel modules: i915
+      }
+    ];
+    kernelConfig = {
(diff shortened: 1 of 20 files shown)


modules/hardware/definitions/dell-latitude/dell-latitude-7230.nix
```diff
@@ -0,0 +1,155 @@
...
+        path = "0000:00:1f.4";
+        vendorId = "8086";
+        productId = "51a3";
+        # Detected kernel driver: i801_smbus
+        # Detected kernel modules: i2c_i801
+      }
+    ];
+    kernelConfig = {
+      stage1.kernelModules = [];
+      stage2.kernelModules = ["i2c_i801" "snd_hda_intel" "snd_sof_pci_intel_tgl" "spi_intel_pci"];
```
brianmcgillion, line 137: `lsmod` shows that these drivers have many dependencies so this list is incomplete and likely will not work without all the dependent drivers. either document that this is just to show the primary drivers, and whoever is making the minimal module list will have to include the others or you would have to traverse the `lsmod` or `modinfo` to extract them all

packages/hardware-scan/default.nix
```diff
@@ -0,0 +1,50 @@
...
+  util-linux,
+  pciutils,
+  usbutils,
+  dmidecode,
+  alejandra,
+  makeWrapper,
+  lib,
+}: let
+  inherit (lib) makeBinPath;
+  pname = "hardware-scan";
```
brianmcgillion, line 17: would you not just use `writeshellApplication` and read in the file?

`text = builtins.readFile ./hardware-scan.sh;`

brianmcgillion:
the `nix fmt `problem is from `modules/hardware/laptop.nix`. also because there is no file extension on `hardware-scan` there is a new WARN emitted:

```
❯ nix fmt
WARN format: no formatter for path: packages/hardware-scan/hardware-scan
```

packages/hardware-scan/default.nix
mbssrc: changed, including file extension so shellcheck picks it up

modules/hardware/definitions/dell-latitude/dell-latitude-7230.nix
mbssrc: There are a number of warnings already (e.g., in --help), where do you want to have this? As comment above?

nesteroff:
That's a great script. It should be a lot easier to add new targets now.

I just wanted to mention that in #686, I added a service for hot-plugging that can also dynamically enumerate input devices, find those that are not defined statically and attach them to the GUIVM. Initially, I needed it so that hot-plugged devices survive a system reboot but it turns out that it can also work for non-USB devices. I'm currently testing it but it looks like everything continues to work on my X1 even when `guivmVirtioInputHostEvdevModule` is commented out.

mbssrc:
> That's a great script. It should be a lot easier to add new targets now.
> 
> I just wanted to mention that in #686, I added a service for hot-plugging that can also dynamically enumerate input devices, find those that are not defined statically and attach them to the GUIVM. Initially, I needed it so that hot-plugged devices survive a system reboot but it turns out that it can also work for non-USB devices. I'm currently testing it but it looks like everything continues to work on my X1 even when `guivmVirtioInputHostEvdevModule` is commented out.

Yes #686 is a good solution - I was actually just testing it now, may comment on it later. Getting rid of the static input device enumeration completely would simplify things a lot. My guess is that we have to keep the evdev module for some of the extra hardware dependencies though...

packages/hardware-scan/hardware-scan.sh
```diff
@@ -0,0 +1,631 @@
+#! /usr/bin/env bash
+# shellcheck shell=bash
```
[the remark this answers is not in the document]
mbssrc: Left this in to make shellcheck happy, could also disable SC2148 but decided to leave this in case someone copies it directly.

milva-unikie:
**Tested on Lenovo-X1**
`nix build github:mbssrc/ghaf/hw-tooling#lenovo-x1-carbon-gen11-debug` and
`nix build github:mbssrc/ghaf/hw-tooling#lenovo-x1-carbon-gen11-debug-installer`

**Issues found:**

I think there is something wrong with camera. With this PR it seems like the camera is not working in any of the vm:s (chromium, element and trusted browser). Restarting chromium did not claim it either.

Some background for camera: There is a known problem on Lenovo-X1 that only one vm can have access to the camera at a time. The vm seems to be chosen randomly at startup between chromium, element and trusted browser. If you restart a vm from host, restarted vm should claim the camera for itself (working at least for the browsers).

**Working:**
- `udevadm verify` showed no fails
- installer works
- test-automation passes
- integrated microphone and speaker work
- 3,5 mm headset works
- all of the keyboard buttons that were working previously are still working (including F1, F2, F3, F5, F6 and PrtSc)
- touchpad, trackpoint and mouse buttons work

Also tried running hardware scan on X1, no problems found.

mbssrc:
> **Tested on Lenovo-X1** `nix build github:mbssrc/ghaf/hw-tooling#lenovo-x1-carbon-gen11-debug` and `nix build github:mbssrc/ghaf/hw-tooling#lenovo-x1-carbon-gen11-debug-installer`
> 
> **Issues found:**
> 
> I think there is something wrong with camera. With this PR it seems like the camera is not working in any of the vm:s (chromium, element and trusted browser). Restarting chromium did not claim it either.
> 
> Some background for camera: There is a known problem on Lenovo-X1 that only one vm can have access to the camera at a time. The vm seems to be chosen randomly at startup between chromium, element and trusted browser. If you restart a vm from host, restarted vm should claim the camera for itself (working at least for the browsers).
> 
> **Working:**
> 
>     * `udevadm verify` showed no fails
> 
>     * installer works
> 
>     * test-automation passes
> 
>     * integrated microphone and speaker work
> 
>     * 3,5 mm headset works
> 
>     * all of the keyboard buttons that were working previously are still working (including F1, F2, F3, F5, F6 and PrtSc)
> 
>     * touchpad, trackpoint and mouse buttons work
> 
> 
> Also tried running hardware scan on X1, no problems found.

Thanks - some part of the USB naming update got lost. Fixed now

milva-unikie:
Camera is working now on Lenovo-X1!

brianmcgillion (approved):

PR #702, opened 2024-07-29
graphics: add greetd login manager

humaidq-tii:
Adds a login manager and a graphical greeter in Ghaf, with option for autologin (by default for now).

This removes redundant services such as ghaf-session, and seatd (as replaced with an official NixOS service option).

## Description of changes

- Adds greetd login manager
  - This runs as a separate user 'greeter' which is in 'video' group.
  - Remove `ghaf-session` service, as sessions are managed by greetd.
- Use gtkgreet as a greeter, customised to match our style guide.
  - We may switch to something more customisable, such as ReGreet.
- Uses labwc compositor to run the greeter
  - Brightness controls, alternative keyboard layouts works on the login screen.
  - Not possible to launch Terminal from greeter user.
  - labwc supports multi-monitor well, as well as `wlr-layer-shell-unstable` for gtkgreet.
- Replace custom seatd service with `services.seatd` as it is now available in NixOS.
- Set wlroots environment variables using `environment.sessionVariables`.
  - This allows other compositor to work properly, but this PR doesn't add any new compositors.
- Fix renderer option for Orin (previously it seemed to fallback correctly even with typo).

I have initially tested greeter with cage. Cage doesn't support `wlr-layer-shell-unstable`, which causes gtkgreet to only show on one screen  Also Cage doesn't seem to set the resolution correctly on the Orin, and generally has crude support for multi-monitors. These issues are resolved when using labwc.

## Checklist for things done

## Testing

- By default, you will be logged in automatically.
- You can log out by clicking Log Out in app drawer, which should bring you to the login screen.
- Setting `ghaf.graphics.labwc.autologinUser = null;` would disable autologin, and show the login screen on boot.
- Login, logout multiple times. Should not cause system to go in bad state.
- In login screen, plug and unplug external monitors. Should not glitch or crash.
  - Login prompt is shown on monitor with cursor, moving cursor to other monitor would move the prompt.
  - Other monitors should only show clock on the centre.
- All applications should be launchable after re-logins.
- Boot time shouldn't be significantly affected.
- Fingerprint authentication should work with both lock screen and login screen. Login screen should not hog access to fingerprint reader after login.
  - With same previous caveats with gtklock.


vunnyso:
Overall PR looks good.

Few observations:
1) As mentioned `You can log out by clicking Log Out in app drawer, which should bring you to the login screen.`
After which it will display `labwc-session` nix store path in drop down which doesn't look pretty. 
image

2) Works fine with external monitor, however login prompt is always shown on X1 device.

3) Boot time details

| Time | Log | Comment |
| ---      | ---    | ------            |
| 13:09:04 | localhost kernel: Linux version 6.8.12      | First Log in `ghaf-host`  |
| 13:09:35 | gui-vm systemd[682]: Started waypipe    | waypipe started in `gui-vm`  |

Approx ~31 Sec vs ~25sec in mainline. 


4) Fingerprint authentication works in lock screen but if I logout and enter username after which need to manually click `Log in` button to unlock. 

image

5. It will be good if system can remember username since we don't have multi users as of now.

vunnyso:
> @vunnyso Thank you for the review. Totally agree with the UI, it isn't pretty. I'm thinking we can move to: https://github.com/rharish101/ReGreet
> 
> > Works fine with external monitor, however login prompt is always shown on X1 device.
> 
> Login prompt should move to external monitor when you move the cursor, it doesn't do that?

No on my setup. Login prompt always stays on X1 display irrespective of cursor position.  

---

> 
> For now, the most important thing is that we have a greetd setup, later it should not be an issue to change graphical greeters.
> 
> I think for a start we should at least merge this, then improve. What do you think?

If we have plan to improve we can go ahead with merge. 

Btw how about delay in boot time as I have mentioned in point 3?

vunnyso:
> > Btw how about delay in boot time as I have mentioned in point 3?
> 
> I did some measurements, with and without autologin. If you measure from after bootloader until the cursor first appears, the delay is about 2s on average.
> 
> For the gtkgreet, it seems to take a second or two more for it to load. You can see a similar delay to waybar.

Can we optimize load time later?

---
> 
> > No on my setup. Login prompt always stays on X1 display irrespective of cursor position.
> 
> The login prompt stays on X1 for gtklock, but for gtkgreet it does move, even with multiple login/logouts.

Alright.

> 
> I did a rebase, and fixed some build issue with autologin set to null.

vunnyso (approved):

PR #711, opened 2024-08-03
lenovo-x1: temporary fix build errors of release and release installer targets to build only

Nerox9:
## Description of changes
The call of the vm by name fails because has "debug" suffix. The call has been modified to use the correct option, following the Lenovo X1 targets. 

However it doesn't solve the functional issue and doesn't work on release targets because of connections are restricted. GIVC will fix it in the future.

## Checklist for things done

## Testing

Build the targets:
#packages.x86_64-linux.lenovo-x1-carbon-gen11-release
#packages.x86_64-linux.lenovo-x1-carbon-gen11-release-installer

You can also check with "#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug" target to be sure does it work.
Run the image and run one of the applications or scripts in the launcher; for example "shutdown".
Observe that the script or application works.

### DIFF
--- a/modules/common/services/desktop.nix
+++ b/modules/common/services/desktop.nix
@@ -41,7 +41,9 @@ in
       graphics = {
         launchers =
           let
-            hostEntry = filter (x: x.name == "ghaf-host-debug") config.ghaf.networking.hosts.entries;
+            hostEntry = filter (
+              x: x.name == "audio-vm" + lib.optionalString config.ghaf.profiles.debug.enable "-debug"
+            ) config.ghaf.networking.hosts.entries;
             hostAddress = head (map (x: x.ip) hostEntry);
             powerControl = pkgs.callPackage ../../../packages/powercontrol { };
             privateSshKeyPath = config.ghaf.security.sshKeys.sshKeyPath;
--- a/modules/microvm/virtualization/microvm/audiovm.nix
+++ b/modules/microvm/virtualization/microvm/audiovm.nix
@@ -121,7 +121,7 @@ let
           systemd.network.networks."10-ethint0".addresses =
             let
               getAudioVmEntry = builtins.filter (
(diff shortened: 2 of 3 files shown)


modules/common/services/desktop.nix
```diff
@@ -30,7 +30,7 @@ in {
       profiles.graphics.compositor = "labwc";
       graphics = {
         launchers = let
-          hostEntry = filter (x: x.name == "ghaf-host-debug") config.ghaf.networking.hosts.entries;
+          hostEntry = filter (x: x.name == "ghaf-host" + lib.optionalString config.ghaf.profiles.debug.enable "-debug") config.ghaf.networking.hosts.entries;
```
mbssrc, line 33: `ghaf-host` will not be reachable from guivm in release at all, any commands must be proxied over admin-vm which requires a 'private' network with the host.

modules/microvm/virtualization/microvm/audiovm.nix
```diff
@@ -112,7 +112,7 @@
         environment.etc = lib.mkIf isGuiVmEnabled {${config.ghaf.security.sshKeys.getAuthKeysFilePathInEtc} = sshKeysHelper.getAuthKeysSource;};
 
         systemd.network.networks."10-ethint0".addresses = let
```
mbssrc, line 114: Isn't this entire entry obsolete?

modules/common/services/desktop.nix
Nerox9: This PR aimed to solve the build error only and quickly. We can fix functionally in GIVC, however if it is needed I can make it with GIVC too.

mbssrc (approved):
Temporary fix, this does not solve any issues for release which may build but will still not work.

brianmcgillion (approved):
Verified that the system evaluates now with these fixes

PR #722, opened 2024-08-09
AudioVM and client configuration with separate VM audio chennels

josa41:
## Description of changes
AudioVM configuration which adds Audio IO channels for VMs (per confiugration).
Clean up of AppVM configuration to add client audioVM connections based on VM name

## Checklist for things done

- [X ] Summary of the proposed changes in the PR description
- [X ] More detailed description in the commit message(s)
- [X ] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [X ] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [X ] Author has run `nix flake check --accept-flake-config` and it passes

### DIFF
--- a/modules/common/services/audio.nix
+++ b/modules/common/services/audio.nix
@@ -43,7 +43,7 @@ in
                 # Enable TCP socket for VMs pulseaudio clients
                 "server.address" = [
                   {
-                    address = "tcp:4713";
+                    address = "tcp:${toString cfg.pulseaudioTcpPort}";
                     "client.access" = "unrestricted";
                   }
                 ];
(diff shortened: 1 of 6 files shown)


modules/common/services/audio.nix
```diff
@@ -77,6 +132,51 @@ in {
       script = ''${pkgs.pulseaudio}/bin/pa-info > /dev/null 2>&1'';
```
mbssrc, line 132: Can we use lingering (see below) instead of the pulseaudio-starter service?

```
systemd.tmpfiles.rules = [
  "f /var/lib/systemd/linger/${config.ghaf.users.accounts.user}"
];
```

modules/common/services/audio.nix
```diff
@@ -77,6 +132,51 @@ in {
...
     };
 
+    # TODO Automate and fix alsa IO ports
+    systemd.services."pipewire-link-starter" = let
```
mbssrc, line 136: I suppose that we want to aim at dynamic support, i.e., supporting several input/output devices such as bluez_input/output, starting with auto-switch and later selection.

UnikieHR:
image

PR #740, opened 2024-08-15
[X1] Turn off the display when system is suspended

vunnyso:
* Addresses SP-4844 - Display turns on in sleep mode when the lid is closed.

## Checklist for things done

## Testing

- Display will get turn off automatically when lid is closed.
- Display will get turn off automatically when we run `sudo systsemctl suspend` on `ghaf-host` 

### Known issues with Qemu
1. As persistent Storage is enabled in `gui-vm` due to which `gui-vm` fails to suspend as virtio filesystem doesn't support suspend, details [here](https://github.com/torvalds/linux/blob/v6.10/fs/fuse/virtio_fs.c#L1058-L1063).
```
gui-vm kernel: virtio-fs: suspend/resume not yet supported
gui-vm kernel: virtio-pci 0000:00:04.0: PM: pci_pm_suspend(): virtio_pci_freeze+0x0/0x50 [virtio_pci] returns -95
gui-vm kernel: virtio-pci 0000:00:04.0: PM: dpm_run_callback(): pci_pm_suspend+0x0/0x170 returns -95
gui-vm kernel: virtio-pci 0000:00:04.0: PM: failed to suspend async: error -95
```

2. If someway we can managed to solve problem 1, then `gui-vm` cannot be resume on keyboard press, way to resume is from qemu monitor socket.

3. If someway we can managed to solve problem 2, then qemu freezes after resume, more details [here](https://gitlab.com/qemu-project/qemu/-/issues/2520)
 
So considering above qemu issues, I think better to turn off display.

### Power improvements

|Check |With PR  | Without PR |
| -------------| ------------- | ------------- |
Battery Drainage | ~2 % per hour  | ~8 % per hour |

Approx 4 times improvement in battery life when lid is closed.

### Next Steps
Try to replicate with `givc` service if feasible.

### DIFF
--- a/modules/hardware/common/qemu.nix
+++ b/modules/hardware/common/qemu.nix
@@ -24,9 +24,6 @@ in
     ghaf.qemu.guivm = optionalAttrs (hasAttr "hardware" config.ghaf) {
       microvm.qemu.extraArgs =
         [
-          # Button
-          "-device"
-          "button"
           # Battery
           "-device"
           "battery"
--- a/modules/host/default.nix
+++ b/modules/host/default.nix
@@ -3,7 +3,7 @@
 #
 # Modules that should be only imported to host
 #
-{ lib, ... }:
+{ lib, pkgs, ... }:
 {
   networking.hostName = lib.mkDefault "ghaf-host";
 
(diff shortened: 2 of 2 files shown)


nesteroff:
I was thinking that maybe we don’t need to suspend the GUIVM at all. Instead, we can simply turn off the display before suspending and turn it back on after resuming using the `wlopm` command.
These commands can be added to a oneshot systemd service that runs before and after the `sleep.target` target on the host. We can use SSH to execute them in the GUIVM from the host, or better yet add them to the RPC service.
This way, we don’t need to take care of the QEMU and virtiofs suspend/resume issues. Also, we will be able to remove the `-device button` from [here](https://github.com/tiiuae/ghaf/blob/d66c087b2ae3d6b02732e68c01f6ad59dfdaabbb/modules/hardware/common/qemu.nix#L28).
Here’s an example of how to turn the display on and off from the host (`pkgs.wlopm` needs to be added to the GUIVM):
```
ssh -t ghaf@gui-vm-debug 'WAYLAND_DISPLAY=/run/user/1000/wayland-0 wlopm --off \*’
ssh -t ghaf@gui-vm-debug 'WAYLAND_DISPLAY=/run/user/1000/wayland-0 wlopm --on \*’
```
@vunnyso what do you think?

vunnyso:
> I was thinking that maybe we don’t need to suspend the GUIVM at all. Instead, we can simply turn off the display before suspending and turn it back on after resuming using the `wlopm` command. These commands can be added to a oneshot systemd service that runs before and after the `sleep.target` target on the host. We can use SSH to execute them in the GUIVM from the host, or better yet add them to the RPC service. This way, we don’t need to take care of the QEMU and virtiofs suspend/resume issues. Also, we will be able to remove the `-device button` from [here](https://github.com/tiiuae/ghaf/blob/d66c087b2ae3d6b02732e68c01f6ad59dfdaabbb/modules/hardware/common/qemu.nix#L28). Here’s an example of how to turn the display on and off from the host (`pkgs.wlopm` needs to be added to the GUIVM):
> 
> ```
> ssh -t ghaf@gui-vm-debug 'WAYLAND_DISPLAY=/run/user/1000/wayland-0 wlopm --off \*’
> ssh -t ghaf@gui-vm-debug 'WAYLAND_DISPLAY=/run/user/1000/wayland-0 wlopm --on \*’
> ```
> 
> @vunnyso what do you think?

Thanks @nesteroff for suggestion, I think its good idea to start with, as I am still struggling to find proper fix from [qemu](https://gitlab.com/qemu-project/qemu/-/issues/2520)

UnikieHR:
image

vunnyso:
Closing, Changes integrated in `givc`
https://github.com/tiiuae/ghaf-givc/pull/17
https://github.com/tiiuae/ghaf-givc/pull/20
https://github.com/tiiuae/ghaf-givc/pull/28

PR #770, opened 2024-09-04
Fix nixos module documentation

Mic92:
## Description of changes

This refactors our x86 target to be exposed as a NixOS module. This makes usage outside of ghaf much easier without having to use expensive "extendModules` function. It also makes usage inside ghaf easier i.e. it gives us a target that we can import in module documentation or we can derive a debug target from the release target without having to use wrapper functions.

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature

### DIFF
--- a/modules/common/networking/hosts.nix
+++ b/modules/common/networking/hosts.nix
@@ -109,12 +109,14 @@ let
 in
 {
   options.ghaf.networking.hosts = {
-    enable = mkOption {
-      type = types.bool;
+    enable = (lib.mkEnableOption "Ghaf hosts file generation") // {
       default = true;
     };
     entries = mkOption {
       type = types.listOf hostsEntrySubmodule;
+      description = ''
+        List of hosts entries.
+      '';
       default = null;
     };
   };
--- a/modules/common/profiles/release.nix
+++ b/modules/common/profiles/release.nix
@@ -8,7 +8,9 @@ let
 in
 {
   options.ghaf.profiles.release = {
-    enable = mkEnableOption "release profile";
+    enable = (mkEnableOption "release profile") // {
+      default = false;
+    };
   };
 
   config = mkIf cfg.enable {
--- a/modules/hardware/definition.nix
+++ b/modules/hardware/definition.nix
@@ -90,7 +90,7 @@ in
       inputDevSubmodule = types.submodule {
         options = {
           name = mkOption {
-            type = types.listOf types.any;
+            type = types.listOf types.raw;
             default = [ ];
             description = ''
               List of input device names. Can either be a string, or a list of strings.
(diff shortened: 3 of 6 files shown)


targets/generic-x86_64/flake-module.nix
```diff
@@ -2,117 +2,113 @@
...
-    (generic-x86 "release" [ ])
-  ];
-in
-{
-  flake = {
-    nixosConfigurations = builtins.listToAttrs (
-      map (t: lib.nameValuePair t.name t.hostConfiguration) targets
-    );
+    nixosConfigurations = {
+      generic-x86_64-release = lib.nixosSystem { modules = [ self.nixosModules.generic-x86-release ]; };
```
Mic92, line 96: In theory this could be done more dry, but the code needed for that seemed longer and more complex, so I didn't.

packages/flake-module.nix
```diff
@@ -26,23 +26,24 @@
...
-          #                 microvm.nixosModules.host
-          #                 lanzaboote.nixosModules.lanzaboote
-          #               ];
-          #           };
-          #         in
-          #           cfg.options;
+          options =
+            let
+              cfg = lib.nixosSystem {
+                # derived from targets/laptop/laptop-configuration-builder.nix + lenovo-x1-carbon-gen10
```
Mic92, line 32: Ideally we would expose a nixos module for `lenovo-x1-carbon-gen10` instead of what we are currently doing.

modules/common/profiles/release.nix
```diff
@@ -8,7 +8,9 @@ let
 in
 {
   options.ghaf.profiles.release = {
-    enable = mkEnableOption "release profile";
+    enable = (mkEnableOption "release profile") // {
```
Mic92, line 11: From no on we would only need one profile and not both.

modules/hardware/definition.nix
```diff
@@ -90,7 +90,7 @@ in
       inputDevSubmodule = types.submodule {
         options = {
           name = mkOption {
-            type = types.listOf types.any;
+            type = types.listOf types.raw;
```
Mic92, line 93: This was a bug uncovered by generating documentation.

modules/common/profiles/release.nix
```diff
@@ -8,7 +8,9 @@ let
...
+      default = true;
```
brianmcgillion, line 12: our default is false. the release builds have some significant issues at the moment so all our testing is on the debug targets

Mic92: fixed.

modules/common/networking/hosts.nix
```diff
@@ -109,12 +109,12 @@ let
 in
 {
   options.ghaf.networking.hosts = {
-    enable = mkOption {
-      type = types.bool;
-      default = true;
-    };
+    enable = lib.mkEnableOption "Ghaf hosts file generation";
```
brianmcgillion, line 112: this will now default to false, needs to be checked that it is set in all places

Mic92: Ah, hang on.

brianmcgillion (approved):

PR #780, opened 2024-09-10
Add test for the installer

remimimimimi:
## Description of changes

Introduce [nixos testing framework](https://nixos.org/manual/nixos/stable/index.html#sec-nixos-tests), by adding test for the installer. 

## Checklist for things done

## Instructions for Testing

For now this test only for the installation of `lenovo-x1-carbon-gen11-debug` but could be extended to other targets. To execute this test run `nix build .#checks.x86_64-linux.installer.driver` and then `./result/bin/nixos-test-driver`. Use this way to utilize kvm, this way it takes up to 10 times less times to run. Other way is to just run `nix build .#checks.x86_64-linux.installer` which will build and run test at build time. To use kvm in this case you'll need to configure nix https://nixos.org/manual/nixos/stable/index.html#sec-running-nixos-tests-requirements, but I couldn't get it working.

Remember that it can eat a lot of memory because system image is located in `/tmp` by default. Without kvm it takes almost all 64gb of ram.

### DIFF
--- a/tests/installer/default.nix
+++ b/tests/installer/default.nix
@@ -1,0 +1,101 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# This is a main test for ghaf. If you want to debug it interactively
+# you can follow this guide https://blog.thalheim.io/2023/01/08/how-to-execute-nixos-tests-interactively-for-debugging/.
+# In other words, call breakpoint() at the point of testScript
+# that you interested in to debug. After that you can build test driver using
+# `nix build .#checks.x86_64-linux.installer.driver` and then run driver
+# `./result/bin/nixos-test-driver`. Wait until the moment logs stop appear,
+# after that you can execute `interact` and finally `new_machine.shell_interact()`.
+# This will allow you to interact with installed ghaf.
+{ pkgs, self }:
+let
+  testConfig = "lenovo-x1-carbon-gen11-debug";
+  system = "x86_64-linux";
+  expectedHostname = "ghaf-host";
+
+  cfg = self.nixosConfigurations.${testConfig};
+
+  testingConfig = cfg.extendModules {
+    modules = [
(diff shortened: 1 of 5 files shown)


leivos-unikie:
Tested 
`nix build .#checks.x86_64-linux.installer.driver`
`./result/bin/nixos-test-driver`

At best it took all 16GB of RAM and about 8/16GB swap on my work laptop.

Tested also adding `breakpoint()` to the testScript before the installation starts and then interacting with the ghaf-installer. I worked fine.

Tested also normal ghaf-installer, installed ghaf on Lenovo-X1 and run ci-test-automation. Tests ok except some could not run because of a bit stale ghaf version. --> Rebase.

leivos-unikie (approved):

tests/installer/default.nix
```diff
@@ -0,0 +1,101 @@
...
+  # FIXME: Only one attribute supported. What about ISO?
+  imagePath = testingConfig.config.system.build.diskoImages + "/disk1.raw.zst";
+  targetPath = "/dev/vdb";
+  installScript = pkgs.callPackage ../../packages/installer { };
+  installerInput = pkgs.lib.strings.escapeNixString "${targetPath}\ny\ny\n";
+in
+pkgs.nixosTest {
+  name = "installer-test";
+  nodes.machine = {
+    virtualisation.emptyDiskImages = [ (1024 * 256) ];
```
Mic92, line 43: Oh, that's quite huge :)
I hope that doesn't break CI.
Maybe it could be made a bit smaller for the test only?
You can introduce also a nixos option that only gets set here.

remimimimimi: It's fine, this is just upper bound for the size of a qcow2 file as it allocates memory only when needs it. After installation `empty0.qcow2` size is only 16GB.

Mic92: I assume this is a sparse file than.

brianmcgillion (approved):

PR #766, opened 2024-08-28
Add bluetooth support

mbssrc:
## Description of changes

Adds initial bluetooth support to audiovm, for the use of bluetooth headsets.

Global changes:
- set `mutableUsers` to false
- enable hardened systemd configurations in audiovm

Bluetooth module:
- add bluetooth service config, service system user, dbus config, hardened config, system user, etc.
- add (temporary) bluetooth manager (blueman), connecting to audiovm dbus via ssh 
- add persistent storage for bluetooth device configs

Functionality added:
- search, add/connect, disconnect, and use bluetooth headsets
- use bluetooth headset (and mic) in chromium

Known issues:
- to use mic in chromium, headset profile needs to be manually switched to an HSP/HFP profile (lower quality)
- (irregular) once in a while a new connection will fail, then after a while pop-up an authorization request 
- annoying blueman connect/disconnect pop-ups that need to be manually closed (ESC) 
- bluetooth headset needs to be (re)enabled after reboot, no autoconnect 

## Checklist for things done

- [x] Summary of the proposed changes in the PR description
- [x] More detailed description in the commit message(s)
- [x] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [ ] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] PR linked to architecture documentation and requirement(s) (ticket id)
- [x] Test procedure described (or includes tests). Select one or more:
  - [x] Tested on Lenovo X1 `x86_64`
  - [ ] Tested on Jetson Orin NX or AGX `aarch64`
  - [ ] Tested on Polarfire `riscv64`
- [x] Author has run `make-checks` and it passes
- [x] All automatic Github Action checks pass - see [actions](https://github.com/tiiuae/ghaf/actions)
- [x] Author has added reviewers and removed PR draft status

## Instructions for Testing

- [x] List all targets that this applies to
Lenovo x1

- [x] Is this a new feature

### Testing Checklist
- [x] List the test steps to verify
  - [ ] Regular audio works as before (note: laptop mic does not work in chromium atm)
  - [ ] Bluetooth manager starts
  - [ ] Bluetooth devices can be searched
  - [ ] Bluetooth devices can be added + connected
  - [ ] Device connection switches sound automatically to headset (high-def sound profile) 
[...]

PR #788, opened 2024-09-16
Add common Ghaf audio configuration for app-vms

josa41:
Fix issue with getting no audio when changing sound IO port

## Description of changes
Add common Ghaf audio setting and configuration for App-vns
Also fixes issue for changing audio IO port

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature

Fixes a bug when playing/recording audio and user unplugs 3.5mm (or usb etc) headset
Audio should re-route automatically to integrated speakers/mic.
There is 1 second delay in the switchin of audio port
Verify recording with shell tool parecord / paplay since some mic recording web pages seemed to cause buggy noise

### DIFF
--- a/modules/common/services/audio.nix
+++ b/modules/common/services/audio.nix
@@ -44,7 +44,7 @@ in
                 # Enable TCP socket for VMs pulseaudio clients
                 "server.address" = [
                   {
-                    address = "tcp:${toString cfg.pulseaudioTcpPort}";
+                    address = "tcp:0.0.0.0:${toString cfg.pulseaudioTcpPort}";
                     "client.access" = "unrestricted";
                   }
                 ];
--- a/modules/microvm/virtualization/microvm/appvm.nix
+++ b/modules/microvm/virtualization/microvm/appvm.nix
@@ -33,6 +33,7 @@ let
             internalIP = index + 100;
           })
 
+          ./common/ghaf-audio.nix
           ./common/storagevm.nix
 
           # To push logs to central location
@@ -71,6 +72,11 @@ let
                   withPolkit = true;
                   withDebug = configHost.ghaf.profiles.debug.enable;
                   withHardenedConfigs = true;
+                };
+
+                ghaf-audio = {
+                  inherit (vm.ghafAudio) enable;
+                  name = "${vm.name}";
                 };
 
                 storagevm = {
@@ -253,6 +259,7 @@ in
               type = types.nullOr types.str;
               default = null;
             };
+            ghafAudio.enable = lib.mkEnableOption "Ghaf application audio support";
             vtpm.enable = lib.mkEnableOption "vTPM support in the virtual machine";
           };
         }
--- a/modules/microvm/virtualization/microvm/common/ghaf-audio.nix
+++ b/modules/microvm/virtualization/microvm/common/ghaf-audio.nix
@@ -1,0 +1,44 @@
+# Copyright 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.ghaf.ghaf-audio;
+  audiovmHost = "audio-vm";
+  audiovmPort = config.ghaf.services.audio.pulseaudioTcpPort;
+  address = "tcp:${audiovmHost}:${toString audiovmPort}";
+  reconnectMs = 1000;
+in
+{
+  options.ghaf.ghaf-audio = with lib; {
+    enable = mkEnableOption "Ghaf audio support for application virtual machine.";
+
+    name = mkOption {
+      description = ''
+        Basename of corresponding virtual machine audio channel.
+      '';
+      type = types.str;
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    security.rtkit.enable = true;
+    users.extraUsers.ghaf.extraGroups = [
+      "audio"
+      "video"
+    ];
+
+    hardware.pulseaudio = {
+      enable = true;
+      extraConfig = ''
+        load-module module-tunnel-sink-new sink_name=${cfg.name}.speaker server=${address} reconnect_interval_ms=${toString reconnectMs}
+        load-module module-tunnel-source-new source_name=${cfg.name}.mic server=${address} reconnect_interval_ms=${toString reconnectMs}
+      '';
+      package = pkgs.pulseaudio-ghaf;
+    };
+  };
+}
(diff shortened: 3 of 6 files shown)


nesteroff:
This seems to be extracted from the [commit](https://github.com/tiiuae/ghaf/pull/766/commits/9e388f01c2710a6beeffc3ba5ad8cbe4ad964c42) I made to get Bluetooth audio working. I just want to point out that I'm using the new `module-tunnel-sink-new` module there with a small patch to enable switching to new devices. Is there any particular reason you decided to stick with the old `module-tunnel-sink`?

josa41:
> This seems to be extracted from the [commit](https://github.com/tiiuae/ghaf/pull/766/commits/9e388f01c2710a6beeffc3ba5ad8cbe4ad964c42) I made to get Bluetooth audio working. I just want to point out that I'm using the new `module-tunnel-sink-new` module there with a small patch to enable switching to new devices. Is there any particular reason you decided to stick with the old `module-tunnel-sink`?

No reason to keep the old module-tunnel-sink except that your change is not in main yet. I need to rebase anyway so i thought it is better to post this in early for comments.

modules/common/services/audio.nix
```diff
@@ -44,7 +44,7 @@ in
                 # Enable TCP socket for VMs pulseaudio clients
                 "server.address" = [
                   {
-                    address = "tcp:${toString cfg.pulseaudioTcpPort}";
+                    address = "tcp:0.0.0.0:${toString cfg.pulseaudioTcpPort}";
```
nesteroff, line 47: It listens on all interfaces by default, but adding 0.0.0.0 disables IPv6. Is that intentional?

modules/microvm/virtualization/microvm/appvm.nix
```diff
@@ -95,7 +95,10 @@ let
                 runWaypipe
                 pkgs.tpm2-tools
                 pkgs.opensc
-              ];
+              ]
+              ++ (lib.optional (
+                  config.ghaf.profiles.debug.enable )
+                    pkgs.pulseaudio );
```
nesteroff, line 101: What's this used for? Setting hardware.pulseaudio.enable to true automatically adds pulseaudio to systemPackages.

modules/microvm/virtualization/microvm/appvm.nix
```diff
@@ -106,6 +109,19 @@ let
...
+
+              hardware.pulseaudio = lib.mkIf vm.ghafAudio {
+                enable = true;
+                extraConfig = ''
+                    load-module module-tunnel-sink-new sink_name=${vm.name}.speaker server=audio-vm:4713 reconnect_interval_ms=1000
+                    load-module module-tunnel-source-new source_name=${vm.name}.mic server=audio-vm:4713 reconnect_interval_ms=1000
+                '';
+                package = pkgs.pulseaudio-ghaf;
+              };
+
```
nesteroff, line 124: Would it make sense to move this to a separate module, like common/vm-audio.nix, and import it here, similar to how we have vm-networking and storage?

nesteroff (commented):
With the latest update, does this PR actually change anything in the audio configuration besides moving things around? The PR description could probably be updated. Either way, it's good to have the audio configuration in one place. I just added a couple of comments in the code.

leivos-unikie:
Tested audio input and output in Slack and chromium with 3.5mm headset and integrated speakers/mic. Works fine and the delay in switching sound device is neglible.

josa41:
> With the latest update, does this PR actually change anything in the audio configuration besides moving things around? The PR description could probably be updated. Either way, it's good to have the audio configuration in one place. I just added a couple of comments in the code.

Yes the change does not actually do any functional changes as is since the same change was merged with bluetooth change.

modules/common/services/audio.nix
josa41: Yes disabling ipv6 was intentional.
I did not know that we need ipv6 for pulseaudio, i can remove the change if it is needed.
Not that we currently do any firewalling for pulseaudio but to me it seems all of our firewalling is done in ipv4 so everyting on ipv6 is wide open.

modules/microvm/virtualization/microvm/appvm.nix
josa41: It could be leftover from when using pipewire instead of pulseaudio.
I can check that thanks.

modules/microvm/virtualization/microvm/appvm.nix
josa41: added ghaf-audio module

nesteroff: Why not move the definition of the ghafAudio option to the module as well and use lib.mkIf cfg.enable? This way it will be consistent with other modules and easier to read. You can take a look at storagevm.nix as an example or pretty much any other module in Ghaf.

leivos-unikie:
I tested this again now also with USB headset. The USB headset worked fine but integrated mic didn't work after unplugging. I noticed that integrated mic activates only if I plug first 3.5mm headset (or boot the machine with 3.5mm headset plugged).

leivos-unikie:
> I noticed that integrated mic activates only if I have plugged and unplugged first 3.5mm headset (or boot the machine with 3.5mm headset plugged).

This issue is in mainline, not caused by this PR.

modules/microvm/virtualization/microvm/appvm.nix
josa41: Added ghaf-audio as a separate module with an option to turn it on/off.

milva-unikie:

**Critical issue**
Using a microphone causes the laptop to not shut down properly. After Shutdown-button is pressed, the screen turns off but the fan stays on. The fan kept running for over 10 minutes before I manually shut it down by pressing the power button of the laptop for 10 seconds. I reproduced this with internal microphone, usb microphone and 3,5 mm microphone.


**Working**

Tested in chromium-vm and in comms-vm (Slack). Firewall in business-vm is blocking Teams from opening, so could not test with that.

- Integrated speakers work.
- 3,5 mm headset works.
- Usb headset works.
- Audio automatically changes to a bluetooth device when one is connected. 
- Audio automatically changes back when bluetooth device is disconnected.
- Did not find any issues when changing between different audio devices.
- Did not find any issues when changing between vm:s.
- Test-automation passes.

Known issues:
- Integrated microphone does not work after boot. It starts working after 3,5 mm headset is plugged in and unplugged.

milva-unikie:
The bug I noted in my previous comment is not caused by this pr. I think this is now good to go in!

brianmcgillion (approved):

PR #791, opened 2024-09-17
desktop: new color scheme, fix cursors

humaidq-tii:
## Description of changes

This implements the new theme as per the new designs.
Also the cursor on the desktop should be consistent with apps.

Regarding opacity: setting alpha channel to 1 for background colour somehow breaks border-radius for GTK applications. Setting it to 0.99 resolves this issue.

## Checklist for things done

- [ ] Change requires full re-installation: no
- [x] Change can be updated with `nixos-rebuild ... switch`: yes, gui-vm will reboot

## Instructions for Testing

  - All targets with labwc desktop
- [ ] Is this a new feature: no
    - Boot, desktop bar should be darker, same as launcher, lock screen, and login screen.
    - Cursor shouldn't change theme when moving across different apps, or desktop and waybar
  - Changes shouldn't impact functionality.

### DIFF
--- a/modules/desktop/graphics/labwc.config.nix
+++ b/modules/desktop/graphics/labwc.config.nix
@@ -12,19 +12,19 @@ let
   audio-ctrl = pkgs.callPackage ../../../packages/audio-ctrl { };
   gtklockStyle = pkgs.writeText "gtklock.css" ''
     window {
-      background: rgba(29, 29, 29, 1);
-      color: #eee;
+      background: rgba(18, 18, 18, 1);
+      color: #fff;
     }
     button {
       box-shadow: none;
       border-radius: 5px;
-      border: 1px solid rgba(255, 255, 255, 0.09);
-      background: rgba(255, 255, 255, 0.06);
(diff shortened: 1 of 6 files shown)


kajusnau:
Looks good to me.
For the colors chosen, would it be beneficial to use only colors defined by Joni in the color palette page in Figma? That way, if we make a color palette package in the future, we'd easily know which color codes represent.
Of course, if the color palette colors aren't set in stone yet, then this doesn't apply.

modules/desktop/graphics/labwc.config.nix
```diff
@@ -187,7 +187,8 @@ let
   '';
 
   environment = ''
-    XCURSOR_THEME=breeze_cursors
+    XCURSOR_THEME=adwaita
```
milva-unikie, line 190: Cursor themes don't work yet, I think this should fix it
```suggestion
    XCURSOR_THEME=Adwaita
```

brianmcgillion (approved):

PR #815, opened 2024-10-01
Add extension to trusted browser to allow opening links in normal browser

tervis-unikie:
## Description of changes

Add browser extension to trusted browser which adds "Open in normal browser" option to pop-up menu for links.

## Checklist for things done

## Instructions for Testing

- [x] Is this a new feature
    * Open an allowed site in trusted browser
    * Right-click on a link
    * Select "Open link in normal browser"
    * Check if the link is actually opened in the normal browser 
    * Open a not-allowed page in trusted browser
    * Right-click anywhere on the "Site cannot be reached" or "403 Access denied" -page
    * Select "Open page in normal browser"
    * Check if the page actually opened in the normal browser

### DIFF
--- a/modules/reference/appvms/business.nix
+++ b/modules/reference/appvms/business.nix
@@ -8,6 +8,7 @@
   ...
 }:
 let
+  inherit (lib) mkIf optionalString;
   #TODO: Move this to a common place
   name = "business";
   tiiVpnAddr = "151.253.154.18";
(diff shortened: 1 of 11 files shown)


packages/open-normal-extension/open_normal.sh
```diff
@@ -0,0 +1,113 @@
...
+fi
+
+# Read the json payload
+LANG=C IFS= read -r -d '' -n "$LEN" JSON
+# Remove json prefix
+PFX="{\"URL\":\""
+URL="${JSON##"$PFX"}"
+# Remove json suffix
+SFX="\"}"
+URL="${URL%%"$SFX"}"
```
slakkala, line 89: For future proofing could also use `jq`, like: `URL="$(dd count=$LEN status=none bs=1 | jq -r .URL)"` (jq would need the package dependency and correct path)

packages/open-normal-extension/open_normal.sh
```diff
@@ -0,0 +1,113 @@
...
+fi
+if [ -z "$B4" ]; then
+    B4=0
+else
+    B4="$(Ord "$B4")"
+fi
+
+# Convert the four bytes to a integer length
+# Using multiply instead of bit shifing as bit shifts broke syntax highlighting
+LEN="$((B1+(B2*256)+(B3*65536)+(B4*16777216)))"
```
slakkala, line 71: Could also use `od`: `read LEN _ < <(od -t u4 -A n -N 4 --endian=little)`

slakkala (commented):
LGTM, dropped a couple of considerations for the script part.

Hoping someone more experienced in nix can take a look at the nix parts.

mbssrc (approved):

modules/reference/appvms/business.nix
```diff
@@ -46,130 +47,160 @@ in
...
+          givc.appvm = {
             enable = true;
-            csdWrapper = "${pkgs.openconnect}/libexec/openconnect/hipreport.sh";
+            name = lib.mkForce "business-vm";
+            applications = [
+              {
+                name = "chromium";
+                command =
+                  "${config.ghaf.givc.appPrefix}/run-waypipe ${config.ghaf.givc.appPrefix}/chromium --enable-features=UseOzonePlatform --ozone-platform=wayland ${config.ghaf.givc.idsExtraArgs}"
+                  + optionalString config.ghaf.givc.enable " --load-extension=${pkgs.open-normal-extension}";
```
mbssrc, line 84: I think here it is too late to check if givc is enabled

packages/open-normal-extension/open_normal.sh
tervis-unikie: Agreed, that would also work, and it would be short and tidy. It would have a dependency though, but Gnu coreutils package is probably already there. Above considerations apply here too.

modules/reference/appvms/business.nix
tervis-unikie: That did cross my mind, but eventually decided to have it there anyway, just in case.
Shall I just remove the optionality of the "--load-extension..." -string?

packages/open-normal-extension/open_normal.sh
tervis-unikie: I considered this a bit, and decided that having jq as dependency and having processing overhead of it is not justified here as input for the script is defined by another script in the same package.

packages/open-normal-extension/open_normal.sh
tervis-unikie: I decided that using od is justified as it makes the script a lot simpler.
I did however ditch the read in this case: `LEN="$(od -t u4 -A n -N 4 --endian=little)"`

modules/reference/appvms/business.nix
tervis-unikie: Removed optionality of the "--load-extension..." -string

johannarautanen:
Tested with platform: **Lenovo X1 Carbon;** flash script
commit: f9c58d74545b909e2ed910c1cc8ccadcc2983401

**Issues:**
-when you opened first Microsoft 365-, Microsoft Outlook- and  Teams-app (login with unikie testuser to unikie network), 
now when open the "Trusted Browser", the extension isn't show at all.
But when the "Trusted Browser" before opening the "microsoft"-app, the extension is visible on the browser.

screenshot-20241011-12-02-08

screenshot-20241011-11-59-35

-When you open allowed site in "Trusted Browser" (TII Jira-page, VPN to access.tii.ae) and click the righ-click the link, there is no "Open in normal browser" to choose. 
And if have pin the "Open in normal browser" visible, the "open in normal browser" cannot be chosen, the option isn't active.
screenshot-20241011-11-55-46
screenshot-20241011-11-56-12



ci-test-automation suite: ok
starting apps: ok
some ad-hoc testing: ok

tervis-unikie:
Thanks for testing!

> **Issues:** -when you opened first Microsoft 365-, Microsoft Outlook- and Teams-app (login with unikie testuser to unikie network), now when open the "Trusted Browser", the extension isn't show at all. But when the "Trusted Browser" before opening the "microsoft"-app, the extension is visible on the browser.

Okay, seems that atm the extension is only loaded when you open trusted browser from the Ghaf-menu (or whatever it is called). I'll have to see how I can add the loading of the extension to other places where trusted browser can be launched.

> -When you open allowed site in "Trusted Browser" (TII Jira-page, VPN to access.tii.ae) and click the righ-click the link, there is no "Open in normal browser" to choose. And if have pin the "Open in normal browser" visible, the "open in normal browser" cannot be chosen, the option isn't active.

In this case you're not right-clicking a link, you're right-clicking the address bar. Current extension only applies to links shown on the actual page. Atm I do not know if it is possible to reasonably add this functionality to the address bar also.

vadika (approved):

tervis-unikie:
Links from outlook, teams and office365 should also have the "Open in normal browser" now.
Please do not test the address-bar, there won't be an option for opening address-bar URL in normal browser atm.

johannarautanen:
Tested with platform: Lenovo X1 Carbon; flash script
commit: cd7eafb66002ba142f40c077e7edd3081b7750a0

**Working:**
-when you right-click on the link(not the address bar) and select "Open in normal browser" -> normal browser opened.
-Links in Outlook, Teams and office365 have now the "Open in normal browser"-option > normal browser opened.
-This time the  "open in normal browser" -extension was available in the "Trusted Browser" although opened after the Teams, outlook and office365-apps.

No new issues found.

ci-test-automation suite: ok
starting apps: ok
some ad-hoc testing: ok

tervis-unikie:
Unfortunately there's no simple/non-hackish way to access the address bar contents in a chromium extension.

However I added the "Open page in normal browser" -option to pop-up menu when right clicking anywhere else than images or links on the page. 
Also clicking the extension icon will open the current tab page in normal browser. 

These make it rather straightforward to open "an access denied" site in normal browser instead.

johannarautanen:
**Tested with platform: Lenovo X1 Carbon**; flash script
commit: ab44145cd20751b8e3fc049a6370688b7718687f

Working:
- you can open the "Trusted Browser"-page in normal browser by right clicking in the page and selecting the "Open in normal browser" else than image 
Screenshot from 2024-10-14 08-06-41

-And when you have tried to open ie. "yle.fi"-page, which cannot be open via tii.ae; when select the "Extensions"-icon > "open in normal browser"-options, then "yle.fi"-page opened in normal browser successfully.
Screenshot from 2024-10-14 08-03-54


-Or when you have tried to open ie. "yle.fi"-page, which cannot be open via tii.ae; when select the pinned "Open in normal browser"-options, then "yle.fi"-page opened in normal browser successfully.
Screenshot from 2024-10-14 08-05-23



No new issues found.

ci-test-automation suite: ok
starting apps: ok
some ad-hoc testing: ok

packages/open-normal-extension/fi.ssrc.open_normal.json
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
```
avnik, line 1: License header could be moved to own `fi.ssrc.open_normal.json.license` or even attributed via `.reuse/dep5` file

avnik (approved):
LGTM for me. I add little notice about license header but treat is a minor

tervis-unikie:
Just rebased on ghaf/main

PR #819, opened 2024-10-04
zfs: enable lz4 compression

Mic92:
This should speed up image creation but also in my experience lz4 compression usually easily keeps up with NVME storage bandwidth so in turn it boost IO performance

## Checklist for things done

  This change would require a full installation. However only if you want to benefit from the effect. It's also possible to enable lz4 at runtime and only use it for new blocks.

## Instructions for Testing

- [ ] Is this a new feature

### DIFF
--- a/modules/disko/disko-ab-partitions.nix
+++ b/modules/disko/disko-ab-partitions.nix
@@ -95,6 +95,7 @@
           type = "zpool";
           rootFsOptions = {
             mountpoint = "none";
+            compression = "lz4";
             acltype = "posixacl";
           };
           datasets = {


Mic92:
My personal motivation for this usually is that I can store more data. I believe there were performance benchmarks in this repository as well? Might be good to apply them to this change.

UnikieHR:
image

brianmcgillion:
https://github.com/tiiuae/ghaf/pull/848/commits/032628305c3ccd2f87c5fe74ce7af3acf5084753#diff-3c738a9508c4b16d85de5b09d6bdf6ed4d8493f3f1575dde7d49b32b3eb6fa8aR93

was added here an tested with the restructured fs layout

PR #850, opened 2024-10-16
[GhafPkgs] Update Audio Control

baz2142:
Add AppVM names support. Now you can see, which stream belongs to which AppVM

## Checklist for things done

## Instructions for Testing

Run Audio Control
Run Chromium and check, that steams appear under "Chromium-appvm" name
Another audio apps, if it is possible

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -215,11 +215,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1727854413,
-        "narHash": "sha256-5nysqAzKvv2Yn2eKAyR+U1zjZ535NSRuKNapICkDwfg=",
+        "lastModified": 1728969248,
+        "narHash": "sha256-UcgTfjWpzC/VfZy6ZNtt1aPZtoqNP6Bk5ffRSaQaM3g=",
         "owner": "tiiuae",
         "repo": "ghafpkgs",
-        "rev": "58ab6179fce5932862707c833b57f4526503399d",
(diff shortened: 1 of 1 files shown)


leivos-unikie:
Tested on Lenovo-X1 (rebuild switch) 
- ci-test-automation pass
- all menu icons work
- audio control tool shows streams by vm name but there is strange note "No streams from the AppVM" although I am listening sound from that particular app-vm
-

leivos-unikie:
Also noticed now that the audio streams stay open although corresponding apps are closed. Is this ok?

baz2142:
> Also noticed now that the audio streams stay open although corresponding apps are closed. Is this ok?

I think, now every AppVM has it's own stream for the all apps. So it's not like every application inside an AppVM creates a stream, but every application inside it goes through the same stream

Is it correct, @josa41?

josa41:
> > Also noticed now that the audio streams stay open although corresponding apps are closed. Is this ok?
> 
> I think, now every AppVM has it's own stream for the all apps. So it's not like every application inside an AppVM creates a stream, but every application inside it goes through the same stream
> 
> Is it correct, @josa41?

Yes in current version the app-vm created the pulseaudio tcp tunnel open even though the app is closed. The actual pulseaudio application audio stream is closed but tcp tunnel that carries the stream stays open.
From audio-vm
`pactl list sink-inputs` or `pactl list source-outputs` should show the actual application audio streams

I have made a patch that would fix the tcp stream so that the connection will get closed along with the app: #852

PR #851, opened 2024-10-16
Enable Wayland security context protocol

nesteroff:
## Description of changes

This enables support for the Wayland security context protocol in Ghaf. I previously added support for it to Labwc and Waypipe and upstreamed those changes. The version of Labwc used in Ghaf already includes it. For Waypipe, I added a patch but it was merged upstream recently. The new version should be released close to the end of the year. After that we should be able to get rid of the Waypipe overlay.

The security context protocol allows window frame colors to be defined on a per-VM basis rather than by application identifier. All settings are defined on the compositor side and AppVMs cannot change them. The same application can now have different border colors when run from different VMs. For example, the trusted browser and the regular one now have different colors, even though they are both Chromium.

Also, enabling the Security Context seems to resolve the issue with window caption rendering reported in SSRCSP-5343.

Technical details:
- Each AppVM now gets its own instance of waypipe and vsockproxy.
- vsockproxy has been updated to prevent one VM from connecting to the port designated for another VM.
- Waypipe borders patch has been updated to draw borders on the GUIVM side. Currently, waypipe border is disabled but it might be useful in the future in case we change the compositor, as not many compositors support security contexts and custom border colors. 

## Checklist for things done

## Instructions for Testing

On Lenovo X1, make sure all apps still work fine after this change and that the window borders have the correct colors. The regular Chromium and Trusted versions should now have different colors as defined in chromium.nix and business.nix.

### DIFF
--- a/modules/desktop/graphics/labwc.config.nix
+++ b/modules/desktop/graphics/labwc.config.nix
@@ -154,6 +154,13 @@ let
           '') cfg.frameColouring
         )
       }
+      ${
+        lib.concatStringsSep "\n" (
+          map (rule: ''
+            <windowRule sandboxAppId="${rule.identifier}" borderColor="${rule.color}" serverDecoration="yes" skipTaskbar="no"  />
+          '') cfg.securityContext
+        )
+      }
     </windowRules>
     <libinput>
       <device category="touchpad"><naturalScroll>yes</naturalScroll></device>
--- a/modules/desktop/graphics/labwc.nix
+++ b/modules/desktop/graphics/labwc.nix
@@ -59,35 +59,27 @@ in
           identifier = "foot";
           colour = "#006305";
         }
-        # TODO these should reference the VM and not the application that is
-        # relayed through waypipe. Ideally this would match using metadata
-        # through Wayland security context.
-        {
-          identifier = "dev.scpp.saca.gala";
-          colour = "#027d7b";
-        }
-        {
-          identifier = "chromium-browser";
-          colour = "#630505";
-        }
-        {
-          identifier = "org.pwmt.zathura";
-          colour = "#122263";
-        }
-        {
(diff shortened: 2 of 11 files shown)


UnikieHR:
image

johannarautanen:
Tested on Lenovo-X1 (flashed to the SSD) (commit: 139e527d5e8ce0fae6571bf7770cda52fbb36c2a)

Working: 
- regular Chromium and Trusted versions have different borders colors 
- all other ghaf apps can be launched
- ci-test automation cases ok

Issues:
- business-app border color is ok, but the header color shouldn't be neon green, not user friendly at all. Hard to read the text in header and see the "minimize-maximize-close"-option.
Screenshot from 2024-10-23 08-31-10

- minor cosmetic thing, the text editor seem to have rounded borders, so are "holes" in corners:
screenshot-20241023-09-55-56

nesteroff:
> * business-app border color is ok, but the header color shouldn't be neon green, not user friendly at all. Hard to read the text in header and see the "minimize-maximize-close"-option.

Thanks for testing! I updated the color to a slightly different shade. Hopefully, it looks better now. As far as I understand, Joni is working on the new colors so we might update them again later.

> * minor cosmetic thing, the text editor seem to have rounded borders, so are "holes" in corners:

I guess it has always been like that but it was overlapped by the waypipe border, which is no longer used. I made a quick workaround to disable the rounded corners. I think a better fix might be to develop a Ghaf GTK theme but this is something we can consider later on.

johannarautanen:
**Tested on Lenovo-X1** (flashed to the SSD) 

Working:

- regular Chromium and Trusted versions have different borders colors
- business-app header color is much better now
- text editor has right angle corners
- all other ghaf apps can be launched
- ci-test automation cases ok

screenshot-20241029-09-14-59

PR #856, opened 2024-10-18
Network hardening using sysctl

gangaram-tii:
## Description of changes
sysctl configuration for network hardening. 
Uses configuration recommended in different security audits.
Configuration and it's recommendation is available in this document:

[Recommended sysctl settings](https://github.com/gangaram-tii/ghaf-debug-tools/blob/main/audit/sysctl/network/references.md)

Sysctl network setting audit report available:
https://github.com/gangaram-tii/ghaf-debug-tools/blob/main/report/hardened_network_audit_report.md

Network Performance impact:
https://github.com/gangaram-tii/ghaf-debug-tools/blob/main/report/perf-iperf3.png
Note: `ghaf+` label is with sysctl hardened network settings.

## Checklist for things done

- [-] Summary of the proposed changes in the PR description
- [-] More detailed description in the commit message(s)
- [-] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [-] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [-] Test procedure described (or includes tests). Select one or more:

## Instructions for Testing

- [-] List all targets that this applies to: all platform targets
- [ ] Is this a new feature: yes
- [-] If it is an improvement how does it impact existing functionality?
   - Test network related functionality e.g. (internet access and inter vm communication).

### DIFF
--- a/modules/common/security/networking.nix
+++ b/modules/common/security/networking.nix
@@ -1,0 +1,201 @@
+# Copyright 2024-2025 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  ...
+}: let
+  cfg = config.ghaf.security.sysctl.network;
+in {
+  ## Options to enable IP security
+  options.ghaf.security.sysctl.network = {
+    disable-all = lib.mkOption {
+      description = ''
+        Disable all sysctl network security configuration.
+      '';
+      type = lib.types.bool;
+      default = false;
+    };
+
+    disable-ipv6 = lib.mkOption {
+      description = ''
+        Disables Internet Protocol v6.
+        It is recommended to reduce attack surface by disabling IPv6 if it is not being used.
+        References:
+        https://www.tenable.com/audits/items/CIS_AlmaLinux_OS_8_Server_v1.0.0_L2.audit:a532993e3ac29c19c0d491463a881303
+        https://linux-audit.com/linux-security-guide-for-hardening-ipv6/
+      '';
+      type = lib.types.bool;
+      default = true;
+    };
+
+    prevent-syn-flooding = lib.mkOption {
+      description = ''
+        Prevent SYN packet flooding.
+        It is recommended to reduce attack surface by disabling IPv6 if it is not being used.
+        References:
+        https://www.tenable.com/audits/items/CIS_AlmaLinux_OS_8_Server_v1.0.0_L2.audit:a532993e3ac29c19c0d491463a881303
+        https://linux-audit.com/linux-security-guide-for-hardening-ipv6/
+      '';
+      type = lib.types.bool;
+      default = true;
+    };
+
+    enable-RFC_1337-protection = lib.mkOption {
+      description = ''
+        Enable RFC 1337 fix to protect against TIME-WAIT assassination attacks or unintended connection resets.
+        References:
+        https://wiki.archlinux.org/title/Talk:Sysctl
+        https://datatracker.ietf.org/doc/html/rfc1337
+      '';
+      type = lib.types.bool;
+      default = true;
+    };
+
+    enable-rp-filter = lib.mkOption {
+      description = ''
+        Enable reverse path filtering to protect against IP address spoofing.
+        References:
+        https://www.tenable.com/audits/items/CIS_Amazon_Linux_2_STIG_v1.0.0_L1.audit:8a082faa85bd75cf16b791cada7d6caf
+        https://docs.vmware.com/en/vRealize-Operations/8.10/com.vmware.vcom.core.doc/GUID-0AAA4D96-5FDE-49A7-8BB3-D7F56C89137C.html
+      '';
+      type = lib.types.bool;
+      default = true;
+    };
+
+    disable-icmp-redirect-acceptance = lib.mkOption {
+      description = ''
+        Disable ICMP redirect acceptance to protect against routing-based exploits.
+        Reference:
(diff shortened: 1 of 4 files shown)


modules/common/security/networking.nix
```diff
@@ -0,0 +1,201 @@
...
+    disable-ipv6 = lib.mkOption {
+      description = ''
+        Disables Internet Protocol v6.
+        It is recommended to reduce attack surface by disabling IPv6 if it is not being used.
+        References:
+        https://www.tenable.com/audits/items/CIS_AlmaLinux_OS_8_Server_v1.0.0_L2.audit:a532993e3ac29c19c0d491463a881303
+        https://linux-audit.com/linux-security-guide-for-hardening-ipv6/
+      '';
+      type = lib.types.bool;
+      default = true;
```
Mic92, line 29: Should this be default to true?

modules/common/security/networking.nix
```diff
@@ -0,0 +1,201 @@
...
+        Enable RFC 1337 fix to protect against TIME-WAIT assassination attacks or unintended connection resets.
+        References:
+        https://wiki.archlinux.org/title/Talk:Sysctl
+        https://datatracker.ietf.org/doc/html/rfc1337
+      '';
+      type = lib.types.bool;
+      default = true;
+    };
+
+    enable-rp-filter = lib.mkOption {
```
Mic92, line 55: Doesn't linux has enabled this by default anyway?

modules/common/security/networking.nix
```diff
@@ -0,0 +1,201 @@
...
+      description = ''
+        Don't accept source routed IP packets to protect internal addressed from outside world.
+        Reference:
+        https://www.tenable.com/audits/items/CIS_Red_Hat_EL5_v2.2.1_L1.audit:0df0d28b8145e0c83c9a9e11c185644a
+      '';
+      type = lib.types.bool;
+      default = true;
+    };
+
+    disable-ping-request = lib.mkOption {
```
Mic92, line 86: This breaks ipv6.

modules/common/security/networking.nix
gangaram-tii: OK, I set default to `false`. In Ghaf by default IPv6 is disabled.
https://github.com/tiiuae/ghaf/blob/main/modules/microvm/virtualization/microvm/common/vm-networking.nix#L24C23-L24C24

modules/common/security/networking.nix
gangaram-tii: It is not enabled by default in Linux. Since it is recommended enable rp_filter in many Linux security audits, so I thought to include it. Now it defaults to `false`.

modules/common/security/networking.nix
gangaram-tii: OK, I set default to `false`. In Ghaf by default IPv6 is disabled.
https://github.com/tiiuae/ghaf/blob/main/modules/microvm/virtualization/microvm/common/vm-networking.nix#L24C23-L24C24

modules/common/security/networking.nix
gangaram-tii: This also defaults to `false` now.

modules/common/security/networking.nix
mbssrc: If we enable IPv6 we should include hardening settings as well. I agree we should support IPv6, but enabling it indiscriminately in all VMs and interfaces may not be the best strategy with our current setup, and afaik enabling it for specific interfaces is still possible with these settings

modules/common/security/networking.nix
mbssrc: Seems this flag is not used? Also, NixOS seems to set it to 2 (loose mode), whereas this option sets it to 1 (strict mode). Not entirely sure about the implications within Ghaf, especially internal network, can you elaborate if strict mode makes sense as a default or not?

gangaram-tii: I studied around this. As far as what I understood, loose mode(default) is fine for VMs which are communicating through internal network. Strict mode should be beneficial in net-vm specially if some one tries to spoof an VM address. In strict mode these packets will be dropped while in loose mode these may be accepted as there is always a valid route for these.

modules/common/security/networking.nix
```diff
@@ -0,0 +1,188 @@
...
+  lib,
+  ...
+}:
+let
+  cfg = config.ghaf.security.sysctl.network;
+in
+{
+  ## Options to enable IP security
+  options.ghaf.security.sysctl.network = {
+    disable-all = lib.mkOption {
```
mbssrc, line 14: Why not the typical `enable` flag that defaults to true?

gangaram-tii: Thanks. Done.

modules/common/security/networking.nix
```diff
@@ -0,0 +1,188 @@
+# Copyright 2024-2025 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  ...
+}:
+let
+  cfg = config.ghaf.security.sysctl.network;
```
vunnyso, line 9: Can add 
```
  inherit (lib) mkOption mkForce mkDefault mkIf types;
```
For better readability.

vunnyso (changes requested):

modules/common/security/networking.nix
gangaram-tii: done

vunnyso (approved):
Approved with prejudice.

Issue #444, opened 2024-01-09
hash mismatch building lenovo-x1-carbon-gen11-debug


henrirosten:
On building Ghaf `lenovo-x1-carbon-gen11-debug` we see occasional build failures:

```
error: builder for '/nix/store/dyw71dxaaxvgyl3mnl5xc54ryqf5kxay-nixos-disk-image.drv' failed with exit code 1;
       last 10 log lines:
       > copying path '/nix/store/3z537aq6rbvfrl59bgfm6ls22z2dyy8z-chromium-120.0.6099.71-sandbox' to 'local'...
       > copying path '/nix/store/bblyj5b3ii8n6v4ra0nb37cmi3lf8rz9-coreutils-9.3' to 'local'...
       > copying path '/nix/store/rpzk0h1d11ciarcjkv1zzgsa1iz8zmv8-libidn-1.41' to 'local'...
       > error: hash mismatch importing path '/nix/store/3z537aq6rbvfrl59bgfm6ls22z2dyy8z-chromium-120.0.6099.71-sandbox';
       >          specified: sha256:01bffjgrkk0saf32pjrgpcx893dksp2mqcbpy8bfn6wrinndd2gs
       >          got:       sha256:1wi876fpb69cv2n8vkzr04mv2p5hbc543mpdci7a2lvk3ghnsr2c
```
More examples are available from the ~'past month' `lenovo-x1-carbon-gen11-debug` build failures in github action pre-merge builds one can look for from: https://github.com/tiiuae/ghaf/actions/workflows/build.yml?query=is%3Afailure.

The issue seems to relate to chromium build (hence only `lenovo-x1-carbon-gen11-debug` is impacted from the set of targets currently build in pre-merge build workflow). Also, having many substituters configured seems to increase the likelihood of the build hitting this issue. For instance, pre-mrege builds configure the following binary caches: https://github.com/tiiuae/ghaf/blob/main/.github/workflows/build.yml#L146-L147.

As a workaround @brianmcgillion temporarily disabled the `lenovo-x1-carbon-gen11-debug` target from the pre-merge build workflow with PR: https://github.com/tiiuae/ghaf/pull/441.

We would like to re-enable the `lenovo-x1-carbon-gen11-debug` in the pre-merge build action, but this issue needs to be resolved first.

PR #874, opened 2024-10-30
build.yml: Use nix-fast-build to build on remote

henrirosten:
## Description of changes

Start using [nix-fast-build](https://github.com/Mic92/nix-fast-build) in the build.yml workflow to execute all build operations completely on the remote end.

Removes the step 'Maximize space available on rootfs' since with `nix-fast-build`, we no longer download the full build closure on the github-hosted runner. Furthermore, with the `--no-download` option, we also skip the download of the final build result (ghaf image), so we no longer need to free up the storage on the github runner.

This change also removes the upload to https://ghaf-dev.cachix.org from the build workflow: it is no longer needed, since builds are already cached on the remote builders' local nix store. Jenkins builds will take care of uploading the build binaries to a separate nix binary cache, maintained with [ghaf-infra](https://github.com/tiiuae/ghaf-infra).

With the above summarized changes, build times on the build.yml workflow will significantly reduce. It should also completely remove the [hash mismatch issues](https://github.com/tiiuae/ghaf/issues/444) on the build workflow as builds and also evaluation operations are executed on the remote.

This change was tested on a fork, as an example, see: https://github.com/henrirosten/ghaf/actions/runs/11587433044. 

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,30 +100,6 @@ jobs:
       group: ${{ github.workflow }}.${{ github.event.pull_request.number || github.ref }}.${{ matrix.arch }}.${{ matrix.target }}
       cancel-in-progress: true
     steps:
-      - name: Maximize space available on rootfs
-        # Why not use https://github.com/easimon/maximize-build-space directly?
-        # The reason is: we want to maximize the space on rootfs, since that's
-        # where the nix store (`/nix/store`) is located. Github action
-        # https://github.com/easimon/maximize-build-space maximizes
-        # the builder space on ${GITHUB_WORKSPACE}, which is not what we need.
-        # Alternatively, we could move the nix store to ${GITHUB_WORKSPACE}
-        # and use https://github.com/easimon/maximize-build-space as such, but
-        # we suspect other tooling (e.g. cachix) would not work well with such
-        # configuration.
-        run: |
-          echo "Available storage before cleanup:"
-          df -h
-          echo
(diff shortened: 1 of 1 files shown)


alextserepov (approved):

.github/workflows/build.yml
```diff
@@ -147,36 +123,24 @@ jobs:
...
+            BUILDER='${{ vars.BUILDER_AARCH }}'
           else
-            echo "Running nix build, with cachix watch-exec"
-            cachix authtoken ${{ secrets.CACHIX_AUTH_TOKEN }}
-            cachix watch-exec -v ghaf-dev -- \
-              nix build .#packages.${{ matrix.arch }}.${{ matrix.target }}
+            echo "::error::Unknown architecture: '${{ matrix.arch }}'"
+            exit 1
           fi
+          NIX_FAST_BUILD_GITREF="1775c732"
```
brianmcgillion, line 140: Do we need to pin the version?

henrirosten: I would prefer that we pin the version. That way, we would be in control when/if we decide to update the `nix-fast-build` version we use, and would not be caught off guard on updates.

brianmcgillion: Sure. I'll try to remember to bump it on every flake lick update, to keep in sync with the dev shell

brianmcgillion (approved):

PR #876, opened 2024-10-30
Desktop: Refactor audio and brightness controls

kajusnau:
## Description of changes

1. **Taskbar Audio Control and LabWC (F1, F2, F3) Hotkeys Now Use Pamixer**
   - The `audio-ctrl` script has been removed, as it was simply a wrapper for existing `pamixer` commands.

2. **Improved Responsiveness of Taskbar Audio and Brightness Controls:**
   - Volume and brightness sliders in the quick settings widget are now more responsive.
   - Volume and brightness popups (displayed near the bottom of the screen) now appear more consistently and react to changes more quickly.

3. **Removed `audio-ctrl` Script**

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature
    - Adjust system volume and display brightness using taskbar sliders and keyboard hotkeys


brianmcgillion (approved):

modules/desktop/graphics/ewwbar.nix
```diff
@@ -356,48 +362,39 @@ let
...
-      }
-
-      mute() {
-          audio-ctrl mut
-      }
-
       get() {
-          volume=$(volume_level)
-          muted=$(is_muted)
+          volume=$(pamixer --get-volume | awk '{print $1 + 0.0}')
```
josa41, line 366: Maybe we could use pamixer argument
`       --get-volume-human
              Get the current volume percentage or the string "muted".
`

modules/desktop/graphics/ewwbar.nix
```diff
@@ -276,29 +278,39 @@ let
...
 
       listen() {
-        inotifywait -m /sys/class/backlight/*/brightness | \
-        while read -r line; do
-            if echo "$line" | grep -q "CLOSE_WRITE"; then
+        prev_brightness=""
+        inotifywait -m -e close_write /sys/class/backlight/*/brightness |
+        while read -r; do
+            get &
+            current_brightness=$(brightnessctl info | grep -oP '(?<=\().+?(?=%)' | awk '{print $1 + 0.0}')
```
josa41, line 292: How many times this loop runs per sec?
It calls external app(s) in a loop, hopefully for not too long time?

modules/desktop/graphics/ewwbar.nix
```diff
@@ -508,10 +505,10 @@ in
                 (sys_slider
                         :header "Volume"
                         :icon {volume.icon}
-                        :icon-onclick "${eww-volume}/bin/eww-volume mute &"
+                        :icon-onclick "PULSE_SERVER=audio-vm:4713 ${pkgs.pamixer}/bin/pamixer --toggle-mute &"
```
josa41, line 508: I think these are duplicate code?
why not use eww-volume script here as before?

modules/desktop/graphics/labwc.config.nix
```diff
@@ -110,13 +109,13 @@ let
         <action name="Execute" command="${pkgs.brightnessctl}/bin/brightnessctl set 5%-" />
       </keybind>
       <keybind key="XF86_AudioRaiseVolume">
-        <action name="Execute" command="${audio-ctrl}/bin/audio-ctrl inc" />
+        <action name="Execute" command="bash -c 'PULSE_SERVER=audio-vm:4713 ${pkgs.pamixer}/bin/pamixer --unmute --increase 5'" />
```
josa41, line 112: If using eww-volume here also is not possible i'd rethink this a bit.
Running bash -c is basically the same as running a script which needs some shell..
Maybe at least make a function for these almost duplicates?

modules/desktop/graphics/ewwbar.nix
```diff
@@ -307,43 +319,37 @@ let
...
+            fi
 
-      is_muted() {
-          audio-ctrl get_mut
+            if ! ${eww} active-windows | grep -q "volume-popup"; then
+              ${eww} open volume-popup
+              ${eww} update volume-popup-visible="true"
+            fi
+            (
+              sleep 2
```
josa41, line 342: Is this really the way to get popups dissapear after timer in eww?
If it is it seems re-usable code to me?

josa41 (commented):
Maybe a bit minor issues..

modules/desktop/graphics/ewwbar.nix
kajusnau: It's quite responsive in use and hasn't caused issues yet.
Volume control (and listening to volume changes) is still less responsive than brightness overall.

modules/desktop/graphics/ewwbar.nix
kajusnau: eww does provide functionality to open a window with a timer **`eww open --duration [time]`**, but no way to reset or cancel the timer once the window is open.
This resulted in jittery behavior, where a user might keep increasing the brightness a few times in a row but the popup would suddenly disappear because the timer runs out and then appear immediately again.

It is reusable code, however tracking the PID of the current timer for that specific window would require storing the PID somewhere externally, which I wanted to avoid for the sake of simplicity and performance.

modules/desktop/graphics/ewwbar.nix
kajusnau: Good point, we can sacrifice a few milliseconds of performance here since it's just a mute switch.

modules/desktop/graphics/ewwbar.nix
kajusnau: We can of course remove the checking if value has changed, but that will cause the brightness popup to appear even if, for example, the brightness is being increased while it's already at 100%

josa41: I'm not an expert in this, i just saw a loop to do some fairly heavy calls for updating which may cause lots of load.
Yes in most cases this is likely to be finished in one loop, not sure if it is possible to make it run for longer period of time and cause lots of resource usage.
In the loop there is get in background process and running same brightnessctl again in the next line.
Would it be possible to combine getting the brightnessctl content in one call and reuse the data?

johannarautanen:
**Tested on Lenovo-X1 (flashed to the SSD)** 

**Working:** 
- Volume and brightness sliders in the quick settings are more responsive when with older version
- Volume and brightness popups now appear more consistently and react to changes more quickly.

- all other ghaf apps can be launched
- ci-test automation cases ok

**Isssue:**

- there are times when the volume or brightness popup isn't visible, but the functionality(volume and brightness) working ok. 
Not easy to reproduced, but when happened when tapping the the volume and brightness buttons and then ie volume popup is not anymore visible, but the F1, F2 and F3 changed the volume. 
And when you tap the brightness F5 and F6, the brightness popup comes visible and after that the volume popup is visible again when changing the volume.

modules/desktop/graphics/ewwbar.nix
kajusnau: In practice, the inside of the while loop gets executed only when a **close_write** event is detected in **/sys/class/backlight/*/brightness**.
This can be checked manually by running **`/nix/store/k5icwhkqvk77zqksi6kld2mjpywql7qg-eww-brightness/bin/eww-brightness listen`** and adjusting display brightness using F5 and F6.

Reusing the data from **brightnessctl** in the **get** call is possible, we could provide it as an argument, but then we'd end up needing an if/else check inside of **get()**.
We would also need to fetch the value before calling **get**, which would cause a minor delay and the overall responsiveness might be affected.

That being said, I'll look into it along with the other changes to be done.

modules/desktop/graphics/ewwbar.nix
kajusnau: I recall trying this a while ago. Just tried again and remembered that the limitation comes from eww  - it requires this value to be of type f64.
In addition, **--get-volume-human** returns "muted" when the volume is muted, which is also not f64 and causes issues when applying it to the slider in eww :sweat_smile:

modules/desktop/graphics/ewwbar.nix
kajusnau: Implemented.
The result is slightly (few ms) better responsiveness.

kajusnau:
> Maybe a bit minor issues..

Updated PR, most of the mentioned issues have been addressed.

modules/desktop/graphics/ewwbar.nix
```diff
@@ -276,29 +278,43 @@ let
...
       }
 
-      set_screen() {
-          brightnessctl set "$1""%" -q
-      }
-
       get() {
-          brightness=$(screen_level)
+          if [ -z "$1" ]; then
+              brightness=$(brightnessctl info | grep -oP '(?<=\().+?(?=%)' | awk '{print $1 + 0.0}')
```
josa41, line 283: Still duplicate code for running the brightnessctl binary.
Maybe have separate internal read() function that only runs the brightnessctl that would "return" the string content for both features.

kajusnau: I'd rather leave this for a more general refactor, if needed.
eww-volume and eww-brightness could both be adjusted to provide listening just for the respective values, rather than an entire JSON object of value, icon etc.

johannarautanen:
Tested on Lenovo-X1 (flashed to the SSD) (Commit: 1950991ce1d4dd59c024725efd4d41ea3301a36b)

Working:

   - Volume and brightness sliders working 
   - Popups appeared when changing volume and brightness
    - all other ghaf apps can be launched
    - ci-test automation cases ok

Notes:
- when the "Terminal"-menu is opened and stayed the screen, the volume and brightness cannot be changed via F1, F2, F3, F5 and F6.  After you really open the Terminal-app or close the Terminal-menu, then hotkeys worked again.
- When you open then the youtube-page via chromium or microsoft teams meeting, then there appeared 3 times volume popups and when you close the page or stop the meeting, again volume popups appeared couple of times.
- Once when "playing" with the "Terminal"-menu and volume popups, then popup stayed on the screen, doesn't go way time being or after lock the screen or closing other apps from the screen. After logout the volume popup disappeared.
Will create separate issues of these notes.

PR #881, opened 2024-11-05
Change to use pulseaudio old tunnelin module

josa41:
There seems to be issues with pulseaudio module-tunnel -new modules The modules cause desync in audio that is not recoverable. The bugs of this in pulseaudio are some 10 years old and seem not to get fixed.

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature
  - [X] List the test steps to verify:

This applies only for audio in pulseaudio tunneling mode.
Requires setting property "ghafAudio.useTunneling = true;"
for some application VM to enable this feature (the default is not use tunneling).


leivos-unikie:
Jon  could you explain more in detail how the desync in audio was noticed and how to test that it has been fixed.

josa41:
We have two ways to provide audio service for appVMs.
The current default way is to let client apps to connect audioVM directly using pulseaudio. This is done by setting environment variable "PULSE_SERVER=audio-vm:4713" to application environment. This makes application to connect directly to pulseaudio over TCP to host "audio-vm" port 4173. This way works fine and it is the default way used by Ghaf at the moment.

The other way for apps to get audio is use local pulseaudio service that tunnels the audio data forward to audiovm. This "backup" way does not need any additional configuration on client apps (as normally pulseaudio is being run on a local system anyway). The problem with this is that there is slight additional pulseaudio audio data buffer on local application VM. Both buffers one on local VM and the other on AudioVM need to stay somewhat in sync. Pulseaudio "new" tunneling module has a few bugs that cause the buffer on our application VM side to grow and there is no way to reduce so more and more delay starts to build up on audio.

To setup a test you need to enable the "backup" audio tunneling mode with setting nix property "ghafAudio.useTunneling = true" on any application VM that you want to test (it does not matter which one). I've usually set the tunneling mode only for bussiness vm so that i can compare the results with chromium vm. The tunneling mode is not enabled for this pull request as we don't want to enable the tunneling unless we really need it for some reason.

I can provide a patch or make a separate branch in my local github that has the tunneling mode enabled for some VM.

Long story short
So while recording audio with microphone the audio recording was being delayed for several seconds. To make it even worse if you record several about 10 second clips of audio. Some audio from the previous recording may end up in to the latter recording. In my oppinion this is due to some audio is being left to buffers along the way. The audio then gets released to the latter recording.

The expected results is that the delay while recording never exceeds 1 second. For this i have to say that normally the delay is in a few milliseconds. In a case if the audio data stream disconnects for some reason the delay to reconnect the audio is 1 second. The disconnect might happen because of network/data transfer error or when switching hardware audio device (connect BT headset for instance).

PR #884, opened 2024-11-06
Add separate audio control channel

josa41:
Add separate audio control channel for audio control data
This allows using unprivileged pulseaudio channel for apps
and channel with more privileged for controlling pulseaudio

## Checklist for things done

- [X] Author has run `make-checks` and it passes

## Instructions for Testing

- [ ] Is this a new feature
No visible change to the user. Controlling audio should work through different TCP port (4714)

### DIFF
--- a/modules/common/services/audio.nix
+++ b/modules/common/services/audio.nix
@@ -24,6 +24,11 @@ in
       default = 4713;
       description = "TCP port used by Pipewire-pulseaudio service";
     };
+    pulseaudioTcpControlPort = mkOption {
+      type = types.int;
+      default = 4714;
+      description = "TCP port used by Pipewire-pulseaudio control";
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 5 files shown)


kajusnau (approved):

leivos-unikie:
Tested on Lenovo-X1
- ci-test-automation passed, performance ok
- didn't notice problems with audio
- audio-vm shows tcp port 4714 LISTENING

I noticed that when I open a youtube video Audio Control shows 2 playback streams for few seconds, then the other stream disappears. What is that other stream? Can it be related to that controlling channel?

josa41:
I've noticed the same, some apps use multiple streams so currently in volume control application all the streams are shown. We have plan to combine the streams from one app under a single volume control in our volume control application but that requires some work. It is basically not possible for us to determine which stream is for which use or why some apps have multiple stream. All we know from the stream is from which VM the stream originates, which application binary name (and PID) uses the stream and possibly but not always the media title name that is being played back.

My guess is that in youtube advertisements are played back using separate stream from the actual media playback. Similar as with teams app all participants and notification sounds are being played back on different streams.

However this issue is not related to this pull request.

brianmcgillion (approved):

PR #892, opened 2024-11-14
Generic initialize of audio devices

josa41:
More generic solution to initialize audio device
Finds default audio device and profile
after going through all profiles resets back to the original default

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature

This fixes a bug with Lenovo X1 audio when booted with 3.5mm headset connected
In the bug wrong audio profile got activated and no audio was heard from either speakers or headset

### DIFF
--- a/modules/common/services/audio.nix
+++ b/modules/common/services/audio.nix
@@ -91,26 +91,76 @@ in
     # Start pipewire on system boot
     systemd.services.pipewire.wantedBy = [ "multi-user.target" ];
 
-    systemd.services."pulseaudio-set-profile" =
+    systemd.services."initialize-audio-profile" =
       let
-        pulse-set-profile = pkgs.writeShellScriptBin "pulseaudio-set-profile" ''
-          ${pkgs.pulseaudio}/bin/pactl set-card-profile "alsa_card.pci-0000_00_07.0-platform-skl_hda_dsp_generic" "HiFi (Headphones, Mic1, Mic2)"
-          ${pkgs.pulseaudio}/bin/pactl set-card-profile "alsa_card.pci-0000_00_07.0-platform-skl_hda_dsp_generic" "HiFi (Mic1, Mic2, Speaker)"
-        '';
+        initialize-audio-profile = pkgs.writeShellApplication {
+          name = "initialize-audio-profile";
+          runtimeInputs = [
+            pkgs.pulseaudio
+            pkgs.jq
+          ];
+          text = ''
+            function setProfile() {
+              if [[ -n "$1" ]] && [[ -n "$2" ]]; then
+                echo "Setting audio device profile: ($1 - $2)"
+                pactl set-card-profile "$1" "$2"
+              fi
+            }
+
+            function findAudioProfiles() {
+              local audio_device_name active_profile profile_list
+              if [[ -n "$1" ]]; then
+                audio_device_name=$(jq --raw-output --argjson index "$1" '.[$index].name' <<< "$pactl_data")
+                active_profile=$(jq --raw-output --argjson index "$1" '.[$index].active_profile ' <<< "$pactl_data")
+                if [[ -n "$audio_device_name" ]]; then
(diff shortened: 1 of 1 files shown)


modules/common/services/audio.nix
```diff
@@ -91,17 +91,70 @@ in
...
-        '';
+        initialize-audio-profile = pkgs.writeShellApplication {
+          name = "initialize-audio-profile";
+          runtimeInputs = [
+            pkgs.pulseaudio
+            pkgs.jq
+          ];
+          text = ''
+            function setProfile() {
+                if [[ $# == 2 ]] && [[ -n "$1" ]] && [[ -n "$2" ]]; then
```
kajusnau, line 104: It should be enough to check that there's exactly two arguments `if [[ $# -eq 2 ]]; then`

modules/common/services/audio.nix
```diff
@@ -91,17 +91,70 @@ in
...
+                    echo "Setting audio device profile: ($1 - $2)"
+                    pactl set-card-profile "$1" "$2"
+                fi
+            }
+
+            function findAudioProfiles() {
+              local audio_device_name active_profile profile_list
+                if [[ $# == 1 ]] && [[ -n "$1" ]]; then
```
kajusnau, line 112: It should be enough to check that there's exactly one argument `if [[ $# -eq 1 ]]; then`

modules/common/services/audio.nix
```diff
@@ -91,17 +91,70 @@ in
...
+                fi
+            done
+          '';
+        };
       in
       {
         enable = true;
-        description = "Force selection of Pulseaudio integrated mic profile";
-        path = [ pulse-set-profile ];
+        description = "Ininitialize default audio device profiles";
```
kajusnau, line 156: Typo

modules/common/services/audio.nix
```diff
@@ -110,7 +163,7 @@ in
           StandardOutput = "journal";
           StandardError = "journal";
```
kajusnau, line 164: Not 100% certain, but doesn't this already default to "journal"? Might be redundant
ref. https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#StandardOutput=

modules/common/services/audio.nix
```diff
@@ -91,17 +91,70 @@ in
...
+            done
+          '';
+        };
       in
       {
         enable = true;
-        description = "Force selection of Pulseaudio integrated mic profile";
-        path = [ pulse-set-profile ];
+        description = "Ininitialize default audio device profiles";
+        path = [ initialize-audio-profile ];
```
kajusnau, line 157: Since this is specified, we probably don't need to call the script using the full nix store path in line 166.
It looks like this line should be removed or line 166 changed to `initialize-audio-profile`?

modules/common/services/audio.nix
josa41: I agree that the checks overlap but i would rather keep the -n checks to catch empty arguments

modules/common/services/audio.nix
josa41: Good point, i removed the path arg

modules/common/services/audio.nix
josa41: It seems you are right. Removed as redundant

kajusnau (approved):

mbssrc (approved):

barnabakos:
1 commit: 67b2bedf0314797c1f4858cdccf7127e4791c842
rebased to current tip of Ghaf main without conflicts: 2cfd015194afebf72c8ff40d21c3e42e94893de0

platform: **Lenovo X1 Carbon; flash script; USB SSD**
- starting apps: ok
- regression / ad-hoc testing: ok

#### works:
- audio heard from headset when booted with 3.5mm audio jack connected

#### notes:
- N/A

#### issues:
- N/A

PR #941, opened 2024-12-20
Update XDG handlers to use GIVC instead of SSH

nesteroff:
## Description of changes

This updates XDG handlers to use GIVC and virtiofs storage instead of SSH and SCP.

For file sharing, this approach uses a shared XDG folder with the following structure: `/shared/xdg/mime/appvm`. Within the XDG folder there is a separate folder for each MIME type. These folders are passed to the VMs that can handle those types as virtiofs devices. Currently, `zathura-vm` is used for handling both PDFs and images. Inside each MIME type folder there are subfolders for each appvm. These subfolders are passed to their respective appvms. This way appvms can share files with `zathura-vm` by simply copying them into they folder.

For example, `chrome-vm` has the `/share/xdg/pdf/chrome-vm` folder mounted as `/run/xdg/pdf` in the guest, but it can’t access the entire mime folder or other appvms files because only `/share/xdg/pdf/chrome-vm` is passed as virtiofs device. However, `zathura-vm` has access to the entire MIME folder, which includes subfolders for all appvms.

The advantage of this approach is that files are shared through simple copying with no network data transfers and appvms cannot access files from other appvms since they are isolated at the virtiofs device level.

## Checklist for things done

## Instructions for Testing

  Lenovo X1
- [x] Is this a new feature
  This doesn't introduce new functionality. It just changes the way XDG handlers work to remove SSH dependency.
  On the Lenovo X1, open PDF and image files (JPG, PNG) either from a USB drive or those downloaded from Chrome. It should still work as expected. Compared to the previous implementation files may open more quickly but this difference is unlikely to be noticeable.


modules/common/security/apparmor/profiles/google-chrome.nix
gngram, line 1: I think using wild card in package path is not safe:

/nix/store/*xdgopenfile/bin/xdgopenfile

gngram (changes requested):

modules/common/security/apparmor/profiles/google-chrome.nix
nesteroff: Thanks for looking at this. The wildcard trick was used in the draft just to quickly make it work with the new module configuration. I guess it would be nice to add more flexibility to the apparmor module in the future so that we don't need to expose the path to the script as a module option. Also, there might be use cases when chrome doesn't need xdg handlers and when xdg handlers are needed for other software. So maybe they need a separate profile. Anyway, it's not within the scope of this PR, so I reverted it to how it was before.

milva-unikie:
The failure in the pre-merge-pipeline Lenovo-X1 tests was likely due to a loose robot finger. No further action is needed for this PR because of that.

milva-unikie:
**Tested on Lenovo-X1** (first `nixos-rebuild switch`, then confirmed same result with new build)

X1 does not boot properly, gets stuck at boot logs. Host, net-vm and a few other vm:s are responding normally.

```
[ghaf@ghaf-host:~]$ microvm -l
admin-vm: current(nixos-system-admin-vm-25.05pre-git)
audio-vm: current(nixos-system-audio-vm-25.05pre-git)
business-vm: current(nixos-system-business-vm-25.05pre-git), not booted: systemctl start microvm@business-vm.service
chrome-vm: current(nixos-system-chrome-vm-25.05pre-git), not booted: systemctl start microvm@chrome-vm.service
comms-vm: current(nixos-system-comms-vm-25.05pre-git), not booted: systemctl start microvm@comms-vm.service
gala-vm: current(nixos-system-gala-vm-25.05pre-git)
gui-vm: current(nixos-system-gui-vm-25.05pre-git), not booted: systemctl start microvm@gui-vm.service
net-vm: current(nixos-system-net-vm-25.05pre-git)
zathura-vm: current(nixos-system-zathura-vm-25.05pre-git), not booted: systemctl start microvm@zathura-vm.service
```

mbssrc:
> **Tested on Lenovo-X1** (first `nixos-rebuild switch`, then confirmed same result with new build)
> 
> X1 does not boot properly, gets stuck at boot logs. Host, net-vm and a few other vm:s are responding normally.
This may be caused by host shared folders missing

nesteroff:
> This may be caused by host shared folders missing

Exactly. In previous versions of microvm.nix, there was 'mkdir -p "$SOURCE"' in the microvm-virtiofsd service but after the latest bump, it no longer creates the source folder. I didn't notice this because I use nixos-rebuild switch all the time. I suppose I can just add all the folders to microvm-host.nix.

brianmcgillion (approved):

nesteroff:
> **Tested on Lenovo-X1** (first `nixos-rebuild switch`, then confirmed same result with new build)
> 
> X1 does not boot properly, gets stuck at boot logs. Host, net-vm and a few other vm:s are responding normally.
> 

Thanks for testing. I've made an update to create shared directories on the host with correct permissions which should fix the issue.

milva-unikie:

- PDF, JPG and PNG files can be opened
- Test-automation passes
- Fails in pre-merge-pipeline are not related to this PR

PR #947, opened 2025-01-13
Parallel nix evaluation in github actions

henrirosten:
## Description of changes

Do not merge this, see: https://github.com/tiiuae/ghaf/pull/947#issuecomment-2589990432

Various updates to Ghaf github actions, most importantly:
- Move the nix evaluation step to its own workflow `eval.yml`, which runs the nix evaluation in parallel, splitting the evaluation to 8 subsets, each of which are executed on a separate github-hosted runner. In testing, this improves the eval action runtime from ~35 minutes down to ~7 minutes. 
- `eval.sh` allows filtering the evaluated targets via the `-t` [option](https://github.com/tiiuae/ghaf/blob/70a3e72f41c2571000b3716948a92539bc341420/.github/eval.sh#L31): the initial version in this PR will evaluate all Ghaf flake outputs that match the default filter '`packages.*`'. 
- Start running the following actions on push or merge to main (and not only in PR): `eval`, `build`, and `check` to make it more apparent if one of those workflows would start failing after merge to main.

## Checklist for things done

## Instructions for Testing

Tested in fork: https://github.com/henrirosten/ghaf/actions/workflows/eval.yml

- [ ] Is this a new feature


joinemm (approved):
LGTM

henrirosten:
Back to draft: let's not merge this yet, I'll add one more change to this PR in a moment.

henrirosten:
- Kept the manual rebase step in `build.yml`, which we need to do since `github.event.pull_request.merge` isn't available for `pull_request_target` triggered events
- Simplify the `build-yml-check` step

.github/eval.sh
```diff
@@ -0,0 +1,182 @@
...
+    done < "$TMPDIR/attrs.txt"
+    printf "}\n" >>"$TMPDIR/eval.nix"
+    echo "[+] Evaluating nix expression:"
+    cat "$TMPDIR/eval.nix"
+    # Evaluate with nix-eval-jobs
+    gcroot="$TMPDIR/gcroot"
+    nix-eval-jobs \
+      --accept-flake-config \
+      --gc-roots-dir "$gcroot" \
+      --force-recurse \
```
Mic92, line 165:
```suggestion
      --force-recurse \
      --check-cache-status \
```

.github/eval.sh
```diff
@@ -0,0 +1,182 @@
...
+        print_err "No flake outputs match filter: '$filter'"; exit 1
+    fi
+    # Remove the '.name' suffix
+    sed -i "s/.name//" "$TMPDIR/outs.txt"
+    # Read the attribute set names
+    grep -oP ".*(?=\.)" "$TMPDIR/outs.txt" | sort | uniq >"$TMPDIR/attrs.txt"
+    # Generate eval expression on the fly
+    printf '%s\n' \
+      "let" \
+      "  flake = builtins.getFlake ("git+file://" + toString ./.);"\
```
Mic92, line 138: Mhm. This will make evaluation impure. I would instead write a new flake.nix into a directory that uses git+file and has flake filtered flake outputs.

.github/eval.sh
Mic92: The actual build has not been implemented yet?

Mic92:
If you want to try out something quickly you can also go for: 

https://github.com/nix-community/nix-github-actions

Don't know how this would scale for this project because we don't quite a few nixos closures here. https://github.com/DeterminateSystems/magic-nix-cache may work but don't know if we would run into too many rate limits here.

.github/eval.sh
henrirosten: Do you mean pre-merge ghaf builds?
They have been runnnig for quite some time, see: https://github.com/tiiuae/ghaf/actions/workflows/build.yml

.github/eval.sh
henrirosten: The eval done here is only to confirm that the evaluation still passes with the changes from the triggering PR. We don't intent to store the evaluation results anywhere.

On building, we run the builds on the r[emote builders with your nix-fast-build](https://github.com/tiiuae/ghaf/blob/489db31b80bc47e114b4a66a1664c095df4f12d4/.github/workflows/build.yml#L141-L146).

Mic92: Ok. I see. In that case impure evaluation wouldn't be a big deal because nix-fast-build would still catch it.

.github/eval.sh
```diff
@@ -0,0 +1,182 @@
...
+
+################################################################################
+
+evaluate () {
+    job="$1"
+    max_jobs="$2"
+    filter="$3"
+    echo "[+] Using filter: '$filter'"
+    # Output all flake output names
+    nix flake show --all-systems --json |\
```
Mic92, line 124: If this is just about evaluation, this will already evaluate everything. You shouldn't need nix-eval-jobs afterwards.

Mic92: It will be also faster because nix-eval-jobs has to serialize drv files to disk.

henrirosten:
Based on https://github.com/tiiuae/ghaf/pull/947#discussion_r1914760549, we should not merge this.
Instead, I'll push a series of PRs to implement the changes from this PR, keeping @Mic92's comments in mind.

henrirosten:
Closing in favor of https://github.com/tiiuae/ghaf/pull/948 and https://github.com/tiiuae/ghaf/pull/949

PR #975, opened 2025-01-27
Restrict systemd services

gngram:
## Description of changes

Apply more restrictions on following Systemd services:
    - dnsmasq
    - firewall
    - enable-ksm
    - generate-shutdown-ramfs
    - kmod-static-node
    - NetworkManager-dispatcher
    - logrotate-checkconf
    - logrotate
    - nscd
    - seatd
    - systemd-journal-catalog-update
    - systemd-journal-flush
    - systemd-random-seed
    - systemd-remount-fs
    - systemd-rfkill
    - systemd-tmpfiles-clean
    - systemd-tmpfiles-setup-dev
    - systemd-tmpfiles-setup
    - systemd-udev-trigger
    - NetworkManager

Removed some disabled services
Cleanup of `dbus` and `greetd`.

## Checklist for things done

## Instructions for Testing

    - All
- [ ] Is this a new feature
      - Test system functionality

### DIFF
--- a/modules/common/systemd/hardened-configs/NetworkManager-dispatcher.nix
+++ b/modules/common/systemd/hardened-configs/NetworkManager-dispatcher.nix
@@ -2,150 +2,64 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 {
-  ##############
-  # Networking #
-  ##############
-
-  # PrivateNetwork=true;
-  # IPAccounting=yes
-  IPAddressDeny = "any";
-  RestrictAddressFamilies = [
-    "AF_PACKET"
-    "AF_NETLINK"
-    "AF_UNIX"
-    "AF_INET"
-    "AF_INET6"
-  ];
-
-  ###############
-  # File system #
-  ###############
-
-  # ProtectHome=true;
-  # ProtectSystem="full";
+  PrivateNetwork = false;
+  IPAccounting = true;
+  ProtectHome = true;
+  ProtectSystem = "strict";
   ProtectProc = "noaccess";
-  # ReadWritePaths=[ "/etc"];
   PrivateTmp = true;
-
-  # Not applicable for the service runs as root
-  # PrivateMounts=true;
-  # ProcSubset="all";
-
-  ###################
-  # User separation #
-  ###################
-
-  # Not applicable for the service runs as root
-  PrivateUsers = true;
-  # DynamicUser=true;
-
-  ###########
-  # Devices #
-  ###########
-
-  # PrivateDevices=false;
-  # DeviceAllow=/dev/null
-
-  ##########
-  # Kernel #
-  ##########
-
+  PrivateMounts = true;
+  ProcSubset = "pid";
+  PrivateDevices = true;
   ProtectKernelTunables = true;
   ProtectKernelModules = true;
   ProtectKernelLogs = true;
-
-  ########
-  # Misc #
-  ########
-
   Delegate = false;
-  # KeyringMode="private";
-  # NoNewPrivileges=true;
+  KeyringMode = "private";
+  NoNewPrivileges = true;
   UMask = 77;
   ProtectHostname = true;
   ProtectClock = true;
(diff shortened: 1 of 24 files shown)


modules/common/systemd/hardened-configs/common/dbus.nix
```diff
@@ -104,50 +44,8 @@
...
-    # "~CAP_SYS_TIME
   ];
 
-  ################
-  # System calls #
-  ################
-
   SystemCallFilter = [
     "@system-service"
     "~@chown"
```
mbssrc, line 51: I don't think mixing white- and blacklisting works?

modules/common/systemd/hardened-configs/common/greetd.nix
```diff
@@ -65,9 +45,6 @@
     "CAP_DAC_READ_SEARCH"
   ];
 
-  ################
-  # System calls #
-  ################
   SystemCallFilter = [
```
mbssrc, line 48: also here

milva-unikie:
Have the comments been addressed, is this ready for testing? A rebase would be nice too.

mbssrc (approved):

leivos-unikie:
Tested on Lenovo-X1
- ci-test-automation bat and gui tests pass
- no issues found in manual exploratory testing
- there are some significant drops in performance metrics:

**ghaf-host**
multi-thread cpu test: over 15% decrease in speed
multi-thread memory read test: over 25% decrease in speed

**VMs**
sysbench tests in VMs: around 50% decrease in cpu and memory tests in net-vm

In addition some minor drops in few other performance tests.

Noticed that fileio read test result is very low but it is the same with ghaf mainline, not caused by this PR.

leivos-unikie:
On Orin-AGX bat tests pass except timesync test.

Time synchronization test fails in the beginning at `Check that time is correct` because ghaf-host does not update its time automatically via NTP / timecyncd.service although the service is running and internet connection available.

Looking for comparison the latest timesync test in prod main pipeline: journalctl logs show that timesync happens 2.5 minutes after boot. 

In this PR timesync does not happen even after waiting 10 min. Time sync starts to work only after manually restarting the service.
```
[ghaf@ghaf-host:~]$ timedatectl -a
               Local time: Thu 1970-01-01 00:02:10 UTC
           Universal time: Thu 1970-01-01 00:02:10 UTC
                 RTC time: Thu 1970-01-01 00:02:10
                Time zone: UTC (UTC, +0000)
System clock synchronized: no
              NTP service: active
          RTC in local TZ: no

[ghaf@ghaf-host:~]$ systemctl restart systemd-timesyncd.service
==== AUTHENTICATING FOR org.freedesktop.systemd1.manage-units ====
Authentication is required to restart 'systemd-timesyncd.service'.
Authenticating as: ghaf
Password: 
==== AUTHENTICATION COMPLETE ====

[ghaf@ghaf-host:~]$ timedatectl -a
               Local time: Thu 2024-12-19 21:27:54 UTC
           Universal time: Thu 2024-12-19 21:27:54 UTC
                 RTC time: Thu 1970-01-01 00:02:25
                Time zone: UTC (UTC, +0000)
System clock synchronized: no
              NTP service: active
          RTC in local TZ: no
```

leivos-unikie:
I didn't tag those performance drops on X1 as a bug but they should be discussed if it is acceptable to ruin performance so much. FYI @brianmcgillion

gngram:
@leivos-unikie 
Lenovo X1:
There was issue with audio-vm, it was not coming up properly, it kept on rebooting. This definitely had some performance impact, I fixed it. I tried to capture the performance number using automation test suits but the numbers doesn't conclude to any result. In multiple runs mainline has varying performance and same is the case with this PR also. Variation is in the same range for both the cases. Can you please verify this at you end?

Nvidia Orin AGX:
I can see time sync issue on main line also. When we flash a fresh image, it doesn't sync. After reboot only it shows the proper time. Same is the case with this PR also.

PR #986, opened 2025-01-31
add audio refresh to the dell laptops

brianmcgillion:
bring the dell laptops inline with the lenovo.

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature

### DIFF
--- a/modules/reference/hardware/dell-latitude/definitions/dell-latitude-7230.nix
+++ b/modules/reference/hardware/dell-latitude/definitions/dell-latitude-7230.nix
@@ -106,6 +106,17 @@
 
   # Audio device for passthrough to audiovm
   audio = {
+    # Force a PCI device reset to the audio device
+    # This is to get the pci hardware device to the default state at shutdown
+    removePciDevice = "0000:00:1f.3";
+
+    # Force a PCI device rescan after resetting a device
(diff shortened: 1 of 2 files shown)


modules/reference/hardware/dell-latitude/definitions/dell-latitude-7230.nix
```diff
@@ -106,6 +106,17 @@
 
   # Audio device for passthrough to audiovm
   audio = {
+    # Force a PCI device reset to the audio device
+    # This is to get the pci hardware device to the default state at shutdown
+    removePciDevice = "0000:00:1f.3";
+
+    # Force a PCI device rescan after resetting a device
+    # to refind the device on host after reset
+    rescanPciDevice = "0000:00:1f.0";
```
josa41, line 115: I'd test first without removePciDevice and rescanPciDevice options, those may not be always needed. With Lenovo X1 the audio device does not reset correctly when it is being passed through and laptop or audiovm shuts down or reboots.
If these are needed for Dell also it seems correct to me.

brianmcgillion:
@vunnyso could you also verify if this is needed.

brianmcgillion:
dont believe that this was required.

PR #1005, opened 2025-02-10
disko: use ext4 and btrfs for debug builds

humaidq-tii:
## Description of changes

This is the first phase to an image based disk installation for Ghaf. The disko image would only be used for debug systems in the future.

This PR removes the A/B partitioning scheme layout with a more simplified one that would match the image based Ghaf system.

- Only 1GB VM size needed instead of 18GB+
- ext4 will be used for the root
- btrfs will be used for the persistence partition (which includes storagevm)
- Removed other unused partitions
- Performance improvements expected (not measured)

A/B Partitioning scheme will be added in the next PR.

## Checklist for things done

- [x] Change requires full re-installation
- [ ] Change can be updated with `nixos-rebuild ... switch`: NO

## Instructions for Testing

Fresh install needed. Build time should be faster.

- [x] List all targets that this applies to: lenovo, dell targets
- [ ] Is this a new feature: no
  - [ ] List the test steps to verify: Install system, persistence should work, system should be stable, persist partition in host should use full space.
- [x] If it is an improvement how does it impact existing functionality? No noticeable impact on user functionality.


brianmcgillion (approved):

johannarautanen:
**Just in case check the native Orin AGX and NX with commit 56294944401933c98fce350af4e5121c01201851**
Working:
- ghaf-screen visible in external display
- applications launched
- automation tests ok
- tested manual tests ok

milva-unikie:
**Tested on Lenovo-X1** (installation with installer & new debug image)

More thorough check with 5629494, but also checked with 80f2bcc.

Issues:
- The maximum size of Shares-folder is only 7.5 G. That is where user is supposed to store all their files. When Shares became full I noticed strange issues including system freezes.

Working:
- Everything else seems to be working correctly.
- Test-automation has one memory related test case that we will need to update. Other than that all tests pass.

Question:
- Is there a way to ensure that Ghaf boots from SSD when there is also Ghaf installed in the internal memory? It seems to randomly decide where to mount the persistent storage from. Sometimes it works correctly and sometimes not.
```
[ghaf@ghaf-host:~]$ lsblk
NAME                                            MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINTS
sda                                               8:0    0 476.9G  0 disk  
├─sda1                                            8:1    0     1M  0 part  
├─sda2                                            8:2    0   500M  0 part  /boot
├─sda3                                            8:3    0    12G  0 part  
│ └─dev-disk-byx2dpartlabel-diskx2ddisk1x2dswap 254:0    0    12G  0 crypt [SWAP]
├─sda4                                            8:4    0    40G  0 part  /etc/shadow
│                                                                          /etc/passwd
│                                                                          /etc/group
│                                                                          /nix/store
│                                                                          /
└─sda5                                            8:5    0   7.5G  0 part  
nvme0n1                                         259:0    0 953.9G  0 disk  
├─nvme0n1p1                                     259:1    0     1M  0 part  
├─nvme0n1p2                                     259:2    0   500M  0 part  
├─nvme0n1p3                                     259:3    0    12G  0 part  
├─nvme0n1p4                                     259:4    0    40G  0 part  
└─nvme0n1p5                                     259:5    0 901.4G  0 part  /persist
```

modules/givc/host.nix
```diff
@@ -52,7 +52,7 @@ in
       adminTlsName = config.ghaf.givc.adminConfig.name;
       adminAddresses = config.ghaf.givc.adminConfig.addresses;
       generatorHostName = hostName;
-      storagePath = "/storagevm/givc";
+      storagePath = "/persist/storagevm/givc";
```
mbssrc, line 55: The `storagevm` in the path `/persist/storagevm` seems to be somewhat unnecessary?

mbssrc: I was generally referring to having the `storagevm` path instead of just `/persist`, unless we want to indicate some differences for types of vm storage? A minor style question

mbssrc: So this is a preparation for the future where the root partition is not ext4 anymore? I was mostly wondering because at the moment there is no usage of `/persist` besides `/persist/storagevm`.

mbssrc (approved):

milva-unikie:
**Tested on Lenovo-X1** (installation with installer & new debug image)

- Shares is now expanded correctly.
- Test-automation needs one change, [PR](https://github.com/tiiuae/ci-test-automation/pull/264) for that is open and should be merged right before this PR.



> Ghaf currently doesn't support dual booting, so there might be unexpected behaviour when loading the persistence partition. For testing it is better to boot from either storage, while the other is erased. If booting external SSD, internal storage shouldn't contain an existing Ghaf install.

Currently you can work around this by renaming `zfspool` to eg. `zfspool2` with one of the installations. It makes testing much easier and faster. 

I tried renaming `persist` since that was the one mounted incorrectly, but it did not fix the issue. If a solution could be found for this in the near future, I would very much appreciate it.

brianmcgillion (approved):

PR #1013, opened 2025-02-13
VPN: Add wireguard-gui service

riskuuse:
Wireguard-gui is a user friendly WireGuard VPN configuration tool. The user can create WireGuard VPN interfaces and turn them up/down. The business-vm and the chrome-vm are used as an example usecase environments.

Normally setting WireGuard requires sudo rights from user. In this case we add polkit rule to allow wireguard-gui execution without sudo and user authentication.

WireGuard VPN configurations are stored to /etc/wireguard/ directory and the storage-vm is utilized to make that directory persistent.

Usage of listOf instead of attrsOf doesn't allows us to modify application virtual machines configuration externally, which may be crucial for certain use cases like wireguard-gui nixos module where we want to be able to inject wireguard-gui application to a certain list of appvms. Transformation from list to attrs and back to list doesn't preserve order. So we add list element index in the attribute value to preserve it. It is a dirty hack that could remain until better implementation of virtual machines module as well as hosts module is introduced.

List of wireguard-gui enabled VMs is written to a text file.

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -510,7 +510,8 @@
         "nixos-hardware": "nixos-hardware",
         "nixpkgs": "nixpkgs",
         "systems": "systems",
-        "treefmt-nix": "treefmt-nix"
+        "treefmt-nix": "treefmt-nix",
+        "wireguard-gui": "wireguard-gui"
       }
     },
     "rust-overlay": {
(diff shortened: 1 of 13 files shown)


enesoztrk:
work in progress https://github.com/tiiuae/ghaf/pull/1099

PR #1022, opened 2025-02-18
Desktop: Fix desktop services not starting after wlr-randr v0.5.0

kajusnau:
## Description of changes

1. **Fixed Launcher and Taskbar Services Not Starting After `wlr-randr` v0.5.0:**  
   - `wlr-randr` v0.5.0 rejects standalone CLI arguments, preventing ghaf-launcher and ewwbar services from starting. 
   - Forced taskbar and launcher services to use Bash as an interpreter in `Exec` commands to ensure proper execution when redirecting output.
   - General note for devs: Systemd units **do not** use shell as an interpreter for `Exec` commands.
     It's therefore recommended to always call a script with no arguments nor options, or use `bash -c` (as shown in this PR) to interpret commands properly.

## Checklist for things done

- [x] Summary of the proposed changes in the PR description
- [x] More detailed description in the commit message(s)
- [x] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [x] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] PR linked to architecture documentation and requirement(s) (ticket id)
- [ ] Test procedure described (or includes tests). Select one or more:
  - [ ] Tested on Lenovo X1 `x86_64`
  - [ ] Tested on Jetson Orin NX or AGX `aarch64`
  - [ ] Tested on Polarfire `riscv64`
- [ ] Author has run `make-checks` and it passes
- [ ] All automatic Github Action checks pass - see [actions](https://github.com/tiiuae/ghaf/actions)
- [x] Author has added reviewers and removed PR draft status
- [ ] Change requires full re-installation
- [x] Change can be updated with `nixos-rebuild ... switch`

## Instructions for Testing

- [ ] List all targets that this applies to:
- [ ] Is this a new feature
  - [x] List the test steps to verify:
    - Verify fix works as expected, taskbar and ghaf launcher work at boot.
    - Verify previously introduced fix for log flooding in headless mode still works and is effective.
- [ ] If it is an improvement how does it impact existing functionality?

PR #1021, opened 2025-02-17
Reduce journal log flood in headless mode

kajusnau:
## Description of changes

1. **Prevent `nwg-drawer` and `eww` services from starting in headless mode:**  
   - Added an `ExecCondition` check using `wlr-randr` to detect headless mode.  
   - If headless mode is detected, the service is not started, preventing repeated restart attempts from filling the logs.  
   - Explicitly added the starting and stopping of `ghaf-launcher` and `ewwbar` services in the `display-event-trigger` script.  

## Checklist for things done

- [x] Summary of the proposed changes in the PR description
- [x] More detailed description in the commit message(s)
- [x] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [x] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] PR linked to architecture documentation and requirement(s) (ticket id)
- [ ] Test procedure described (or includes tests). Select one or more:
  - [ ] Tested on Lenovo X1 `x86_64`
  - [ ] Tested on Jetson Orin NX or AGX `aarch64`
  - [ ] Tested on Polarfire `riscv64`
- [ ] Author has run `make-checks` and it passes
- [ ] All automatic Github Action checks pass - see [actions](https://github.com/tiiuae/ghaf/actions)
- [x] Author has added reviewers and removed PR draft status
- [ ] Change requires full re-installation
- [x] Change can be updated with `nixos-rebuild ... switch`

## Instructions for Testing

- [ ] List all targets that this applies to:
- [ ] Is this a new feature
  - [ ] List the test steps to verify:
- [ ] If it is an improvement how does it impact existing functionality?

PR #1029, opened 2025-02-21
Fix journal log flood in headless mode

kajusnau:
## Description of changes

**1. Follow Up Fix To Reduce Log Flooding In Headless Mode:** 
   - Log flooding was previously fixed in #1021 but broke again after #1022.
   - This PR fixes log flooding in headless mode caused by these services: `ewwbar`, `audio-control`, and `ghaf-launcher`.

## Checklist for things done

## Instructions for Testing

- [ ] Is this a new feature


brianmcgillion (approved):

milva-unikie:
Lenovo-X1:
- The Lenovo X1 failed to log out via the GUI in the pre-merge pipeline.
- I reran the test locally, and it passed -> most likely there was a random glitch.
- Did a quick sanity check with X1, everything seems to be in order.

Orins:
- Based on pre-merge pipeline results, the log flooding on Orins has been fixed.
- It would probably be a good idea to also boot Orins with a display and check that they work properly.

kajusnau:
> First tried with Orin AGX wasn't able to reproduced the issue, then tried with Orin NX and doing all kinds of testing launching application and check the power control... if i remember right, use doing logout and external display didn't "awake" and serial connection didn't work either. There have been earlier different kind issue with Logout, but attached the log file here just in case. Need to restart the NX to be able to get the logs: [logout_display_stayed_black.txt](https://github.com/user-attachments/files/18937278/logout_display_stayed_black.txt)

Went through the logs, I wasn't able to find the cause of the black screen issue as described. 
The log is quite lengthy and the timestamps are jumbled so identifying where and what is going wrong is not easy.

I did find that some services attempt to start but fail due to various dependencies not being available on Nvidia targets. The services:
- blueman-applet
- audio-control
These should probably be disabled when GIVC is not available, but this goes a bit beyond the scope of this PR.

Also noticed that nm-applet service also fails to start with no display connected, so that will be disabled.

kajusnau:
Update:
- Disabled **nm-applet** service in headless mode
- Removed some references to a widget that's no longer in use in `ewwbar`
- Removed the **ExecStopPost** unsubscribe call from ewwbar service config - this call always fails due to audio-control service already being stopped via the relationship between `ewwbar.service` and `audio-control.service`

johannarautanen:
**Checked with native the Orin AGX and NX, build used from USB memory stick**
Working:
- tested manual test cases ok
- automation test cases ok
- no black screen screens seen during testing

PR #1041, opened 2025-02-28
Fix file manager refresh issue

nesteroff:
## Description of changes

This is a workaround for SSRCSP-5894. The problem occurs when the “Unsafe chrome-vm share” is open in the file manager and the user downloads a file in Chrome and saves it there. In this case the file manager does not display the downloaded file. It is very confusing since it looks as if the file has not been downloaded or has been saved to the wrong location.

The reason for this behavior is that the file manager on Linux uses the inotify API to watch for file changes. However, it turns out that virtiofs doesn’t propagate inotify events to the guest. There was an [attempt](https://lore.kernel.org/all/CAOQ4uxieK3KpY7pf0YTKcrNHW7rnTATTDZdK9L4Mqy32cDwV8w@mail.gmail.com/T/) to add support for this but as far as I understand it has not been merged into the mainline.

As a workaround, this PR adds a small script that monitors inotify events for the shared folder on the host and sends them to the guivm over vsock. In the guivm, when the script receives a notification from the host it rewrites the modification time of the directory to the same value triggering an inotify event which causes the file manager to refresh the folder contents.

As an alternative implementation we could try using polling in the guivm instead of passing inotify events from the host. Many file managers seem to use the GFileMonitor class from glib which relies on inotify on Linux but it can be updated to use polling for virtiofs folders as it already has something similar for NFS. However, polling has its drawbacks and modifying glib or implementing a plugin for it might be quite complex.

## Checklist for things done

## Instructions for Testing

  Lenovo X1
- [x] Is this a new feature
    - Open "Unsafe chrome-vm share" folder in the file manager
    - Download a file from Chrome to that directory
    - Make sure the list of files refreshes automatically and the new file is in the list
  Fixes SSRCSP-5894

### DIFF
--- a/modules/microvm/virtualization/microvm/microvm-host.nix
+++ b/modules/microvm/virtualization/microvm/microvm-host.nix
@@ -25,7 +25,7 @@ let
       config.ghaf.hardware.definition.audio.rescanPciDevice
     else
       config.ghaf.hardware.definition.audio.removePciDevice;
-
+  vinotify = pkgs.callPackage ../../../../packages/vinotify { };
 in
 {
   imports = [
@@ -46,6 +46,10 @@ in
         '';
         type = types.listOf types.str;
         default = [ ];
+      };
+
+      inotifyPassthrough = mkEnableOption "inotify passthrough" // {
+        default = true;
       };
     };
   };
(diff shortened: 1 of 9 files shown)


milva-unikie:
Tested on Lenovo-X1 (`nixos-rebuild switch`)
- Works as described in Testing instructions
- Test-automation passes

Notes:
- Orin-NX failed shutdown in pre-merge-pipeline. That issue is not caused by this PR.

PR #1074, opened 2025-03-18
lenovo-x1-gen11-hardening: build image with dm-verity

humaidq-tii:
## Description of changes

This pull request builds upon #1005.

This pull request adds a new partitioning scheme in a new `lenovo-x1-gen11-hardening` target, where it would use the new image-based disk partitioning scheme. All other targets aren't yet updated, as this partitioning scheme is still experimental and to eventually be used for the release targets. For testing we enable it for the extras debug target too, as the release variant might not be functional. The goal of this target is also to be a testing ground for newer hardening techniques.

The new partitioning scheme is similar to what is introduced in #1005, but it uses EROFS for the verity tree and root partitions. This is a read-only file system that will make the nix-store completely immutable, allowing us to verify that the system image hasn't been tampered with. EROFS is a modern and fast read-only filesystem, which would also help avoid unintentional tampering of the system.

Some refactoring has been done to allow multiple partitioning schemes without them affecting each other.

- Enables dm-verity.
  - Set default to kernel panic on any corrupted block.
- Added new image-based partitioning scheme.
  - Image generated by systemd-repart with dm-verity roothash included.
  - Uses EROFS for the verity tree and root (nix-store) partition.
- Added preliminary boilerplate for systemd-sysupdate (OTA system updates).
  - The code will require more work in order for it to work especially with the boot loader.
- Importing disko module now assumes that it must be enabled.
  - You cannot import disko partitioning scheme without using it, same with the new image-based partitioning scheme.
  - This is as certain options that are imported/inherited from disko (upstream) conflicts with repart NixOS module.
- Made some code generic (such as the `laptop-configuration-builder.nix` and the ghaf installer script).
- No disk encryption yet, also images aren't signed with secure boot keys.
- Enable QR code for kernel panics for all x86_64, this allows us to to see when dm-verity fails.
- Remove disk hardware definition as it isn't used for both dm-verity (repart) and debug (disko) builds.

## Checklist for things done

- [x] Change requires full re-installation
- [ ] Change can be updated with `nixos-rebuild ... switch`: **NO**

## Instructions for Testing

  - [ ] AGX
  - [ ] NX
  - [x] x86_64: `lenovo-x1-gen11-hardening` only
- [x] Is this a new feature: Yes, as it adds integrity checking which is also a security feature
    - Build `.#lenovo-x1-gen11-hardening-debug-installer` (do not test release).
    - Installer works properly.
    - Run `sudo veritysetup status root` on ghaf-host, should show status verified.
    - Open file manager, the Shares folder should be around 940GB or larger, indicating that the data partition expanded to fill disk after boot.
    - Build a normal debug build (not the extras). Should work as normal (specifically the installer script and boot).
  - Yes, nixos-rebuild will no longer work on the extras targets.


modules/partitioning/flake-module.nix
```diff
@@ -6,7 +6,14 @@
     disko-debug-partition.imports = [
       inputs.disko.nixosModules.disko
       ./disko-debug-partition.nix
-      ./disko-postboot.nix
+      ./btrfs-postboot.nix
+    ];
+    verity-release-partition.imports = [
+      inputs.disko.nixosModules.disko
```
brianmcgillion, line 12: did you mean to include disko here in the verity release?

milva-unikie:
### Issues
**lenovo-x1-extras-debug** (Should this image work?)
- Does not boot to the user account creation screen. The Ghaf logo is shown with Lenovo logo during boot but after that the screen goes black.
- The image is quite large (17.8 GB). Could it be compressed?
- Flash script does not support uncompressed .raw images.
```
[milla@nixos:~/Code/ghaf]$ sudo ./packages/pkgs-by-name/flash-script/flash.sh -d /dev/sdc -i result/ghaf_0.0.1.raw 
Found result/ghaf_0.0.1.raw...
Unsupported file format
```

### Working
**lenovo-x1-extras-debug-installer**
- Installer works
- `sudo veritysetup status root` shows `status: verified`
- Shares folder data partition has expanded
- Test-automation mostly passes, the few failures are not related to this PR (see Notes)

**lenovo-x1-carbon-gen11-debug-installer**
- Installer works
- Boots after installation
- Test-automation passes

**lenovo-x1-carbon-gen11-debug**
- Boots
- Test-automation passes

### Notes
- There seem to be issues with network passthrough in the extras image. At least chrome-vm and comms-vm do not have internet connection even thought the laptop is connected to both wifi and ethernet. This issue also exists in main so I will create a separate bug about it.

brianmcgillion (approved):

modules/common/systemd/base.nix
```diff
@@ -59,6 +59,7 @@ let
         inherit (cfg) withUkify;
         withUserDb = cfg.withHomed;
         withUtmp = cfg.withJournal || cfg.withAudit;
+        withSysupdate = true;
```
mbssrc, line 62: Can we make this an option? This and the other flags are only necessary on host, so we can move this to the `ghaf.systemd` config in `microvm-host.nix`

modules/hardware/x86_64-generic/x86_64-linux.nix
```diff
@@ -23,8 +23,11 @@ in
     hardware.enableAllFirmware = true;
 
     boot = {
-      # Enable normal Linux console on the display
-      kernelParams = [ "console=tty0" ];
+      # Enable normal Linux console on the display, and QR code kernel panic
+      kernelParams = [
+        "console=tty0"
+        "drm.panic_screen=qr_code"
```
mbssrc, line 29: Should also be guarded with verity flag

modules/partitioning/btrfs-postboot.nix
```diff
@@ -47,9 +45,13 @@ let
       parted -s -a opt "$PARENT_DISK" "resizepart $PARTNUM 100%"
     '';
   };
+
+  enable =
```
mbssrc, line 49: Don't we want to keep both options available: with verity and without?

modules/partitioning/verity-partition.nix
```diff
@@ -0,0 +1,137 @@
...
+{
+  config,
+  lib,
+  pkgs,
+  modulesPath,
+  ...
+}:
+
+let
+  roothashPlaceholder = "61fe0f0c98eff2a595dd2f63a5e481a0a25387261fa9e34c37e3a4910edf32b8";
```
mbssrc, line 12: does lib.fakeSha256 work as well?

modules/partitioning/verity-partition.nix
```diff
@@ -0,0 +1,137 @@
...
+
+        # Compress the image
+        ${pkgs.zstd}/bin/zstd --compress $out/*raw
+        rm $out/*raw
+      '';
+    });
+
+    image.repart.split = cfg.split;
+
+    systemd.enableEmergencyMode = debugEnable;
```
mbssrc, line 57: This should be the default - `ghaf.systemd.withDebug = config.ghaf.profiles.debug.enable;`  in microvm-host which in turn enables `enableEmergencyMode = cfg.withDebug;` in `systemd/base.nix`. Does this not work?

modules/partitioning/verity-partition.nix
```diff
@@ -0,0 +1,137 @@
...
+        grub.enable = false;
+        systemd-boot.enable = lib.mkForce false;
+      };
+
+      # Enable dm-verity and compress initrd
+      initrd = {
+        systemd = {
+          enable = true;
+          dmVerity.enable = true;
+          emergencyAccess = debugEnable;
```
mbssrc, line 76: same as above

modules/hardware/common/input.nix
```diff
@@ -8,10 +8,6 @@ let
 in
 {
   config = {
```
mbssrc, line 10: Disko works fine without setting `config.disko.devices.disk.disk1.device` then? If so, we can remove this from the hardware config as well (`modules/hardware/definition.nix` and all hardware profiles)

mbssrc (approved):

msaarine:
**Testing**
1) (.#lenovo-x1-gen11-hardening-debug-installer)
Built, installed & tested those things mentioned in PR (sudo veritysetup status root, & shares folder size) & run automated bat tests(35) for the image 

- Building & Installation worked fine.
- The system seemed to work as expected, the given command was supported and status was 'verified'.
- File manager, the Shares folder showd to be around 940GB 
- Automated bat tests passed.

2) (.#lenovo-x1-carbon-gen11-debug-installer)
Built, Installed & tested a normal build and & run automated bat tests(35) for the image

- Building worked fine  ( .#lenovo-x1-carbon-gen11-debug-installer)
- Installation (**Please Check Note below!**)
- The system seemed to work as expected, Automated bat tests passed.

**Note:**
The only thing that I want to mention here is the installation after the 'hardening' image was used.

When I tried to install 'a normal build  'lenovo-x1-carbon-gen11-debug-installer' I faced some problems I have not seen before.
When installation was done, there is a step where it is advised like :">>Please remove installation media and re-boot"

Usually in my installations, I have always removed the SSD device and then executed 'sudo reboot' in that installation step.
Now when I unplugged the SSD, I got error like:
**I/O error, dev loop0, sector 3117570 op 0x0:(READ) flags 0x800 phys_seg 84 prio class 0
... (picture attached)**
and 'sudo reboot' after that  'did nothing'

That was surpise for me. I was told that this is ok and happens sometimes. I have been working with these issues only a short period of time so maybe I just haven't faced this before this.
To avoid that situation, I used a bit different process (instead of just removing the SDD): 

- sudo shutdown now
- Removed the media after shutdown
- Powered the PC
-->I ended up to normal situation to set username/passwd etc and the SW worked just fine

So if that functionality is known issue, I did not capture regression/non-working implementation during my testing.
PR1074-old-iso-installer-after-

PR #1085, opened 2025-03-19
Refactor: Imports structure

mbssrc:
## Description of changes

Make more flake-modules, and various changes. To be discussed.

## Checklist for things done

## Instructions for Testing

  - [ ] AGX
  - [ ] NX
  - [ ] x86_64
- [ ] Is this a new feature


modules/profiles/flake-module.nix
```diff
@@ -18,8 +18,10 @@
     # speciic profiles that are needed for certain classes of devices should be included below.
     # This can be on a category basis or integrated into an existing category if it has a common base
     profiles-laptop.imports = [
-      ./laptop-x86.nix
+      inputs.self.nixosModules.microvm
       inputs.self.nixosModules.hw-laptop
+      inputs.self.nixosModules.disko-debug-partition
```
brianmcgillion, line 23: do not add disko here it is not used in all laptop configurations only in the debug ones

targets/laptop/flake-module.nix
brianmcgillion, line 1: as the disko is not included this invalidates all changes to this file

brianmcgillion (changes requested):

modules/hardware/flake-module.nix
```diff
@@ -3,11 +3,17 @@
 {
 
   flake.nixosModules = {
-    laptop.imports = [
+    hw-laptop.imports = [
```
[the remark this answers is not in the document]
mbssrc: made the naming in line with `hw-x86_64-generic`. could change all names to `<arch>-generic/laptop`

modules/profiles/flake-module.nix
mbssrc: as of now (pre PR#1074), it is. same as current implementation with commonModules. also, the enable flag is now in reference profile, so it can be turned on/off in the settings

mbssrc: My thinking here is that we sync the targets identifier and the conditional modules definition, i.e. target = hw + platform profile identifier, with platform profile {mvp, ...} + {debug, release, any-other}. This would allow to move towards a generic interface for a laptop-builder and target definition. Not quite sure if a disk layout configuration at this level is necessary? And could we not just make the namespaces not overlapping? Then, instead of relying on conditional imports we control it via config settings, e.g. using -debug/release or even additional flags

brianmcgillion: there are a min of 3 levels of config that need to be supported. 

1. the hardware config (e.g. x1)
2. the profile config (e.g. mvp)
3. the target config ( with customizations available specific to this target definition)

The current laptop targets/laptop/flake-module.nix supports this.

brianmcgillion (approved):

PR #1104, opened 2025-03-25
desktop: add COSMIC Epoch DE

kajusnau:
## Description of Changes

This PR introduces COSMIC as a new desktop environment option for Ghaf, replacing LabWC if enabled.
Notable changes:

1. **Integration of COSMIC DE (Alpha 6)** using NixOS options (available in `nixpkgs` as of April 2025).
2. **Dedicated COSMIC Nix configuration** in `cosmic.nix`:
   - Disables certain default features for compatibility with Ghaf.
   - Overrides first-party COSMIC packages to remove unsupported features in Ghaf.
   - Uses `swayidle` instead of COSMIC's idle manager.
3. **Ghaf-COSMIC system-wide desktop configuration** in `cosmic.config.yaml`:
   - `cosmic.config.yaml` is expanded into `/run/current-system/sw/share/cosmic` by `ghaf-cosmic-config` package.
   - Helper script `cosmic-config-to-yaml.sh` available to convert local `.config/cosmic` into YAML format for easier maintenance.
   -  `/run/current-system/sw/share/cosmic` - system default config
      `.config/cosmic` - user config
   - `/run/current-system/sw/share/cosmic` is used as the default config for fresh installations.
     Any changes made by the user after the initial installation will be saved in  `.config/cosmic` and prioritized over system defaults.
     This means the user can install Ghaf, log into Cosmic and proceed to change keybindings, desktop layout, etc. freely.
4. **Replacement of desktop utilities** with COSMIC-native applications:
   - File Manager: `pcmanfm` → `cosmic-files`
   - Archiving: `xarchiver` → context menu in `cosmic-files`
   - Text Editor: `gnome-text-editor` → `cosmic-edit`
   - Terminal: `foot` → `cosmic-term`
   - Screenshot: `grim`, `slurp` → `cosmic-screenshot`
5. **Apps launched via waypipe**:
   - Added VM name prefixes to app title bars to help identify application origin.
   - This is needed because colored app borders are not available in COSMIC at this time.
     For more info see **Known Issues** section. 

### Known Limitations / Issues

- COSMIC does not support forced Server Side Decorations:
  - This is in contrast to LabWC, which supports forcing system title bars and borders for all apps
  - Due to this limitation, the Ghaf-specific colored borders for apps running under different VMs are not supported
  - In an attempt to mitigate this, apps running over waypipe will now have a VM prefix in the title bar (if available) in the following format: `[vm name] <app title>`
[...]

PR #1125, opened 2025-04-03
feat: power manager module, refactor ghaf-powercontrol

kajusnau:
This PR introduces a new `power-manager` module to override default logind power management and refactors `ghaf-powercontrol` for improved functionality and robustness.

## Description of changes

**1. Add power-manager module:**
   - Implement suspend signal interception and handling.
   - Implement shutdown/reboot signal interception and handling.
   - Suspend and shutdown/reboot listener services can be enabled individually through nix options.
   - Use `ghaf-powercontrol` to override power state management.

The power-manager module provides better control over system power management by intercepting logind signals and using `ghaf-powercontrol` for custom handling of suspend, shutdown, and reboot operations.
With this module enabled, the user should be able to simply call `systemctl suspend/poweroff/reboot`, regardless of whether the DE is running on the host or in a VM.
This module is expected to be first used in COSMIC (#1104)
This PR also reduces the scope of the COSMIC PR.

**2. Refactor ghaf-powercontrol:**
   - Refactor to handle any Wayland compositor by iterating over available `wayland-[0-9]` sockets.
   - Fix premature exit during suspend if `wlopm` fails.

## Known issues/Notes using `power-manager` module:
- Suspension limitations:
  - Currently, the system continues to suspend even after `ghaf-powercontrol suspend` is called.
  - If running on a host:
    - This might result in two suspension requests in a row.
  - If running on a VM:
    - `ghaf-powercontrol suspend` will execute first, then, if woken up by the user:
    - a non-functional VM suspend, which gets cancelled almost immediately
    - The end result is that the system suspends, and, if woken up by the user, attempts to suspend again, but wakes up immediately.
- Root execution:
  - `power-manager` module adds system services which run as root.
  - This means `ghaf-powercontrol`, which is called by these services, will also run as root, which may cause it to fail to find and turn off the display as part of the suspension procedures.
  This might result in the display showing a static image while the system is suspended

## Testing instructions
As power-manager is not enabled by default, only `ghaf-powercontrol` is affected by this PR. Therefore testing should:
1. Verify no regression can be observed in power control behavior:
   - Suspend, shutdown, and reboot all function as before wherever `ghaf-powercontrol` is used (e.g. LabWC ewwbar, swayidle autosuspend, etc.)

## Checklist for things done

  - [ ] Tested on Dell Latitude `x86_64`

## Instructions for Testing

  - [x] AGX
  - [x] NX
  - [x] x86_64
- [x] Is this a new feature
  - Minor bugfix in `ghaf-powercontrol`

### DIFF
--- a/modules/desktop/graphics/power-manager.nix
+++ b/modules/desktop/graphics/power-manager.nix
@@ -1,0 +1,148 @@
+# Copyright 2022-2025 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+# This module overrides logind power management using ghaf-powercontrol.
+# The specific operations overriden are:
+# - Suspend
+# - Shutdown
+# - Reboot
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  cfg = config.ghaf.graphics.power-manager;
+  inherit (lib)
+    mkIf
+    mkOption
+    types
+    ;
+
+  ghaf-powercontrol = pkgs.ghaf-powercontrol.override { ghafConfig = config.ghaf; };
+
+  logindSuspendListener = pkgs.writeShellApplication {
+    # This listener monitors systemd's suspend signals and delays the suspend process
+    # While the process is delayed, we use ghaf-powercontrol to handle the suspend operation instead
+    # TODO: Investigate if suspension can be cancelled entirely if caught. Some notes on that:
+    # - `--mode=block` does not work, as it will block suspension entirely and no signal will be sent
+    # - `--mode=delay` works, but the system will still suspend after the delay
+    name = "logind-suspend-listener";
+    runtimeInputs = [
+      pkgs.dbus
+      pkgs.systemd
+      ghaf-powercontrol
+    ];
+    text = ''
+      systemd-inhibit --what=sleep --who="ghaf-powercontrol" \
+        --why="Handling ghaf suspend" --mode=delay \
+          dbus-monitor --system "type='signal',interface='org.freedesktop.login1.Manager',member='PrepareForSleep'" | \
+            while read -r line; do
+              if echo "$line" | grep -q "boolean true"; then
+                echo "Found prepare for sleep signal"
+                echo "Suspending via ghaf-powercontrol"
+                ghaf-powercontrol suspend
+              fi
+            done
+    '';
+  };
+
+  logindShutdownListener = pkgs.writeShellApplication {
+    # This listener monitors systemd's shutdown/reboot signals and delays the process
+    # While the process is delayed, we use ghaf-powercontrol to handle the shutdown/reboot operation instead
+    name = "logind-shutdown-listener";
+    runtimeInputs = [
+      pkgs.dbus
+      pkgs.systemd
+      ghaf-powercontrol
+    ];
+    text = ''
+      systemd-inhibit --what=shutdown --who="ghaf-powercontrol" \
+        --why="Handling ghaf shutdown/reboot" --mode=delay \
+          dbus-monitor --system "type='signal',interface='org.freedesktop.login1.Manager',member='PrepareForShutdownWithMetadata'" | \
+            while read -r line; do
+              if echo "$line" | grep -q "boolean true"; then
+                echo "Found prepare for shutdown signal. Checking type..."
+                while read -r subline; do
+                    if echo "$subline" | grep -q "reboot"; then
+                        echo "Found type: reboot"
+                        echo "Rebooting via ghaf-powercontrol"
+                        ghaf-powercontrol reboot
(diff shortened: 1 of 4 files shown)


modules/desktop/graphics/power-manager.nix
```diff
@@ -0,0 +1,144 @@
...
+# - Reboot
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  cfg = config.ghaf.graphics.power-manager;
+
```
vunnyso, line 16: Can add?
```
  inherit (lib)
    mkIf
    mkOption
    types;
```

milva-unikie:

Issues:
- Suspend from the Power menu does not work. After clicking the Suspend button the display stays on but freezes. When mouse is moved the screen locks. The system never actually suspends.
- Similar behavior in the Lock screen when trying to suspend. The display freezes for a while but the system is not suspended.

Working:
- Reboot and Shutdown work from both the Power menu and from the Lock screen.
- Swayidle auto-suspend works
- The laptop is suspended when the lid is closed
- Test-automation passes

johannarautanen:
**Checked with native Orin AGX and NX,** flashed to the memory stick:
Issue:
- Suspend doesn't work: screen goes black and stay that way, not able to "wake up" anymore. Need to reboot the Orin's.
Happened via Power menu and Lock screen
There appeared texts on the serial connection, when suspend activated: 
Screenshot from 2025-04-04 09-02-21
 
Working:
- Lock, logout, reboot and shutdown working the in power menu
-  reboot and shutdown working in Lock Screen
- automated tests ok

milva-unikie:

Everything is working!

- Suspend, Reboot and Shutdown work from both the Power menu and from the Lock screen
- Swayidle auto-suspend works
- The laptop is suspended when the lid is closed
- Test-automation passes

Notes:
- Native Orin-NX boot test failed in the pre-merge pipeline. I retriggered the test in dev setup and it passed with no issues. All x86 targets are still building as I am writing this comment.

johannarautanen:
**Checked with cross compile AGX**, flashed to the memory stick:
Issue:
Suspend doesn't work: screen goes black and stay that way, not able to "wake up" anymore. Need to reboot the Orin's.

kajusnau:
> **Checked again with native Orin AGX and NX,** flashed to the memory stick: Issue: Suspend doesn't work: screen goes black and stay that way, not able to "wake up" anymore. Need to reboot the Orin's.
> 
> On the main build when select the "suspend" there appeared "lock screen" and you have use the device after entering the "ghaf" And before that "suspend->lock screen"-issue appeared the "suspend" works ok.

> **Checked with cross compile AGX**, flashed to the memory stick: Issue: Suspend doesn't work: screen goes black and stay that way, not able to "wake up" anymore. Need to reboot the Orin's.

I suspect this was the actual behavior:

1. Screens turn off, system suspends
2. System wakes up, screens stay turned off

I pushed a minor change that adds the following after waking up:
- 1 second sleep
- Sanity check request to turn off displays (they should already be off, but this may allow `wlopm` to "catch its' breath" before attempting to turn displays back on)
This seemed to do the trick in my manual test but more testing and logs needed next week

johannarautanen:
**Tried the native AGX wit the latest commit:**
Cannot wake up the Orin after suspend, external screen stays black and serial connections isn't responding.
Screenshot from 2025-04-07 06-47-55
[suspend.txt](https://github.com/user-attachments/files/19625570/suspend.txt)

kajusnau:
Based on @johannarautanen 's findings, I've decided to disable suspending on Orins for the time being.

It seems that before this PR there were already issues on main that caused suspending to not work at all.
This PR seems to fix that issue, as suspending is working as intended, but, in turn, waking up the device seems impossible.

I recommend proceeding with this PR as is, with suspending disabled on Orins, and taking another look at SSRCSP-5888 to either modify it or open a new bug ticket.

johannarautanen:
**Checked with cross compile Orin AGX and NX**
Working:
- Lock, logout, reboot and shutdown working the in power menu
- Suspend is disabled from power menu and lock screen. 
When you select Suspend from Power menu, there appeared note: 
Screenshot from 2025-04-07 13-17-19

In Lock screen when you select Suspend, nothing happened, there is still lock screen available and you able to open the lock screen without problems.
- reboot and shutdown working in Lock Screen
- automated tests ok

johannarautanen:
**Checked with cross native Orin AGX and NX**
Working:
- Lock, logout, reboot and shutdown working the in power menu
- Suspend is disabled from power menu and lock screen.
When you select Suspend from Power menu, there appeared note:
Screenshot from 2025-04-07 13-17-19

In Lock screen when you select Suspend, nothing happened, there is still lock screen available and you able to open the lock screen without problems.

- reboot and shutdown worked in Lock Screen
- automated tests ok

PR #1149, opened 2025-04-11
Update vhotplug to fix issues with multiple devices with the same VID/PID

nesteroff:
## Description of Changes
This fixes an issue with hot-plugging that can potentially cause a race condition and system crash.

When the system is booted from a USB SSD drive, it remains connected to the host because vhotplug detects it as a boot device and doesn't pass it to the GUIVM. When a second USB drive is connected, it gets passed to the GUIVM. If the user then unplugs it, the vhotplug service tries to remove it from the GUIVM but at the same time QEMU tries to find the device by VID/PID and reconnect it. If both drives are the same model, QEMU finds the boot drive and connect it to the GUIVM causing a system crash.

This issue is difficult to reproduce because in most cases the vhotplug service removes the drive from the GUIVM faster than QEMU can detect and reconnect it. However, Samuli managed to consistently reproduce the crash.

To fix the issue, the updated version of vhotplug uses the device bus and address which are unique to each device.

This change is also required for any scenario involving multiple identical USB devices of any type as relying on VID/PID doesn't allow to distinguish between them reliably.

The only drawback to this approach is that devices passed using bus and address are not automatically reconnected after a system suspend/resume cycle. A small QEMU patch has been created to resolve this.

### Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Improvement / Refactor

## Related Issues / Tickets

SSRCSP-6350

## Checklist

## Testing Instructions

Testing instructions are in SSRCSP-6350.

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`

### Installation Method
- [x] Other: 
A reboot might be needed when updated using nixos-rebuild.

### Test Steps To Verify:
1. Check that regular hot-plugging scenarios still work as expected.
2. Check that system doesn't crash using test steps from SSRCSP-6350.

### DIFF
--- a/overlays/custom-packages/qemu/default.nix
+++ b/overlays/custom-packages/qemu/default.nix
@@ -15,6 +15,7 @@ prev.qemu_kvm.overrideAttrs (
     patches = prev.patches ++ [
       ./acpi-devices-passthrough-qemu-8.1.patch
       ./0001-ivshmem-flat-memory-support.patch
+      ./usb-host-enable-autoscan-for-bus-addr.patch
     ];
   })
   // {
--- a/overlays/custom-packages/qemu/usb-host-enable-autoscan-for-bus-addr.patch
+++ b/overlays/custom-packages/qemu/usb-host-enable-autoscan-for-bus-addr.patch
@@ -1,0 +1,46 @@
+From 88687c9939c5d744b1f7224d77943099b659ab17 Mon Sep 17 00:00:00 2001
+From: Yuri Nesterov <yuriy.nesterov@unikie.com>
+Date: Wed, 9 Apr 2025 22:24:52 +0300
+Subject: [PATCH] usb-host: enable autoscan for bus+addr to survive host
+ suspend/resume
+
+Currently, there is a special case for USB devices added using the
+hostbus= and hostaddr= properties to avoid adding them to the hotplug
+watchlist, since the address changes every time the device is plugged
+in. However, when the host system goes into suspend and then resumes,
+those devices are disconnected from the guest with the "no device"
+error but their address stays the same. Enabling autoscan and adding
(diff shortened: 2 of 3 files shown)


brianmcgillion (approved):

vunnyso (approved):

leivos-unikie:
Connecting cable keyboard and mouse via Bluecloud USB hub before connecting USB storage was essential part of the procedure getting system crash. 

It seems that also the type of Bluecloud USB hub is critical.  I tested now first with a Bluecloud hub with C-type connector and was not able to reproduce the issue with ghaf mainline. I think the bug requires A-type USB hub to reproduce.

https://jira.tii.ae/browse/SSRCSP-6350

nesteroff:
> It seems that also the type of Bluecloud USB hub is critical. I tested now first with a Bluecloud hub with C-type connector and was not able to reproduce the issue with ghaf mainline. I think the bug requires A-type USB hub to reproduce.
> 

I also noticed that it's harder to reproduce the issue with the current version of the mainline although it’s still possible with some patience. However, there is a simple way to trigger it:
- Boot from a USB SSD.
- Connect another USB SSD of the same model and make sure it's mounted in the GUIVM.
- Stop the vhotplug service on the host: systemctl stop vhotplug
- Disconnect the second USB SSD.
- Try to access it by browsing files or something like that.
- The system crashes.

msaarine:
Tested OK on 'Lenovo X1 Carbon'.
- I did not manage to get the failure happen (several trials, both 2 methods tried) (SW without this fix failed on my test setup)
-bat tests OK, USB stick/SSD mounts OK.

PR #927, opened 2024-12-10
Add flash scripts as build result

remimimimimi:
## Description of changes

Flash script is changing every now and then and each Ghaf binary images must be flashed to bootable drive with correct version of the flash.sh, so provide flash scripts alongside with images.

## Checklist for things done

- [x] Summary of the proposed changes in the PR description
- [ ] More detailed description in the commit message(s)
- [x] Commits are squashed into relevant entities - avoid a lot of minimal dev time commits in the PR
- [ ] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] PR linked to architecture documentation and requirement(s) (ticket id)
- [ ] Test procedure described (or includes tests). Select one or more:
  - [ ] Tested on Lenovo X1 `x86_64`
  - [ ] Tested on Jetson Orin NX or AGX `aarch64`
  - [ ] Tested on Polarfire `riscv64`
- [ ] Author has run `make-checks` and it passes
- [ ] All automatic Github Action checks pass - see [actions](https://github.com/tiiuae/ghaf/actions)
- [ ] Author has added reviewers and removed PR draft status
- [ ] Change requires full re-installation
- [ ] Change can be updated with `nixos-rebuild ... switch`

## Instructions for Testing

1. Verify that installer still works.
2. Check that all laptop and nvidia-jetson-orin packages result directory contains image and flash script(s).

PR #1196, opened 2025-05-12
Rebase: Flash script as build result

juliuskoskela:
## Description of Changes

Bundle flash-script with image build. For example:

```
nix build .#lenovo-x1-carbon-gen11-debug
```

Will no produce

```
result/
  flash-script/bin/flash-script
  image/disk1.raw.zst
```

Instead of only the image.

### Type of Change
- [x] Improvement / Refactor

## Related Issues / Tickets

https://github.com/tiiuae/ghaf/pull/927

## Checklist

### Applicable Targets
- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`

### Installation Method
- [x] Other: 

### Test Steps To Verify:

1. ...

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -53,5 +53,21 @@ nixpkgs.lib.extend (
         in
         lib.elem system platforms
       );
+
+    genPkgWithFlashScript =
+      pkg: system:
+      let
+        pkgs = import nixpkgs { inherit system; };
+      in
+      pkgs.linkFarm "ghaf-image" [
+        {
+          name = "image";
+          path = pkg;
+        }
+        {
+          name = "flash-script";
+          path = pkgs.callPackage ./packages/flash { };
+        }
+      ];
   }
 )
--- a/lib.nix
+++ b/lib.nix
@@ -66,7 +66,7 @@ nixpkgs.lib.extend (
         }
         {
           name = "flash-script";
-          path = pkgs.callPackage ./packages/flash { };
+          path = pkgs.callPackage ./packages/pkgs-by-name/flash-script/package.nix { };
         }
       ];
   }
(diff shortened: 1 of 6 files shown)


brianmcgillion:
overall would it make more sense to add this to the devshell and also to have it exposed directly as a package in the outputs? you need the script to be downloaded/installed once and technically it is still generic as this PR does not "fill in" the target package that is to be flashed. so it I think it should be a one time build/download and installed when setting up a system.

juliuskoskela:
> overall would it make more sense to add this to the devshell and also to have it exposed directly as a package in the outputs? you need the script to be downloaded/installed once and technically it is still generic as this PR does not "fill in" the target package that is to be flashed. so it I think it should be a one time build/download and installed when setting up a system.

I agree. So basically the same way it is done for Jetson targets. The original ticket reads:

```
Bootable drives for Ghaf devices are flashed with https://github.com/tiiuae/ghaf/blob/main/packages/flash/flash.sh 

Flash script is changing every now and then and each Ghaf binary images must be flashed to bootable drive with correct version of the flash.sh.

Currently flash.sh is not part of the build output. This is a problem for at least 

    User trial users
    Testers
    Open source users trying to download Ghaf binary images and flashing to their devices

Please provide flash.sh script as build output. Store it to the same folder with the target binary image.
```

So should we change the requirements slightly and just generate `${target}-flash-script` targets for all targets and leave the base targets (that produce the image) as is?

kajusnau:
> Flash script is now part of the devshell. Should this be closed? @brianmcgillion @juliuskoskela

Found this in my involved PRs, closing as it doesn't seem relevant anymore.

PR #1215, opened 2025-05-22
usb: Added a common stub for external usb devices

vunnyso:
## Description of Changes
- Created a common list of external USB devices for x86_64 architecture, which can be assigned to any VM as required.
- Added support for a few Xbox controllers.

### Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`

### Installation Method
- [ ] Other: 

### Test Steps To Verify:

1. ..

### DIFF
--- a/modules/development/debug-tools.nix
+++ b/modules/development/debug-tools.nix
@@ -63,6 +63,8 @@ in
         pkgs.grpcurl
 
         pkgs.lshw
+        # Tools for joysticks and serial peripherals
+        pkgs.linuxConsoleTools
 
         #(config.boot.kernelPackages.perf.override {python3 = pkgs.python311;})
         sysbench-test-script
--- a/modules/hardware/common/default.nix
+++ b/modules/hardware/common/default.nix
@@ -4,6 +4,7 @@
   imports = [
     ./usb/internal.nix
     ./usb/external.nix
+    ./usb/external-devices.nix
     ./usb/vhotplug.nix
     ./devices.nix
     ./input.nix
--- a/modules/hardware/common/usb/external-devices.nix
+++ b/modules/hardware/common/usb/external-devices.nix
@@ -1,0 +1,33 @@
+# Copyright 2022-2025 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+_: {
+  config = {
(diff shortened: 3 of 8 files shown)


vunnyso:
Moved to FMO https://github.com/tiiuae/ghaf-fmo-laptop/pull/106/

mbssrc:
This is a good intermediate step, think we can merge this for now.

The problem is that we have two competing mechanisms for the USB passthrough, and should IMO unify the two to minimize problems with USB definition overall, especially downstream.

My suggestion is the following:
- remove external devices from the hardware definition, and keep only static internal devices in the hardware definition. All additional rules for external devices (like the ones we have here) can be managed by vhotplug, as they are not hardware dependent
- collect internal devices from hardware definition and combine the rules with vhotplug, so we can remove the qemu arguments and udev rules for internal devices as well
- generate a unified vhotplug ruleset based on profile (e.g., MVP) and internal hardware devices for a target, by creating a module with an interface that makes it easy to define USB rules per profile

mbssrc:
> > This is a good intermediate step, think we can merge this for now.
> > The problem is that we have two competing mechanisms for the USB passthrough, and should IMO unify the two to minimize problems with USB definition overall, especially downstream.
> > My suggestion is the following:
> > 
> > * remove external devices from the hardware definition, and keep only static internal devices in the hardware definition. All additional rules for external devices (like the ones we have here) can be managed by vhotplug, as they are not hardware dependent
> > * collect internal devices from hardware definition and combine the rules with vhotplug, so we can remove the qemu arguments and udev rules for internal devices as well
> > * generate a unified vhotplug ruleset based on profile (e.g., MVP) and internal hardware devices for a target, by creating a module with an interface that makes it easy to define USB rules per profile
> 
> I completely agree @mbssrc, even I have noticed issues in downstream projects between vhotplug and USB passthrough with Qemu args. I will try to incorporate your suggestions in upcoming PR.

Awesome - please sync with @gngram who is working on this as well

brianmcgillion (approved):

PR #1232, opened 2025-06-04
lenovo-x1-gen11: Add TPM-backed encryption for the persist partition

hros-tii:
## Description of Changes

This PR builds upon https://github.com/tiiuae/ghaf/pull/1074 which was merged in main as https://github.com/tiiuae/ghaf/commit/3f25f7338e6a69bdb1bcc4a79bb165b8077d0173. It builds on top of the dm-verity protected rootfs. 

The persist partition is encrypted using the TPM as a LUKS token. An initial step takes place where persist is enrolled to the TPM and a PIN is set. A recovery passphrase is also added to the LUKS header which can optionally be saved and used to later access the partition contents in case of decryption failure. The user is then prompted for the PIN during the boot sequence.

A new user profile is added (mvp-user-trial-hardening). For now selecting this profile only enables the feature described here. It is automatically enabled by the `lenovo-x1-gen11-hardening-debug` target along with dm-verity.

### Limitations

- Other partitions are not encrypted, namely rootfs and swap.

This could be implemented in a similar way as for persist, but some modifications could be made for user convenience, to avoid having to input a different PIN for each partition. The decryption step would also need to be moved from user space to  initrd.

- Plymouth is disabled, to display the PIN input prompt

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

- [x] Tested on Lenovo X1 `x86_64` Gen11

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`

### Installation Method

### Test Steps To Verify:

1. Build an installer image for the `lenovo-x1-gen11-hardening-debug-installer` target.
2. Flash to x86_64 laptop
3. First boot should stop halfway to prompt for the PIN. 
4. Press Enter once (no input) to clear the passphrase prompt
5. Enter the PIN that will be used to decrypt the partition.
6. Press Enter a few times to complete the enrollment.
7. System should boot normally. On host system run `dmsetup status` or `mount` to confirm the persist partition is mapped with dm-crypt.

### DIFF
--- a/modules/partitioning/verity-partition.nix
+++ b/modules/partitioning/verity-partition.nix
@@ -103,6 +103,15 @@ in
     # System is now immutable
     system.switch.enable = false;
 
+    swapDevices = [
+      {
+        device =
+          if config.ghaf.storage.encryption.enable then "/dev/mapper/swap" else "/dev/disk/by-partlabel/swap";
+        discardPolicy = "both";
+        options = [ "nofail" ];
+      }
+    ];
+
     fileSystems =
       let
         tmpfsConfig = {
@@ -123,7 +132,11 @@ in
             partConf = config.image.repart.partitions."50-persist".repartConfig;
           in
           {
-            device = "/dev/disk/by-partuuid/${partConf.UUID}";
+            device =
+              if config.ghaf.storage.encryption.enable then
+                "/dev/mapper/persist"
+              else
+                "/dev/disk/by-partuuid/${partConf.UUID}";
             fsType = partConf.Format;
           };
       }
--- a/modules/common/security/default.nix
+++ b/modules/common/security/default.nix
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 {
   imports = [
+    ./disk-encryption.nix
     ./sshkeys.nix
     ./apparmor
     ./audit
--- a/modules/common/security/disk-encryption.nix
+++ b/modules/common/security/disk-encryption.nix
@@ -1,0 +1,214 @@
+# Copyright 2022-2025 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  lib,
+  config,
+  pkgs,
+  utils,
+  ...
+}:
+let
+  inherit (lib)
+    types
+    mkOption
+    mkEnableOption
+    mkIf
+    ;
+  cfg = config.ghaf.storage.encryption;
+
+  partitions =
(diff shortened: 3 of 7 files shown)


vunnyso (changes requested):
Thanks @hros-tii for nice feature integration. 

Few findings when I have tested it.
1. Persist size is reduced to `3G` and its not using entire disk space compared to `lenovo-x1-gen11-hardening` 
image
2. On every boot system is asking for PIN which can be prevented.
3.  PR need to be fix the `run-checks` and rebase your PR with latest.

modules/partitioning/verity-partition.nix
```diff
@@ -96,6 +102,12 @@ in
     # System is now immutable
     system.switch.enable = false;
 
+    swapDevices = [
```
vunnyso, line 105: It will be nice it encrypt `swap` partition as well.

vunnyso:
> Thanks @vunnyso for the review. Especially item 1) is a nice catch. I'll push some changes to address your comments.
> 
> > On every boot system is asking for PIN which can be prevented.
> 
> This is on purpose to prevent the partition to be decrypted automatically, in the event of a stolen device for example. Do you mean that the PIN should be remembered for a small period of time after the first input? I agree that this is a tradeoff with usability to ask the user to enter a password twice (although the two passwords can be identical), but it gives the best assurance security-wise.

Thanks @hros-tii. I mean entering a PIN at every boot is not user-friendly. It would be beneficial if we could make PIN as persistent by trying different TPM configuration registers or another security mechanism. For an additional layer of security, we can enroll mandatory Yubikey presence if needed.

modules/partitioning/verity-sysupdate.nix
```diff
@@ -11,7 +11,7 @@ let
   cfg = config.ghaf.partitioning.verity;
 in
 {
-  config = lib.mkIf cfg.enable {
+  config = lib.mkIf cfg.sysupdate {
```
vunnyso, line 14: No need to check for `cfg.enable` now?

vunnyso (approved):

vunnyso:
Tested on Lenovo-X1 Gen11, `filesystem` snapshot remains intact. Basic testing looks fine.

| Branch | File system snapshot |
| ----------- | ----------- |
| **mainline**|  image |
| **This PR**| image |

PR #1302, opened 2025-08-05
Changes to the VM target

hros-tii:
## Description of Changes

This PR contains some adjustments that were made to the `vm` target while working on the TPM storage feature. We are posting them in a PR in case there is interest to merge parts into upstream. Changes include:

- added a non-graphical qemu target for testing Ghaf on machines without a display output
- increased the resources allocated to qemu host
- set the `useNixStoreImage` option, which disables mounting the top-level nix store inside the qemu ghaf-host with an overlay. This was causing issues because it is then mounted the same way inside the microvm. A temporary disk image is now created when starting the ghaf-host with a copy of the nix store.
- enabled a few reference VMs for testing
- a VMware disk image can also be created instead of qemu (useful on MacOS)

We found this setup helpful to experiment with host services, inter-vm communication, or the TPM. However device passthrough is still not implemented for this qemu target (I think it could be done by running hardware-scan inside qemu and assigning the PCI indexes).

### Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [x] QEMU virtual machine `x86_64`

### Test Steps To Verify:

`nix run .#vm-debug` or `nix run .#vm-debug-nogui`

### DIFF
--- a/modules/microvm/modules.nix
+++ b/modules/microvm/modules.nix
@@ -185,23 +185,25 @@ in
       netvm.extraModules =
         optionals cfg.netvm.enable [
           serviceModules.logging
+          serviceModules.givc
           commonModule
+          managedUserAccounts
         ]
         ++ optionals (cfg.netvm.enable && fullVirtualization) [
           deviceModules.netvmPCIPassthroughModule
           kernelConfigs.netvm
           firmwareModule
           serviceModules.wifi
-          serviceModules.givc
           serviceModules.audit
           referenceServiceModule
-          managedUserAccounts
         ];
       # Audiovm modules
       audiovm.extraModules =
         optionals cfg.audiovm.enable [
           serviceModules.logging
+          serviceModules.givc
           commonModule
+          managedUserAccounts
         ]
         ++ optionals (cfg.audiovm.enable && fullVirtualization) [
           deviceModules.audiovmPCIPassthroughModule
@@ -210,16 +212,16 @@ in
           qemuModules.audiovm
           serviceModules.audio
           serviceModules.audit
-          serviceModules.givc
           serviceModules.bluetooth
           serviceModules.xpadneo
-          managedUserAccounts
         ];
       # Guivm modules
       guivm.extraModules =
         optionals cfg.guivm.enable [
           serviceModules.logging
+          serviceModules.givc
           commonModule
+          managedUserAccounts
         ]
         ++ optionals (cfg.guivm.enable && fullVirtualization) [
           deviceModules.guivmPCIPassthroughModule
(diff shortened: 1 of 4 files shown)


modules/microvm/modules.nix
```diff
@@ -185,23 +185,25 @@ in
       netvm.extraModules =
         optionals cfg.netvm.enable [
           serviceModules.logging
+          serviceModules.givc
```
brianmcgillion, line 188: im surprised this build on aarch64, did not know that our givc was ready for these targets. so lets test it and see if it blows up :)

brianmcgillion (approved):
LGTM. really nice work. thank you

targets/vm/flake-module.nix
```diff
@@ -68,18 +88,35 @@ let
...
+      name = "${format}-${variant}";
       package = hostConfiguration.config.system.build.${hostConfiguration.config.formatAttr};
     };
   targets = [
-    (vm "debug")
-    (vm "release")
+    (vm "vm" "debug" true)
+    (vm "vm" "debug-nogui" false)
+    (vm "vm" "release" true)
+    (vm "vmware" "debug" true)
```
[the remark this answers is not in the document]
brianmcgillion: the PR message states for support while running/developing on a mac.

modules/microvm/modules.nix
brianmcgillion: Thanks @emrahbillur this is great!

enesoztrk (commented):
It would be better if you squash some dev commits

hros2:
> `Creating Nix store image...` takes relatively long time (~100 sec on my machine), also when the vm has already been launched before and ghaf-host.qcow2 exists. Is this expected?

For some reason nixpkgs does not cache the qcow2 image between the vm invocations when `useNixStoreImage` is used:
```
        The Nix store image is built just-in-time right before the VM is
        started. Because it does not produce another derivation, the image is
        not cached between invocations and never lands in the store or binary
        cache.
```
They don't have an option to enable this at the moment. It would need a patch to nixpkgs, to change this part https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/virtualisation/qemu-vm.nix#L174

hros2:
> Currently gui of vm-debug is not very useful because ghaf user cannot log in and there is no testuser. Probably this is problem for another PR.
> 
> (Is GUI still running on ghaf-host, not in gui-vm?)

Yes the desktop is running on ghaf-host because the GPU is not passed through. It was already the case before on the `vm` target since the fullVirtualisation is not enabled. For the features we were working on, it was still useful because they weren't related to the GUI. 
I tried experimenting a bit with passthrough of virtio devices but it is not so straightforward, so at the moment it is still better to test device passthrough on a real device.

PR #1312, opened 2025-08-11
Firewall blacklisting mechanism & testing

enesoztrk:
## Description of Changes
* SSH tarpit to slow or trap unauthorized SSH attempts
* Packet marking and redirection for blacklisted traffic
* Blacklist management via ipset and iptables
* Ensures required firewall kernel modules are included for all the targets.
* Includes tests for tarpit, marking, redirection, and blacklist enforcement

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`

### Installation Method
- [ ] Other: 

### Test Steps To Verify:

1. Network-related services should work
2. Launch an attack using the command `hping3 -S -p 22 -i u10000 -c 20 ${netVmIp}` 
3. Check the blacklist on net-vm with `ipset list BLACKLIST`. You should see attacker's IP in the members section.
4. Attempt to connect from the attacker's device to the Ghaf laptop via SSH. The connection should hang.

### DIFF
--- a/docs/src/content/docs/ghaf/releases/ghaf-25.08.mdx
+++ b/docs/src/content/docs/ghaf/releases/ghaf-25.08.mdx
@@ -132,4 +132,3 @@ Fixed bugs that were presentin the ghaf-25.06 release:
 Released images are available at https://archive.vedenemo.dev/ghaf-25.08/
 
 Download the required image and use the following instructions: [Build and Run](/ghaf/dev/ref/build_and_run).
-
--- a/modules/common/firewall/firewall.nix
+++ b/modules/common/firewall/firewall.nix
@@ -14,14 +14,51 @@ let
     mkIf
     strings
     concatMapStringsSep
-    optionalString
+    concatStringsSep
     mkBefore
+    optionals
+    mkEnableOption
     ;
-
-  sshTcpPort = 22;
+  blackListName = "BLACKLIST";
+  blacklistMarkNum = "8";
   cfg = config.ghaf.firewall;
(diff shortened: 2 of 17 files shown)


vunnyso:
Overall, it looks good. Although I'm not a networking expert, I think we can do one round of testing.

milva-unikie:
**Issues**
- Automated Orin tests are failing. They try to make too many SSH connections in a minute. The pre-merge tests were fine since those only include some of the cases. When running the whole BAT suite we reach the limit of [10 connections/minute](https://github.com/enesoztrk/ghaf/blob/ab9e0ba554a4d7b7aa45e98755c3eac93b0726a3/modules/common/security/ssh-tarpit/default.nix#L59).
- There are two options:
  - [Might take a while] Wait while we look into the Orin tests and try to reduce the number of required connections. Currently the connection is often closed and reopened between cases.
  - [Faster solution] Increase the limit to 20 connections/minute, at least temporarily. With the limit set to 20 all Orin tests passed (15 was not enough).

**Tested on Lenovo-X1 and Orin-NX**
- Networking seems to work fine
- Attacker gets blacklisted for one hour, after that connecting works again
- Connections from other IP addresses work while one IP address is blacklisted

vunnyso (approved):

PR #1416, opened 2025-09-16
Active Directory

mbssrc:
## Description of Changes

Add AD enrollment for systemd-homed, and initial support for AD users managed with SSSD.
Ghaf docs come with a separate PR.

#### AD configuration
The AD domains options are defined in `modules/common/users/active-directory.nix`, and the test domain defined in `modules/common/users/ad-users.nix`. This should move to a provisioning file when available.

Aside from a shortlist of sanity checks, it configures the krb.conf (for krb tooling) and DNS for AD domains on all machines.
This _should_ not impact any other network connectivity - in testing that worked fine. It allows configuration of machines without AD computer join and some future features, but can be removed from global config if it makes problems.

The majority of flags has been tested with single (working) domain configuration. Note that there are several important distinctions between local (sssd) and per-domain settings. Any other configuration aside from the test domain should be 
evaluated (e.g. different providers and subsequent settings), as each one has their own implications. 

Most notably, the limitation is the user setup script and subsequent tooling.

#### User provision

Ghaf now provides 3 mechanisms for user provisioning:
1. **Managed users**: These are traditional NixOs users defined by configuration. For more details refer to the `mutable-users` enable option in `modules/common/users/profiles.nix`. This includes the "ghaf" admin user.
2. **Homed users**: This is the default, local homed user profile as before. Can be configured locally or details fetched from AD. As of now, it doesn't do much as we enforce single users and fixed uid/gid.
3. **Active Directory users**: These users are fetched from the AD domain the computer is enrolled in.

**Provisioning Script**:
The user provisioning script has been updated and moved to a package to accommodate all necessary functions and get rid of long scripts in .nix files, and be available as cli command. 
This script is a temporary solution - this should be properly implemented (not bash). However, this can be used until the user enrollment flow is clear, e.g. admin logging into UI in which case we can make this a proper UI application.

- It supports creation of local users (+ optional AD query), and domain joining
- It (quite heavily) limits the choice of AD configuration (e.g. LDAP+GSSAPI). For details dm me
- For AD related tasks, it automatically checks connectivity to the AD and allows to either use a cable of connect to WiFi

**Notes for CI/CD testing**:

The systemd service names for automated test setup have changed:
- for automated user removal: `systemctl start user-provision-remove.service`
- for automated user creation: `systemctl start user-provision-test.service`
- user-provision can also be used as cli tool, homectl also works as before

#### AD User functionality (new)
Here a quick (possibly incomplete) list of features currently supported with "Active Directory users". 
Note that there are a plethora of limitations by cosmic-greeter.

- login: with compliant username (john.smith -> johnsmith, no domain name needed). For current test setup we go with RFC2307 compliance. Notably: the user has to enter the username every time in cosmic-greeter, and press enter before entering the password. Messages are confusing or missing.
- offline login: enabled by default, user passwords are cached (defined by SSSD parameter, defaults to 7 days). Note that the first login must be done with connection to the backend, which should still be available due to previous domain join.
- Home creation: homes are automatically created. No home encryption is enabled as of yet.
- Login restrictions: in the test setup, all domain users are allowed to login. This can be limited via backend (e.g. policy + enforcing GPO access control).
- Password update: works via command line 'passwd' command. cosmic-greeter does not support it, which also makes ADs 'change password on next login' option not possible via GUI
- Desktop usability: _should_ work the same as homed user

#### Shortlist of possible future improvements

- Password change policy from AD not supported by cosmic-greeter
- Add FIDO2 support + MFA
- Support remote data (e.g. autofs+smb with local cache, or other)
- Evaluate FS encryption with SSSD
- Add sudo provider / AD sudo support
- Support PAC service module for SSSD (upstream?)
- Support TPM sealing
- More AD controls and refinement
 
### Other changes with this patch
- remove waypipe ssh
- persist log storage only when enabled
- re-organize file structure
- update mutable users logic (see option `config.ghaf.users.profile.mutable-users.enable` for details)

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other: 

### Test Steps To Verify:

1. Regression test for simple local user creation.
2. Get AD credentials to be able to join machines (DM me)
3. Create local user account with AD query
4. Rebuild and flash with homed profile disabled and ad-users enabled
5. Test machine enrollment
6. Test AD user functionality

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -33,7 +33,8 @@ path = [
   "overlays/custom-packages/cosmic/cosmic-greeter/0001-Replace-fallback-background-with-Ghaf-default.patch",
   "overlays/custom-packages/cosmic/cosmic-applets/*.patch",
   "overlays/custom-packages/system76-scheduler/0001-fix-add-missing-loop-in-process-scheduler-refresh-ta.patch",
-  "overlays/custom-packages/cosmic/cosmic-greeter/0001-Fix-softlock.patch"
+  "overlays/custom-packages/cosmic/cosmic-greeter/0001-Fix-softlock.patch",
+  "overlays/custom-packages/cosmic/cosmic-greeter/0002-fix-username-handle-empty-usernames.patch"
 ]
 
 [[annotations]]
--- a/overlays/custom-packages/cosmic/cosmic-greeter/0002-fix-username-handle-empty-usernames.patch
+++ b/overlays/custom-packages/cosmic/cosmic-greeter/0002-fix-username-handle-empty-usernames.patch
@@ -1,0 +1,64 @@
+From 029608aded74b2bb303656b04a5137a9afa434c7 Mon Sep 17 00:00:00 2001
+From: Manuel Bluhm <manuel@ssrc.tii.ae>
+Date: Fri, 19 Dec 2025 16:23:36 +0400
+Subject: [PATCH] fix(username): handle empty usernames
+
+Default to prompting username if no user was found, and create
+session only if username is available.
+
+Signed-off-by: Manuel Bluhm <manuel@ssrc.tii.ae>
+---
+ src/greeter.rs | 16 +++++++++++-----
+ 1 file changed, 11 insertions(+), 5 deletions(-)
+
+diff --git a/src/greeter.rs b/src/greeter.rs
+index d917e79..25c1bac 100644
+--- a/src/greeter.rs
++++ b/src/greeter.rs
+@@ -1152,6 +1152,7 @@ impl cosmic::Application for App {
+             .or_else(|| session_names.first().cloned())
+             .unwrap_or_default();
+         let data_idx = Some(0);
++        let entering_name = username.is_empty();
+         let selected_username = NameIndexPair { username, data_idx };
+         let accessibility = Accessibility {
+             helper: cosmic_settings_daemon_config::greeter::GreeterAccessibilityState::config()
+@@ -1171,7 +1172,7 @@ impl cosmic::Application for App {
+             dialog_page_opt: None,
+             dropdown_opt: None,
+             heartbeat_handle: None,
+-            entering_name: false,
++            entering_name,
+             accessibility,
+             theme_builder: Default::default(),
+             randr_list: None,
+@@ -1324,10 +1325,12 @@ impl cosmic::Application for App {
+             Message::Socket(socket_state) => {
+                 self.socket_state = socket_state;
+                 if let SocketState::Open = &self.socket_state {
+-                    // When socket is opened, send create session
+-                    self.send_request(Request::CreateSession {
+-                        username: self.selected_username.username.clone(),
+-                    });
++                    if !self.selected_username.username.is_empty() {
++                        // When socket is opened and username is set, send create session
++                        self.send_request(Request::CreateSession {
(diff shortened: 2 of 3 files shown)


mbssrc:
> ## Tested on Darter Pro (new images)
> ### With homed user enabled (this PR as-is)
> 
>     * Enrolling a local user works
> 
>     * Enrolling an AD user to homed works (required changing the DNS provider IP address)
> 
>     * Automated tests pass with [Active Directory support ci-test-automation#584](https://github.com/tiiuae/ci-test-automation/pull/584)
> 
> 
> ### With AD user enabled
> 
> In `modules/reference/profiles/mvp-user-trial.nix`
> 
> ```
>       # Setup user profiles
>       users.profile = {
>         homed-user.enable = false;
>         ad-users.enable = true;
>         mutable-users.enable = false;
>       };
> ```
> 
>     * Login to AD user works
> 
> 
> **Issues** (I tried to check that none of these were not addressed in the description)
> 
>     * Login on the second boot does not work without internet connection.
> 
>     * The first login attempt seems to always fail. I can see that the password I am typing is correct, but it always fails on first try.
> 
>     * Gui-vm name is `ghaf@ghaf-4257395465` and `janedoe@ghaf-4257395465` instead of `ghaf@gui-vm` and `janedoe@gui-vm`. It is now the same as net-vm.

1. Is that after logging in successfully on first boot? After first login, credentials should be cached offline for the defined period (7 days by default)
2. Will see if it can be solved neatly
3. On purpose - each machine is registered with its hostname in AD - so cannot be gui-vm but has to be the unique machine name

milva-unikie:
>     1. Is that after logging in successfully on first boot? After first login, credentials should be cached offline for the defined period (7 days by default)
Yes. I logged in the first time with an Ethernet connection. Then disconnected the Ethernet cable, rebooted and tried to login again. It did not work, I tried entering the username and password multiple times.

mbssrc:
Changes (+rebased):

- admin user (ghaf): by default, the admin user is now a system user with UID 901. A new flag enables UI login and changes the UID to 1001 (as before), and allows login. This effectively removes the ghaf user from the greeter if the respective flag is not enabled, and removes the admin user enumeration and subsequent session start by the greeter (which then requires their password, and will throw an error even if username is changed as the session is already initiated)

- Fix offline login: setting SSSD's pam verbosity to 0 as the subsequent offline cache message is not well received by the greeter and causes it to hang. Note: please use only a single user for UI login, do not login with both (online will work, offline will fail). This will later be enforced via backend, but is still possible in this test configuration as it requires an extra step

- Fix cosmic-greeter: By default, a session is opened even when no username is available, which causes various errors. The patch enforces that (1) session establishment requires a username (2) username cannot be empty (hangs indefinitely), and (3) username is prompted if none is found

- Execution condition of user provisioning is now again dependent on whether a homed user is available (homed-user) or machine is enrolled (ad-users)

juliuskoskela:
I've tested on X1 and provisioning via AD works as intended including offline login.

brianmcgillion (approved):

PR #1433, opened 2025-09-25
Add tls_config support to alloy server

everton-dematos:
## Description of Changes

Enable mTLS support for the Alloy log forwarder when sending logs to Loki. Improvements can be listed as follows:
- Supports per-target mTLS profiles (custom CA, SNI, min TLS version, and client cert/key), enabling easy personalization for different SIEMs or gateways with their own PKI and policy requirements.
- Allows using a custom CA (internal PKI), which is typical in SIEM environments and zero-trust networks.
- Client identity available (mTLS-ready).
- Transport is TLS-encrypted with explicit policy (no silent downgrade).

### Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [ ] System 76 `x86_64`

### Installation Method
- [ ] Other: 

### Test Steps To Verify:

1. Check if the logs are being uploaded to remote server (e.g., Grafana)
2. On `admin-vm`, run the following command to dump a live graph of all running Alloy components: `curl -s http://127.0.0.1:12345/api/v0/web/components`
3. Look for the `loki.write` block labeled `remote`. You should see: `referencesTo` includes `local.file.tls_cert` and `local.file.tls_key` (and `local.file.tls_ca` if you set a custom CA). This confirms the `tls_config` is wired in.

### DIFF
--- a/modules/common/logging/server.nix
+++ b/modules/common/logging/server.nix
@@ -6,34 +6,76 @@
   ...
 }:
 let
+  inherit (lib)
+    mkIf
+    mkEnableOption
+    mkOption
+    types
+    optionalString
+    optionals
+    ;
+  inherit (lib.strings) hasPrefix;
   cfg = config.ghaf.logging.server;
 in
 {
   options.ghaf.logging.server = {
-    enable = lib.mkEnableOption "Enable logs aggregator server";
-    endpoint = lib.mkOption {
+    enable = mkEnableOption "Enable logs aggregator server";
+    endpoint = mkOption {
       description = ''
         Assign endpoint url value to the alloy.service running in
         admin-vm. This endpoint URL will include protocol, upstream
         address along with port value.
       '';
-      type = lib.types.nullOr lib.types.str;
+      type = types.nullOr types.str;
       default = null;
     };
-    identifierFilePath = lib.mkOption {
+    identifierFilePath = mkOption {
       description = ''
         This configuration option used to specify the identifier file path.
         The identifier file will be text file which have unique identification
         value per machine so that when logs will be uploaded to cloud
         we can identify its origin.
       '';
-      type = lib.types.nullOr lib.types.path;
+      type = types.nullOr types.path;
       example = "/etc/common/device-id";
       default = "/etc/common/device-id";
     };
+
+    tls = {
+      caFile = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "Optional CA bundle for server verification (e.g., /etc/givc/ca-cert.pem). If null, use system CAs.";
+      };
+      certFile = mkOption {
+        type = types.nullOr types.str;
+        default = "/etc/givc/cert.pem";
+        description = "Client certificate (PEM) used for mTLS.";
+      };
+      keyFile = mkOption {
+        type = types.nullOr types.str;
+        default = "/etc/givc/key.pem";
+        description = "Client private key (PEM) used for mTLS.";
+      };
+      serverName = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "Expected TLS server_name (SNI), e.g., loki.example.com (optional).";
+      };
+      minVersion = mkOption {
+        type = types.nullOr (
+          types.enum [
+            "TLS12"
+            "TLS13"
+          ]
+        );
+        default = "TLS12";
+        description = "Minimum TLS version for the outbound connection.";
+      };
+    };
   };
 
-  config = lib.mkIf config.ghaf.logging.server.enable {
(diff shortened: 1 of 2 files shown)


modules/common/logging/server.nix
```diff
@@ -31,6 +31,40 @@ in
       example = "/etc/common/device-id";
       default = "/etc/common/device-id";
     };
+
+    tls = {
+      enable = lib.mkEnableOption "Enable mTLS for Alloy to Loki";
```
vunnyso, line 36: `mkEnableOption` automatically prepends `Whether to enable` before the [description](https://noogle.dev/f/lib/mkEnableOption)

modules/common/logging/server.nix
```diff
@@ -31,6 +31,40 @@ in
...
+      caFile = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        description = "Optional CA bundle for server verification (e.g., /etc/givc/ca-cert.pem). If null, use system CAs.";
+      };
+      certFile = lib.mkOption {
```
vunnyso, line 42: Better to add below snippet in let section and remove adding `lib` prefix.
```nix
 inherit (lib)
    mkEnableOption
    mkOption
    types
    optionalString
    ;
 ```

File: modules/microvm/sysvms/adminvm.nix [hunk 3]
```diff
@@ -63,6 +63,14 @@ let
             logging = {
               server = {
                 inherit (configHost.ghaf.logging) enable;
+                tls = {
+                  enable = true;
+                  caFile = null;
```
vunnyso, line 68: Any reason for not using caFile?
```nix
caFile  = "/etc/givc/ca-cert.pem";
```

vunnyso (changes requested):
Overall good, few minor comments.

modules/common/logging/server.nix
```diff
@@ -6,34 +6,77 @@
...
         we can identify its origin.
       '';
-      type = lib.types.nullOr lib.types.path;
+      type = types.nullOr types.path;
       example = "/etc/common/device-id";
       default = "/etc/common/device-id";
     };
+
+    tls = {
+      enable = mkEnableOption "mTLS for Alloy to Loki connection";
```
mbssrc, line 45: Why an enable flag if it is a requirement?

modules/common/logging/server.nix
everton-dematos: Done.

modules/common/logging/server.nix
everton-dematos: Done.

modules/common/logging/server.nix
```diff
@@ -6,34 +6,77 @@
   ...
```
mbssrc, line 6: Isn't the client configuration missing for mTLS?

modules/microvm/sysvms/adminvm.nix
everton-dematos: For the current setup, I thought of keeping `tls.caFile = null` as our Loki/Grafana uses a public CA (Let’s Encrypt) and validates against the system trust store.
If we change SIEMs or use a private certificate, we can set `tls.caFile` for that setup.

modules/common/logging/server.nix
```diff
@@ -6,34 +6,77 @@
...
+        description = "Expected TLS server_name (SNI), e.g., loki.example.com (optional).";
+      };
+      minVersion = mkOption {
+        type = types.nullOr (
+          types.enum [
+            "TLS12"
+            "TLS13"
+          ]
+        );
+        default = "TLS12";
```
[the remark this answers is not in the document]
everton-dematos: I set a minimum TLS 1.2 to drop TLS 1.0/1.1 while still allowing TLS 1.3. It does not force 1.2. If the server offers 1.3, 1.3 will be established. 
There is no legacy dependency on 1.2. If we want to require 1.3 everywhere, I can bump the default to `"TLS13"`.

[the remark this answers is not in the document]
everton-dematos: In this example, it was using 1.2, then I forced an interruption, it restored with 1.3.

Screenshot from 2025-09-26 15-53-31

milva-unikie:
**Tested on Darter Pro** (`nixos-rebuild switch`)
- Grafana displays logs normally
- `loki.write` block has correct references

PR #1434, opened 2025-09-25
power: allow system vms to shutdown gracefully, preserve audio on shutdown

kajusnau:
## Description of Changes

1. **Added pipewire state dir to persistent storage:**
   - Allows audio state to be saved between reboots

2. **Added gracefulShutdown option to Ghaf common options:**
   - Enabled by default
   - Only affects system-vms
   - If enabled, tells power-manager module to handle shutdown gracefully instead of using microvm's built-in shutdown logic

3. **Power manager module changes:**
   - Host power manager module config now includes overriding microvm's default `ExecStop` behavior for system VMs
   - With this change, when the host is shutting down or a system-vm shutdown is requested, the host will start the respective vm's poweroff.target, which, in turn, will allow the VM to gracefully shutdown
   - The host will wait until the VM process is fully stopped before proceeding with shutdown, or send SIGTERM to the VM if `TimeoutStopSec` is elapsed
     

With these changes, the general shutdown process is as follows:

1. User issues gui-vm shutdown request
2. gui-vm starts shutdown procedures, part of which is signaling the host to shutdown
3. Host receives shutdown signal from gui-vm, starts shutdown procedures
4. Host shutdown procedures tell its' active services to stop via their `ExecStop` binaries
5. All system VMs running under the host will run their `ExecStop` binaries, which now include starting `poweroff.target` on the respective VM
6. Once all `microvm@` services are stopped, host will continue shutting down.

Additionally, these changes allow the display backlight state in gui-vm to be properly stored on shutdown and restored on the next boot.

**Some notes:**
- _After_ this PR, if changes to gui-vm are detected, `nixos-rebuild switch` will attempt to stop `microvm@gui-vm.service`, which, in turn, will trigger the host to shutdown. This is **intended and not unexpected** behavior, although not perfect. My recommendation is to use `nixos-rebuild boot` instead.

### Type of Change
- [x] New Feature
- [x] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [x] Can be updated with `nixos-rebuild ... switch` power-cycle recommended before testing
- [ ] Other: 

### Test Steps To Verify:

Test backlight and audio state restoration:
1. Boot into Ghaf and login
7. Adjust volume and brightness to some values you can easily remember
8. Reboot or shutdown the system
9. Boot into Ghaf and login
10. Verify volume and brightness values are restored

Test (any) system vm shutdown
`audio-vm` is a good choice here
1. Boot into Ghaf, no need to login
2. Open two separate terminals, both in ghaf-host
3. In the first, run `journalctl -f -u microvm@<sysvm-name>.service`
4. In the second, run `sudo systemctl stop microvm@<sysvm-name>.service`
5. Observe the shutdown process in the first terminal
11. Verify the VM shutdown successfully with `systemctl status microvm@<sysvm-name>.service` - there should be no non-zero exit codes in the execution of `ExecStop`

### DIFF
--- a/modules/common/services/power.nix
+++ b/modules/common/services/power.nix
@@ -362,8 +362,12 @@ in
     };
     host = {
       enable = mkEnableOption ''
-        host power management profile. This profile manages the hosts pre- and post-suspend actions to coordinate guest
-        suspend actions and devices
+        Host power management profile. This profile manages the host's pre- and post-suspend actions
+        to coordinate guest suspend actions and devices.
+
+        Additionally, if a system VM has `ghaf.gracefulShutdown = true`, enabling this host profile
+        allows the host to override the VM's default microvm ExecStop logic, starting
+        the guest's `poweroff.target` and waiting for the VM process to exit.
       '';
     };
   };
@@ -456,14 +460,14 @@ in
 
       # Shutdown displays early before suspend
       powerManagement = {
+        # This is misleading, as it is executed only on suspend, not shutdown
         powerDownCommands = lib.mkBefore ''
-          ${lib.getExe ghaf-powercontrol} turn-off-displays '*'
+          ${getExe ghaf-powercontrol} turn-off-displays '*'
         '';
       };
 
       # Override systemd actions for suspend, poweroff, and reboot
       systemd.services = optionalAttrs (cfg.vm.enable && useGivc) {
-
         # Replace systemd-sleep with GIVC suspend
         systemd-suspend.serviceConfig.ExecStart = [
           ""
@@ -474,12 +478,9 @@ in
         gui-shutdown-interceptor = {
           description = "Ghaf GUI shutdown interceptor";
           wantedBy = [
-            "reboot.target"
-            "poweroff.target"
+            "shutdown.target"
           ];
           before = [
-            "reboot.target"
-            "poweroff.target"
             "shutdown.target"
           ];
           unitConfig.DefaultDependencies = false;
@@ -488,17 +489,27 @@ in
             ExecStart = "${getExe guest-shutdown-interceptor}";
           };
         };
-
       };
 
       # Logind configuration for desktop
-      services.logind.settings.Login = {
-        HandleLidSwitch = if cfg.allowSuspend then "suspend" else "ignore";
-        KillUserProcesses = true;
-        IdleAction = "lock";
-        UserStopDelaySec = 0;
-        HoldoffTimeoutSec = 20;
-      };
+      services.logind.settings.Login =
+        let
+          lidEvent = if cfg.allowSuspend then "suspend" else "lock";
+        in
+        mkDefault {
+          HandleLidSwitch = lidEvent;
+          HandleLidSwitchDocked = lidEvent;
+          HandleLidSwitchExternalPower = lidEvent;
+          HandleSuspendKey = "ignore";
+          HandleHibernateKey = "ignore";
+          HandlePowerKey = "ignore";
+          HandlePowerKeyLongPress = "ignore";
+          KillUserProcesses = true;
+          IdleAction = "lock";
+          IdleActionSec = "10min";
+          UserStopDelaySec = 0;
+          HoldoffTimeoutSec = 20;
+        };
     })
 
     # Host power management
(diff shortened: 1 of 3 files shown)


kajusnau:
### Note for QA
Please wait for any possible comments from reviewers and discussions to be resolved before testing.
Changes are still likely to happen

modules/common/services/power.nix
```diff
@@ -229,25 +230,51 @@ let
...
-      fi
-    '';
+    text =
+      let
+        normalizeUnit = u: if hasSuffix ".service" u || hasSuffix ".target" u then u else "${u}*";
+        startCmds = concatStringsSep "\n" (
+          map (unit: ''
+            echo "Starting ${unit}..."
+            ${lib.getExe' pkgs.systemd "systemctl"} start ${normalizeUnit unit} || true
+          '') cfg.gui.shutdown.startUnits
```
vunnyso, line 248: Could we make it more adaptable to handle `shutdown.startUnits` for other VMs as well? We're experiencing a similar issue with `audio-vm`, where some services are not shutting down properly, preventing  proper system shutdown.

modules/common/services/power.nix
```diff
@@ -229,25 +230,51 @@ let
...
+                  echo "Found prepare for shutdown signal. Checking type..."
+                  while read -r subline; do
+                      if echo "$subline" | grep -q "reboot"; then
+                          ${startCmds}
+                          ${stopCmds}
+                          ${getExe guest-power-actions} reboot
+                      elif echo "$subline" | grep -q "poweroff"; then
+                          ${startCmds}
+                          ${stopCmds}
+                          ${getExe guest-power-actions} poweroff
```
vunnyso, line 272: Do we need to handle else case with `exit 0` ?

modules/common/services/power.nix
```diff
@@ -229,25 +230,51 @@ let
...
+        startCmds = concatStringsSep "\n" (
+          map (unit: ''
+            echo "Starting ${unit}..."
+            ${lib.getExe' pkgs.systemd "systemctl"} start ${normalizeUnit unit} || true
+          '') cfg.gui.shutdown.startUnits
+        );
+        stopCmds = concatStringsSep "\n" (
+          map (unit: ''
+            echo "Stopping ${unit}..."
+            ${lib.getExe' pkgs.systemd "systemctl"} stop ${normalizeUnit unit} || true
```
vunnyso, line 253:
```suggestion
            ${getExe' pkgs.systemd "systemctl"} stop ${normalizeUnit unit} || true
```

modules/common/services/power.nix
kajusnau: The types of shutdown sent via this signal are "power-off", "reboot", "halt", "kexec" or "soft-reboot".
I think for now we can safely ignore "halt", "kexec", and "soft-reboot", as I haven't noticed those signals being sent by any gui-vm component as of now.

We don't want to `exit 0` because that would stop the execution of the service, so if anything other than poweroff or reboot is requested, the shutdown would stop being inhibited. This would mean any follow-up requests to shutdown would be let through and gui-vm would obey.

Ref. https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.login1.html#Signals

modules/common/services/power.nix
kajusnau: Outside the scope of this PR, but we could potentially declare some helper services in gui-vm to start some units in other VMs, for example:
```
services."poweroff-audio-vm" = {
  enable = true;
  description = "Allow clean shutdown for audio-vm";
  serviceConfig = {
    Type = "oneshot";
    ExecStart = "${lib.getExe' pkgs.givc-cli "givc-cli"} ${config.ghaf.givc.cliArgs} start service --vm 'audio-vm' poweroff.target";
  };
};
```

modules/common/services/power.nix
```diff
@@ -456,49 +512,55 @@ in
 
       # Shutdown displays early before suspend
       powerManagement = {
+        # This is misleading, as it is executed only on suspend, not shutdown
```
mbssrc, line 515: This is true - the description in nixpkgs is wrong. It could be adjusted easily here or in nixpkgs to be run on shutdown as well by adding the correct targets

nesteroff:
Regarding the stopUnits workaround, if I understand the problem correctly the microvm shutdown command doesn’t work so all VMs just get terminated with SIGTERM during host shutdown. The guivm is where the shutdown is initiated so it starts shutting down but also gets killed with SIGTERM partway through once the host begins stopping systemd services. As a result, some services in guivm don’t shut down cleanly and we need to force-stop them with stopUnits before the host terminates guivm.

What if we used something like this to wait until guivm fully shuts down and all services in the guest are stopped:
```
systemd.services."microvm@gui-vm".serviceConfig.ExecStop = ''
  /bin/sh -c 'echo "QEMU PID=$MAINPID"; \
    while kill -0 $MAINPID 2>/dev/null; do \
      sleep 1; \
    done; \
    echo "QEMU PID=$MAINPID stopped"'
'';
```

We could also forward the shutdown command to other VMs using givc and apply the same idea to wait for them to stop gracefully. Would that approach make sense?

kajusnau:
> What if we used something like this to wait until guivm fully shuts down and all services in the guest are stopped:
> 
> ```
> systemd.services."microvm@gui-vm".serviceConfig.ExecStop = ''
>   /bin/sh -c 'echo "QEMU PID=$MAINPID"; \
>     while kill -0 $MAINPID 2>/dev/null; do \
>       sleep 1; \
>     done; \
>     echo "QEMU PID=$MAINPID stopped"'
> '';
> ```
> 
> We could also forward the shutdown command to other VMs using givc and apply the same idea to wait for them to stop gracefully. Would that approach make sense?


I like the idea but this would require some decent amount of work on givc side

Could another alternative be to simply hook a new service into the `shutdown.target` on the host and configure a shutdown procedure for the VMs?
Maybe this service could then issue `givc` calls to trigger `poweroff.target` on all VMs in a certain order, and wait for the corresponding `microvm@<vm>.services` to stop?

Anyway, I think this may be outside the scope of this PR

nesteroff:
Probably there are two separate issues here: problems with forwarding shutdown from the guivm to the host and graceful VM shutdown. Personally, I haven’t noticed issues with the first one but VMs don’t seem to shut down cleanly on my system.

From what I can see, microvm tries to shut them down by sending a Ctrl+Alt+Del sequence and then waiting for the socket to disconnect:
https://github.com/microvm-nix/microvm.nix/blob/main/lib/runners/qemu.nix#L341

However, this doesn’t work in our case. They use socat to wait for input from stdio, which ends up being /dev/null when running under systemd. Because of that, the command terminates immediately, systemd sees the process as still active and then kills it with SIGTERM. The same happens with guivm but since it initiates the shutdown, it goes through part of the process before receiving SIGTERM.

Waiting for the QEMU process to exit helps prevent early guivm termination and it also fixes brightness issues because it stops systemd-backlight during shutdown and other services as well. It's interesting that VMs shut down cleanly on your end. I’m not sure why there’s a difference. Anyway, just sharing what I noticed during my previous testing.

kajusnau:
> It's interesting that VMs shut down cleanly on your end. I’m not sure why there’s a difference. Anyway, just sharing what I noticed during my previous testing.

In my case I was issuing `systemctl start poweroff.target` on the VM itself, and allowing it to shutdown, rather than stopping the microvm@<vm>.service from the host side

kajusnau:
> After adding code to wait for the QEMU process to stop, guivm terminates gracefully, brightness is saved and restored on next boot

Interesting! I’ll come back to this PR next week and try to rework it with your suggestion

nesteroff:
> Interesting! I’ll come back to this PR next week and try to rework it with your suggestion

Thanks and sorry if this causes extra work. I guess it would be nice to fix the shutdown in microvm.nix, maybe by using qemu-guest-agent or something like that but it’s more complex and probably out of scope for now.

vunnyso (approved):
LGTM

kajusnau:
NixOS power-management module adjustment PR open at https://github.com/NixOS/nixpkgs/pull/447426

modules/common/services/power.nix
kajusnau: https://github.com/NixOS/nixpkgs/pull/447426

kajusnau:
Updated: after a discussion with Manuel, added guards to check for givc for both gracefulShutdown option and the host microvm services adjustments

mbssrc (approved):

PR #1444, opened 2025-10-03
Enable TLS for alloy client to server

everton-dematos:
## Description of Changes

- Switch Alloy clients to `https` and add `tls_config` (CA/cert/key, TLS ≥1.2) delivered via systemd `LoadCredential`. 
- On `admin-vm`, keep Alloy behind `stunnel`, which terminates TLS and enforces mTLS on the public port. 
- The `stunnel` was used as a lightweight, low-dep TLS terminator vs heavier proxies.

Before, there was no encryption in the `alloy` client to server communication:

Screenshot from 2025-10-03 15-53-07

After this patch:

Screenshot from 2025-10-03 15-55-35

- The `stunnel` service performance was measured for 10 minutes after boot:
  - Average CPU usage:   0.04%
  - Max CPU usage:       1.18% at t=26s
  - Average Memory:      4.25 MiB
  - Max Memory:          7.57 MiB at t=2s
stunnel_metrics_stunnel_20251003_104729

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [ ] System 76 `x86_64`

### Installation Method
- [ ] Other: 

### Test Steps To Verify:

1. Check if the logs are being uploaded to remote server (e.g., Grafana)
2. The alloy service in both clients (all VMs except admin-vm), and server (admin-vm) should be working fine: `systemctl status alloy.service `
3. The admin-vm should have the `stunnel` service working fine: `systemctl status stunnel.service`
4. (Optional) It is possible to add the `tshark` package to a specific VM to record the traffic and analyze the output `pcap` file:
- For example, we can add  `++ [ pkgs.wireshark-cli ]` into `gui-vm` - https://github.com/tiiuae/ghaf/blob/main/modules/microvm/sysvms/guivm.nix#L209
- Then, it is possible to monitor, from `gui-vm`, the logs being sent to `admin-vm`: `sudo tshark -i any -p -s 0 -f 'tcp port 9999 and host 192.168.100.5' \
  -a duration:300 \
  -w /tmp/loki_to_adminvm_$(date +%Y%m%d_%H%M%S).pcapng`
- You can use wireshark to analyze the output pcap

### DIFF
--- a/modules/common/logging/client.nix
+++ b/modules/common/logging/client.nix
@@ -6,28 +6,68 @@
   ...
 }:
 let
+  inherit (lib)
+    mkIf
+    mkEnableOption
+    mkOption
+    types
+    optionalString
+    ;
   cfg = config.ghaf.logging.client;
   inherit (config.ghaf.logging) listener;
 in
 {
   options.ghaf.logging.client = {
-    enable = lib.mkEnableOption "Enable the alloy client service";
-    endpoint = lib.mkOption {
+    enable = mkEnableOption "Alloy client service";
+    endpoint = mkOption {
       description = ''
         Assign endpoint url value to the alloy.service running in
         different log producers. This endpoint URL will include
         protocol, upstream, address along with port value.
       '';
-      type = lib.types.str;
-      default = "http://${listener.address}:${toString listener.port}/loki/api/v1/push";
+      type = types.str;
+      default = "https://${listener.address}:${toString listener.port}/loki/api/v1/push";
+    };
+
+    tls = {
+      caFile = mkOption {
+        type = types.nullOr types.path;
+        default = "/etc/givc/ca-cert.pem";
+        description = "CA bundle used to verify the admin-vm TLS terminator certificate.";
+      };
+      certFile = mkOption {
+        type = types.nullOr types.path;
+        default = "/etc/givc/cert.pem";
+        description = "Client certificate (PEM) used for mTLS to the admin-vm.";
+      };
+      keyFile = mkOption {
+        type = types.nullOr types.path;
+        default = "/etc/givc/key.pem";
+        description = "Client private key (PEM) used for mTLS to the admin-vm.";
+      };
+      minVersion = mkOption {
+        type = types.nullOr (
+          types.enum [
+            "TLS12"
+            "TLS13"
+          ]
+        );
+        default = "TLS12";
+        description = "Minimum TLS version for the outbound connection.";
+      };
     };
   };
-  config = lib.mkIf cfg.enable {
+
+  config = mkIf cfg.enable {
 
     assertions = [
       {
         assertion = listener.address != "";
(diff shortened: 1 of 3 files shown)


modules/common/logging/client.nix
```diff
@@ -6,29 +6,69 @@
...
       '';
-      type = lib.types.str;
-      default = "http://${listener.address}:${toString listener.port}/loki/api/v1/push";
+      type = types.str;
+      default = "https://${listener.address}:${toString listener.port}/loki/api/v1/push";
+    };
+
+    tls = {
+      caFile = mkOption {
+        type = types.nullOr types.str;
```
vunnyso, line 34: As `caFile` is using path to file, so it will be good to `path` instead of `str`
```suggestion
        type = types.nullOr types.path;
```
Same goes to `certFile` and `KeyFile`

modules/common/logging/client.nix
```diff
@@ -6,29 +6,69 @@
...
+    optionalString
+    ;
   cfg = config.ghaf.logging.client;
   inherit (config.ghaf.logging) listener;
 in
 {
   options.ghaf.logging.client = {
-    enable = lib.mkEnableOption "Enable the alloy client service";
-    endpoint = lib.mkOption {
+    enable = mkEnableOption "Enable the alloy client service";
```
vunnyso, line 21: `mkEnableOption` automatically prepends `Whether to enable` before the [description](https://noogle.dev/f/lib/mkEnableOption)

modules/common/logging/server.nix
```diff
@@ -72,6 +77,19 @@ in
         default = "TLS12";
         description = "Minimum TLS version for the outbound connection.";
       };
+
+      terminator = {
+        backendPort = mkOption {
+          type = types.port;
+          default = 3101;
+          description = "HTTP backend port for Alloy when TLS terminator is enabled.";
```
vunnyso, line 85: `terminator` is HTTP?

modules/common/logging/client.nix
everton-dematos: Done. I changed for both `client` and `server`.

modules/common/logging/client.nix
everton-dematos: Done.

modules/common/logging/server.nix
everton-dematos: Yes, the terminator’s backend is plain HTTP.

`stunnel` takes HTTPS/mTLS on the public port, terminates TLS there, and forwards the request to Alloy over HTTP on `backendPort`. That hop is intended to be local-only (same VM), so TLS is only needed on the public side.

PR #1452, opened 2025-10-07
chrome: revert dedicated profiles for apps and browser

kajusnau:
## Description of Changes

1. **Disable use of `--profile-directory` flag for chrome:**
   - Affects `business-vm` and `comms-vm`
   - Sharing chrome tabs to e.g. Teams is now the same as it used to be
   - Fixes https://jira.tii.ae/browse/SSRCSP-7292 but re-introduces https://jira.tii.ae/browse/SSRCSP-7164

### Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [x] Requires full re-installation - Clean install or wipe `/home/appuser/.config/google-chrome` in the respective VM
- [ ] Other: 

### Test Steps To Verify:

1. Test chrome tab sharing functionality in Teams, Trusted Browser, Slack, or wherever else.

### DIFF
--- a/modules/reference/appvms/business.nix
+++ b/modules/reference/appvms/business.nix
@@ -141,7 +141,7 @@ in
           description = "Isolated Trusted Browsing";
           packages = [ trustedBrowserWrapper ];
           icon = "thorium-browser";
-          command = "trusted-browser-wrapper --profile-directory=TrustedBrowserProfile";
+          command = "trusted-browser-wrapper";
           givcArgs = [
             "url"
           ];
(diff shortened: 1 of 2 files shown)


modules/reference/appvms/business.nix
```diff
@@ -141,7 +141,7 @@ in
           description = "Isolated Trusted Browsing";
           packages = [ trustedBrowserWrapper ];
           icon = "thorium-browser";
-          command = "trusted-browser-wrapper --profile-directory=TrustedBrowserProfile";
```
vunnyso, line 144: If feasible, implementing it through `mkOption` would be beneficial as it would facilitate easier future transitions if needed.

kajusnau: We currently don't expose options for any app-vm declarations

While this could be added as an option to `modules/reference/programs/google-chrome.nix`, I don't think it's worth it:
- We'd need a way to declare the profile names within the business/comms vm
- The final declaration would likely end up quite a bit messier in the end
- The same functionality should be added to `modules/reference/programs/chromium.nix` on the off-chance we end up switching to chromium

Personally I don't see a future need for this as it was quite dirty to begin with.

If other reviewers agree that an option should be added, I can look into it :saluting_face:

vunnyso: Alright, let's revisit this in the future if required.

vunnyso (approved):

PR #1531, opened 2025-11-04
Fix atomic user creation to prevent boot failure on interruption

vadika:
- Add cleanup of partial systemd-homed state at script start
- Add ExecStopPost cleanup handler for interrupted service
- Ensure user creation is atomic: either fully created or start fresh
- Fixes boot failure when setup-ghaf-user is interrupted (e.g., Ctrl-Alt-Del)

Previously, if the system was rebooted during user creation, partial state in /var/lib/systemd/home/* could prevent subsequent boots because the lock file wasn't created but user data was partially written.

### Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [ ] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

 
 If system is rebooted during 1st time boot when the user creation wizard is running, it should not lock up on next boot.

### DIFF
--- a/modules/common/users/desktop.nix
+++ b/modules/common/users/desktop.nix
@@ -181,6 +181,25 @@ in
                   # Consider 'systemctl stop' as normal exit
                   trap 'exit 0' TERM
 
+                  # Clean up any partial state from previous interrupted runs
+                  # This ensures atomic behavior - either user is fully created or we start fresh
+                  if [ ! -f /var/lib/nixos/user.lock ]; then
+                    echo "Cleaning up any partial user creation state..."
+                    # Remove any partial homed user records
+                    if [ -d /var/lib/systemd/home ]; then
+                      for user_file in /var/lib/systemd/home/*.home; do
+                        [ -e "$user_file" ] || continue
+                        if [ -f "$user_file" ]; then
+                          username=$(basename "$user_file" .home)
+                          echo "Removing incomplete user: $username"
+                          homectl remove "$username" 2>/dev/null || true
+                        fi
+                      done
+                      # Clean up any remaining files
+                      rm -rf /var/lib/systemd/home/* 2>/dev/null || true
+                    fi
+                  fi
+
                   brightnessctl set 100%
 
                   SETUP_COMPLETE=false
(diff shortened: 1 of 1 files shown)


modules/common/users/desktop.nix
```diff
@@ -181,6 +181,25 @@ in
...
+                          username=$(basename "$user_file" .home)
+                          echo "Removing incomplete user: $username"
+                          homectl remove "$username" 2>/dev/null || true
+                        fi
+                      done
+                      # Clean up any remaining files
+                      rm -rf /var/lib/systemd/home/* 2>/dev/null || true
+                    fi
+                  fi
+
```
kajusnau: At a glance, this snippet is similar to `cleanupScript`. Would it be possible to just run `cleanupScript` instead and keep it dry?

I would even prefer we run this snippet (or `cleanupScript`) as part of **ExecStartPre** for this service

modules/common/users/desktop.nix
```diff
@@ -327,6 +369,7 @@ in
                 TTYReset = true;
                 TTYVHangup = true;
                 ExecStart = "${userSetupScript}/bin/setup-ghaf-user";
+                ExecStopPost = "${cleanupScript}/bin/cleanup-ghaf-user";
```
kajusnau: I would prefer we run this as part of **ExecStartPre** for this service instead.
That way we should cover cases where a force shutdown happens and state is dirty due to **ExecStopPost** failing to execute, which relies entirely on a clean shutdown.

The condition check on line 362 may prevent this approach, in which case it may be possible to do the cleanup (and check for existing account) as part of [**ExecCondition**](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#ExecCondition=)?

To make it short, IMO it makes more sense to cleanup any possible dirty states before the user creation is started _only_, as relying on **ExecStopPost** is iffy 🤔

kajusnau (changes requested):
Great change! Some notes

- Would like to keep it dry

- I think it may be better for the flow to be **check for existing user -> cleanup if necessary -> create user**

modules/common/users/desktop.nix
kajusnau: If going with this approach, the cleanup operations in `userSetupScript` could be removed.

PR #1607, opened 2025-12-02
Firmware control

brianmcgillion:
### Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [ ] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

1. ...

### DIFF
--- a/modules/development/debug-tools.nix
+++ b/modules/development/debug-tools.nix
@@ -9,7 +9,6 @@
 let
   cfg = config.ghaf.development.debug.tools;
 
-  rm-linux-bootmgrs = pkgs.callPackage ./scripts/rm_linux_bootmgr_entries.nix { };
   sysbench-test-script = pkgs.callPackage ./scripts/sysbench_test.nix { };
   sysbench-fileio-test-script = pkgs.callPackage ./scripts/sysbench_fileio_test.nix { };
   nvpmodel-check = pkgs.callPackage ./scripts/nvpmodel_check.nix { };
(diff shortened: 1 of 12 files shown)


modules/development/dt-camera.nix
```diff
@@ -0,0 +1,26 @@
...
+  options.ghaf.development.debug.tools.camera = {
+    enable = lib.mkEnableOption "Camera Debugging Tools";
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages =
+      lib.optionals (config.nixpkgs.hostPlatform.system == "x86_64-linux")
+        (rmDesktopEntries [
+          pkgs.v4l-utils
+          pkgs.ffmpeg_7-full
```
vunnyso: `ffmpeg_7` is primarily utilized for audio and video playback purposes. Should we categorize it under camera tools? Additionally, it might be more advantageous to use `ffmpeg` instead, ensuring we always have the latest version unless there is a specific requirement for the full version.

brianmcgillion: I just moved it from debug tools common to here. it is used for testing the camera in the business vm. If we want to change the version we can do that in a follow on PR. the idea here was to just move things around and not duplicate where not needed. if we can use the ffmpeg_7 that would be good. as it will save another 400 MB

PR #1634, opened 2025-12-12
(feat): Enable full disk encryption

vunnyso:
## Description of Changes
This change modifies the disk partitioning scheme to use `LVM on LUKS`, compared to the previous scheme which used `LUKS on LVM`. The advantages of the new scheme include reduced layout complexity and the ability to enable full disk encryption. With this approach, there is no need to encrypt different partitions separately.

### Encryption
Disk encryption will be enabled for both debug and release images.  To avoid entering PIN or password on every reboot. The PIN or password requirement applies exclusively to release images.

#### **Authentication:**
In debug builds, the boot process remains unchanged. However in release builds, users will be prompted to set a PIN or password during the first boot. For subsequent boots, they will need to enter the same PIN or password to unlock the system.

***
**Partitioning layout details:**
----

**Encrypted area highlighted.**

**With this PR**:
image

**With this PR** (if `storage.encryption.enable = false`):
image

**With mainline**:
image

***

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

1. Make sure `Partitioning layout` matches as mentioned in `Description of Changes`
2. Please test with debug and release builds.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -98,6 +98,7 @@ export default defineConfig({
                         "ghaf/dev/ref/wireguard-gui",
                         "ghaf/dev/ref/yubikey",
                         "ghaf/dev/ref/ghaf-vms",
+                        "ghaf/dev/ref/disk-encryption",
                       ],
                     },
                     {
--- a/docs/src/content/docs/ghaf/dev/ref/disk-encryption.mdx
+++ b/docs/src/content/docs/ghaf/dev/ref/disk-encryption.mdx
@@ -1,0 +1,40 @@
+---
+title: Disk Encryption
+description: How Ghaf uses LVM on LUKS for flexible and secure full disk encryption.
+---
+
+# LVM on LUKS for Full Disk Encryption
+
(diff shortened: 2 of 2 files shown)


kajusnau (commented):
Great work!

Not my area of expertise so I'll leave proper review to others, just leaving a small suggestion. 🍻

brianmcgillion:
2 things though. you say in the PR description it is only for release but it is enabled in mvp-user-trial. so on all images. which is great actually. it should be on by default for all to get it tested.

Also you mention that it is a pin. is is a pin or a password?

milva-unikie:
(Vunny will check the image size, not ready for testing yet)

vadika:
This needs to be refactored to be harmonized with A/B updates scheme. It need to use systemd-repart for partitioning!!!

PR #1697, opened 2026-01-22
feat(about): add about-ghaf application

mbssrc:
## Description of Changes

Add a local "Getting Started" application to the laptop. It is a local website that entails information about the laptop and its features. Useful for users to understand what they are looking at.

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

Please proof-read and verify the tutorials.

### DIFF
--- a/packages/pkgs-by-name/ghaf-intro/index.html
+++ b/packages/pkgs-by-name/ghaf-intro/index.html
@@ -9,18 +9,19 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Welcome to Ghaf</title>
     <style>
+      /* Ghaf dark theme - matches docs/src/styles/custom.css */
       :root {
-        --bg-primary: #0a0a0f;
-        --bg-secondary: #12121a;
-        --bg-card: #1a1a25;
-        --text-primary: #e8e8ed;
-        --text-secondary: #b8b8c8;
-        --accent-green: #218838;
-        --accent-blue: #337aff;
-        --accent-red: #9c0000;
-        --accent-orange: #ffa500;
-        --accent-purple: #122263;
-        --border-color: #2a2a3a;
+        --bg-primary: #0d1310;
+        --bg-secondary: #141f1a;
+        --bg-card: #2d3c34;
+        --text-primary: #ffffff;
+        --text-secondary: #c4ccc8;
+        --accent-green: #007f3b;
+        --accent-blue: #3b82f6;
+        --accent-red: #ef4444;
+        --accent-orange: #f59e0b;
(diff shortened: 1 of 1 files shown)


milva-unikie:
**Tested on Darter Pro**
- Skimmed through the `Getting Started` guide. Looks good!
- Darter Pro and Orin tests have passed in the pre-merge pipeline so I would say this PR is good to go in. Lenovo X1 and Dell 7330 tests are still running, but there should be no issues with them.

packages/pkgs-by-name/ghaf-intro/index.html
```diff
@@ -0,0 +1,3014 @@
...
+        </p>
+
+        <p>
+          When one application is compromised on a traditional system, the
+          attacker typically gains access to everything that user can access:
+          emails, documents, browser sessions, saved passwords, and the ability
+          to persist across reboots.
+        </p>
+
+        <h3>Ghaf's Multi-Layer Defense</h3>
```
kajusnau: For this section I have a couple notes:
- Using (arbitrary?) colors here IMO doesn't add anything to the overall presentation, and can even clash with some of our existing color schemes. e.g. The Green used for "layer 1" is almost identical to the business VM green, red in "Result" is almost the same as chrome VM red
- If these are actual layers would it help to represent them as such? As a sort of stack
- The result is not a layer, so it would help if it was a visually distinct element

kajusnau (changes requested):
Really nice looking page! Easy to navigate, clear wording, good tutorials.

Some objective, some subjective notes in my review, as well as some general notes:

- I think the use of colors is a bit excessive in general. In some cases it makes some text more difficult to read (light text against a light-colored background), in other cases it clashes with pre-existing color definitions, as I mention in the comments.
- Depending on the target audience, it could be argued that the language should be more formal:
   ```
   can’t -> cannot
   won’t -> will not
   don’t -> do not
   doesn’t -> does not
   didn’t -> did not
   isn’t -> is not
   etc
   ```
   
   
Late edit:
I tested https://github.com/tiiuae/ghaf/commit/7db1c4a91c19a10cbdb6e401f4900b204efaff26, so maybe not everything applies.

kajusnau (approved):
Approving so as not to block the PR in case we need to move forward quickly.

PR #1742, opened 2026-02-10
feat(jetson-orin): add generic LUKS flash flow

vadika:
## Description of Changes

Opt-in LUKS root filesystem encryption for Jetson Orin (`ghaf.hardware.nvidia.orin.diskEncryption.*`), with initrd unlock via `/dev/mapper/cryptroot`.

The root is encrypted at **image build time** (`sdimage.nix`, `cryptsetup reencrypt --encrypt`), so flashing is unattended and writes an already-encrypted APP partition. `*-luks-*` targets unlock with a device-unique key from OP-TEE: the first boot rotates away from the manufacturing passphrase and re-encrypts the volume key, and later boots unlock without user input. `userPassphrase` is available as a testing-only alternative.

Supporting changes:

- The encrypted container ships on two partition tables — the raw sd-image on USB (MBR) and the NVIDIA flash on eMMC/NVMe (GPT) — which necessarily carry different PARTUUIDs. The root is therefore identified by the pinned LUKS2 **header** UUID (`diskEncryption.luksUuid`, `readOnly`), identical on both media and independent of `mmcblk`/`nvme`/`sda` enumeration. The ESP is identified by its FAT label.
- New `flashScriptOverrides.deviceDisk` / `deviceDiskRootfsPartition` / `deviceDiskEspPartition`, replacing the single hardcoded `rootfsDevice`. No defaults — rootfs storage varies per carrier board, and an assertion requires all three. This lets Orin NX flash to NVMe while still supporting USB boot.
- LUKS and dm-verity are mutually exclusive root strategies (verity owns `fileSystems."/"` as a tmpfs overlay). An assertion rejects the combination, and the verity targets no longer generate `-luks` variants.
- LUKS targets stay on `upstream-6-6` and add the kernel config (`BLK_DEV_DM`, `DM_CRYPT`, related dm/crypto symbols) needed for device-mapper at boot.
- Docs updated for the LUKS build and flash flow.

Hardware-verified on an Orin AGX devkit (eMMC) and Orin NX (NVMe): the board resolves `/dev/disk/by-uuid/<uuid>`, LUKS unlocks to `/dev/mapper/cryptroot`, the first-boot resize grows the encrypted filesystem, and the system boots to `ghaf-host login:`.

### Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [x] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

### Applicable Targets

- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

Four configurations are worth covering: AGX/eMMC and NX/USB with the device-unique key, a non-LUKS target as a regression check, and `userPassphrase` via a manual override. Full matrix with per-case commands and expected output is in [this comment](https://github.com/tiiuae/ghaf/pull/1742#issuecomment-4927222231).

The short version, for `nvidia-jetson-orin-agx-debug-luks`:

1. Build the image and flash script:
   ```sh
   nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64
   nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64-flash-script
   ```
2. Put the board in recovery mode and flash. No passphrase prompt:
   ```sh
   ```
3. Watch the first boot on the serial console. It rotates to the device-unique key and re-encrypts the root (~1 min/GB on a 32GB AGX), then reaches `ghaf-host login:` without ever prompting.
4. Verify the encrypted root:
   ```sh
   findmnt /                                    # /dev/mapper/cryptroot
   sudo blkid /dev/mmcblk0p2                    # TYPE="crypto_LUKS"
   sudo cryptsetup luksDump /dev/mmcblk0p2      # luksDiskDeviceUniqueKey token
   df -h /                                      # grown to fill the disk
   ```
   The manufacturing passphrase must no longer unlock the disk:
   ```sh
   echo ghaf | sudo cryptsetup open --test-passphrase /dev/mmcblk0p2   # must fail
   ```
5. Reboot. It must be fast — no re-encryption, no prompt — and the resize skips via a marker on the ESP:
   ```sh
   sudo journalctl -b -u resize-partitions      # "Resize already performed, skipping."
   ```

### DIFF
--- a/modules/profiles/orin.nix
+++ b/modules/profiles/orin.nix
@@ -29,6 +29,38 @@
 let
   cfg = config.ghaf.profiles.orin;
   hostGlobalConfig = config.ghaf.global-config;
+  ensureSystemProfile = pkgs.writeShellApplication {
+    name = "ghaf-ensure-system-profile";
+    runtimeInputs = with pkgs; [
+      coreutils
+      nix
+    ];
+    text = ''
+      profile=/nix/var/nix/profiles/system
+      registration=/nix-path-registration
+      current_system=$(readlink -f /run/current-system)
+      generation_link=/nix/var/nix/profiles/system-1-link
+
+      if [ -z "$current_system" ] || [ ! -e "$current_system" ]; then
+        echo "Current system closure is unavailable" >&2
+        exit 1
+      fi
+
+      if [ ! -f "$registration" ] && [ -L "$profile" ] && [ "$(readlink -f "$profile")" = "$current_system" ]; then
+        exit 0
+      fi
+
+      if [ -f "$registration" ]; then
+        nix-store --load-db < "$registration"
+        rm -f "$registration"
+        touch /etc/NIXOS
+      fi
+
+      mkdir -p /nix/var/nix/profiles
+      ln -sfn "$current_system" "$generation_link"
+      ln -sfn system-1-link "$profile"
+    '';
+  };
 in
 {
   _file = ./orin.nix;
@@ -238,6 +270,17 @@ in
     environment.sessionVariables = {
       GBM_BACKEND = "dri";
       __EGL_VENDOR_LIBRARY_FILENAMES = "/run/opengl-driver/share/glvnd/egl_vendor.d/50_mesa.json";
+    };
+
+    systemd.services.ghaf-ensure-system-profile = {
+      description = "Ensure persistent NixOS system profile exists";
+      wantedBy = [ "multi-user.target" ];
+      after = [ "local-fs.target" ];
+      before = [ "nix-gc.service" ];
+      serviceConfig = {
+        Type = "oneshot";
+        ExecStart = "${ensureSystemProfile}/bin/ghaf-ensure-system-profile";
+      };
     };
 
     # Cosmic on orin
--- a/modules/reference/hardware/jetpack/agx/orin-agx-industrial.nix
+++ b/modules/reference/hardware/jetpack/agx/orin-agx-industrial.nix
@@ -24,7 +24,11 @@
         agx.enableNetvmWlanPCIPassthrough = true;
         carrierBoard = "devkit";
         # AGX industrial devkit boots rootfs from eMMC.
-        flashScriptOverrides.rootfsDevice = "mmcblk0p1";
+        flashScriptOverrides = {
+          deviceDisk = "mmcblk0";
+          deviceDiskEspPartition = "mmcblk0p1";
+          deviceDiskRootfsPartition = "mmcblk0p2";
+        };
       };
 
       # Net VM hardware-specific modules - use hardware.definition for composition model
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
@@ -5,12 +5,27 @@
 {
   lib,
   config,
+  options,
   pkgs,
   ...
 }:
 let
   cfg = config.ghaf.hardware.nvidia.orin;
 
+  # verity-volume.nix owns fileSystems."/" (tmpfs overlay) when enabled, so the
+  # diskEncryption ext4 root override below must stand down to avoid a
+  # conflicting mkForce. Defensive `options ?` check so this evaluates even in
+  # module sets where the verity option is absent. An assertion below rejects
+  # the combination outright.
+  verityEnabled =
+    (options ? ghaf.partitioning.verity.enable) && config.ghaf.partitioning.verity.enable;
+  luksDiskKeyDescription = "luksDiskDeviceUniqueKey";
+  inherit (cfg.diskEncryption) luksUuid;
+
+  # ESP is referenced by its FAT label, which both media share. Fall back to the
+  # sd-image default when config.sdImage is absent (e.g. verity image builds),
+  # so this evaluates in module sets that do not use the sd-image format.
+  espLabel = config.sdImage.firmwarePartitionName or "FIRMWARE";
   rtcSeedAnchorPath = "/var/lib/systemd/timesync/clock";
   rtcSeedMaxAheadSeconds = 180 * 24 * 60 * 60;
   rtcSeedMinEpochSeconds = 1704067200; # 2024-01-01T00:00:00Z
@@ -228,6 +243,9 @@ let
         # Avoid interactive prompt in gen_ekb.py by providing UEFI auth key.
         printf '%s' "0x00000000000000000000000000000000" > auth.key
 
+        # Used for disk encryption.
+        printf '%s' "0x00000000000000000000000000000000" > sym2_t234.key
+
                 openssl req -x509 -newkey rsa:2048 -sha256 -nodes \
                   -keyout ek-rsa-key.pem -out ek-rsa.pem \
                   -subj "/CN=Jetson Orin fTPM EK RSA/O=Ghaf" \
@@ -246,6 +264,7 @@ let
           -chip t234 \
           -oem_k1_key oem_k1.key \
           -in_auth_key auth.key \
+          -in_sym_key2 sym2_t234.key \
           -in_ftpm_rsa_ek_cert ek-rsa.der \
           -in_ftpm_ec_ek_cert ek-ecc.der \
           -out "$out/eks_t234.img"
@@ -257,6 +276,310 @@ let
     mkIf
     types
     ;
+
+  resizepartitionsScript = pkgs.writeShellApplication {
+    name = "resize-partitions-cmds";
+    runtimeInputs = with pkgs; [
+      gptfdisk
+      parted
+      cryptsetup
+      util-linux
+      e2fsprogs
+      coreutils
+      systemd
+    ];
+    text = ''
+      RESIZE_MARKER=".ghaf-resize-done"
+      ESP_MOUNT="/mnt-esp"
+      ESP_DEVICE="/dev/disk/by-label/${espLabel}"
+
+      # Resolve the root partition by the pinned LUKS header UUID rather than a
+      # hardcoded device name or a partition-table PARTUUID (which differs between
+      # the USB sd-image (MBR) and the eMMC/NVMe flash (GPT)). blkid matches the
+      # crypto_LUKS container by its own UUID, identical on both media.
+      PART_DEV=""
+      for _ in $(seq 1 30); do
+        for d in $(blkid -o device -t UUID=${luksUuid} 2>/dev/null); do
+          ${lib.optionalString cfg.diskEncryption.enable ''
+            [ "$(blkid -o value -s TYPE "$d" 2>/dev/null)" = "crypto_LUKS" ] || continue
+          ''}
+          PART_DEV="$d"
+          break
+        done
+        if [ -n "$PART_DEV" ]; then break; fi
+        sleep 1
+      done
+
+      if [ -z "$PART_DEV" ]; then
+        echo "Root partition (LUKS UUID=${luksUuid}) not found, skipping resize."
+        exit 0
+      fi
+
+      # Derive disk and partition number from the resolved node, so mmcblk/nvme/sda all work.
+      real_dev=$(readlink -f "$PART_DEV")
+      base_dev=$(basename "$real_dev")
+      PART_NUM=$(cat "/sys/class/block/$base_dev/partition")
+      DISK="/dev/$(basename "$(readlink -f "/sys/class/block/$base_dev/..")")"
+
+      RESIZE_TARGET="$PART_DEV"
+      ${lib.optionalString cfg.diskEncryption.enable ''
+        MAPPER_NAME="${cfg.diskEncryption.mapperName}"
+        RESIZE_TARGET="/dev/mapper/$MAPPER_NAME"
+      ''}
+
+      mkdir -p "$ESP_MOUNT"
+      if [[ -b $ESP_DEVICE ]]; then
+        if mount "$ESP_DEVICE" "$ESP_MOUNT"; then
+          if [ -f "$ESP_MOUNT/$RESIZE_MARKER" ]; then
+            echo "Resize already performed, skipping."
+            umount "$ESP_MOUNT"
+            exit 0
+          fi
+          umount "$ESP_MOUNT"
+        else
+          echo "Failed to mount ESP $ESP_DEVICE, continuing without marker check."
+        fi
+      else
+        echo "ESP partition not found, continuing without marker check."
+      fi
+
+      for _ in {1..30}; do
+        [ -b "$RESIZE_TARGET" ] && break
+        sleep 1
+      done
+
+      if [ ! -b "$RESIZE_TARGET" ]; then
+        echo "Target device $RESIZE_TARGET not found, skipping."
+        exit 0
+      fi
+
+      echo "Fixing GPT..."
+      sgdisk -e "$DISK" || true
+
+      echo "Resizing partition $PART_NUM to 100%..."
+      # Use resizepart which works on busy partitions by using the BLKPG_RESIZE_PARTITION ioctl
+      parted -s "$DISK" resizepart "$PART_NUM" 100%
+
+      partprobe "$DISK" || true
+      udevadm settle || true
+
+      ${lib.optionalString cfg.diskEncryption.enable ''
+        echo "Resizing LUKS container $MAPPER_NAME..."
+        ${
+          # deviceUniqueKey and userPassphrase are mutually exclusive and exactly
+          # one is set (assertion below), so this if/else always resizes.
+          if cfg.diskEncryption.deviceUniqueKey.enable then
+            ''
+              cryptsetup resize --key-description ${luksDiskKeyDescription} "$MAPPER_NAME"
+            ''
+          else
+            ''
+              # Use systemd-ask-password to handle prompts in a non-interactive environment
+              PASSPHRASE=$(systemd-ask-password --timeout=60 "Enter passphrase for resizing LUKS container:")
+              if [ -n "$PASSPHRASE" ]; then
+                echo "$PASSPHRASE" | cryptsetup resize -v "$MAPPER_NAME" --key-file=-
+              else
+                echo "No passphrase entered, LUKS resize might have failed."
+              fi
+            ''
+        }
+        echo "LUKS status for $MAPPER_NAME after resize:"
+        cryptsetup status "$MAPPER_NAME"
+      ''}
+
+      echo "Resizing filesystem on $RESIZE_TARGET..."
+      # resize2fs may require a filesystem check before resizing
+      e2fsck -fy "$RESIZE_TARGET" || true
+      resize2fs "$RESIZE_TARGET"
+
+      # Persist completion on the ESP so later initrd boots can skip the
+      # service without mounting the root filesystem and racing fsck.
+      if [[ -b $ESP_DEVICE ]] && mount "$ESP_DEVICE" "$ESP_MOUNT"; then
+        touch "$ESP_MOUNT/$RESIZE_MARKER"
+        umount "$ESP_MOUNT"
+      else
+        echo "Failed to persist resize marker on ESP."
+      fi
+    '';
+  };
+
+  preDiskUniqueKeyScript = pkgs.writeShellApplication {
+    name = "pre-disk-unique-key";
+    runtimeInputs = with pkgs; [
+      cryptsetup
+      coreutils
+      gnused
+      keyutils
+      nvidia-jetpack.nvLuksSrv
+      gnugrep
+      util-linux
+    ];
+    text = ''
+      handle_error() {
+        if ! nvluks-srv-app -n; then
+          printf "Note: Attempt to stop the nvluks service failed (it may have already been stopped). Proceeding with device reboot.\n"
+        fi
+
+        printf "Encountered unexpected/unrecoverable error with disk encryption.\n"
+        printf "Rebooting in 10 seconds...\n"
+        sleep 10
+        reboot
+      }
+
+      wait_for_luks_device() {
+        luksDev=$1
+
+        for _ in {0..5}; do
+          if cryptsetup isLuks "$luksDev"; then
+            return 0
+          fi
+          sleep 1
+        done
+
+        printf "error: [%s] is not a luks device OR device is not ready\n" "$luksDev"
+        handle_error
+      }
+
+      sanitize_string() {
+        local dirty_str="$1"
+
+        echo "$dirty_str" | tr '[:upper:]' '[:lower:]' | tr -cd 'a-z0-9'
+      }
+
+      read_luksDev_serial() {
+        local luksDev=$1
+        local luksSerial
+
+        luksSerial=$(sanitize_string "$(udevadm info --query=property --name="$luksDev" | sed -n 's/^ID_SERIAL=//p')")
+        if [[ -z "$luksSerial" ]]; then
+           printf "error: Failed to get luksDev serial\n"
+           handle_error
+        else
+          echo "$luksSerial"
+        fi
+      }
+
+      switch_to_use_device_unique_key() {
+        uniqueKeyDescription=$1
+        defaultKey=$2
+        luksDev=$3
+
+        printf "Switching to use device unique key. This might take a bit..\n"
+        if ! printf "%s" "$defaultKey" | cryptsetup luksAddKey --new-key-description "$uniqueKeyDescription" --key-file=- "$luksDev"; then
+          printf "error: Failed to set unique key\n"
+          handle_error
+        fi
+
+        if ! printf "%s" "$defaultKey" | cryptsetup luksRemoveKey --key-file=- "$luksDev"; then
+          printf "error: Failed to remove default key\n"
+          handle_error
+        fi
+
+        # luksAddKey/luksRemoveKey only swap keyslots; the volume key that
+        # actually encrypts the data is still the one the manufacturing
+        # passphrase protected. Reencrypt derives a fresh volume key from the
+        # device-unique key, so a leaked manufacturing passphrase grants nothing.
+        printf "Note: Re-encryption may take 1 minute for every 1 GB of data ...\n"
+        if ! cryptsetup reencrypt --key-description "$uniqueKeyDescription" "$luksDev"; then
+           printf "error: Re-encryption failed\n"
+           handle_error
+        fi
+      }
+
+      # Resolve the LUKS partition by its pinned header UUID and verify it is a
+      # crypto_LUKS payload, independent of mmcblk/nvme/sda enumeration and of the
+      # partition-table scheme (MBR on USB, GPT on eMMC/NVMe).
+      luksDev=""
+      for _ in $(seq 1 30); do
+        for d in $(blkid -o device -t UUID=${luksUuid} 2>/dev/null); do
+          [ "$(blkid -o value -s TYPE "$d" 2>/dev/null)" = "crypto_LUKS" ] || continue
+          luksDev="$d"
+          break
+        done
+        if [ -n "$luksDev" ]; then break; fi
+        sleep 1
+      done
+      if [[ -z "$luksDev" ]]; then
+        printf "error: LUKS root partition (UUID=%s) not found\n" "${luksUuid}"
+        handle_error
+      fi
+      defaultManufactureKey=${cfg.diskEncryption.deviceUniqueKey.deviceManufacturerPassphrase}
+      luksDiskKeyKeyringDescription=${luksDiskKeyDescription}
+
+      wait_for_luks_device "$luksDev"
+
+      luksDevSerial=$(read_luksDev_serial "$luksDev")
+
+      luksKeyKeyringID=$(printf "%s" "$(nvluks-srv-app -c "$luksDevSerial" -u | tr -d '\n')" | keyctl padd user "$luksDiskKeyKeyringDescription" @u)
+
+      # Lock/prevent any other queries to luks key
+      if ! nvluks-srv-app -n; then
+        printf "error: Unable to stop nvluks service\n";
+        handle_error
+      fi
+
+      # Granting write access, because otherwise key is not removable
+      if ! keyctl setperm "$luksKeyKeyringID"  0x3f070000; then
+        printf "error: Unable modify setperm\n"
+        handle_error
+      fi
+
+      if ! keyctl describe "$luksKeyKeyringID" > /dev/null 2>&1; then
+        printf "error: Unable to add luks key to keyring\n"
+        handle_error
+      fi
+
+
+      # First boot is determined by checking whether the LUKS keyring description exists.
+      # Do not use cryptsetup --test-passphrase for this check, because it succeeds
+      # if the device can be unlocked by any available mechanism, including tokens/keyring.
+      if ! cryptsetup luksDump "$luksDev" 2>/dev/null | grep -Fq ${luksDiskKeyDescription}; then
+        switch_to_use_device_unique_key "$luksDiskKeyKeyringDescription" "$defaultManufactureKey" "$luksDev"
+      fi
+
+      if ! cryptsetup token add --key-description "$luksDiskKeyKeyringDescription" "$luksDev" > /dev/null 2>&1; then
+        printf "error: Unable to add cryptrsetup token\n"
+        handle_error
+      fi
+    '';
+  };
+
+  postDiskUniqueKeyScript = pkgs.writeShellApplication {
+    name = "post-disk-unique-key";
+    runtimeInputs = with pkgs; [
+      coreutils
+      keyutils
+    ];
+    text = ''
+      if ! keyctl revoke "$(keyctl search @u user ${luksDiskKeyDescription})"; then
+        printf "warn: Unable to revoke a key from keyring\n"
+      fi
+    '';
+  };
+
+  # Shared by the deviceDisk, deviceDiskRootfsPartition and deviceDiskEspPartition
+  # options declared below.
+  deviceDiskDescription = ''
+    Rootfs disk and its ESP/rootfs partitions, as kernel device names. The
+    rootfs partition is passed to NVIDIA's flash.sh as the trailing argument
+    and is consumed by the partition layout in the .conf file referenced by
+    configFileName, deciding where the APP partition lands. Downstream callers
+    (e.g. disk-encryption modules) read these to know the rootfs location at
+    build time.
+
+    Orin NX booting from USB, for example:
+
+    ```nix
+    flashScriptOverrides.deviceDisk = "sda";
+    flashScriptOverrides.deviceDiskEspPartition = "sda1";
+    flashScriptOverrides.deviceDiskRootfsPartition = "sda2";
+    ```
+
+    No default: rootfs storage is not common to all carrier boards, so every
+    per-SoM module must set all three explicitly. An assertion in the config
+    block enforces non-empty values.
+  '';
+
 in
 {
   _file = ./jetson-orin.nix;
@@ -273,19 +596,20 @@ in
       default = "";
     };
 
-    flashScriptOverrides.rootfsDevice = mkOption {
-      description = ''
-        Rootfs device passed to NVIDIA's flash.sh as the trailing argument
-        (e.g. "mmcblk0p1" for eMMC/SD, "nvme0n1p1" for NVMe). This is
-        consumed by the partition layout in the .conf file referenced by
-        configFileName and decides where the APP partition lands. Downstream
-        callers (e.g. disk-encryption modules) can read this to know the
-        rootfs location at build time.
+    flashScriptOverrides.deviceDisk = mkOption {
+      description = deviceDiskDescription;
+      type = types.str;
+      default = "";
+    };
 
-        No default: rootfs storage is not common to all carrier boards, so
-        every per-SoM module must set this explicitly. An assertion in the
-        config block enforces a non-empty value.
-      '';
+    flashScriptOverrides.deviceDiskRootfsPartition = mkOption {
+      description = deviceDiskDescription;
+      type = types.str;
+      default = "";
+    };
+
+    flashScriptOverrides.deviceDiskEspPartition = mkOption {
+      description = deviceDiskDescription;
       type = types.str;
       default = "";
     };
@@ -328,18 +652,120 @@ in
       default = true;
     };
 
+    diskEncryption = {
+      enable = mkEnableOption "generic LUKS root filesystem encryption for eMMC APP partition";
+
+      deviceUniqueKey = {
+        enable = mkEnableOption ''
+          disk decryption using a unique hardware key fetched from the OP-TEE.
+
+          On the first boot, the key (initially encrypted with a manufacturer key)
+          is rotated and re-encrypted with a device-unique key. Note that this
+          initial setup makes the first boot significantly slower.
+
+          This method provides an unattended boot process and does not require
+          user input to unlock the drive'';
+
+        deviceManufacturerPassphrase = mkOption {
+          description = ''
+            The temporary passphrase used to decrypt the device disk at the first
+            boot. Once the key is rotated to a device-unique key, this passphrase
+            is no longer needed for subsequent unlocks.
+          '';
+          type = types.str;
+          default = "ghaf";
+        };
+      };
+
+      userPassphrase = {
+        enable = mkEnableOption ''
+          a semi-manual passphrase for disk encryption. During device boot, the
+          device will prompt the console for the user to input the password.
+
+          Note: This option is primarily intended for testing purposes'';
+
+        passphrase = mkOption {
+          description = "The manual passphrase used to encrypt the disk.";
+          type = types.str;
+          default = "ghaf";
+        };
+      };
+
+      mode = mkOption {
+        description = "Disk encryption mode for Jetson root filesystem";
+        type = types.enum [ "generic-luks-passphrase" ];
+        default = "generic-luks-passphrase";
+      };
+
+      mapperName = mkOption {
+        description = "Mapped device name used by initrd after LUKS unlock";
+        type = types.str;
+        default = "cryptroot";
+      };
+
+      luksUuid = mkOption {
+        description = ''
+          Pinned LUKS2 header UUID of the encrypted root. Set on the container at
+          image build time (`sdimage.nix`, `cryptsetup reencrypt --uuid`) and
+          referenced at runtime as `/dev/disk/by-uuid/<uuid>` by the crypttab
+          device, the `pre-disk-unique-key` udev rule and the resize scan.
+
+          The header UUID rather than a partition PARTUUID, because the same
+          container ships on two partition tables — the raw sd-image on USB
+          (MBR) and the NVIDIA flash on eMMC/NVMe (GPT) — which necessarily carry
+          different PARTUUIDs. The header lives inside the container, so it is
+          identical on both media and independent of the kernel device name
+          (mmcblk/nvme/sda). It also survives the first-boot device-unique-key
+          rekey, which is a keyslot operation, not a header rewrite.
+
+          Read-only: two drives holding a verbatim ghaf flash carry the same UUID,
+          so a runtime scan cannot tell them apart and picks the first match.
+          Disambiguating needs a per-install-unique UUID, left as follow-up.
+        '';
+        type = types.str;
+        readOnly = true;
+        default = "0ada0e5b-0e5b-4e5b-8e5b-0000000000a9";
+      };
+    };
   };
 
   config = mkIf cfg.enable {
     assertions = [
       {
-        assertion = cfg.flashScriptOverrides.rootfsDevice != "";
+        assertion =
+          cfg.flashScriptOverrides.deviceDisk != ""
+          && cfg.flashScriptOverrides.deviceDiskRootfsPartition != ""
+          && cfg.flashScriptOverrides.deviceDiskEspPartition != "";
         message = ''
-          ghaf.hardware.nvidia.orin.flashScriptOverrides.rootfsDevice must be
-          set explicitly (e.g. "mmcblk0p1" for eMMC/SD, "nvme0n1p1" for NVMe).
+          ghaf.hardware.nvidia.orin.flashScriptOverrides.deviceDisk* must be
+          set explicitly (e.g. "mmcblk0" for eMMC/SD, "nvme0n1" for NVMe).
           The default is intentionally empty because rootfs storage varies per
           carrier board; the per-SoM module must declare it to avoid silently
           flashing to the wrong device.
+        '';
+      }
+      {
+        assertion =
+          if cfg.diskEncryption.enable then
+            (cfg.diskEncryption.deviceUniqueKey.enable != cfg.diskEncryption.userPassphrase.enable)
+          else
+            true;
+        message = ''
+          Disk encryption is enabled, but the unlock method is misconfigured.
+          You must enable exactly one of:
+          - 'diskEncryption.deviceUniqueKey.enable'
+          - 'diskEncryption.userPassphrase.enable'
+
+          Note: They are mutually exclusive and cannot both be false.
+        '';
+      }
+      {
+        assertion = !(cfg.diskEncryption.enable && verityEnabled);
+        message = ''
+          ghaf.hardware.nvidia.orin.diskEncryption.enable and
+          ghaf.partitioning.verity.enable are mutually exclusive root strategies:
+          verity owns fileSystems."/" as a tmpfs overlay, LUKS as an ext4 root on
+          /dev/mapper/${cfg.diskEncryption.mapperName}. Enable at most one.
         '';
       }
     ];
@@ -349,13 +775,13 @@ in
     hardware.nvidia-jetpack.firmware.eksFile = "${firmwareEkbImage}/eks_t234.img";
     hardware.nvidia-jetpack.kernel.version = "${cfg.kernelVersion}";
     # jetpack-nixos hardcodes the trailing rootfs device as mmcblk0p1; replay
-    # the same default here but route it through cfg.flashScriptOverrides.rootfsDevice
-    # so per-SoM modules (e.g. orin-nx → nvme0n1p1) only set the option, not
+    # the same default here but route it through cfg.flashScriptOverrides.deviceDiskRootfsPartition
+    # so per-SoM modules (e.g. orin-nx → nvme0n1p2) only set the option, not
     # the whole flashArgs list. mkOverride 75 beats jetpack-nixos's plain
     # assignment (prio 100) while leaving room for downstream mkForce (prio 50).
     hardware.nvidia-jetpack.flashScriptOverrides.flashArgs = lib.mkOverride 75 [
       config.hardware.nvidia-jetpack.flashScriptOverrides.configFileName
-      cfg.flashScriptOverrides.rootfsDevice
+      cfg.flashScriptOverrides.deviceDiskRootfsPartition
     ];
     nixpkgs.hostPlatform.system = "aarch64-linux";
 
@@ -366,6 +792,15 @@ in
 
     ghaf.global-config.givc.enable = true;
     ghaf.global-config.logging.enable = true;
+
+    environment.systemPackages = with pkgs; [
+      gptfdisk
+      parted
+      cryptsetup
+      util-linux
+      e2fsprogs
+    ];
+
     ghaf.hardware = {
       aarch64.systemd-boot-dtb.enable = true;
       passthrough = {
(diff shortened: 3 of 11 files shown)


Mic92 (approved):

leivos-unikie:
I did
`nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64-flash-script`
and flashed Orin AGX with that.

During flashing it asked for LUKS passphrase and flash finished fine but then at reboot it froze:

```
Jetson System firmware version 36.4.4-8f205d6ed58a date 1980-01-01T00:00:00+00:0
0
ESC   to enter Setup.
F11   to enter Boot Manager Menu.
Enter to continue boot.
......EFI stub: Booting Linux Kernel...
EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path
EFI stub: Using DTB from configuration table
EFI stub: Exiting boot services...
��debugfs initialized
��I/TC: Reserved shared memory is disabled
I/TC: Dynamic shared memory is enabled
I/TC: Normal World virtualization support is disabled
I/TC: Asynchronous notifications are disabled
��[    4.395456] i2c i2c-0: of_i2c: modalias failure on /bus@0/i2c@3160000/prod-settings
[    4.395741] i2c i2c-0: Failed to create I2C device for /bus@0/i2c@3160000/prod-settings
[    4.398564] i2c i2c-2: of_i2c: modalias failure on /bus@0/i2c@3180000/prod-settings
[    4.398785] i2c i2c-2: Failed to create I2C device for /bus@0/i2c@3180000/prod-settings
[    4.401212] i2c i2c-3: of_i2c: modalias failure on /bus@0/i2c@3190000/prod-settings
[    4.401417] i2c i2c-3: Failed to create I2C device for /bus@0/i2c@3190000/prod-settings
[    4.403777] i2c i2c-5: of_i2c: modalias failure on /bus@0/i2c@31b0000/prod-settings
[    4.403977] i2c i2c-5: Failed to create I2C device for /bus@0/i2c@31b0000/prod-settings
[    4.407180] i2c i2c-6: of_i2c: modalias failure on /bus@0/i2c@31c0000/prod-settings
[    4.407396] i2c i2c-6: Failed to create I2C device for /bus@0/i2c@31c0000/prod-settings
[    4.410216] i2c i2c-8: of_i2c: modalias failure on /bus@0/i2c@31e0000/prod-settings
[    4.410419] i2c i2c-8: Failed to create I2C device for /bus@0/i2c@31e0000/prod-settings
[    4.413200] i2c i2c-1: of_i2c: modalias failure on /bus@0/i2c@c240000/prod-settings
[    4.413406] i2c i2c-1: Failed to create I2C device for /bus@0/i2c@c240000/prod-settings
[    4.416774] i2c i2c-7: of_i2c: modalias failure on /bus@0/i2c@c250000/prod-settings
[    4.422530] i2c i2c-7: Failed to create I2C device for /bus@0/i2c@c250000/prod-settings
Booting initrd of NixOS 26.05 (Yarara) (Initrd).
[    7.466255] systemd[1]: systemd-cryptsetup@cryptroot.service: Service has no ExecStart=, ExecStop=, or SuccessAction=. Refusing.
[    7.490377] systemd[1]: Failed to isolate default target: Unit systemd-cryptsetup@cryptroot.service has a bad unit file setting.
[!!!!!!] Failed to isolate default target.
[    7.504911] systemd[1]: Freezing execution.
```

leivos-unikie:
I tried with this
`nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64-flash-script`

Flash succeeded but I didn't notice any request for typing LUKS passphrase. After flash boot fails and the device keeps on rebooting forever.

```
[  OK  ] Created slice Slice /system/modprobe.
[  OK  ] Created slice Slice /system/systemd-cryptsetup.
[  OK  ] Started Dispatch Password Requests to Console Directory Watch.
         Expecting device /dev/mapper/cryptroot...
         Expecting device /dev/mmcblk0p1...
[  OK  ] Reached target Path Units.
[  OK  ] Reached target Slice Units.
[  OK  ] Reached target Swaps.
[  OK  ] Reached target Timer Units.
[  OK  ] Listening on D-Bus System Message Bus Socket.
[  OK  ] Listening on Journal Socket (/dev/log).
[  OK  ] Listening on Journal Sockets.
[  OK  ] Listening on udev Control Socket.
[  OK  ] Listening on udev Kernel Socket.
[  OK  ] Reached target Socket Units.
         Mounting Kernel Configuration File System...
         Starting Journal Service...
         Starting Load Kernel Modules...
         Starting Create Static Device Nodes in /dev gracefully...
         Starting Coldplug All udev Devices...
[  OK  ] Mounted Kernel Configuration File System.
[  OK  ] Finished Load Kernel Modules.
         Starting Apply Kernel Variables...
[  OK  ] Finished Create Static Device Nodes in /dev gracefully.
         Starting Create Static Device Nodes in /dev...
[  OK  ] Finished Apply Kernel Variables.
[  OK  ] Finished Create Static Device Nodes in /dev.
[  OK  ] Reached target Preparation for Local File Systems.
[  OK  ] Reached target Local File Systems.
         Starting Rule-based Manager for Device Events and Files...
[  OK  ] Started Journal Service.
         Starting Create System Files and Directories...
[  OK  ] Started Rule-based Manager for Device Events and Files.
[  OK  ] Finished Create System Files and Directories.
[  OK  ] Finished Coldplug All udev Devices.
[  OK  ] Found device /dev/mmcblk0p1.
         Starting Service for device unique key disk encryption...
[    6.704631] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
[    6.704925] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
[    6.705188] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
         Starting Virtual Console Setup...
[  OK  ] Finished Virtual Console Setup.
[   12.797438] pre-disk-unique-key[316]: error: [/dev/mmcblk0p1] is not a luks device OR device is not ready
[   12.800860] pre-disk-unique-key[316]: Encountered unexpected/unrecoverable error with disk encryption.
[   12.801855] pre-disk-unique-key[316]: Rebooting in 10 second...
[   20.810112] pre-disk-unique-key[316]: Rebooting now
[  OK  ] Removed slice Slice /system/modprobe.
[  OK  ] Stopped target Path Units.
[  OK  ] Stopped target Slice Units.
[  OK  ] Stopped target Socket Units.
[  OK  ] Stopped Dispatch Password Requests to Console Directory Watch.
[  OK  ] Stopped target Swaps.
[  OK  ] Stopped target Timer Units.
[  OK  ] Removed slice Slice /system/systemd-cryptsetup.
[  OK  ] Stopped Apply Kernel Variables.
[  OK  ] Stopped Load Kernel Modules.
[  OK  ] Reached target Unmount All Filesystems.
         Starting Resize partitions to fill the disk on first boot...
[  OK  ] Stopped Create System Files and Directories.
[  OK  ] Stopped target Local File Systems.
[  OK  ] Stopped target Preparation for Local File Systems.
[  OK  ] Stopped Create Static Device Nodes in /dev.
[  OK  ] Stopped Create Static Device Nodes in /dev gracefully.
[  OK  ] Reached target System Shutdown.
[  OK  ] Reached target Late Shutdown Services.
[  OK  ] Finished System Reboot.
[  OK  ] Reached target System Reboot.
[   23.652251] pre-disk-unique-key[495]: nvluks-srv-app: TEEC_InvokeCommand failed 0xffff0001 origin 0x4
[   23.653274] pre-disk-unique-key[493]: add_key: Invalid argument
[   23.675354] resize-partitions-cmds[498]: mount: /mnt-esp: unknown filesystem type 'crypto_LUKS'.
[   23.675988] resize-partitions-cmds[498]:        dmesg(1) may have more information after failed mount system call.
[   23.699811] systemd-shutdown[1]: Failed to set watchdog hardware timeout to 10min: Invalid argument
[   23.678617] resize-partitions-cmds[360]: Failed to mount ESP /dev/mmcblk0p2, continuing without marker check.
[   24.030157] reboot: Restart����ing system with comma��Shutdown state requested 1
```

leivos-unikie:
Tested flashing with 
`nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64-flash-script`
built just before that change 2 hour ago. (Was it rebase or also something else, didn't notice.)

Flash script didn't ask any LUKS passphrase. After first flash it ended up in eternal reboot loop. After second flash it booted to ghaf login (with some errors): 
```
[FAILED] Failed to start Grow the root partition and filesystem to fill the SD card.
See 'systemctl status expand-root-partition.service' for details.
         Starting Grow the root partition and filesystem to fill the SD card...
         Starting Create SUID/SGID Wrappers...
         Starting Rebuild Journal Catalog...
         Starting Load/Save OS Random Seed...
         Starting Record System Boot/Shutdown in UTMP...
[FAILED] Failed to start Grow the root partition and filesystem to fill the SD card.
See 'systemctl status expand-root-partition.service' for details.
[  OK  ] Finished Rebuild Journal Catalog.
[  OK  ] Finished Load/Save OS Random Seed.
[  OK  ] Finished Create SUID/SGID Wrappers.
[  964.291149] tegra-audio-graph-card sound: ASoC: failed to register soundcard -16
[  964.300059] tegra-audio-graph-card sound: error -EBUSY: parse error
[  OK  ] Finished Record System Boot/Shutdown in UTMP.
[  OK  ] Created slice Slice /system/ghaf-seed-time-from-rtc.
[  OK  ] Reached target First Boot Complete.
[  OK  ] Listening on Load/Save RF Kill Switch Status /dev/rfkill Watch.
         Starting Grow the root partition and filesystem to fill the SD card...
         Starting Create SUID/SGID Wrappers...
         Starting Update Boot Loader Random Seed...
         Starting Save Transient machine-id to Disk...
         Starting Virtual Console Setup...
         Starting Load/Save RF Kill Switch Status...
[FAILED] Failed to start Grow the root partition and filesystem to fill the SD card.
See 'systemctl status expand-root-partition.service' for details
```

```
[ghaf@ghaf-host:~]$ systemctl list-units --state=failed
  UNIT                          LOAD      ACTIVE SUB    DESCRIPTION                                               
● expand-root-partition.service loaded    failed failed Grow the root partition and filesystem to fill the SD card
● resize-partitions.service     not-found failed failed resize-partitions.service
```

Now built from the current state of the PR and continuing to flash again...

leivos-unikie:
Tested again with latest
`nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64-flash-script`

Didn't see any LUKS passphrase request.

I noticed now that Orin AGX requires quite a lot of time at the first boot
`[   20.876562] pre-disk-unique-key[316]: Note: Re-encryption may take 1 minute for every 1 GB of data ...`

I assume that if this in intercepted e.g. by power off it might render the ghaf unbootable (or stuck in reboot loop).

After waiting ~16 minutes Orin AGX finally booted to ghaf login. Boot had few problems:
`[FAILED] Failed to start Grow the root partition and filesystem to fill the SD card.`


```
[ghaf@ghaf-host:~]$ systemctl list-units --state=failed
  UNIT                          LOAD      ACTIVE SUB    DESCRIPTION                                               
● expand-root-partition.service loaded    failed failed Grow the root partition and filesystem to fill the SD card
● resize-partitions.service     not-found failed failed resize-partitions.service
```

vadika:
> ```
> See 'systemctl status expand-root-partition.service' for details
> ```

what are the logs saying?

leivos-unikie:
> > ```
> > See 'systemctl status expand-root-partition.service' for details
> > ```
> 
> what are the logs saying?

```
[ghaf@ghaf-host:~]$ systemctl status expand-root-partition.service
× expand-root-partition.service - Grow the root partition and filesystem to fill the SD card
     Loaded: loaded (/etc/systemd/system/expand-root-partition.service; enabled; preset: ignored)
     Active: failed (Result: exit-code) since Tue 2026-03-17 20:11:56 UTC; 33min ago
 Invocation: 9e000ff9e4954ebdb26da93a92d4c75d
    Process: 2220 ExecStart=/nix/store/ldrzxpx64rrgl084yvahinxlbavyqwg5-unit-script-expand-root-partition-start/bin/expand-root-partition-start (code=exited, status=1/FAILURE)
   Main PID: 2220 (code=exited, status=1/FAILURE)
         IP: 0B in, 0B out
         IO: 0B read, 0B written
   Mem peak: 2.7M
        CPU: 35ms

Mar 17 20:11:55 ghaf-host systemd[1]: Starting Grow the root partition and filesystem to fill the SD card...
Mar 17 20:11:55 ghaf-host expand-root-partition-start[2242]: sfdisk: no disk device specified
Mar 17 20:11:56 ghaf-host systemd[1]: expand-root-partition.service: Main process exited, code=exited, status=1/FAILURE
Mar 17 20:11:56 ghaf-host systemd[1]: expand-root-partition.service: Failed with result 'exit-code'.
Mar 17 20:11:56 ghaf-host systemd[1]: Failed to start Grow the root partition and filesystem to fill the SD card.


```

vadika:
Rebased onto main. Two fixes on top:

- Disable the stock `expand-root-partition.service` on encrypted targets — it fails with `sfdisk: no disk device specified` on a `/dev/mapper/cryptroot` root (the initrd resize already grows the LUKS partition).
- Resolve the LUKS root/ESP by **PARTUUID** instead of hardcoded `mmcblk0`/`sda`, so it works regardless of mmcblk/nvme/sda enumeration. This also corrects the partition the resize/unlock scripts were pointing at.

Note: device-unique-key targets unlock via fTPM, so no passphrase prompt is expected.

milva-unikie:
**Tested on Orin AGX**

`nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64-flash-script`

and then flashed with

`sudo ./result/bin/flash-orin-agx-devkit`

Flashing worked but initial boot went into emergency mode

```
Booting initrd of NixOS 26.11 (Zokor) (Initrd).
[  OK  ] Created slice Slice /system/modprobe.
[  OK  ] Created slice Slice /system/systemd-cryptsetup.
[  OK  ] Started Dispatch Password Requests to Console Directory Watch.
         Expecting device /dev/disk/by-partuuid/2178694e-02...
         Expecting device /dev/mapper/cryptroot...
[  OK  ] Reached target Path Units.
[  OK  ] Reached target Slice Units.
[  OK  ] Reached target Swaps.
[  OK  ] Reached target Timer Units.
[  OK  ] Listening on D-Bus System Message Bus Socket.
[  OK  ] Listening on Journal Socket (/dev/log).
[  OK  ] Listening on Journal Sockets.
[  OK  ] Listening on udev Control Socket.
[  OK  ] Listening on udev Kernel Socket.
[  OK  ] Reached target Socket Units.
         Mounting Kernel Configuration File System...
         Starting Journal Service...
         Starting Load Kernel Modules...
         Starting Create Static Device Nodes in /dev gracefully...
         Starting Coldplug All udev Devices...
[  OK  ] Mounted Kernel Configuration File System.
[  OK  ] Finished Load Kernel Modules.
[  OK  ] Finished Create Static Device Nodes in /dev gracefully.
         Starting Apply Kernel Variables...
         Starting Create Static Device Nodes in /dev...
[  OK  ] Finished Apply Kernel Variables.
[  OK  ] Finished Create Static Device Nodes in /dev.
[  OK  ] Reached target Preparation for Local File Systems.
[  OK  ] Reached target Local File Systems.
         Starting Rule-based Manager for Device Events and Files...
[  OK  ] Started Journal Service.
         Starting Create System Files and Directories...
[  OK  ] Finished Create System Files and Directories.
[  OK  ] Started Rule-based Manager for Device Events and Files.
[  OK  ] Finished Coldplug All udev Devices.
         Starting Virtual Console Setup...
[    8.296856] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
[    8.297108] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
[    8.297365] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
[  OK  ] Finished Virtual Console Setup.
[ TIME ] Timed out waiting for device /dev/disk/by-partuuid/2178694e-02.
[DEPEND] Dependency failed for Cryptography Setup for cryptroot.
[DEPEND] Dependency failed for /dev/mapper/cryptroot.
[DEPEND] Dependency failed for File System Check on /dev/mapper/cryptroot.
[DEPEND] Dependency failed for /sysroot.
[DEPEND] Dependency failed for /sysroot/run.
[DEPEND] Dependency failed for Initrd File Systems.
[DEPEND] Dependency failed for Find NixOS closure.
[DEPEND] Dependency failed for Initrd Root File System.
[DEPEND] Dependency failed for Initrd Root Device.
[DEPEND] Dependency failed for Local Encrypted Volumes.
[  OK  ] Stopped Dispatch Password Requests to Console Directory Watch.
[  OK  ] Started Emergency Shell.
[  OK  ] Reached target Emergency Mode.
         Starting Resize partitions to fill the disk on first boot...
         Starting Create Volatile Files and Directories in the Real Root...
[  OK  ] Finished Create Volatile Files and Directories in the Real Root.
You are in emergency mode. After logging in, type "journalctl -xb" to view
system logs, "systemctl reboot" to reboot, or "exit"
to continue bootup.
Press Enter for system maintenance
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -10,7 +10,25 @@
...
+  # (mmcblk0 / nvme0n1 / sda) and survives LUKS encryption (it lives in the
+  # partition table, not the filesystem). This replaces the previous
+  # somType-based "/dev/mmcblk0pX" / "/dev/sdaX" hardcodes, which broke whenever
+  # the rootfs landed on a different controller (e.g. NX flashes to NVMe).
+  #
+  # KNOWN LIMITATION (twin drive): if two drives both hold a verbatim ghaf flash
+  # they carry the same PARTUUID, so the runtime scan cannot tell them apart and
+  # picks the first match. Disambiguating that case needs either boot-device
+  # detection or a per-install-unique firmwarePartitionID, left as follow-up.
+  luksDiskId = lib.toLower (
```
TanelDettenborn: Change required: At least in your use case, PARTUUID does not work for internal drives. For example, on AGX and eMMC, the partition table is created independently. It is not created by the Nix code, but by the Nvidia BSP, which uses a randomly generated UUID at flash time.

I am uncertain if the PARTUUID can be defined via a command-line option in the Nvidia BSP. While that could be investigated, I would propose looking into Ghaf's [rootfsDevice](https://github.com/tiiuae/ghaf/blob/main/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix#L276) option instead.

If we want to stick with a UUID approach, we could use the filesystem UUID, which can be defined directly via a Nix option.

modules/reference/hardware/jetpack/nvidia-jetson-orin/sdimage.nix
```diff
@@ -18,6 +18,22 @@
...
+  rootfsExtraSlackMiB = 64;
+
+  cryptsetup =
+    (pkgs.callPackage "${toString pkgs.path}/pkgs/by-name/cr/cryptsetup/package.nix" { }).overrideAttrs
+      (oldAttrs: {
+        # /run/cryptsetup (the upstream default) is not writable inside the
+        # image build, and /build only exists on sandboxed-to-/build builders.
+        # /tmp is writable in every nix build environment.
+        configureFlags = oldAttrs.configureFlags ++ [
+          "--with-luks2-lock-path=/tmp/cryptsetup"
```
TanelDettenborn, line 38: Change required: [7ce5db7567010147b59e44ec8d55e7d7a145edbf](https://github.com/tiiuae/ghaf/pull/1742/changes/7ce5db7567010147b59e44ec8d55e7d7a145edbf)-commit message. It mentions flashing, but this is just syntactic sugar - cryptsetup is actually used during image generation, not during the flash process itself.

Also, I am wondering if we should use Nix's `$TMPDIR` environment variable here instead?

modules/reference/hardware/jetpack/nvidia-jetson-orin/sdimage.nix
```diff
@@ -59,6 +81,123 @@
...
+            # Keep the plaintext file one LUKS data offset larger than the
+            # ext4 payload so reencrypt does not collapse the final image size
+            # together with the resized filesystem.
+            truncate -s "$(( TARGET_PAYLOAD_BYTES + LUKS_DATA_OFFSET_BYTES ))" "$ROOT_IMAGE"
+
+            echo "Encrypting extracted root image with LUKS2 ..."
+            # cryptsetup is built with --with-luks2-lock-path=/tmp/cryptsetup;
+            # newer releases refuse to take the reencryption lock if that
+            # directory does not already exist.
+            mkdir -p /tmp/cryptsetup
```
TanelDettenborn: If this change is needed, the cryptsetup-lock path should be stored in a Nix variable.

During testing, I found that - at least with the current version - cryptsetup will automatically create sub-directories as long as the parent directory is writable.

TanelDettenborn (changes requested):
A couple of small things. Otherwise, looks good!

milva-unikie:
**Tested on Orin AGX**

Still same result, boots to emergency mode on initial boot.

<details>

<summary>Full logs</summary>

`nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64-flash-script`

and then flashed with

`sudo ./result/bin/flash-orin-agx-devkit`

Flashing worked but initial boot went into emergency mode

```
Booting initrd of NixOS 26.11 (Zokor) (Initrd).
[  OK  ] Created slice Slice /system/modprobe.
[  OK  ] Created slice Slice /system/systemd-cryptsetup.
[  OK  ] Started Dispatch Password Requests to Console Directory Watch.
         Expecting device /dev/disk/by-partuuid/2178694e-02...
         Expecting device /dev/mapper/cryptroot...
[  OK  ] Reached target Path Units.
[  OK  ] Reached target Slice Units.
[  OK  ] Reached target Swaps.
[  OK  ] Reached target Timer Units.
[  OK  ] Listening on D-Bus System Message Bus Socket.
[  OK  ] Listening on Journal Socket (/dev/log).
[  OK  ] Listening on Journal Sockets.
[  OK  ] Listening on udev Control Socket.
[  OK  ] Listening on udev Kernel Socket.
[  OK  ] Reached target Socket Units.
         Mounting Kernel Configuration File System...
         Starting Journal Service...
         Starting Load Kernel Modules...
         Starting Create Static Device Nodes in /dev gracefully...
         Starting Coldplug All udev Devices...
[  OK  ] Mounted Kernel Configuration File System.
[  OK  ] Finished Load Kernel Modules.
         Starting Apply Kernel Variables...
[  OK  ] Finished Create Static Device Nodes in /dev gracefully.
         Starting Create Static Device Nodes in /dev...
[  OK  ] Finished Apply Kernel Variables.
[  OK  ] Finished Create Static Device Nodes in /dev.
[  OK  ] Reached target Preparation for Local File Systems.
[  OK  ] Reached target Local File Systems.
         Starting Rule-based Manager for Device Events and Files...
[  OK  ] Started Journal Service.
         Starting Create System Files and Directories...
[  OK  ] Started Rule-based Manager for Device Events and Files.
[  OK  ] Finished Create System Files and Directories.
[  OK  ] Finished Coldplug All udev Devices.
         Starting Virtual Console Setup...
[    6.174965] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
[    6.175230] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
[    6.175488] tegra-xusb-padctl 3520000.padctl: Failed to create device link (0x180) with 1-0008
[  OK  ] Finished Virtual Console Setup.
[ TIME ] Timed out waiting for device /dev/disk/by-partuuid/2178694e-02.
[DEPEND] Dependency failed for Cryptography Setup for cryptroot.
[DEPEND] Dependency failed for /dev/mapper/cryptroot.
[DEPEND] Dependency failed for Initrd Root Device.
[DEPEND] Dependency failed for /sysroot.
[DEPEND] Dependency failed for Initrd Root File System.
[DEPEND] Dependency failed for /sysroot/run.
[DEPEND] Dependency failed for Initrd File Systems.
[DEPEND] Dependency failed for Find NixOS closure.
[DEPEND] Dependency failed for File System Check on /dev/mapper/cryptroot.
[DEPEND] Dependency failed for Local Encrypted Volumes.
[  OK  ] Stopped Dispatch Password Requests to Console Directory Watch.
[  OK  ] Started Emergency Shell.
[  OK  ] Reached target Emergency Mode.
         Starting Resize partitions to fill the disk on first boot...
         Starting Create Volatile Files and Directories in the Real Root...
[  OK  ] Finished Create Volatile Files and Directories in the Real Root.
You are in emergency mode. After logging in, type "journalctl -xb" to view
system logs, "systemctl reboot" to reboot, or "exit"
to continue bootup.
Press Enter for system maintenance
(or press Control-D to continue):
```

</details>

milva-unikie:
**Tested on Orin AGX**

Device finally boots. No issues found.

1. Built `nix build .#nvidia-jetson-orin-agx-debug-luks-from-x86_64-flash-script`.
2. Flashed with `sudo ./result/bin/flash-orin-agx-devkit`. After flashing device booted automatically, first boot took 15 minutes because

```
[   18.387314] pre-disk-unique-key[319]: Note: Re-encryption may take 1 minute for every 1 GB of data ...
Finished, time 14m36s,   12 GiB written, speed  14.6 MiB/s
```

3. Ran automated tests, no issues.
4. Rebooted and ran automated tests again, still no issues.

modules/profiles/orin.nix
```diff
@@ -29,6 +29,40 @@
...
   cfg = config.ghaf.profiles.orin;
   hostGlobalConfig = config.ghaf.global-config;
+  ensureSystemProfile = pkgs.writeShellApplication {
+    name = "ghaf-ensure-system-profile";
+    runtimeInputs = with pkgs; [
+      coreutils
+      nix
+    ];
+    text = ''
+      set -euo pipefail
```
kajusnau: Can be omitted, see https://ryantm.github.io/nixpkgs/builders/trivial-builders/#trivial-builder-writeShellApplication

modules/profiles/orin.nix
```diff
@@ -29,6 +29,40 @@
...
+
+      profile=/nix/var/nix/profiles/system
+      registration=/nix-path-registration
+      current_system=$(${pkgs.coreutils}/bin/readlink -f /run/current-system)
```
kajusnau: `coreutils` is in `runtimeInputs`, so `readlink` should be available in `PATH`

modules/profiles/orin.nix
```diff
@@ -29,6 +29,40 @@
...
+      generation_link=/nix/var/nix/profiles/system-1-link
+
+      if [ -z "$current_system" ] || [ ! -e "$current_system" ]; then
+        echo "Current system closure is unavailable" >&2
+        exit 1
+      fi
+
+      if [ ! -f "$registration" ] && [ -L "$profile" ] && [ "$(${pkgs.coreutils}/bin/readlink -f "$profile")" = "$current_system" ]; then
```
kajusnau: Same as above

modules/profiles/orin.nix
```diff
@@ -29,6 +29,40 @@
...
+        exit 0
+      fi
+
+      if [ -f "$registration" ]; then
+        ${pkgs.nix}/bin/nix-store --load-db < "$registration"
```
kajusnau: `nix` is in `runtimeInputs`, so `nix-store` should be available in `PATH`

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -228,6 +249,10 @@ let
         # Avoid interactive prompt in gen_ekb.py by providing UEFI auth key.
         printf '%s' "0x00000000000000000000000000000000" > auth.key
 
+        # Used for disk encryption.
+        # printf '%s' "0x00000000000000000000000000000000" > sym_t234.key
```
kajusnau: Stray commented line?

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+      echo "Resizing partition $PART_NUM to 100%..."
+      # Use resizepart which works on busy partitions by using the BLKPG_RESIZE_PARTITION ioctl
+      parted -s "$DISK" resizepart "$PART_NUM" 100%
+
+      # Re-read partition table and wait for udev
+      partprobe "$DISK" || true
+      udevadm settle || true
+
+      ${lib.optionalString cfg.diskEncryption.enable ''
+        echo "Resizing LUKS container $MAPPER_NAME..."
```
kajusnau, line 367: This print is misleading if  `cfg.diskEncryption.deviceUniqueKey.enable == false` (below `optionalString` code snippet won't run)

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+           # Use systemd-ask-password to handle prompts in a non-interactive environment
+          PASSPHRASE=$(systemd-ask-password --timeout=60 "Enter passphrase for resizing LUKS container:")
+          if [ -n "$PASSPHRASE" ]; then
+            echo "$PASSPHRASE" | cryptsetup resize -v "$MAPPER_NAME" --key-file=-
+          else
+             echo "No passphrase entered, LUKS resize might have failed."
+          fi
+        ''}
+        echo "LUKS status for $MAPPER_NAME after resize:"
+        cryptsetup status "$MAPPER_NAME"
```
kajusnau, line 387: Same comment as above

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+      gawk
+      gnused
+      keyutils
+      nvidia-jetpack.nvLuksSrv
+      gnugrep
+      util-linux
+    ];
+    text = ''
+      handle_error()
+      {
```
kajusnau: Stylistic comment:
I think in Ghaf bash scripts we usually have the opening curly bracket on the same line as the function name. This stylistic choice is inconsistent in this script.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+        if ! nvluks-srv-app -n; then
+          printf "Note: Attempt to stop the nvluks service failed (it may have already been stopped). Proceeding with device reboot.\n"
+        fi
+
+        printf "Encountered unexpected/unrecoverable error with disk encryption.\n"
+        printf "Rebooting in 10 second...\n"
```
kajusnau: 10 seconds

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+        sleep 8
+        printf "\nRebooting now\n"
+        sleep 2
```
kajusnau: probably don't need to sleep here

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+        reboot now
```
kajusnau: AFAIK `reboot` doesn't support `now`? https://man.archlinux.org/man/reboot.8.en
`reboot` is probably enough

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+        # 2. Verify device readiness
+
+        # Following is a basic sanity check
+        for _ in {0..5}
+        do
+          if cryptsetup isLuks "$luksDev"; then
+            return 0
+          fi
+          sleep 1
+        done
```
kajusnau: Inconsistent syntax compared to other instances of `for ...`
```suggestion
        for _ in {0..5}; do
          if cryptsetup isLuks "$luksDev"; then
            return 0
          fi
          sleep 1
        done
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+        if [[ ! "$(cryptsetup --version | gawk '{print $2}')" =~ ^2\.2\.[0-9]+$ ]]; then
+          printf "Note: Re-encryption may take 1 minute for every 1 GB of data ...\n"
+          if ! cryptsetup reencrypt --key-description "$uniqueKeyDescription" "$luksDev"; then
+             printf "error: Re-encryption failed\n"
+             handle_error
+          fi
+        fi
+      }
+
+      # Convinience variables
```
kajusnau: Convenience

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+          printf "error: Failed to set unique key\n"
+          handle_error
+        fi
+
+        if ! printf "%s" "$defaultKey" | cryptsetup luksRemoveKey --key-file=- "$luksDev"; then
+          printf "error: Failed to remove default key\n"
+          handle_error
+        fi
+
+        # Re-encrypt only if cryptsetup version is NOT 2.2.XX
```
kajusnau: Maybe I missed something, but why?

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+        reboot now
+      }
+
+      does_luks_device_ready_exist()
+      {
+        luksDev=$1
+
+        # Following serving two purpose
+        # 1. A basic sanity check
+        # 2. Verify device readiness
```
kajusnau: The function name could use a cleanup to describe the purpose a little more accurately

Also it seems like Purpose # 2 in the comment is never fulfilled?

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +283,299 @@ let
...
+      if ! keyctl describe "$luksKeyKeyringID" > /dev/null 2>&1; then
+        printf "error: Unable to add luks key to keyring\n"
+        handle_error
+      fi
+
+
+      # First boot is determined by checking whether the LUKS keyring description exists.
+      # Do not use cryptsetup --test-passphrase for this check, because it succeeds
+      # if the device can be unlocked by any available mechanism, including tokens/keyring.
+      if ! cryptsetup luksDump "$luksDev" 2>/dev/null |grep -Fq ${luksDiskKeyDescription}; then
```
kajusnau:
```suggestion
      if ! cryptsetup luksDump "$luksDev" 2>/dev/null | grep -Fq ${luksDiskKeyDescription}; then
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -290,6 +618,22 @@ in
       default = "";
     };
 
+    flashScriptOverrides.deviceDiskRootfsPartition = mkOption {
+      description = ''
+        See deviceDisk option description
+      '';
```
kajusnau: Not a strong opinion but I think this is bad practice when we could just as easily have the description available in a nix var and reuse it for all 3 options

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -328,34 +672,98 @@ in
...
+          On the first boot, the key (initially encrypted with a manufacturer key)
+          is rotated and re-encrypted with a device-unique key. Note that this
+          initial setup makes the first boot significantly slower.
+
+          This method provides an unattended boot process and does not require user input to unlock the drive.";
+
+        deviceManufacturePassphrase = mkOption {
+          description = "The temporary passphrase used to decrypt the device disk at the first boot.
+            Once the key is rotated to a device-unique key,
+            this passphrase is no longer needed for subsequent unlocks.";
```
kajusnau: Should probably use single quotes for this multi-line description

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -328,34 +672,98 @@ in
...
+      deviceUniqueKey = {
+        enable = mkEnableOption "Enable disk decryption using a unique hardware key fetched from the OP-TEE.
+
+          On the first boot, the key (initially encrypted with a manufacturer key)
+          is rotated and re-encrypted with a device-unique key. Note that this
+          initial setup makes the first boot significantly slower.
+
+          This method provides an unattended boot process and does not require user input to unlock the drive.";
+
+        deviceManufacturePassphrase = mkOption {
```
kajusnau:
```suggestion
        deviceManufacturerPassphrase = mkOption {
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -328,34 +672,98 @@ in
...
+          description = "The temporary passphrase used to decrypt the device disk at the first boot.
+            Once the key is rotated to a device-unique key,
+            this passphrase is no longer needed for subsequent unlocks.";
+          type = types.str;
+          default = "ghaf";
+        };
+      };
+
+      userPassphrase = {
+        enable = mkEnableOption "Enables a semi-manual passphrase for disk encryption.
```
kajusnau: Should probably use single quotes for this multi-line description

Also `mkEnableOption` prepends "Whether to enable" implicitly, so:

```suggestion
        enable = mkEnableOption "setting a passphrase for disk encryption during initial boot. ...
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -328,34 +672,98 @@ in
...
   };
 
   config = mkIf cfg.enable {
     assertions = [
       {
-        assertion = cfg.flashScriptOverrides.rootfsDevice != "";
+        assertion =
+          cfg.flashScriptOverrides.deviceDisk != ""
+          || cfg.flashScriptOverrides.deviceDiskRootfsPartition != ""
+          || cfg.flashScriptOverrides.deviceDiskEspPartition != "";
```
kajusnau: Shouldn't all of these options be set? Therefore:
```suggestion
          cfg.flashScriptOverrides.deviceDisk != ""
          && cfg.flashScriptOverrides.deviceDiskRootfsPartition != ""
          && cfg.flashScriptOverrides.deviceDiskEspPartition != "";
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -419,9 +837,198 @@ in
...
+        };
+      };
+
+      systemd.initrdBin = [
+        pkgs.gawk
+        pkgs.gnused
+        pkgs.keyutils
+        pkgs.nvidia-jetpack.nvLuksSrv
+        pkgs.gnugrep
+      ];
```
kajusnau:
```suggestion
      systemd.initrdBin = with pkgs; [
        gawk
        gnused
        keyutils
        nvidia-jetpack.nvLuksSrv
        gnugrep
      ];
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -419,9 +837,198 @@ in
...
+        ++ lib.optionals cfg.diskEncryption.deviceUniqueKey.enable [
+          preDiskUniqueKeyScript
+          postDiskUniqueKeyScript
+        ];
+
+      services.udev.rules = ''
+        ${lib.optionalString cfg.diskEncryption.deviceUniqueKey.enable ''
+          ACTION=="add", SUBSYSTEM=="block", ENV{ID_FS_UUID}=="${luksUuid}", TAG+="systemd", ENV{SYSTEMD_WANTS}+="pre-disk-unique-key.service"
+        ''}
+      '';
```
kajusnau: Can we simplify to?:

```suggestion
      services.udev.rules = lib.optionalString cfg.diskEncryption.deviceUniqueKey.enable ''
        ACTION=="add", SUBSYSTEM=="block", ENV{ID_FS_UUID}=="${luksUuid}", TAG+="systemd", ENV{SYSTEMD_WANTS}+="pre-disk-unique-key.service"
      '';
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -419,9 +837,198 @@ in
...
+        serviceConfig = {
+          Type = "oneshot";
+          RemainAfterExit = true;
+          ExecStart = "${resizepartitionsScript}/bin/resize-partitions-cmds";
+          StandardInput = "tty";
+          StandardOutput = "journal+console";
+          StandardError = "journal+console";
+          KeyringMode = "shared";
+        };
+      };
```
kajusnau: Would prefer to have all of these services under 

```nix
systemd.services = lib.mkMerge [
  ...
];
```

See [modules/common/services/power.nix line 905](https://github.com/tiiuae/ghaf/blob/80ddbae0b70980772ac49d705d58ef39820625c4/modules/common/services/power.nix#L905)

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -5,12 +5,33 @@
...
   cfg = config.ghaf.hardware.nvidia.orin;
 
+  # LUKS full-disk encryption and dm-verity are mutually exclusive root
+  # strategies: verity-volume.nix owns fileSystems."/" (tmpfs overlay) when
+  # enabled, so the diskEncryption ext4 root override below must stand down to
+  # avoid a conflicting mkForce. Defensive `options ?` check so this evaluates
+  # even in module sets where the verity option is absent.
+  verityEnabled =
+    (options ? ghaf.partitioning.verity.enable) && config.ghaf.partitioning.verity.enable;
+  luksDiskKeyDescription = "luksDiskDeviceUniqueKey";
```
kajusnau, line 22: Should this be an `assertion` instead?

modules/reference/hardware/jetpack/nvidia-jetson-orin/luks-ids.nix
```diff
@@ -0,0 +1,25 @@
...
+# sd-image blob) and is independent of the kernel device name (mmcblk/nvme/sda).
+# It also survives the first-boot device-unique-key rekey (keyslot op, not a
+# header rewrite).
+#
+# KNOWN LIMITATION (twin drive): two drives each holding a verbatim ghaf flash
+# carry the same LUKS UUID, so a runtime scan cannot tell them apart and picks
+# the first match. Disambiguating needs a per-install-unique UUID, left as
+# follow-up.
+{
+  luksUuid = "0ada0e5b-0e5b-4e5b-8e5b-0000000000a9";
```
kajusnau: Why not have this as a `readOnly` option?

kajusnau (commented):
I think this could use a cleanup before merge 😀 

P.S.: blame brian... 😶‍🌫️

vadika:
> I think this could use a cleanup before merge 😀
> 
> P.S.: blame brian... 😶‍🌫️

it definitely could, thanks for comments!

modules/profiles/orin.nix
vadika: Dropped, `writeShellApplication` already sets it.

modules/profiles/orin.nix
vadika: Fixed, plain `readlink` now.

modules/profiles/orin.nix
vadika: Fixed.

modules/profiles/orin.nix
vadika: Fixed, plain `nix-store` now.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Yes, removed.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Fixed, `foo() {` everywhere in this script now.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Fixed.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Collapsed to one `sleep 10`.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Right, `reboot` now.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Applied.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Removed, the block already has a real comment below it.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Applied.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Applied. Also dropped `gawk` — its only user was the cryptsetup version check, now gone.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Applied.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: It is not — the two `optionalString`s are siblings, both nested inside the outer `optionalString cfg.diskEncryption.enable`. When `deviceUniqueKey.enable` is false, `userPassphrase.enable` is true (an assertion enforces exactly one of them), so the resize below still runs and the print is accurate. Left as is.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Same — `cryptsetup status` is inside the outer `diskEncryption.enable` block, not the `deviceUniqueKey` one. It always runs when encryption is on.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: The keyslot swap above (`luksAddKey` + `luksRemoveKey`) does not change the volume key that actually encrypts the data — that key is still the one the manufacturing passphrase protected, so anyone who captured that passphrase during production could have derived it. `cryptsetup reencrypt` derives a fresh volume key from the device-unique key, which is what makes the rotation meaningful.

The `2.2.x` guard was dead code — `cryptsetup reencrypt` only landed in 2.4.0, and nixpkgs pins a much newer release — so I dropped the version check (and `gawk` with it) and moved the rationale into a comment.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Renamed to `wait_for_luks_device`. Purpose #2 was the whole point of the loop and #1 was just the same `isLuks` call, so I dropped the comment rather than pretend there were two things.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Agreed. Now one `deviceDiskDescription` binding shared by all three options, and it lists all three as a set.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Switched to `''` and reflowed.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Applied, renamed everywhere (`jetson-orin.nix` + `sdimage.nix`).

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Applied to both `deviceUniqueKey.enable` and `userPassphrase.enable`.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Good catch — with `||` a single non-empty value satisfied the assertion. `&&` now.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Merged into one attrset: `resize-partitions` unconditionally, the three device-unique-key services via `// lib.optionalAttrs cfg.diskEncryption.deviceUniqueKey.enable {...}`. That also drops the three `mkIf`s.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Added one. The `verityEnabled` binding still guards the `fileSystems."/"` override, because without it the two `mkForce`s collide and you get a module-system conflict trace instead of the assertion message. The assertion is what tells you why.

modules/reference/hardware/jetpack/nvidia-jetson-orin/luks-ids.nix
vadika: Done — it is now `ghaf.hardware.nvidia.orin.diskEncryption.luksUuid`, `readOnly`, with the rationale and the twin-drive caveat in its description. `luks-ids.nix` is gone; `sdimage.nix` reads it from `config`.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -85,7 +197,95 @@ in
...
       ];
+
+      systemd.services.resize-partitions = {
+        description = "Resize partitions to fill the disk on first boot";
+        wantedBy = [ "initrd.target" ];
+        before = [
+          "sysroot.mount"
+          "initrd-root-fs.target"
+        ];
+        after = [ "cryptsetup.target" ];
```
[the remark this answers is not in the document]
vadika: `resize-partitions` now has `after = [ "initrd-root-device.target" "cryptsetup.target" ]`.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -357,7 +467,118 @@ in
...
+
+      supportedFilesystems = [ "ext4" ];
+
+      systemd.services."systemd-cryptsetup@${cfg.diskEncryption.mapperName}" = {
+        overrideStrategy = "asDropin";
+        unitConfig.After = [
+          "initrd-root-device.target"
+          "cryptsetup.target"
+        ];
+      };
```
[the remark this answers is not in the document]
vadika: The `systemd-cryptsetup@cryptroot` drop-in is gone — thanks for testing. The ordering it was adding now lives on `resize-partitions` directly.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
```diff
@@ -166,15 +167,69 @@ let
...
+
+              if [ "$GHAF_LUKS_PASSPHRASE" != "$GHAF_LUKS_PASSPHRASE_CONFIRM" ]; then
+                echo "Passphrases do not match. Try again."
+                continue
+              fi
+
+              break
+            done
+          else
+            GHAF_SKIP_LUKS_ENCRYPTION=1
```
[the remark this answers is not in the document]
vadika: This is obsolete now: the interactive prompt is gone. Encryption moved out of the flash script into the sd-image build (`sdimage.nix`, `cryptsetup reencrypt --encrypt`), so the flash script only writes an already-encrypted APP image and never sees a passphrase. The build-time passphrase is a Nix option (`diskEncryption.deviceUniqueKey.deviceManufacturerPassphrase` / `userPassphrase.passphrase`); for `deviceUniqueKey` it is a manufacturing passphrase that the device rotates away from on first boot.

vadika:
Pushed the cleanup pass. All review comments addressed and resolved, except the two about the resize prints — those two `optionalString`s are siblings under `diskEncryption.enable`, not nested under `deviceUniqueKey.enable`, so the prints are accurate. Left open in case I am misreading you.

Three things beyond the comments, while I was in there:

- `sdimage.nix` had `GHAF_LUKS_CHECK_GEOMETRY` / `GHAF_LUKS_VALIDATE_IMAGE` / `GHAF_LUKS_VALIDATE_FULL_FSCK` debug blocks with a missing `fi`, so the validate block was unreachable unless the geometry check was also on, and `rm -f "$GHAF_LUKS_PASSPHRASE_FILE"` sat inside the same dead branch. All three env-var blocks are gone (~45 lines) and the passphrase file is always removed.
- `flashScriptOverrides.deviceDisk*` assertion used `||` where it meant `&&`, so setting any one of the three satisfied it.
- Dropped the `cryptsetup` `2.2.x` version gate around `reencrypt`. `reencrypt` landed in 2.4.0 and nixpkgs pins far newer, so the branch never fired; the comment now explains why the reencrypt is needed at all (keyslot swap does not rotate the volume key).

Net -53 lines. `nix fmt` clean, and `nvidia-jetson-orin-{agx,nx}-debug{,-luks}` toplevel + sdImage all evaluate.

vadika:
## Testing

Four configurations worth covering. Everything below is cross-compiled from an x86_64 host; drop the `-from-x86_64` suffix to build natively on the Jetson.

| # | Target | Unlock | Rootfs |
|---|--------|--------|--------|
| 1 | `nvidia-jetson-orin-agx-debug-luks` | device-unique key | eMMC (`mmcblk0`) |
| 2 | `nvidia-jetson-orin-nx-debug-luks` | device-unique key | USB (`sda`) |
| 3 | `nvidia-jetson-orin-agx-debug` | none | eMMC |
| 4 | AGX + `userPassphrase` override | console prompt | eMMC |

3 is the regression case: it shares the `flashScriptOverrides.deviceDisk*` options, the reworked `boot.initrd.systemd.services` attrset and the `ghaf-ensure-system-profile` script with the LUKS path, and it must keep booting unchanged.

### Build and flash (1, 2, 3)

```sh
nix build .#<target>-from-x86_64
nix build .#<target>-from-x86_64-flash-script
```

Put the board in recovery mode, then for `-luks` targets:

```sh
```

and for non-LUKS:

```sh
cd result && sudo ./flash.sh
```

No passphrase prompt at flash time in either case — the root is encrypted during the image build, and the flash script writes the encrypted APP partition as-is.

### First boot, LUKS targets (1, 2)

Watch the serial console. The first boot rotates from the manufacturing passphrase to the OP-TEE device-unique key and re-encrypts the root, which takes roughly a minute per GB — a 32GB AGX sits on `Switching to use device unique key` for a long while. This is the change most worth eyeballing, since the `2.2.x` version gate that used to skip `cryptsetup reencrypt` is gone.

It should reach `ghaf-host login:` without ever prompting, and then:

```sh
lsblk                                   # rootfs partition -> crypt -> /
findmnt /                               # /dev/mapper/cryptroot
sudo blkid /dev/mmcblk0p2               # TYPE="crypto_LUKS"  (sda2 on NX)
ls -l /dev/disk/by-uuid/0ada0e5b-0e5b-4e5b-8e5b-0000000000a9
sudo cryptsetup luksDump /dev/mmcblk0p2 | grep -A2 Tokens   # luksDiskDeviceUniqueKey
df -h /                                 # grown to fill the disk
```

`luksDump` must show the `luksDiskDeviceUniqueKey` token and no remaining passphrase keyslot. If the manufacturing passphrase (`ghaf`) still unlocks the disk, the rekey silently failed:

```sh
echo ghaf | sudo cryptsetup open --test-passphrase /dev/mmcblk0p2   # must fail
```

### Second boot (1, 2)

```sh
sudo reboot
```

Must be fast — no re-encryption, no prompt. The resize is idempotent and skips via a marker on the ESP:

```sh
sudo journalctl -b -u resize-partitions   # "Resize already performed, skipping."
```

Reboot once more if you want to confirm the marker survives.

### Non-LUKS regression (3)

Boots to login as before. Additionally:

```sh
systemctl status ghaf-ensure-system-profile   # active (exited), no PATH errors
readlink -f /nix/var/nix/profiles/system      # matches /run/current-system
```

That service lost its `${pkgs.coreutils}/bin/` prefixes in this pass, so a broken `PATH` would show up here and nowhere else.

### userPassphrase (4)

No flake target generates this — it is the testing-only unlock mode. Build with an override:

```nix
ghaf.hardware.nvidia.orin.diskEncryption = {
  enable = true;
  deviceUniqueKey.enable = false;
  userPassphrase.enable = true;
  userPassphrase.passphrase = "ghaf";
};
```

Initrd prompts on the serial console for the passphrase; the first-boot resize prompts a second time. Both take the value of `userPassphrase.passphrase`.

### NX on NVMe

The NX target defaults to USB. For NVMe, override all three:

```nix
ghaf.hardware.nvidia.orin.flashScriptOverrides = {
  deviceDisk = "nvme0n1";
  deviceDiskEspPartition = "nvme0n1p1";
  deviceDiskRootfsPartition = "nvme0n1p2";
};
```

Setting only some of them is now an eval error rather than a silent flash to the wrong device — worth confirming by dropping one and checking that `nix eval` fails with the `deviceDisk*` assertion.

vadika:
The red `build` and `Analyze` checks here are not from this branch — `actions/checkout` v7 refuses fork checkout under `pull_request_target`, and `codeql-action/init` was left behind its siblings. Both landed on `main` after this branch's base: the ten `build` jobs were green on `ec0c30f2` and fail on the next push with no workflow change in between.

Fix is #2032. `eval` and `jenkins-pre-merge` are the checks that exercise this branch.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -257,6 +276,309 @@ let
     mkIf
     types
     ;
+
+  # Shared by deviceDisk, deviceDiskRootfsPartition and deviceDiskEspPartition.
+  deviceDiskDescription = ''
```
kajusnau: Great, this is exactly what I had in mind with my original comment.
The only nitpick I have with this is that it should probably be defined closer to where it's used, e.g. line 585, where options are declared

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
kajusnau: Maybe I'm still missing something here, but how will line 396 be reached if 391 evaluates to `false`?

Line 391 opens an optional string that closes at line 407, so anything within that block will never be reached/included if `cfg.diskEncryption.enable == false`?

modules/reference/hardware/jetpack/agx/orin-agx-industrial.nix
```diff
@@ -24,7 +24,9 @@
         agx.enableNetvmWlanPCIPassthrough = true;
         carrierBoard = "devkit";
         # AGX industrial devkit boots rootfs from eMMC.
-        flashScriptOverrides.rootfsDevice = "mmcblk0p1";
+        flashScriptOverrides.deviceDisk = "mmcblk0";
+        flashScriptOverrides.deviceDiskEspPartition = "mmcblk0p1";
+        flashScriptOverrides.deviceDiskRootfsPartition = "mmcblk0p2";
```
kajusnau: Not strictly necessary, but a readability suggestion:

```suggestion
        flashScriptOverrides = {
          deviceDisk = "mmcblk0";
          deviceDiskEspPartition = "mmcblk0p1";
          deviceDiskRootfsPartition = "mmcblk0p2";
        };
```

kajusnau (approved):

brianmcgillion (approved):

milva-unikie:
I was just finishing testing luks image on NX
- SSH connection to the device does not work at all, only serial to host works
- `resize-partitions.service` and `pre-disk-unique-key.service` fail

Also noticed that building cross-compiled luks images does not work without aarch builder. They fail with
```
> Encrypting extracted root image with LUKS2 ...
       > /nix/store/x93ybh3d28isxdc4c1icgy4s4gncdlfx-stdenv-linux/setup: line 1869: /nix/store/cc29f0qjsmv0fr59xcpr9xl0wwjg7ags-cryptsetup-aarch64-unknown-linux-gnu-2.8.6-bin/bin/cryptsetup: cannot execute binary file: Exec format error
```

I will open bugs for these issues.

PR #1837, opened 2026-03-18
SPIRE integration

gngram:
## Description of Changes

This PR refactors  SPIRE (formerly referred as SPIFFE module) server and agent implementation and reduces reduces the memory footprint of the SPIRE binaries.

- Includes an option to use custom spire package.
- A custom spire package(spire-min) is overlay-ed by stripping of unused cloud-specific modules to reduce memory foot print.
- standardized naming conventions by renaming remaining spiffe references to spire.
- Server health report published, server health is checked before initiating agent to avoid failure when server is not ready.
- implemented a generic agent registration flow based on the specific node attestation method used by each VM.
- Updated common configurations to allow the server to dynamically retrieve VM lists, per-VM workloads, and attestation methods.
- per-VM workload registration, earlier there was no separation of workloads.
- single global config to enable or disable the spiffe identity issuance using spire

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

1. check the status of spire-agent service in each vm(except admin), it should be running and without any error.
2.  check the status of spire-server service in admin vm, it should be running and without any error.

### DIFF
--- a/modules/common/security/spire/agent.nix
+++ b/modules/common/security/spire/agent.nix
@@ -1,0 +1,162 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.ghaf.security.spire.agent;
+  dataDir = "/var/lib/spire/agent";
+  runtimeDataDir = "/run/spire/agent";
+  inherit (lib)
+    getExe
+    mkIf
+    mkOption
+    mkEnableOption
+    optionalString
+    types
+    ;
+  spire-package = config.ghaf.common.spire.package;
+  healthCheckPort = toString config.ghaf.common.spire.server.healthCheckPort;
+
+  joinTokenConf = optionalString (cfg.nodeAttestationMode == "join_token") ''
+    join_token_file = "${cfg.settings.join_token.token}"
+  '';
+  joinTokenPlugin = optionalString (cfg.nodeAttestationMode == "join_token") ''
+    NodeAttestor "join_token" {
+      plugin_data {}
+    }
+  '';
+  agentConf = ''
+    agent {
+      data_dir = "${dataDir}"
+      log_level = "${cfg.logLevel}"
+      server_address = "${config.ghaf.common.spire.server.address}"
+      server_port = ${toString config.ghaf.common.spire.server.port}
+      trust_domain = "${config.ghaf.common.spire.server.trustDomain}"
+      trust_bundle_path = "${cfg.trustBundlePath}"
+      socket_path = "${runtimeDataDir}/agent.sock"
+      ${joinTokenConf}
+    }
+
+    plugins {
+      ${joinTokenPlugin}
+
+      WorkloadAttestor "unix" {
+        plugin_data {}
+      }
+      KeyManager "disk" {
+        plugin_data {
+          directory = "${dataDir}/keys"
+        }
+      }
+    }
+  '';
+
+  server-health = pkgs.writeShellApplication {
+    name = "spire-server-healthcheck";
+
+    runtimeInputs = [
+      pkgs.curl
+    ];
+
+    text = ''
+      SERVER_URL="http://${config.ghaf.common.spire.server.address}:${healthCheckPort}/ready"
+      MODE="${cfg.nodeAttestationMode}"
+
+      echo "Checking SPIRE Server readiness at $SERVER_URL..."
+
+      until curl --fail --silent "$SERVER_URL" > /dev/null 2>&1; do
+        echo "SPIRE Server is not ready yet. Retrying in 3 seconds..."
+        sleep 3
+      done
+
+      if [ "$MODE" == "join_token" ]; then
+        while [ ! -e "${cfg.settings.join_token.token}" ]; do
+          echo "Waiting for server token..."
+          sleep 1
+        done
+      fi
+      echo "SPIRE Server is ready! Starting SPIRE Agent..."
+    '';
+  };
+in
+{
+  _file = ./agent.nix;
+
+  options.ghaf.security.spire.agent = {
+    enable = mkEnableOption "SPIRE agent";
+    nodeAttestationMode = mkOption {
+      type = types.spireNodeAttestationMode;
+      default = "join_token";
+      description = "Node attestation mode";
+    };
+    workloads = mkOption {
+      type = types.spireWorkloads;
+      default = [ ];
+      description = "List of workloads for this spire agent";
+    };
+    trustBundlePath = mkOption {
+      type = types.path;
+      default = "/etc/common/spire/bundle.pem";
+      description = "Path to the SPIRE trust bundle PEM file (used to verify the server during bootstrap)";
+    };
+    logLevel = mkOption {
+      type = types.str;
+      default = "INFO";
+      description = "SPIRE server log level";
+    };
+    settings = {
+      join_token = {
+        token = mkOption {
+          type = types.path;
+          default = "/etc/common/spire/tokens/${config.networking.hostName}.token";
+          description = "SPIRE server log level";
+        };
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.etc."spire/agent.conf".text = agentConf;
+    services.spire.agent = {
+      enable = true;
+      package = spire-package;
+      configFile = "/etc/spire/agent.conf";
+    };
+
+    environment.systemPackages = [ spire-package ];
+    systemd = {
+      services = {
+        spire-agent = {
+          requires = [ "network-online.target" ];
+          after = [
+            "network-online.target"
+          ];
+
+          unitConfig = {
+            RequiresMountsFor = [ "/etc/common" ];
+          };
+
+          serviceConfig = {
+            ExecStartPre = getExe server-health;
+            NoNewPrivileges = true;
+            PrivateTmp = true;
+            ProtectSystem = "strict";
+            ProtectHome = true;
+            ReadWritePaths = [
+              "${dataDir}"
+              "${runtimeDataDir}"
+            ];
+          };
+        };
+      };
+      tmpfiles.rules = [
+        "d /run/spire 0755 root root - -"
+        "d ${runtimeDataDir} 2750 spire-agent spire-agent - -"
+      ];
+    };
+  };
+}
--- a/modules/common/security/spire/server.nix
+++ b/modules/common/security/spire/server.nix
@@ -1,0 +1,266 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.ghaf.security.spire.server;
+  inherit (lib)
+    filterAttrs
+    getExe
+    mkIf
+    mkOption
+    mkEnableOption
+    types
+    optionalString
+    concatMapStringsSep
+    ;
+  runtimeDataDir = "/run/spire/server";
+  tokenDir = "/etc/common/spire/tokens";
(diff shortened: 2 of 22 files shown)


kajusnau (commented):
Some basic comments, also suggest using `getExe` throughout the patch, since it's already used in some places but not all.

overlays/custom-packages/spire-mini/0001-removed-cloud-infra-to-reduce-memory-footprint.patch
```diff
@@ -0,0 +1,265 @@
+From c00cf4f1dd5a3fedd7dc701828d476efe1f09e4a Mon Sep 17 00:00:00 2001
+From: Ganga Ram <Ganga.Ram@tii.ae>
+Date: Tue, 17 Feb 2026 12:46:42 +0400
+Subject: [PATCH] removed cloud infra to reduce memory footprint
```
vunnyso: Can we can try to push this patch to upstream?

vadika (changes requested):
## Review

Nine issues found — two architectural, seven code-level.

---

### 1. PR scope: module definitions only, no deployment wiring

The PR title and description frame this as module/packaging/plumbing work rebased on mainline. The `lib/` and `modules/common/security/spire/` additions are correct scope. However the second commit adds deployment wiring across 10+ VM files that *set* options rather than *define* them:

- `modules/microvm/host/microvm-host.nix`
- `modules/microvm/sysvms/adminvm-base.nix`, `adminvm.nix`
- `modules/microvm/sysvms/appvm-base.nix`
- `modules/microvm/sysvms/audiovm-base.nix`, `audiovm.nix`
- `modules/microvm/sysvms/guivm-base.nix`, `guivm.nix`
- `modules/microvm/sysvms/idsvm/idsvm.nix`
- `modules/microvm/sysvms/netvm-base.nix`, `netvm.nix`
- `modules/microvm/appvm.nix`

These belong in a follow-up PR. This PR should stop at `lib/`, `modules/common/security/spire/`, `modules/common/common.nix`, and `lib/global-config.nix`.

---

### 2. Deleted file had unresolved merge conflict markers

`modules/common/services/yubikey.nix.orig` — deleted here, but the file contained `<<<<<<< HEAD` / `=======` / `>>>>>>>` markers. It was tracked with an unresolved conflict, which means it was never properly resolved before being staged. The deletion is correct; just noting this should not have been committed in the first place.

---

### 3. Duplicate `awk` line — `modules/common/security/spire/server.nix`

Inside `spireGenerateJoinTokensApp`:

```sh
token=$(echo "$output" | awk '/^Token:/ {print $2}')

# Extract the token from the successful output
token=$(echo "$output" | awk '/^Token:/ {print $2}')
```

Verbatim duplicate. The first assignment is immediately overwritten. Remove one of them along with the dead comment.

---

### 4. `x509pop` referenced but absent from the enum — `lib/default.nix` + `create-workload-entries.nix`

`lib/default.nix` declares:
```nix
spireNodeAttestationMode = lib.types.enum [ "join_token" ];
```

`modules/common/security/spire/create-workload-entries.nix` branches on:
```nix
if (agentCfg.nodeAttestationMode == "x509pop") then ...
```

`"x509pop"` cannot pass Nix type validation, so this branch is permanently unreachable and silently dead. Either add `"x509pop"` to the enum, or remove the dead branch.

---

### 5. Join token files are world-readable — `modules/common/security/spire/server.nix`

```sh
chmod 0644 "$tokenFile"
```

Join tokens are single-use bootstrap credentials. `0644` allows any local user to read them before the agent consumes the token. At a minimum `0640` with root:spire-agent ownership. Consider `0600` root-only if the agent service runs as root.

---

### 6. Health-check port 8080 hardcoded in three places

- `modules/common/security/spire/server.nix` — `bind_port = "8080"` in `serverConf`
- `modules/common/security/spire/server.nix` — `8080` in `networking.firewall.allowedTCPPorts`
- `modules/common/security/spire/agent.nix` — `SERVER_URL="http://...:8080/ready"`

If the health port ever changes, all three need updating and it's easy to miss the agent-side. Should be a shared `let` binding (e.g. `healthCheckPort = 8080`) or a configurable option alongside `server.port`.

---

### 7. `PrivateTmp = false` needs justification — `modules/common/security/spire/server.nix`

```nix
serviceConfig = {
  PrivateTmp = false;
  ...
};
```

`false` is the default — this is a no-op on a vanilla system, but it explicitly overrides any stricter default the upstream nixpkgs `services.spire.server` module might set. That is a security regression by overriding. Either remove it (if not needed) or add a comment explaining why it must be unset (e.g., interaction with tmpfiles-managed socket dirs).

---

### 8. `spire-publish-bundle` missing `RemainAfterExit` — `modules/common/security/spire/server.nix`

`spire-generate-join-tokens` and `spire-create-workload-entries` are both `Type = "oneshot"` with `RemainAfterExit = true`, so dependents can order against their completion. `spire-publish-bundle` is also `Type = "oneshot"` but lacks `RemainAfterExit`, meaning systemd considers it inactive after it exits. Agents waiting on the bundle file have no reliable ordering anchor. Likely contributes to the reported Orin failures. Add `RemainAfterExit = true`.

---

### 9. Wrong package in health-check `runtimeInputs` — `modules/common/security/spire/agent.nix`

```nix
runtimeInputs = [
  pkgs.curl
  pkgs.spire-agent
];
```

`pkgs.spire-agent` is not a top-level nixpkgs attribute — `spire-agent` is a binary inside `pkgs.spire`. This will fail to evaluate. Use `spire-package` which is already in scope in the same `let` block, and ensures the custom/overridden package is used.

vadika (approved):
LGTM!

modules/microvm/host/microvm-host.nix
```diff
@@ -128,12 +128,21 @@ in
...
+          spire.agents = lib.mkIf config.ghaf.security.spire.agent.enable {
+            ghaf-host = {
+              inherit (config.ghaf.security.spire.agent) nodeAttestationMode workloads;
+            };
+          };
+        };
+
+        security.spire.agent = {
+          inherit (config.ghaf.global-config.spire) enable;
+          logLevel = if config.ghaf.globalConfig.spire.debug then "DEBUG" else "INFO";
```
everton-dematos: Shouldn't it be `global-config` instead of `globalConfig`?
When I tried to enable it and build, I got the following error: `error: attribute 'globalConfig' missing`
Then, when I changed to `global-config`, it passed

everton-dematos (changes requested):

modules/microvm/host/microvm-host.nix
gngram: Thanks, I had fixed it in other PR, but missed it here. Will fix it.

everton-dematos (approved):
Tested on `lenovo-x1-carbon-gen10-debug` 

I changed https://github.com/gngram/ghaf/blob/pull/spire_module_update/lib/global-config.nix#L343 to `true` in order to test the last added commit (https://github.com/tiiuae/ghaf/pull/1837/changes/0e8733db22c488970f0863bc4821eafff3bf6d40).

Everything is working fine, considering this PR is the basic SPIRE support.

SPIRE Agents on VMs:
```
$ systemctl status spire-agent.service 
● spire-agent.service - SPIRE agent
     Loaded: loaded (/etc/systemd/system/spire-agent.service; enabled; preset: ignored)
     Active: active (running) since Mon 2026-04-06 12:36:52 +04; 2min 28s ago
 Invocation: d67553bff6f64dc6a5c7fc68b8462115
    Process: 1357 ExecStartPre=/nix/store/f89pr34j4vs0zfr3w2wv8r8znj2z6hfi-spire-server-healthcheck/bin/spire-server-healthcheck (code=exited, status=0/SUCCESS)
   Main PID: 1370 (spire-agent)
         IP: 51.1K in, 29.5K out
         IO: 0B read, 0B written
      Tasks: 8 (limit: 1121)
     Memory: 46.2M (peak: 65M)
        CPU: 285ms
     CGroup: /system.slice/spire-agent.service
             └─1370 /nix/store/yi5ijaqp4k1z2wz4a2k1wsdw9zfllhhm-spire-1.14.4/bin/spire-agent run --config=/etc/spire/agent.conf --expandEnv
```

SPIRE Server on `admin-vm`:
```
$ systemctl status spire-server.service 
● spire-server.service - SPIRE Server
     Loaded: loaded (/etc/systemd/system/spire-server.service; enabled; preset: ignored)
     Active: active (running) since Mon 2026-04-06 12:36:45 +04; 3min 9s ago
 Invocation: 49ec85e96381411c9ff5184f5de9d2c4
       Docs: https://spiffe.io/docs/latest/deploying/spire_server/
   Main PID: 1130 (spire-server)
         IP: 267.1K in, 458.8K out
         IO: 1.4M read, 2.3M written
      Tasks: 9 (limit: 518)
     Memory: 30.5M (peak: 56M, swap: 936K, swap peak: 1.7M)
        CPU: 1.685s
     CGroup: /system.slice/spire-server.service
             └─1130 /nix/store/yi5ijaqp4k1z2wz4a2k1wsdw9zfllhhm-spire-1.14.4/bin/spire-server run --config=/etc/spire/server.conf --expandEnv
```

leivos-unikie:
Everton already checked the services with spire enabled on gen10.

Also checked this with spire disabled on lenovo-x1-carbon-gen11-debug-installer, nothing seems to be broken.

modules/common/security/spire/agent.nix
```diff
@@ -0,0 +1,162 @@
...
+      default = "join_token";
+      description = "Node attestation mode";
+    };
+    workloads = mkOption {
+      type = types.spireWorkloads;
+      default = [ ];
+      description = "List of workloads for this spire agent";
+    };
+    trustBundlePath = mkOption {
+      type = types.str;
```
vunnyso:
```suggestion
      type = types.path;
```

modules/common/security/spire/server.nix
```diff
@@ -0,0 +1,267 @@
...
+    enable = mkEnableOption "SPIRE server";
+
+    logLevel = mkOption {
+      type = types.str;
+      default = "INFO";
+      description = "SPIRE server log level";
+    };
+
+    trustBundlePath = mkOption {
+      type = types.str;
```
vunnyso:
```suggestion
      type = types.path;
```

modules/common/security/spire/server.nix
```diff
@@ -0,0 +1,267 @@
...
+        };
+        spire-publish-bundle = {
+          description = "Publish SPIRE trust bundle (PoC)";
+          wantedBy = [ "multi-user.target" ];
+          after = [ "spire-server.service" ];
+          wants = [ "spire-server.service" ];
+
+          serviceConfig = {
+            Type = "oneshot";
+            #RemainAfterExit = true;
```
vunnyso: Can remove deadcode?

overlays/custom-packages/spire/default.nix
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{ prev }:
+prev.spire.overrideAttrs (_old: rec {
+  version = "1.14.4";
```
vunnyso, line 7: Could you add a brief comment of why we specifically need to use version `1.14.4`?

modules/common/security/spire/agent.nix
gngram: done

modules/common/security/spire/server.nix
gngram: done

modules/common/security/spire/server.nix
gngram: done

overlays/custom-packages/spire/default.nix
gngram: done

brianmcgillion (approved):

PR #1845, opened 2026-03-22
overlay hunt: remove unused libfm overlay

vadika:
## Summary
- Remove the libfm overlay — it patches a library that nothing in Ghaf depends on
- libfm is the core library of PCManFM file manager, but Ghaf uses COSMIC desktop with `cosmic-files` which has no libfm dependency
- Verified: `nix why-depends nixpkgs#cosmic-files nixpkgs#libfm` → "does not depend on"
- Zero references to `pcmanfm` or `libfm` anywhere in the Ghaf codebase
- As a library overlay, this was needlessly changing the derivation hash of any transitive dependent, forcing rebuilds from source

## Test plan
- [ ] Verify `nix build .#lenovo-x1-carbon-gen11-debug` succeeds (libfm not in closure)
- [ ] Verify COSMIC file manager still works (it uses cosmic-files, not pcmanfm/libfm)


kajusnau (approved):
Good to remove IMO

nesteroff (approved):
I believe support for PCManFM was removed in https://github.com/tiiuae/ghaf/commit/9997d215c1541abfc2d9d902fcfcd19383587a15 so the overlay is no longer needed. It was used to make the file manager automatically refresh folder contents when they were changed from a VM.

brianmcgillion (approved):

PR #1850, opened 2026-03-23
cross: drop gfortran from fftw in cross-compilation overlay

vadika:
## Summary

- Remove gfortran from `fftwFloat.nativeBuildInputs` in the cross-compilation overlay
- PipeWire → fftwFloat (fftw-single) → gfortran dependency chain forces cross-compilation of the entire Fortran toolchain for aarch64, which is not in the binary cache
- Ghaf does not use FFTW Fortran bindings; the Fortran wrapper is only generated when docs are enabled (`--disable-doc` is already set for cross builds)

## Test plan

- [ ] `nix build --dry-run .#packages.x86_64-linux.nvidia-jetson-orin-agx-debug-from-x86_64` no longer includes `aarch64-unknown-linux-gnu-gfortran` in derivations to build
- [ ] Full cross build of Jetson Orin target succeeds
- [ ] Audio (PipeWire) functionality unaffected on target

### DIFF
--- a/overlays/cross-compilation/default.nix
+++ b/overlays/cross-compilation/default.nix
@@ -4,6 +4,16 @@
 # This overlay is for specific fixes needed only to enable cross-compilation.
 #
 (final: prev: {
+  # Remove gfortran from FFTW to avoid cross-compiling the entire Fortran
+  # toolchain. FFTW is pulled in by PipeWire for audio processing. The Fortran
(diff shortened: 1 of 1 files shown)


avnik (approved):
Looks good. But maybe also add
`.overlay { withDocs = false; }`? to strip doc generation entriely?

brianmcgillion (approved):

brianmcgillion (approved):

PR #1713, opened 2026-01-27
fix(orin): enroll UEFI secure boot keys from certs

vadika:
Summary

  - Generate ESLs from repo PK/KEK/db certs at build time for Jetson Orin.
  - Wire ESLs into jetpack‑nixos UEFI secure boot enrollment.
  - Enable the new secure boot module by default for Orin targets.

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

- [ ] Clear summary in PR description
- [ ] Detailed and meaningful commit message(s)
- [ ] Commits are logically organized and squashed if appropriate
- [ ] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] Author has run `make-checks` and it passes
- [ ] All automatic GitHub Action checks pass - see [actions](https://github.com/tiiuae/ghaf/actions)
- [ ] Author has added reviewers and removed PR draft status

### Applicable Targets
- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [ ] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [ ] System 76 `x86_64`

### Installation Method
- [ ] Requires full re-installation
- [ ] Can be updated with `nixos-rebuild ... switch`
- [ ] Other:

### Test Steps To Verify:

Build, flash and reboot -- SB should be enabled on device

PR #1862, opened 2026-03-27
feat(jetson-orin): A/B verity boot with secure boot and OTA updates

Mic92:
## Description

A/B verity boot end-to-end on Jetson AGX Orin, with UEFI Secure Boot and OTA update support.

**Verity boot**: UKI-based boot with dm-verity root filesystem on erofs (lz4hc compressed). First-boot service creates swap/persist/B-slot placeholder on the device.

**Secure Boot**: Cherry-picks #1713 for UEFI key enrollment, then adds flash-time EFI signing with sbsign. Private keys stay out of the Nix store — the flash script reads them from `SECURE_BOOT_SIGNING_KEY_DIR` at runtime (falls back to embedded dev-keys path). OTA update UKIs are signed at build time in debug builds (asserts on `debugEnable`); in production the OTA server signs images before distribution.

**OTA updates**: Enable givc on Orin with upstream ghaf-givc (PR tiiuae/ghaf-givc#372 merged). The manifest includes `unpacked_size` so ota-update can create and resize LVM slots on demand.

~Depends on https://github.com/tiiuae/ghaf-givc/pull/372 (merged)~

## Verified on Jetson AGX Orin devkit

Tested on physical Jetson AGX Orin devkit, cross-compiled from x86_64-linux.

- ✅ Flash with signed EFIs — boots with `Secure Boot: enabled (user)`
- ✅ First-boot creates swap (4G), persist (~39G), B-slot placeholder
- ✅ `ota-update image install` writes signed UKI + root + verity to auto-created LVs
- ✅ `bootctl set-oneshot` → reboot into slot B (26.03.2) — accepted by Secure Boot
- ✅ Slot B shows `active: true` after reboot
- ✅ `ota-update image remove` recycles old slot
- ✅ Unsigned UKI rejected by UEFI firmware (tested with previous unsigned build)

### Known limitation

OTA update UKI signing uses `uki-signing-key-dir` which places dev private keys in the Nix store. This is gated by an assertion to debug builds only. In production, the OTA server must sign UKIs before distribution — this integration is not yet implemented. See `secureboot.mdx` for details.

### Build and flash
```bash
nix build .#nvidia-jetson-orin-agx-verity-debug-from-x86_64-flash-script

# Put Jetson in recovery mode (hold recovery button + press reset)
# Verify: lsusb | grep "0955:7023"
```

### Verify boot (serial console or SSH)
```bash
bootctl status
# Expected: Secure Boot: enabled (user)

ota-update image status
# Expected: Slot A active, root ~6.2G + verity ~0.1G
```

### Test OTA update
```bash
# Build update image (bump version first)
echo "26.03.2" > .version
nix build .#nvidia-jetson-orin-agx-verity-debug-from-x86_64-ghafImage

# Copy to device and install
scp result/* root@<jetson-ip>:/tmp/
ssh root@<jetson-ip> ota-update image install --manifest /tmp/ghaf_26.03.2_*.manifest

# Switch to slot B and reboot
ssh root@<jetson-ip> "bootctl set-oneshot ghaf-26.03.2-*.efi && reboot"

# After reboot, verify slot B is active
ota-update image status
# Expected: slot B (26.03.2) active: true
```


vunnyso (dismissed):

milva-unikie:
Does not build when rebased to mainline, this should be fixed before testing
```
13:11:50  error:
13:11:50         … while evaluating the attribute 'optionalValue.value'
13:11:50           at /nix/store/fvb3fhm8c193i6692v5223bxpa5w2nv7-source/lib/modules.nix:1297:5:
13:11:50           1296|
13:11:50           1297|     optionalValue = if isDefined then { value = mergedValue; } else { };
13:11:50               |     ^
13:11:50           1298|   };
13:11:50  
13:11:50         … while evaluating a branch condition
13:11:50           at /nix/store/fvb3fhm8c193i6692v5223bxpa5w2nv7-source/lib/modules.nix:1297:21:
13:11:50           1296|
13:11:50           1297|     optionalValue = if isDefined then { value = mergedValue; } else { };
13:11:50               |                     ^
13:11:50           1298|   };
13:11:50  
13:11:50         … while evaluating definitions from `/nix/store/bdlxvwrzjcl4s4vabyga52y82asjf3sz-source/targets/nvidia-jetson-orin/flake-module.nix':
13:11:50  
13:11:50         (stack trace truncated; use '--show-trace' to show the full, detailed trace)
13:11:50  
13:11:50         error: The option `ghaf.partitioning.verity-volume' does not exist. Definition values:
13:11:50         - In `/nix/store/fvb3fhm8c193i6692v5223bxpa5w2nv7-source/flake.nix':
13:11:50             {
13:11:50               enable = true;
13:11:50               uki-signing-key-dir = {
13:11:50                 _type = "if";
13:11:50                 condition = true;
13:11:50             ...
13:11:50  
13:11:50         Did you mean `ghaf.partitioning.verity'?
```

leivos-unikie:
> Does not build when rebased to mainline, this should be fixed before testing
> 
> ```
> 13:11:50  error:
> 13:11:50         … while evaluating the attribute 'optionalValue.value'
> 13:11:50           at /nix/store/fvb3fhm8c193i6692v5223bxpa5w2nv7-source/lib/modules.nix:1297:5:
> 13:11:50           1296|
> 13:11:50           1297|     optionalValue = if isDefined then { value = mergedValue; } else { };
> 13:11:50               |     ^
> 13:11:50           1298|   };
> 13:11:50  
> 13:11:50         … while evaluating a branch condition
> 13:11:50           at /nix/store/fvb3fhm8c193i6692v5223bxpa5w2nv7-source/lib/modules.nix:1297:21:
> 13:11:50           1296|
> 13:11:50           1297|     optionalValue = if isDefined then { value = mergedValue; } else { };
> 13:11:50               |                     ^
> 13:11:50           1298|   };
> 13:11:50  
> 13:11:50         … while evaluating definitions from `/nix/store/bdlxvwrzjcl4s4vabyga52y82asjf3sz-source/targets/nvidia-jetson-orin/flake-module.nix':
> 13:11:50  
> 13:11:50         (stack trace truncated; use '--show-trace' to show the full, detailed trace)
> 13:11:50  
> 13:11:50         error: The option `ghaf.partitioning.verity-volume' does not exist. Definition values:
> 13:11:50         - In `/nix/store/fvb3fhm8c193i6692v5223bxpa5w2nv7-source/flake.nix':
> 13:11:50             {
> 13:11:50               enable = true;
> 13:11:50               uki-signing-key-dir = {
> 13:11:50                 _type = "if";
> 13:11:50                 condition = true;
> 13:11:50             ...
> 13:11:50  
> 13:11:50         Did you mean `ghaf.partitioning.verity'?
> ```

Checked also that attempt to rebase manually produces merge conflict at flake.lock and if choosing the main version build fails.

Mic92:
I can work on a fix when my access to Jenkins is restored.

flake.nix
```diff
@@ -156,7 +156,7 @@
     };
 
     nix-store-veritysetup-generator = {
-      url = "github:tiiuae/ghaf-nix-store-veritysetup-generator/ghaf";
+      url = "github:Mic92/ghaf-nix-store-veritysetup-generator/fix-cross-compilation";
```
avnik: Could be reverted, I merged your fix there

avnik (changes requested):
Only one request!
Your cross fix on veritysetup-generator is merged, so you could unstage this change and update lockfile instead

flake.nix
Mic92: @avnik done. The bump is still needed because it hasn't been bump in ghaf yet.

avnik (approved):
I believe here no changes, except flake.lock bumping.
So let go with it

leivos-unikie:
Tested on Orin AGX
- ota-update was not available on net-vm
- did ssh to ghaf-host and ran `ota-update image install --manifest /persist/sysupdate/ghaf_26.06.2_*.manifest` there

nix-gc.service failed in reboot to updated version, otherwise all good

See the details:
[PR1862_testing.txt](https://github.com/user-attachments/files/28631166/PR1862_testing.txt)

leivos-unikie:
Added bug label because of that nix-gc.service.

Can somebody comment if that is to be considered blocking issue?

avnik (approved):
LGTM

Also regarding nix-gc, it should be disabled for a/b updated systems, because /nix is readonly

Mic92:
@leivos-unikie The ota-update is only available from the host because it needs to write to the filesystem of the whole machine. However I don't think this is a particular concern at the moment because this is replaced by an OCI updater as far as I know.

Mic92:
Rebased onto main. The veritysetup-generator bump and the givc commit are gone, both already covered by main. The flash-time signing now sits on top of the new secureboot.nix from the orin-nx work, only adding the signingKeyDir option. Verity debug targets enroll the dev certs, release keeps the production certs from modules/secureboot/keys. Also disabled nix.gc on verity targets since the store is read-only anyway. The plain AGX/NX sd-image targets are bit-identical to main (checked with nix-diff).

avnik (approved):

avnik:
@brianmcgillion Brian, could you take a look/merge, before it out of sync again

milva-unikie:
**Tested on Orin-AGX**

Samuli is on holiday, so I tested using the same steps he used a few weeks ago. Everything works, `nix-gc.service` no longer fails.

brianmcgillion (approved):

PR #1864, opened 2026-03-30
feat(active-directory): GPO-based access control with SSSD

juliuskoskela:
## Description of Changes

Enable AD/SSSD GPO-based access control so that only users explicitly granted
"Allow log on locally" rights via Group Policy can log into Ghaf devices through
greetd/cosmic-greeter.

### Changes

- **Enable AD profile for trial target** — switch `mvp-user-trial` from
  `homed-user` to `ad-users` profile and update the DNS provider IP.
- **GPO enforcement in SSSD** — change `gpoAccessControl` from `"permissive"`
  to `"enforcing"`, enable `ad_gpo_implicit_deny`, set `ad_site`, and map
  `greetd` + `cosmic-greeter` to interactive logon rights via
  `ad_gpo_map_interactive`.
- **PAM strict access** — enable `sssdStrictAccess` on the display manager PAM
  service so that SSSD account-phase deny decisions (from GPO evaluation) are
  actually enforced at login time.

### AD Server-Side Setup Performed

On the domain controller (`vm-ghaf-dev-dc.ghaf-test.com`):

1. Created GPO **"Ghaf Device Login Restriction"** linked to the domain.
2. Configured **Computer Configuration → Policies → Windows Settings → Security
   Settings → Local Policies → User Rights Assignment → "Allow log on locally"**
   to include only the allowed test user (`johnsmith`).
3. Verified GPO link is enabled and security filtering includes
   `Authenticated Users`.

### Known Issue Under Investigation

With `gpoAccessControl = "enforcing"` and the GPO configured on the AD side, a
denied user (`janedoe`) was still able to log in. Investigation identified two
potential root causes:

1. **Client-side (addressed in this PR):** Without `sssdStrictAccess = true`,
   NixOS PAM may not enforce the SSSD account-phase deny for display manager
   login paths. This PR adds that setting.
2. **Server-side (under investigation):** The computer object may be in the
   default `Computers` container rather than a proper OU — GPOs cannot be linked
   to the default container. A verification checklist has been prepared covering:
   - Computer object OU placement
   - GPO link and security filtering
   - Inheritance blocking
   - `gpresult /r` validation on the DC

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [ ] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

1. Build and flash `lenovo-x1-carbon-gen11-debug` target.
2. Boot device, complete provisioning TUI, and join `ghaf-test.com` domain.
3. Reboot after domain join.
4. **Allowed user:** Log in as `johnsmith` via cosmic-greeter — login should succeed.
5. **Denied user:** Attempt to log in as `janedoe` via cosmic-greeter — login should be denied.
6. If denial does not work, collect logs:
   ```bash
   journalctl -u sssd -b | grep -i gpo
   journalctl -b | grep -iE 'pam|sss|greetd|cosmic'
   ```
7. Verify GPO cache is populated:
   ```bash
   ls -la /var/lib/sss/gpo_cache/
   ```

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -164,6 +164,7 @@ export default defineConfig({
                         "ghaf/dev/guides/extending-targets",
                         "ghaf/dev/guides/downstream-setup",
                         "ghaf/dev/guides/migration",
+                        "ghaf/dev/guides/ad-login",
                       ],
                     },
                     {
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -165,6 +165,7 @@ export default defineConfig({
                         "ghaf/dev/guides/extending-targets",
                         "ghaf/dev/guides/downstream-setup",
                         "ghaf/dev/guides/migration",
+                        "ghaf/dev/guides/ad-login",
                       ],
                     },
                     {
--- a/docs/src/content/docs/ghaf/dev/guides/ad-login.mdx
+++ b/docs/src/content/docs/ghaf/dev/guides/ad-login.mdx
@@ -1,0 +1,340 @@
+---
+title: "Active Directory Login with GPO Access Control"
+---
+{/*
+SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+SPDX-License-Identifier: CC-BY-SA-4.0
+*/}
+
+# Active Directory Login with GPO Access Control
+
+This guide covers configuring Ghaf devices for Active Directory (AD) authentication with Group Policy Object (GPO) based access control, so that only authorized AD users can log in.
+
+## Overview
+
+Ghaf uses SSSD (System Security Services Daemon) to integrate with Active Directory. SSSD handles user authentication via Kerberos, identity lookups via LDAP, and access control via GPO evaluation.
+
+### GPO Access Control
+
+SSSD's `ad_gpo_access_control` setting determines how the client handles Windows GPO "logon rights" policies:
+
+| Setting | Behavior |
+|---------|----------|
+| `disabled` | Ignores GPO entirely — any domain user can log in |
+| `permissive` | Evaluates GPO but only logs results, does not block |
+| `enforcing` | Strictly enforces GPO — denies login if user lacks rights |
+
+When set to `enforcing`, SSSD:
+
+1. Queries AD for GPOs linked to the computer object
+2. Parses the "User Rights Assignment" section
+3. Checks for "Allow log on locally" (`SeInteractiveLogonRight`)
+4. Denies login if the user is not allowed (or is explicitly denied)
+
+## Prerequisites
+
+- An Active Directory domain controller with DNS configured
+- Domain Admin or delegated GPO management permissions
+- Test users created in AD (at least one allowed, one denied)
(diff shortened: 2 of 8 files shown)


juliuskoskela:
Rebased onto current `main` (`deb1e64d`) and force-pushed — this clears the ~176-commit drift and the merge conflict.

### Conflict resolution

Only `modules/common/services/sssd.nix` conflicted, against 86c1ebfa `fix(sssd): Use the pam login not greetd`. The resolution keeps `main`'s `login`-by-default `displayManagerService` semantics and layers this PR's multi-service strict-access map on top.

### Correction to the first version of this comment

I originally flagged the absent in-tree `ghaf.users.active-directory.domains` definition as a gap needing a home before merge. **That was wrong, and I retract it.**

a6d031af `fix(ad-users): harden AD profile setup` removed the hardcoded `ghaf-test.com` block *and in the same commit added its replacement to the documentation* — `docs/src/content/docs/ghaf/dev/guides/ad-login.mdx` gained an explicit `domains` configuration as a stated prerequisite, with a complete worked example preserving every GPO setting: `gpoAccessControl = "enforcing"`, `ad_site`, `ad_gpo_implicit_deny`, `ad_gpo_map_interactive`.

So the domain moved from a hardcoded lab value in a shared module to documented per-deployment target configuration — the right shape for a repo that should not ship any particular site's DC address. The `domains != { }` assertion is the guard for that design, not a defect: enabling `ad-users` without a domain fails fast and tells you to supply one.

I reached the wrong conclusion by reading only the `.nix` half of that commit's diff.

### What I do still think is worth addressing

**The description is stale.** Its first bullet — *"switch `mvp-user-trial` from `homed-user` to `ad-users` profile and update the DNS provider IP"* — is not in the diff (`mvp-user-trial.nix` is not among the 8 changed files), and per the above it deliberately should not be. This is what led me to misread the design, and it will mislead other reviewers the same way. Worth rewriting before review.

**The AD path has no evaluation coverage.** No in-tree configuration sets `ad-users.enable = true` — the only setting is `false`, at `modules/reference/profiles/mvp-user-trial.nix:26` — so CI never evaluates this code path. Green CI here does not indicate that the AD modules so much as evaluate. An eval-only check against a dummy domain (no lab, no hardware, no real DC) would close that, and would have surfaced the 86c1ebfa conflict automatically rather than after four months of drift.

**Enforcement is still unverified.** The problem noted in the description — a denied user still able to log in — remains open. Worth flagging for anyone picking it up: it cannot be reproduced without a target configured with a real domain *and* `gpoAccessControl = "enforcing"`, since that option defaults to `permissive`. A partial configuration authenticates fine and silently enforces nothing.

### Verification

`nix eval .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug.drvPath` succeeds on the rebased tree. That confirms the rebase does not break the default (homed-user) build; it does not exercise the AD path.

PR #1872, opened 2026-04-01
feat(nvidia): dual sd-image + two-stage initrd flash for Jetson Orin

vadika:
Minimize ghaf's Jetson Orin two-stage initrd flash module against upstream `jetpack-nixos`, and ship it alongside the legacy sd-image flash so both are selectable per target.

## Background

The earlier version of this PR reimplemented `flashInitrd` end-to-end inside ghaf, including the per-(L4T, SoC) DTS overlay paths, gadget identity constants, RCM boot script, and module list. Research during review revealed that upstream `jetpack-nixos` already owns all of that via its own `flashInitrd` and the `mkRcmBootScript` helper consumed inside `device-pkgs/`. Ghaf was duplicating, not extending — every new Jetson SoC (Thor, future Orin variants) would have meant re-adding entries to ghaf's `forceXusbPeripheralDts.overridePaths` and re-validating the gadget module list.

## What changed in this iteration

- **Upstream `tiiuae/jetpack-nixos/march-rebase`** gained a `flashScriptOverrides.postFlashInitrdCommands` hook (separate commit on that fork). The hook splices arbitrary shell into `jetpack-init` between the firmware-flash success branch and the final reboot.
- **`flake.lock`** bumped to that patched `jetpack-nixos` rev.
- **`modules/reference/hardware/jetpack/nvidia-jetson-orin/initrd-flash.nix`** shrank from 614 lines. It now:
  - sets the new `postFlashInitrdCommands` hook to the device-side Phase 2 mass-storage handoff
  - sets `additionalInitrdFlashModules = [ "usb_f_mass_storage" ]` (gated on `!onlyQSPI`)
  - overrides `initrdFlashScript` and `flashScript` with a thin wrapper built inside `overrideScope`, where Phase 1 delegates to `_jprev.initrdFlashScript` (upstream's pre-override Phase 1) and Phase 2 follows on the host side
- **Deleted from ghaf**: `forceXusbPeripheralDts`, `forceXusbPeripheralDtbo`, `spiModules`/`usbModules` selection, `manufacturer`/`product`/`serialnumber` constants, the inline `jetpack-init` script, and the inline `rcmScript` definition.

## Dual flash flow (this revision)

Both flash flows now ship side by side, selected per target:

- New option `ghaf.hardware.nvidia.orin.flashScriptOverrides.method` — `enum [ "initrd" "sdimage" ]`, default `initrd`.
- `initrd-flash.nix` / `flash-images.nix` config gated on `method == "initrd"`.
- `sdimage.nix` + `partition-template.nix` restored (from `main`), gated on `method == "sdimage"`, and **injected into the sd-image flash targets via `extendModules` — never imported by the base config**, so upstream `sd-image.nix` stays out of initrd configs (its top-level `imports` can't be `mkIf`-gated).
- Targets per SoC:
  - `-flash-script` / `-flash-qspi` — **sd-image** flow (dev/CI: `dd` to fast external USB/SD; faster than pushing over the device's USB2 gadget)
  - `-flash-initrd` / `-flash-initrd-qspi` / `-flash-images` — **two-stage initrd** flow (production: flash internal eMMC). The previous PR's `-flash-script` (= initrd) is renamed to `-flash-initrd`.
- **Verity** targets are excluded from all flash targets. This also fixes a pre-existing `fileSystems."/boot"` conflict for verity configs (`flash-images` labels it `FIRMWARE` vs `verity-image` `ESP`, both unforced) by gating `flash-images.nix` on `!verity`.
- **Phase 2 auto-reboot:** the flash-initrd's console shares the same gadget ACM serial the Phase 2 protocol uses, so `IMAGES_DONE` is unreliable under `mass_storage` load, and the `mass_storage` LUN holds the target drive open, blocking upstream's post-flash reboot. Replaced the fragile teardown with a write-idle fallback (field 7 of `/sys/block/<dev>/stat`; idle for `WRITE_IDLE_SECS=20` = done, cannot fire mid-`dd` since the counter keeps advancing) plus an **unconditional reboot** after the wait loop. Supersedes the old "host never sends `IMAGES_DONE` on failure" teardown.

## What stayed

- `flash-images.nix` — still produces `esp.img.zst` + `root.img.zst` (+ `flash-manifest.json`) and the `ghaf-ensure-system-profile` first-boot service
- `flashScriptOverrides.onlyQSPI` mode and option declaration
- USB-only transport (no NFS+IPv6 setup needed on host or CI — important for portability across container, workstation, and CI/CD flash contexts)
- `meta.platforms = [ "x86_64-linux" ]` on the host script

## Testing instructions

All Nix commands run in the dev container (x86_64 host, aarch64 cross via the remote builder/cache):
`docker compose exec --user dev -w /workspace/ghaf nixos-dev bash -lc "<cmd>"`.

Eval + build (no hardware):

```
nix flake show --all-systems --accept-flake-config          # every Orin variant listed
nix build .#nvidia-jetson-orin-agx-debug-from-x86_64-flash-initrd
nix build .#nvidia-jetson-orin-agx-debug-from-x86_64-flash-script
```

Flash (AGX Orin in RCM; connect the flash USB-C **directly to the host**, not through a hub — hub chains fail RCM config with `error -110`):

```
# production — two-stage initrd → internal eMMC (auto-reboots on success)
sudo ./result/bin/initrd-flash-ghaf-host --target=emmc

# dev/CI — sd-image → external USB/SD

# firmware-only variants: build -flash-initrd-qspi / -flash-qspi, run the same bins
```

`--target` accepts `emmc` | `nvme` | `usb`. Watch progress on the board's console UART.

## Test plan

- [x] `nix flake show --all-systems --accept-flake-config` clean — every Orin variant listed
- [x] `nix build` of all five flash targets (`-flash-script`, `-flash-qspi`, `-flash-initrd`, `-flash-initrd-qspi`, `-flash-images`) succeeds
- [x] sd-image and initrd `-flash-script`/`-flash-initrd` resolve to distinct derivations
- [x] verity target `toplevel` + `-ghafImage` eval cleanly; verity emits no flash targets
- [x] `nix build .#checks.x86_64-linux.pre-commit` clean; `reuse lint` compliant
- [x] **AGX Orin hardware:** two-stage initrd full flash → eMMC → auto-reboots into NixOS userland
- [x] **AGX Orin hardware:** sd-image full flash → boots to login
- [x] **AGX Orin hardware:** `onlyQSPI` firmware-only, both flows
- [ ] AGX64 / Orin NX hardware — pending hardware availability

## Companion change

The hook this PR depends on: <https://github.com/tiiuae/jetpack-nixos/commit/f440892dad860e531513deddc47ec2147ea0f656> on branch `march-rebase`.


leivos-unikie:
Tested flashing Orin AGX. Flash output seems ok and produces working ghaf.

However, the two stage output of nix build .#nvidia-jetson-orin-agx-debug-from-x86_64
esp.img.zst  root.img.zst
requires updated packages/pkgs-by-name/flash-script/flash.sh for test automation to manage flashing of two-fold images on USB SSDs.

**Before this can be merged it should be verified that test-automation is able to handle two stage images.**

[PR1872_2-stage_initrd_flash_logs.txt](https://github.com/user-attachments/files/26567625/PR1872_2-stage_initrd_flash_logs.txt)

henrirosten:
I support @leivos-unikie’s comment above that this still needs discussion on how do we integrate with CI test automation.

Currently, the hardware-test automation assumes that a target exposes:
  - one canonical flashable image artifact, for example a single image file
  - one automation-facing flasher entrypoint, currently the ghaf [flash-script](https://github.com/tiiuae/ghaf/blob/ea7fb142a2b10a007f34375cac7252f72f0e7551/packages/pkgs-by-name/flash-script/package.nix#L14)
  
This PR changes Orin from that shape to a multi-file flash set such as `esp.img.zst` + `root.img.zst`, with also requiring a target-specific flasher.

That's why for this PR, the pre-merge pipeline now [fails for all Orin variants](https://ci-prod.vedenemo.dev/job/ghaf-pre-merge/1835/), as it tries to use the [flash-script and assumes one image artifact](https://ci-prod.vedenemo.dev/job/ghaf-hw-test/21611/console).

Before merge, we need the PR to make a few things explicit:
1. which output files external automation is expected to use for flashing
2. which flasher package the automation is expected to run to flash those files
3. what UEFI signing applies to in this model: specifically, which artifact is the UEFI-signing target, and how Secure Boot is expected to work with the new flashing flow

The main point is that external CI should not have to guess which files belong together or which flashing path it should use for various Ghaf targets.

If Ghaf can emit a small metadata file listing the required artifacts and the flasher to use, that would make CI integration much easier.

leivos-unikie:
Tested on Orin AGX

1.
- Built full two stage flash-script
- Flashed, flashing finished with:
```
  ...
  [device] IMAGES_DONE
  [device] IMAGES_DONE
  [device] IMAGES_DONE
  [device] IMAGES_DONE
ERROR: Timed out waiting for: Flashing platform firmware successful
Flash complete. Device is rebooting.
```
- After reboot ghaf was running ok

2.
- Built qpsi-script
- Flashed, flashing finished with:
```
Finished flashing device
Flashing platform firmware successful.
Device reported flashing succeeded. Device should be rebooting.
QSPI flash complete. Device is rebooting.
```
- Booted latest (06.05.2026) nightly orin agx ghaf image from USB SSD
- Boot ok

brianmcgillion (approved):

vadika:
Addressing @henrirosten / @leivos-unikie's CI-integration questions in the last push (62e53aec):

**1. Output files** — `nix build .#packages.x86_64-linux.<target>-flash-images` (new attribute) produces a single directory containing:
- `flash-manifest.json` — versioned (`schema_version: 1`) machine-readable description of the artifact set.
- `esp.img.zst` — FAT32 ESP (256 MiB, label `FIRMWARE`).
- `root.img.zst` — ext4 rootfs (auto-sized, label `NIXOS_ROOT`).

For the QSPI-only variant, the same `-flash-images` attribute exists but contains only the manifest with `transport: "qspi-only"` and an empty `artifacts` array — so automation can detect the variant by reading the manifest rather than special-casing target names.

**2. Flasher entrypoint** — `nix build .#packages.x86_64-linux.<target>-flash-script` (for the full 2-stage flow) or `<target>-flash-qspi` (firmware-only). The manifest's `flasher.entrypoint` field (`bin/initrd-flash-<host>`) and `flasher.system` (`x86_64-linux`) confirm the binding. The convention is that the same `<target>` attribute prefix groups `-flash-images` with its flasher — no cross-naming.

The flasher now accepts `--target=emmc|nvme|usb` (default `emmc`) to select the destination drive on the device. The manifest's `flasher.target_drives` lists the supported values so automation can drive the flag without hardcoding the set:

```json
{
  "schema_version": 1,
  "host_name": "ghaf-host",
  "transport": "initrd-mass-storage",
  "flasher": {
    "entrypoint": "bin/initrd-flash-ghaf-host",
    "system": "x86_64-linux",
    "target_drives": ["emmc", "nvme", "usb"]
  },
  "artifacts": [
    { "name": "esp.img.zst",  "role": "esp",  "compression": "zstd",
      "partition": { "gpt_type": "EF00", "label": "FIRMWARE",   "size_mib": 256 } },
    { "name": "root.img.zst", "role": "root", "compression": "zstd",
      "partition": { "gpt_type": "8300", "label": "NIXOS_ROOT", "size_mib": "auto" } }
  ],
  "signing": {
    "firmware_pkc_signed": false,
    "esp_secureboot": false,
    "root_dmverity": false
  }
}
```

**3. UEFI / Secure Boot** — `signing.firmware_pkc_signed: true` iff `hardware.nvidia-jetpack.firmware.secureBoot.pkcFile != null`. When that is set, upstream `jetpack-nixos` signs the platform-firmware blob (boot.img + bootloader variants) with the PKC before it is shipped into the initrd; the flasher just consumes the already-signed blob. The ESP image emitted by ghaf is **not** UEFI-Secure-Boot–signed today (`esp_secureboot: false` is a hard-coded flag in this schema reserved for when that work lands), and the root image is **not** dm-verity-protected (`root_dmverity: false` likewise). The manifest reports the current posture so consumers do not need to re-sign anything; if any of those flags later change to `true`, automation can react.

Schema bump rules: any breaking change to field names / removal increments `schema_version` so consumers can fail loudly. Additive fields keep the version stable. The schema itself is documented in the comment block at the top of `modules/reference/hardware/jetpack/nvidia-jetson-orin/flash-images.nix`.

Two follow-ups already filed:
- jetpack-nixos PR #20 makes the host-side initrd-flash serial timeouts configurable (the hardcoded 4-minute serial-enumerate wait was the root cause of "Timed out waiting for: Flashing platform firmware successful" in @leivos-unikie's test 1). I'll bump `flake.lock` here and set the longer timeouts once that lands.
- The `-flash-images` attribute is intentionally provided for every cross variant including QSPI so automation can use one uniform discovery path.

If the manifest schema is missing anything you need to drive the test automation, please call it out before merge so we can fix it here rather than chase a schema bump.

vadika:
Follow-up for CI workflow (4ba9c3bd): the flasher now accepts `--flash-images=DIR` so the script binary and the artifact set can be built/shipped independently.

```
# Jenkins (build host):
nix build .#packages.x86_64-linux.<target>-flash-images
scp -r result/ test-agent:/tmp/flash-images-<sha>/

# Test agent:
nix build .#packages.x86_64-linux.<target>-flash-script  # cached via substituter
result/bin/initrd-flash-<target> --target=usb --flash-images=/tmp/flash-images-<sha>
```

Validation happens before any partition is touched:
- `schema_version` must match what the script was built against (currently `1`).
- `host_name` in the manifest must match the script's target (refuses to flash mismatched images).
- `transport` must match the flasher variant (`initrd-mass-storage` for `-flash-script`, `qspi-only` for `-flash-qspi`).
- Required artifacts (`esp.img.zst`, `root.img.zst`) must be present in the directory.

leivos-unikie:
I checked that the flash script is trying to flash also AGX QSPI or at least still expects USB connection to the device in addition to flashing ghaf images on USB SSD connected to the flashing PC.

When I start the flash script with just USB SSD connected to laptop it fails 
```
sudo ./flash-script/bin/initrd-flash-ghaf-host --target=usb --flash-images=flash-images
[sudo] password for leivos: 
Flash target drive: usb
Flash images:       flash-images
###############################################################################
# L4T BSP Information:
# R36 , REVISION: 5.0
###############################################################################
ECID is 
Error: probing the target board failed.
       Make sure the target board is connected through 
       USB port and is in recovery mode.
```

but when I start it with also USB-C cable connected to Orin AGX flash port and Orin being set to recovery mode flash starts.
[two-stage_flash_to_usb_ssd.txt](https://github.com/user-attachments/files/28830014/two-stage_flash_to_usb_ssd.txt)


This is not the way automated setups can handle it. Dependency for USB connection to the target should be ripped off from the flash script.

leivos-unikie:
Tried booting AGX from USB SSD after that hybrid flash (qspi + usb ssd)
It failed at
`Timed out waiting for device /dev/disk/by-label/NIXOS_SD.`

[usb_ssd_boot_after_two_stage_flash.txt](https://github.com/user-attachments/files/28835676/usb_ssd_boot_after_two_stage_flash.txt)

milva-unikie:
@vadika please update the PR description with latest Testing Instructions. They seem to be out of date and include a lot of unnecessary steps. It is impossible to find the correct information among all these comments.

In particular:
1. How to flash the internal memory of Orin.
2. How to flash Ghaf image to an external SSD (required for automated tests).

----
**Automated test status**

Currently `jenkins-pre-merge` tests are failing for Orins with error
```
Multi-image hardware tests are not implemented for nvidia-jetson-orin-agx-debug; found 2 images
```
----

I also have one question. I built `nvidia-jetson-orin-agx-debug-from-x86_64-flash-initrd`, but I am not able to use it for flashing because my SSD is connected as `/dev/sbd`. Is there really now way to change the device at the moment?
```
[]$ sudo ./result/bin/initrd-flash-ghaf-host --help
Usage: initrd-flash-ghaf-host [--target=emmc|nvme|usb] [--flash-images=DIR]

  --target         Destination drive for the NixOS rootfs (default: emmc).
                   'emmc' targets /dev/mmcblk0 on the device.
                   'nvme' targets /dev/nvme0n1 (requires an M.2 NVMe present).
                   'usb'  targets /dev/sda    (requires a USB drive attached
                                                to a host-mode USB port).
                   The chosen target must be detected on the device; otherwise
                   the script aborts before any partition is touched.
```

PR #1875, opened 2026-04-01
Fix intel-laptop target

nesteroff:
## Description of Changes

Fix multiple issues in the Intel laptop target.

### Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Improvement / Refactor

## Related Issues / Tickets

https://jira.tii.ae/browse/SSRCSP-8182
https://jira.tii.ae/browse/SSRCSP-8189
https://jira.tii.ae/browse/SSRCSP-8019
https://jira.tii.ae/browse/SSRCSP-8176
https://jira.tii.ae/browse/SSRCSP-8059

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:
Test procedures can be found in the tickets addressed by this PR:
https://jira.tii.ae/browse/SSRCSP-8182
https://jira.tii.ae/browse/SSRCSP-8189
https://jira.tii.ae/browse/SSRCSP-8019
https://jira.tii.ae/browse/SSRCSP-8176
https://jira.tii.ae/browse/SSRCSP-8059


modules/common/services/power.nix
```diff
@@ -250,6 +250,42 @@ let
...
+      matched=0
+
+      while IFS= read -r enabled_model; do
+        if [ "$enabled_model" = "$board_vendor $board_name" ]; then
+          matched=1
+          break
+        fi
+      done <<'EOF'
+      ${lib.concatStringsSep "\n" cfg.suspend.s2idleModels}
+      EOF
```
kajusnau: I think this would be easier to read and maintain with a simple array assignment before any processing

```suggestion
      s2idle_models=(${lib.concatMapStringsSep " " (lib.replaceStrings [ " " ] [ "-" ] )cfg.suspend.s2idleModels)})
      for model in "''${s2idle_models[@]}"; do
        if [ "$model" = "$board_vendor-$board_name" ]; then
          matched=1
          break
        fi
      done
```

modules/common/services/power.nix
```diff
@@ -250,6 +250,42 @@ let
...
   };
 
+  host-set-mem-sleep = pkgs.writeShellApplication {
+    name = "host-set-mem-sleep";
+    runtimeInputs = [
+      pkgs.coreutils
+      pkgs.gnugrep
+    ];
+    text = ''
+      set -eu
```
kajusnau: These options are already [set by default](https://ryantm.github.io/nixpkgs/builders/trivial-builders/#trivial-builder-writeShellApplication), along with `pipefail`.

If `pipefail` is causing problems, it can be excluded from `writeShellApplication` `bashOptions` attr:
```nix
  host-set-mem-sleep = pkgs.writeShellApplication {
    name = "host-set-mem-sleep";
    bashOptions = [ "errexit" "nounset" ];
    runtimeInputs = [
      pkgs.coreutils
      pkgs.gnugrep
    ];
    ...
```

modules/common/services/power.nix
```diff
@@ -250,6 +250,42 @@ let
...
+      EOF
+
+      if [ "$matched" -ne 1 ]; then
+        exit 0
+      fi
+
+      if ! grep -qw "s2idle" /sys/power/mem_sleep; then
+        echo "s2idle is not supported on this system"
+        exit 0
+      fi
```
kajusnau, line 274: Shouldn't this be an error?
At least from a config perspective, we'd know that the selected model shouldn't be part of `cfg.suspend.s2idleModels`

Also I recommend clearly printing the offending model (exact format can be adjusted)

```suggestion
      if ! grep -qw "s2idle" /sys/power/mem_sleep; then
        echo "s2idle is not supported on this system [$board_vendor $board_name]"
        exit 1
      fi
```

modules/common/services/power.nix
```diff
@@ -601,6 +658,25 @@ in
...
+                DefaultDependencies = false;
+                ConditionPathExists = [
+                  "/sys/power/mem_sleep"
+                  "/sys/class/dmi/id/board_vendor"
+                  "/sys/class/dmi/id/board_name"
+                ];
+              };
+              serviceConfig = {
+                Type = "oneshot";
+                ExecStart = getExe host-set-mem-sleep;
```
kajusnau: Probably works as-is, since the service runs as `root`, but we may want to ensure proper privileges with `+` prefix ([ref](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Command%20lines)), since we'll be writing to `/sys/power/mem_sleep`

```suggestion
                ExecStart = "+''${getExe host-set-mem-sleep}";
```

modules/common/services/power.nix
```diff
@@ -601,6 +658,25 @@ in
         };
 
         services = mkMerge [
+          (optionalAttrs (cfg.suspend.mode == "auto" && cfg.suspend.s2idleModels != [ ]) {
+            set-mem-sleep = {
+              description = "Set s2idle mem_sleep mode for selected models";
```
kajusnau: Nitpicky suggestion to change the description to something that would make more sense in a systemd unit context for strings like `Started <Description>` and `Failed to start <Description>`

```suggestion
              description = "Automatic s2idle mem_sleep Mode Selection";
```

modules/common/services/power.nix
```diff
@@ -250,6 +250,42 @@ let
...
+
+      if ! grep -qw "s2idle" /sys/power/mem_sleep; then
+        echo "s2idle is not supported on this system"
+        exit 0
+      fi
+
+      echo "Enabling s2idle mode"
+      printf '%s' "s2idle" > /sys/power/mem_sleep
+    '';
+  };
```
kajusnau, line 279: Summary of above proposed changes all in one spot, along with further simplification removing the need for `matching` check (not tested so please double check 🙂 ):

```suggestion
  host-set-mem-sleep = pkgs.writeShellApplication {
    name = "host-set-mem-sleep";
    runtimeInputs = [
      pkgs.coreutils
      pkgs.gnugrep
    ];
    text = ''
      board_vendor="$(cat /sys/class/dmi/id/board_vendor)"
      board_name="$(cat /sys/class/dmi/id/board_name)"
      s2idle_models=(${lib.concatMapStringsSep " " (lib.replaceStrings [ " " ] [ "-" ] )cfg.suspend.s2idleModels)})

      if ! printf '%s\n' "''${s2idle_models[@]}" | grep -iqxF "$board_vendor-$board_name"; then
        exit 0
      fi

      if ! grep -qw "s2idle" /sys/power/mem_sleep; then
        echo "s2idle is not supported on this system [$board_vendor $board_name]"
        exit 1
      fi

      echo "Enabling s2idle mode"
      printf '%s' "s2idle" > /sys/power/mem_sleep
    '';
  };
```

kajusnau (changes requested):
Clever fixes and workaround!
Some improvement suggestions

modules/reference/hardware/intel-laptop/extra-config-host.nix
```diff
@@ -0,0 +1,12 @@
...
+# SPDX-License-Identifier: Apache-2.0
+{
+  lib,
+  ...
+}:
+{
+  ghaf.services.power-manager.suspend = {
+    mode = lib.mkDefault "auto";
+    s2idleModels = [ "System76 Darter Pro" ];
+  };
```
kajusnau, line 11: Since the `set-mem-sleep` service is very lightweight, IMO this could just be a global default, and then we could remove the s2idle setting from all darp11 targets in `targets/laptop/flake-module.nix`

modules/common/services/power.nix
nesteroff: Thanks! I like the cleanup suggestions. The only thing I'm not sure about is the direct array assignment, since DMI data can contain special characters that could break it, so I  kept the heredoc version and simplified it a bit.

modules/reference/hardware/intel-laptop/extra-config-host.nix
nesteroff: Sure, but maybe we should do it in a follow-up after this change is tested thoroughly to make sure it works fine.

modules/common/services/power.nix
kajusnau: Nice, I wasn't aware of the possible special chars, makes sense to keep the previous approach.

milva-unikie:
**Tested**
- `intel-laptop-debug` on Darter Pro and Lenovo X1
- `intel-laptop-low-mem-debug` on Dell 7330

All linked bugs have been fixed.

kajusnau (approved):

brianmcgillion (approved):

brianmcgillion:
needs a rebase

nesteroff:
> needs a rebase

Rebased onto the latest main. It should be up to date now.

PR #1886, opened 2026-04-07
Apply firewall dynamic rules through policy update using iptables

gngram:
## Description of Changes
Apply firewall dynamic rules through policy update using iptables
- Current nftable rules applied through policy update breaks the existing firewall chain arc.

### Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

1. Boot the system and connect to internet, wait for 5 minutes to get the policy update.
2. Try to open  'https://www.instant-gaming.com/' in Google Chrome, you should not be able to access it.

### DIFF
--- a/modules/common/firewall/firewall.nix
+++ b/modules/common/firewall/firewall.nix
@@ -24,6 +24,7 @@ let
   blackListName = "BLACKLIST";
   blacklistMarkNum = "8";
   cfg = config.ghaf.firewall;
+  rulePath = "/etc/firewall/rules/iptables.rules";
 
   blacklistRuleType = types.listOf (
     types.submodule {
@@ -520,21 +521,37 @@ in
       ];
 
     ghaf.givc.policyClient.policies = mkIf cfg.updater.enable {
-      firewall-rules =
-        let
-          rulePath = "/etc/firewall/rules/fw.nft";
-        in
-        {
-          dest = rulePath;
-          updater = {
-            url = "https://raw.githubusercontent.com/tiiuae/ghaf-policies/deploy/vm-policies/firewall-rules/fw.nft";
-            poll_interval_secs = 300;
-          };
-
-          script = pkgs.writeShellScript "apply-nftables" ''
-            ${pkgs.nftables}/bin/nft -f ${rulePath}
-          '';
+      firewall-rules = {
+        dest = rulePath;
+        updater = {
+          url = "https://raw.githubusercontent.com/tiiuae/ghaf-policies/deploy/vm-policies/firewall-rules/iptables.rules";
+          poll_interval_secs = 300;
         };
+      };
+    };
+    systemd = mkIf cfg.updater.enable {
+      paths.apply-dynamic-firewall-rules = {
+        description = "Watch dynamic firewall rules file for changes";
+        wantedBy = [ "multi-user.target" ];
+        pathConfig = {
+          PathModified = rulePath;
+        };
+      };
+
+      services.apply-dynamic-firewall-rules = {
+        description = "Apply dynamic firewall rules";
+        wantedBy = [ "multi-user.target" ];
+        after = [ "firewall.service" ];
+        serviceConfig.Type = "oneshot";
+        script = ''
+          if [ -f "${rulePath}" ]; then
+            echo "Dynamic firewall rules found at ${rulePath}, applying..."
+            ${pkgs.iptables}/bin/iptables-restore --noflush < "${rulePath}" || echo "ERROR: Failed to apply dynamic firewall rules."
+          else
+            echo "No dynamic firewall rules found at ${rulePath}, skipping."
+          fi
+        '';
+      };
     };
   };
 }


modules/common/firewall/firewall.nix
```diff
@@ -493,17 +493,17 @@ in
...
           updater = {
-            url = "https://raw.githubusercontent.com/tiiuae/ghaf-policies/deploy/vm-policies/firewall-rules/fw.nft";
+            url = "https://raw.githubusercontent.com/tiiuae/ghaf-policies/deploy/vm-policies/firewall-rules/iptables.rules";
             poll_interval_secs = 300;
           };
 
-          script = pkgs.writeShellScript "apply-nftables" ''
-            ${pkgs.nftables}/bin/nft -f ${rulePath}
+          script = pkgs.writeShellScript "apply-iptables" ''
+            ${pkgs.coreutils}/bin/cat ${rulePath} | ${pkgs.iptables}/bin/iptables-restore --noflush
```
enesoztrk:
```suggestion
           ${pkgs.iptables}/bin/iptables-restore --noflush < ${rulePath}
```

enesoztrk (changes requested):
Should we introduce dedicated `ghaf-fw-policy-*` chains (`input`, `forward`, `output`, `pre-nat,` `post-nat`, etc.) that are flushed and repopulated by the policy updater, rather than having it restore directly into the existing firewall chains — keeping the base firewall rules managed by `firewall.nix` untouched while allowing live rule updates via` iptables-restore --noflush`?

modules/common/firewall/firewall.nix
gngram: Done!

gngram:
> Should we introduce dedicated `ghaf-fw-policy-*` chains (`input`, `forward`, `output`, `pre-nat,` `post-nat`, etc.) that are flushed and repopulated by the policy updater, rather than having it restore directly into the existing firewall chains — keeping the base firewall rules managed by `firewall.nix` untouched while allowing live rule updates via` iptables-restore --noflush`?

Indeed it's a good idea to have a separate chain for dynamic rules. I think we should create a task to adapt nftable based chaining along with introduction of a new chain for dynamic rules.

vadika:
> Should we introduce dedicated `ghaf-fw-policy-*` chains (`input`, `forward`, `output`, `pre-nat,` `post-nat`, etc.) that are flushed and repopulated by the policy updater, rather than having it restore directly into the existing firewall chains — keeping the base firewall rules managed by `firewall.nix` untouched while allowing live rule updates via` iptables-restore --noflush`?

I'm  for landing the dedicated ghaf-fw-policy-* chains. Two reasons:
 1) --noflush against the base chains gives unbounded rule growth on every poll unless the policy file is hand-crafted to delete its own prior rules, which is fragile;
2) without isolated chains there's no trust boundary between firewall.nix and the remote policy source.

enesoztrk (approved):
IMO, we should do one round of testing

milva-unikie:

Policy is enforced only on the first boot.

**Steps to reproduce**
1. Boot & connect to Wi-Fi
2. Wait until www.instant-gaming.com/ becomes unreachable in chrome-vm
```
[ghaf@chrome-vm:~]$ ping www.instant-gaming.com
PING www.instant-gaming.com (104.18.4.124) 56(84) bytes of data.
From chrome-vm (192.168.100.102) icmp_seq=1 Destination Port Unreachable
ping: sendmsg: Operation not permitted
```
3. Reboot
4. Ping www.instant-gaming.com/ in chrome-vm, his time it works and waiting for 5 minutes does not change that
```
[ghaf@chrome-vm:~]$ ping www.instant-gaming.com
PING www.instant-gaming.com (104.18.4.124) 56(84) bytes of data.
64 bytes from 104.18.4.124: icmp_seq=1 ttl=57 time=11.2 ms
```

gngram:
@milva-unikie, please verify it again.

milva-unikie:
> @milva-unikie, please verify it again.

Tested again with a new image, the policy is still not working on the second boot.

gngram:
Instead of appending dynamic rules in the chain I created a separate service to apply it. After reboot, I can not access the website. Also I searched the website in Google search, and when I open the site link from the search result, some time it loads only text from landing page but links on the page are blocked. Probably it is either loading from cache or from CDN but actual content is not accessible.

milva-unikie:
**Tested on Darter Pro**
- Issue has been fixed, website is not accessible on the second boot

enesoztrk (approved):

brianmcgillion (approved):

PR #1887, opened 2026-04-09
fix(power): keep display on during suspension; virtiofs: enable cache

kajusnau:
## Virtiofs changes:

Explicitly change the virtiofs cache setting from the default `auto` to `always` for `/nix/store` shares in Ghaf guests.

### Concerns:
Aggressive caching may introduce coherency issues on mutable shares, especially when the filesystem does **not** have exclusive access to the directory.
The Nix store, however, is immutable by design - paths are content-addressed and never modified in-place. This makes the coherency concern a non-issue, as there's nothing to go stale.

Reference manual for these options can be found here
https://gitlab.com/virtio-fs/virtiofsd
https://gitlab.com/virtio-fs/virtiofsd/-/blob/main/src/passthrough/mod.rs#L95

### Impact:
Expected benefits:
- Improved IO performance for repeated access to commonly shared libs - especially true for GUI libs in gui-vm
- Reduced virtiofsd CPU overhead - by default, the FUSE protocol uses close-to-open consistency. This means that any cached contents of the file are invalidated the next time that file is opened.
   With `cache` set to `always`, repeated reads are served from the guest memory directly, rather than through the FUSE client.
   
What this will **not** do:
- Improve guest cold boot times
- Speed up **first** time access of commonly shared libs
- Bring up general system responsiveness to the level of `storeDisk` - but it should be noticeably improved overall

## Suspension changes:

Dirty and hopefully temporary "fix" for recent suspension issues
It seems that i915 driver timeout occurs only if the built-in display is disabled before suspension.
This patch changes suspension behavior - suspending now does not turn off the display, but sets the backlight to 0% brightness via `brightnessctl s 0`

This behavior _may_ be explained by the fact that `wlopm --off` triggers a full DRM atomic disable that tears down i915 power wells, while `brightnessctl s 0` only writes to a PWM register that i915 never sees.

Tested manually - 15ish suspension cycles - no i915 driver crash.
Note: I did manual tests on a gen11 X1 with an OLED panel. All of our other targets use typical LCD panels, which may behave differently - for example, it may be so that "0% backlight brightness" does not result in a fully black "off" display, but rather a very dimly lit one.

### Type of Change
- [x] New Feature
- [x] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

#### Virtiofs changes:
1. Ensure https://jira.tii.ae/browse/SSRCSP-8196 is **NOT** re-introduced
2. (Optional) Test and verify perf improvement similar to https://github.com/tiiuae/ghaf/pull/1817#issuecomment-4038757738
3. Test memory usage in affected VMs - all App VMs, gui-vm, net-vm, and audio-vm.
   Memory usage may increase overtime as VMs continue to run and `/nix/store` paths are being cached.
   Ideally, we should test the memory footprint of net-vm (sensitive to OOM issues) and gui-vm (more lib lookups than other VMs):
   1. Boot into Ghaf, wait for system to settle and inspect memory usage in net-vm and gui-vm
   2. Use the system as usual - open some apps, browse the net, etc. - for 2-3 hours
   3. Inspect memory usage in net-vm and gui-vm, verify no major OOM issues
   
#### Suspension changes:
1. Try to suspend any laptop device by either closing the lid or suspending manually
2. Verify device display is practically off - black screen, indistinguishable from being actually powered off
3. Verify device resumes from suspension without a graphics driver timeout
4. Repeat steps 1 and 3 - at least 10 times, letting the device sleep for at least 1 minute each time
5. Verify no issues even after multiple suspension attempts
6. Verify battery usage is not much higher during suspension compared to mainline.

### DIFF
--- a/modules/common/services/power.nix
+++ b/modules/common/services/power.nix
@@ -131,7 +131,7 @@ let
               ${givc-cli} start service --vm "$vm_name" suspend.target &
               # Wait until suspend is active
               ${getExe pkgs.wait-for-unit} ${config.ghaf.networking.hosts.admin-vm.ipv4} 9001 \
-              "$vm_name" systemd-suspend.service 5 \
+              "$vm_name" systemd-suspend.service 10 \
               activating start
               ;;
             resume)
@@ -492,8 +492,7 @@ in
       # Shutdown displays early before suspend
       powerManagement = {
         powerDownCommands = lib.mkBefore ''
-          ${getExe ghaf-powercontrol} turn-on-displays '*'
-          ${getExe ghaf-powercontrol} turn-off-displays '*'
+          ${getExe ghaf-powercontrol} fake-turn-off-displays '*'
         '';
       };
 
@@ -536,11 +535,12 @@ in
           serviceConfig = {
             Type = "oneshot";
             RemainAfterExit = true;
-            ExecStop = "${getExe ghaf-powercontrol} turn-on-displays '*'";
+            ExecStop = "${getExe ghaf-powercontrol} fake-turn-on-displays '*'";
           };
         };
       };
 
+      services.upower.ignoreLid = true;
       # Logind configuration for desktop
       services.logind.settings.Login =
         let
@@ -549,7 +549,6 @@ in
         mkDefault {
           HandleLidSwitch = lidEvent;
           HandleLidSwitchDocked = "ignore";
-          HandleLidSwitchExternalPower = lidEvent;
           # Below keys are usually handled by host anyway
           HandleSuspendKey = "ignore";
           HandleHibernateKey = "ignore";
@@ -562,6 +561,7 @@ in
 
     # Host power management
     (mkIf cfg.host.enable {
+      services.upower.ignoreLid = true;
       # Host still handles power buttons in most situations
       services.logind.settings.Login = {
         HandleLidSwitch = mkDefault "ignore";
--- a/modules/microvm/sysvms/appvm-base.nix
+++ b/modules/microvm/sysvms/appvm-base.nix
@@ -343,6 +343,7 @@ in
             source = "/nix/store";
             mountPoint = "/nix/.ro-store";
             proto = "virtiofs";
+            cache = "always";
           }
         ];
 
--- a/modules/microvm/sysvms/audiovm-base.nix
+++ b/modules/microvm/sysvms/audiovm-base.nix
@@ -218,6 +218,7 @@ in
         source = "/nix/store";
         mountPoint = "/nix/.ro-store";
         proto = "virtiofs";
+        cache = "always";
       }
     ];
 
--- a/modules/microvm/sysvms/guivm-base.nix
+++ b/modules/microvm/sysvms/guivm-base.nix
@@ -382,6 +382,7 @@ in
         source = "/nix/store";
         mountPoint = "/nix/.ro-store";
         proto = "virtiofs";
+        cache = "always";
       }
     ];
 
--- a/modules/microvm/sysvms/netvm-base.nix
+++ b/modules/microvm/sysvms/netvm-base.nix
@@ -230,6 +230,7 @@ in
         source = "/nix/store";
         mountPoint = "/nix/.ro-store";
         proto = "virtiofs";
+        cache = "always";
       }
     ];
 
--- a/packages/ghaf-powercontrol/package.nix
+++ b/packages/ghaf-powercontrol/package.nix
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 {
   writeShellApplication,
+  brightnessctl,
   lib,
   systemd,
   wlopm,
@@ -28,6 +29,7 @@ writeShellApplication {
     libnotify
     toybox
     jq
+    brightnessctl
   ]
   ++ (lib.optional useGivc givc-cli);
 
@@ -77,12 +79,21 @@ writeShellApplication {
         return 1
       fi
 
-      if [ "$action" != "on" ] && [ "$action" != "off" ]; then
-        echo "Error: First argument must be 'on' or 'off'"
-        return 1
-      fi
-
-      local cmd="wlopm --$action '$display_name'"
+      case "$action" in
+        on|off)
+          local cmd="wlopm --$action '$display_name'"
+          ;;
+        fake-off)
+          local cmd="brightnessctl -c backlight -s s 0"
+          ;;
+        fake-on)
+          local cmd="brightnessctl -c backlight -r || brightnessctl -c backlight s 100%"
+          ;;
+        *)
+          echo "Error: First argument must be one of 'on, off, fake-on, fake-off'"
+          exit 1
+          ;;
+      esac
 
       echo "Attempting to turn displays '$display_name' $action..."
 
@@ -147,6 +158,12 @@ writeShellApplication {
         ;;
       turn-on-displays)
         try_toggle_displays on "$2"
+        ;;
+      fake-turn-off-displays)
+        try_toggle_displays fake-off "$2"
+        ;;
+      fake-turn-on-displays)
+        try_toggle_displays fake-on "$2"
         ;;
       help|--help)
         help_msg


vunnyso (approved):

kajusnau:
Combined with #1888

leivos-unikie:
Tested on system76-darp11-b-debug-installer (no encryption or sb enabled)

1.
Tried to reproduce https://jira.tii.ae/browse/SSRCSP-8196
- Second file added to Unsafe chrome-vm share can be seen in COSMIC Files after user either
    - clicks File -> Reload folder (F5)
    - switches between directories
    - restarts COSMIC Files
- I think this is the way it used to work before f528b27ef8fd17ec76c68822d9fd09239b1ed3ba mentioned in the bug ticket.
- But it could be argued if it is acceptable that updating COSMIC Files view requires F5. In local vm directories (non shared) the added file gets immediately updated in the COSMIC Files directory view.

2.
Ran "Measure time to launch COSMIC Settings" few times
(the first data point is from older tests)
image

For comparison "Measure time to launch COSMIC Settings" results on system76-darp11-b-debug from ghaf-nightly-perftest pipeline
image

This PR seems to introduce slight decrease in launch time also on darter.

leivos-unikie:
More results with system76-darp11-b-debug-installer...

3.
Monitored memory consumption over 1h (ignored swap here). First 2 data points are before opening any apps. Then opened all apps/tools via Applications. Only memory usage in gui-vm and ghaf-host is significantly affected by running apps but no OOM or other problems with memory found. Initial relative memory consumption of zathura-vm is quite high ( >90%). Probably a bit too little memory assigned for that in the first place.
image

leivos-unikie:
With less than 10 repeats of suspension by power menu / closing the lid I got darter-pro to a stage where the display does not anymore wake up after opening lid and pushing all buttons. Ssh works
[gui_journalctl.log](https://github.com/user-attachments/files/26627468/gui_journalctl.log)

leivos-unikie:
Battery usage during suspension on lenovo-x1 - OK

power_Automatic suspension

power_GUI Suspend and wake up

leivos-unikie:
"This patch changes suspension behavior - suspending now does not turn off the display, but sets the backlight to 0% brightness via brightnessctl s 0"

Is there any danger that if the device runs out of power while suspended that the brightness value would be 0 at the next boot?

kajusnau:
> With less than 10 repeats of suspension by power menu / closing the lid I got darter-pro to a stage where the display does not anymore wake up after opening lid and pushing all buttons. Ssh works [gui_journalctl.log](https://github.com/user-attachments/files/26627468/gui_journalctl.log)

Interesting.. I see no i915 driver crash in the logs, so that's a good sign. It may be so that the resume action service did not run for some reason. Will investigate next week on darp11


> Is there any danger that if the device runs out of power while suspended that the brightness value would be 0 at the next boot?

It's technically possible only if the device transitions from suspension (0% brightness) straight to shutdown, at which point brightness value is stored for the next boot. However, transitioning from suspension to shutdown without waking up in between is unlikely in real use.
It's also possible that the device simply powers off when the battery runs out - in this case I think the proper shutdown procedure will not occur, therefore the brightness value will not be stored for the next boot. This means that the brightness will be reset to approx 5%. This behavior should be the same, no matter if suspension had occurred before or not.

kajusnau:
Update:

Could not reproduce exact behavior described above with `system76-darp11-b-debug` target

From the logs provided by Samuli, I could see one major issue - `resume-actions.service` in gui-vm failed to run properly, therefore the backlight brightness was never restored, though the system was up and running

I made the following changes:
- Adjusted suspension backlight action to be more precise, targeting only `backlight` class devices
- Added a failsafe to set backlight brightness to 100% in case restoration from saved state fails
- Adjusted wait time for `systemd-suspend` units to be active in guests from 5 to 10 seconds
- Set `services.upower.ignoreLid` to `false` explicitly, which _may_ help reduce unintended suspension/wakeup events from being triggered when the lid is closed/opened

Re-testing is needed. If the same issue occurs on darp11, let's evaluate the frequency and verify it does **not** appear on other targets. If that's the case, I propose to move forward with this patch to fix suspension on all other targets.

Known issues from my own testing, though these don't seem to be related to this PR:
- Audio VM sometimes fails to "suspend" cleanly - It looks like host waits 5 seconds for the `systemd-suspend` unit to be active, but this is not always enough time
- GUI suspension (no lid closing) seems to be a lot more consistent. Lid close suspension sometimes triggers multiple suspend->resume->suspend cycles. Cause unknown
- `cosmic-greeter` _sometimes_ crashes after resume, which can be identified by a gray screen. It seems to restart cleanly in 30 seconds or so, with no progress being lost.

leivos-unikie:
Tested with lenovo-x1-carbon-gen11-debug-installer (no encryption or sb enabled)

 - Verified that https://jira.tii.ae/browse/SSRCSP-8196 is fixed. Behaves the same way as reported on darter-pro.
 
 - Launching all apps and browsing does not cause dramatic changes in memory consumption. However, using Falcon is an exception, it causes gui-vm available memory to deep dive. The following plot shows vms available memory development while opening all apps one by one. The dip in gui-vm occured after giving this prompt for Falcon: "456-4=?"
X1__mem_recording__LenovoX1-1__Log_memory__mem_avail

milva-unikie:
**Tested on Darter Pro** (`system76-darp11-b-debug-installer`)

I suspended the laptop around 20 times using different methods (different suspend buttons / closing the lid). It woke up after every time. Sometimes a bit slowly, but it was still fully functional afterward.


> However, using Falcon is an exception, it causes gui-vm available memory to deep dive.

I believe this same behavior also exists in mainline. Falcon has always used a lot of memory.

leivos-unikie:
Tested GUI suspension multiple times on lenovo-x1 (lenovo-x1-carbon-gen11-debug-installer)

**Suspension via power menu or lock screen**
- Display shuts down always immediately when clicking suspend but the time for the device to reach actual suspended state depends on apps running
- When nearly all ghaf apps were running it took consistently >30 sec to reach suspended state where the power led starts to pulsate. Usually during this time (between display shutting down and power led starting to pulsate) it is not possible to wake up the device (managed to do that only once in ~10 attempts).
- After the device is really suspended (power led pulsating) wake up works always very smoothly without significant delays
- If no apps are running the device reached suspended state much faster, in 5 sec.
- In case it takes longer to reach suspended state it is possible to operate blindly (from lock screen) for a short time after display has gone off. For example I could log in and write some text into open terminal window after display had swithed off.

leivos-unikie:
> I believe this same behavior also exists in mainline. Falcon has always used a lot of memory.

Yes.

leivos-unikie:
Tested suspension **by lid close** multiple times on lenovo-x1 (lenovo-x1-carbon-gen11-debug-installer)

- Device usually wakes up promptly when the lid is opened
- Once device woke up when lid was opened but display stayed off, then immediately device went again to suspend. Had to press some button to wake up.
- After staying ~15min suspended display didn't turn on in wake up. Had to reboot. In this case I had also USB mouse/keyboard receiver plugged to the laptop which might have some effect. gui-vm logs:
[gui-vm_2026-04-14 12_49_05.txt](https://github.com/user-attachments/files/26707894/gui-vm_2026-04-14.12_49_05.txt)
- Repeated the issue after 12min suspend without USB mouse/keyboard receiver. This time I had all (default) ghaf apps running. Display fails to wake up. gui-vm logs:
[gui-vm_2026-04-14 14_00_00.txt](https://github.com/user-attachments/files/26707917/gui-vm_2026-04-14.14_00_00.txt)

kajusnau:
> Tested suspension **by lid close** multiple times on lenovo-x1 (lenovo-x1-carbon-gen11-debug-installer)
> 
> * Device usually wakes up promptly when the lid is opened
> * Once device woke up when lid was opened but display stayed off, then immediately device went again to suspend. Had to press some button to wake up.
> * After staying ~15min suspended display didn't turn on in wake up. Had to reboot. In this case I had also USB mouse/keyboard receiver plugged to the laptop which might have some effect. gui-vm logs:
>   [gui-vm_2026-04-14 12_49_05.txt](https://github.com/user-attachments/files/26707894/gui-vm_2026-04-14.12_49_05.txt)
> * Repeated the issue after 12min suspend without USB mouse/keyboard receiver. This time I had all (default) ghaf apps running. Display fails to wake up. gui-vm logs:
>   [gui-vm_2026-04-14 14_00_00.txt](https://github.com/user-attachments/files/26707917/gui-vm_2026-04-14.14_00_00.txt)

Not a i915 GPU driver crash
Seems like backlight brightness fails to restore due to `ghaf-powercontrol[9822]: Error: Could not determine user session UI`, which may indicate a session crash.

I propose we move forward with merging this patch as is, and open a separate bug ticket for this issue.

leivos-unikie:
> > Tested suspension **by lid close** multiple times on lenovo-x1 (lenovo-x1-carbon-gen11-debug-installer)
> > 
> > * Device usually wakes up promptly when the lid is opened
> > * Once device woke up when lid was opened but display stayed off, then immediately device went again to suspend. Had to press some button to wake up.
> > * After staying ~15min suspended display didn't turn on in wake up. Had to reboot. In this case I had also USB mouse/keyboard receiver plugged to the laptop which might have some effect. gui-vm logs:
> >   [gui-vm_2026-04-14 12_49_05.txt](https://github.com/user-attachments/files/26707894/gui-vm_2026-04-14.12_49_05.txt)
> > * Repeated the issue after 12min suspend without USB mouse/keyboard receiver. This time I had all (default) ghaf apps running. Display fails to wake up. gui-vm logs:
> >   [gui-vm_2026-04-14 14_00_00.txt](https://github.com/user-attachments/files/26707917/gui-vm_2026-04-14.14_00_00.txt)
> 
> Not a i915 GPU driver crash Seems like backlight brightness fails to restore due to `ghaf-powercontrol[9822]: Error: Could not determine user session UI`, which may indicate a session crash.
> 
> I propose we move forward with merging this patch as is, and open a separate bug ticket for this issue.

Dealing this in separate PR is ok for me.

I was able to debug the state a bit more, found out that display brightness value was 0. Robot automation was able login and do all desktop stuff although display was dark.

brianmcgillion (approved):

PR #1888, opened 2026-04-09
fix(power): keep display on during suspension

kajusnau:
## Description of Changes

Dirty and hopefully temporary "fix" for recent suspension issues
It seems that i915 driver timeout occurs only if the built-in display is disabled before suspension.
This patch changes suspension behavior - suspending now does not turn off the display, but sets the backlight to 0% brightness via `brightnessctl s 0`

This behavior _may_ be explained by the fact that `wlopm --off` triggers a full DRM atomic disable that tears down i915 power wells, while `brightnessctl s 0` only writes to a PWM register that i915 never sees.

Tested manually - 15ish suspension cycles - no i915 driver crash.
Note: I did manual tests on a gen11 X1 with an OLED panel. All of our other targets use typical LCD panels, which may behave differently - for example, it may be so that "0% backlight brightness" does not result in a fully black "off" display, but rather a very dimly lit one.

### Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Improvement / Refactor

## Related Issues / Tickets
Fixes:
- https://jira.tii.ae/browse/SSRCSP-8219
- https://jira.tii.ae/browse/SSRCSP-8271

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

1. Try to suspend any laptop device by either closing the lid or suspending manually
2. Verify device display is practically off - black screen, indistinguishable from being actually powered off
3. Verify device resumes from suspension without a graphics driver timeout
4. Repeat steps 1 and 3 - at least 10 times, letting the device sleep for at least 1 minute each time
5. Verify no issues even after multiple suspension attempts
6. Verify battery usage is not much higher during suspension compared to mainline.

### DIFF
--- a/modules/common/services/power.nix
+++ b/modules/common/services/power.nix
@@ -492,8 +492,7 @@ in
       # Shutdown displays early before suspend
       powerManagement = {
         powerDownCommands = lib.mkBefore ''
-          ${getExe ghaf-powercontrol} turn-on-displays '*'
-          ${getExe ghaf-powercontrol} turn-off-displays '*'
+          ${getExe ghaf-powercontrol} fake-turn-off-displays '*'
         '';
       };
 
@@ -536,7 +535,7 @@ in
           serviceConfig = {
             Type = "oneshot";
             RemainAfterExit = true;
-            ExecStop = "${getExe ghaf-powercontrol} turn-on-displays '*'";
+            ExecStop = "${getExe ghaf-powercontrol} fake-turn-on-displays '*'";
           };
         };
       };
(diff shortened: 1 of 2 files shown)


vunnyso (approved):

vunnyso:
Will it be good to combine it with https://github.com/tiiuae/ghaf/pull/1887 so that testing becomes easier?

kajusnau:
> Will it be good to combine it with #1887 so that testing becomes easier?

Good idea, closing this one

PR #1909, opened 2026-04-27
feat(installer): rework to TUI, improve UX

kajusnau:
### Installer
- Reworked ghaf-installer to TUI using Gum as a framework (same as user provisioning)
- The installer now starts immediately when booting from the installation media
- The user is lightly _discouraged_ from dropping down into the shell, but of course can choose to exit the installer and do so anyway
   - Most common actions (install, disk wipe, reboot, reboot into system firmware, shutdown) are all part of the installer now
   - Exiting the installer must be done consciously by exiting from the main menu
- Enabled Ghaf graphical boot, Ghaf logo will be shown as on normal installed images

### Docs
- Significantly reworked the installer doc to match the new flow
- Expanded the installer doc based on generic NixOS doc
- Added a Getting Started section to the README
- Removed old references and added new ones in README

### New general approach to installing Ghaf:
1. Build Ghaf installer
   `nix build .#intel-laptop-debug-installer`
2. Flash installer to your installation media of choice:
   `sudo ghaf-flash -d /dev/sdb -i result/iso/ghaf.iso -f`
3. Plug in the installation media to the target device and boot from it
4. Wait for the Ghaf installer TUI to show up, follow on-screen instructions

(Optional) Drop into the shell if necessary

### Type of Change
- [x] New Feature
- [ ] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [x] Other: Changes target the installer

### Test Steps To Verify:

1. Evaluate new functionality
2. Verify all previous actions work (disk encryption, secure boot, disk wipe, etc.)


kajusnau:
Waiting for feedback and reviews before requesting testing
Requesting feedback from people who I believe have good first-hand experience with installing and re-installing Ghaf.

PS. any way to preview the changes to `.mdx` files with full rendering?

milva-unikie:
I tried the new installer yesterday on Darter Pro and I like it. The installer now guides the user through the process and it is easy to understand what is happening at each step.

It is good that the user can’t accidentally choose to install the system on the same device the installer is running from. For Secure Boot, there is a warning if the laptop is not in setup mode (and a hint for the user on how to fix it). The GUI is clean and fits well with the user creation. Ghaf loading screen is also a nice addition.

Nice work on both usability and design!

kajusnau:
Update:
Fixed the Installation Complete screen's `Return to the main menu` option to work as expected

rodrigopinotii:
I just checked on the sys76 and well done, nice to see that there are improvements and that users are getting guidance on how to set up their machines. It's clear and easy to understand. From my side, only some observations:

"What would you like to do" page —> I'm not sure why "Wipe disk only" is here. What's the use case for this? If there's a good reason for it (like wiping a device before decommissioning), it would help to explain that to the user. Right now it just feels a bit repetitive.
"Select Target Device" page —> "Target device" is a bit confusing. Something like "Select the disk where you want to perform the installation" would be much clearer. The user should not have to stop and think about what "target device" means.
"Ghaf Installation Complete" page —> I do not really see the point of having "Return to the main menu" here. The installation is done, the user just wants to reboot and start using the laptop. Having a second option just adds unnecessary confusion — someone might even wonder "wait, did something go wrong?" I would suggest to have  "Reboot" as the main action a let the user use inmediatlly the beauty of ghaf

kajusnau:
Thanks for the feedback!

> "What would you like to do" page —> I'm not sure why "Wipe disk only" is here. 

Ghaf is sensitive to situations where multiple installations exist (e.g. internal and external drive both have Ghaf installed), so I think it's helpful the user can wipe the internal drive with this utility at any time.
The wording could be better, though, so I will change that - `wipe` will be generally replaced by `erase`

>"Select Target Device" page —> "Target device" is a bit confusing. Something like "Select the disk where you want to perform the installation" would be much clearer.

Agreed. The exact wording will not be that, but will be improved. The working title is "Select Installation Disk" followed by a "Select a disk: ..." prompt

> "Ghaf Installation Complete" page —> I do not really see the point of having "Return to the main menu" here. The installation is done, the user just wants to reboot and start using the laptop. Having a second option just adds unnecessary confusion — someone might even wonder "wait, did something go wrong?" I would suggest to have "Reboot" as the main action a let the user use inmediatlly the beauty of ghaf

I disagree about "Return to the main menu". I think it's a perfectly valid option the user may want to choose. As an example, let's say I followed the installation prompts and installed Ghaf, but then at the install complete step I notice that I installed the OS on `/dev/nvme1` instead of `/dev/nvme0`. Simple mistake but I don't want to reboot just to restart the entire process again.
As for the possible options in the Install Complete screen, I propose the following:
<details>
<summary>Updated Install Complete options</summary>
image
</details>
Should there be a countdown before the device reboots at this stage? It may be helpful to give the user enough time to remove the installation media.
Yes agree, good idea

kajusnau:
Update:

- Added a "Reboot (recommended)" pre-selected option at the end of the installation process
   - Selecting this will start a 10s timer and ask the user to remove the installation media before the timer runs out
   - The system will reboot when the timer runs out
- Reworded all instances of `wipe` to `erase` instead
- Adjusted wording of disk erase process based on feedback
- Improved the visual progress indicator during installation
- Refactored secure boot enrollment - core functionality is the same, but needs testing to confirm it works as expected

brianmcgillion (approved):

PR #1938, opened 2026-05-11
fix(qemu): use host clock for lid polling

eyad-tii:
## Description of Changes

The QEMU lid button polls host procfs state to mirror the physical lid into gui-vm. Using `QEMU_CLOCK_VIRTUAL` can delay the poll after host s2idle resume, leaving gui-vm with a stale closed lid state long enough for logind to trigger another suspend.

Use `QEMU_CLOCK_HOST` so the poll timer follows host elapsed time and expires promptly after the QEMU process is thawed.

### Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Improvement / Refactor

## Related Issues / Tickets

Fixes https://jira.tii.ae/browse/SSRCSP-8402

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

1. With the patched build, close the lid, then open it after the laptop suspends.
2. After lid is opened, the host resumes and stays awake.

### DIFF
--- a/packages/pkgs-by-name/ghaf-qemu/patches/0004-hw-acpi-Introduce-the-QEMU-lid-button.patch
+++ b/packages/pkgs-by-name/ghaf-qemu/patches/0004-hw-acpi-Introduce-the-QEMU-lid-button.patch
@@ -430,16 +430,16 @@ index 0000000000..dfe86af713
 +        Object *obj = object_resolve_path_type("", TYPE_ACPI_DEVICE_IF, NULL);
 +        acpi_send_event(DEVICE(obj), ACPI_BUTTON_CHANGE_STATUS);
 +    }
-+    timer_mod(s->probe_state_timer, qemu_clock_get_ms(QEMU_CLOCK_VIRTUAL) +
++    timer_mod(s->probe_state_timer, qemu_clock_get_ms(QEMU_CLOCK_HOST) +
 +              s->probe_state_interval);
 +}
 +
 +static void button_probe_state_timer_init(BUTTONState *s)
 +{
 +    if (s->enable_procfs && s->probe_state_interval > 0) {
-+        s->probe_state_timer = timer_new_ms(QEMU_CLOCK_VIRTUAL,
++        s->probe_state_timer = timer_new_ms(QEMU_CLOCK_HOST,
 +                                            button_probe_state, s);
-+        timer_mod(s->probe_state_timer, qemu_clock_get_ms(QEMU_CLOCK_VIRTUAL) +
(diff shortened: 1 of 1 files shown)


kajusnau (approved):
Great find 👏 

I think the same change could be applied also to
- `packages/pkgs-by-name/ghaf-qemu/patches/0002-hw-acpi-Introduce-the-QEMU-Battery.patch`
- `packages/pkgs-by-name/ghaf-qemu/patches/0003-hw-acpi-Introduce-the-QEMU-AC-adapter.patch`

But these can be addressed later

milva-unikie:
**Tested on Darter Pro** (new image) **and Lenovo X1** (`nixos-rebuild`)

I am still able to reproduce the double suspension on both laptops.

- On Lenovo X1 it keeps happening when lid is closed for a few minutes.
  - Logs: [x1-gui-vm.log](https://github.com/user-attachments/files/27624270/x1-gui-vm.log), [x1-host.log](https://github.com/user-attachments/files/27624271/x1-host.log). I closed the lid at 09:37:36, opened it at 09:41:04 and then woke the laptop up again by moving mouse at 09:41:32.
- On Darter Pro I was only able to reproduce when the lid was closed for a longer time (15+ mins). I tried to keep the lid closed for 15 minutes twice and the double suspension happened both times.
  - Logs: [darp-gui-vm.log](https://github.com/user-attachments/files/27624359/darp-gui-vm.log), [darp-host.log](https://github.com/user-attachments/files/27624360/darp-host.log). I closed the lid at 08:11:22, opened it at 08:32:26 and then woke the laptop up again from the power button at 08:33:07.

I used `system76-darp11-b-debug` and `lenovo-x1-carbon-gen11-debug-installer` images.

kajusnau:
Even though testing results are negative, I propose we move forward with this - the scope of this patch is to fix QEMU lid polling, which has been done. It doesn't fully fix the double suspension issue, but it slightly improves it on darter pro, and seems to be the correct approach anyway.

Edit:
Alternatively, I could take this patch into my WIP workaround for the double suspension

vadika:
Lets not merge this patch -- using host clocks may cause deadlocks, thats why the most of the virtual drivers use virtualclocks.

vadika (changes requested):
Using host clocks can cause deadlocks, so I think we should't change them.

eyad-tii:
Closing since we won't use host clocks and issues with double suspend were resolved with #1945

PR #1948, opened 2026-05-14
fix(orin-nx): flash the p3768 devkit to NVMe instead of eMMC

vadika:
## Problem

The Orin NX flash profile was wired for the Xavier NX devkit (`p3509-a02` carrier, eMMC), but:
- the Orin NX SOM has **no eMMC**
- `orin-nx.nix`'s `deviceTree` already targets the **p3768** devkit

## Fix

Three coupled changes:

- **`orin-nx.nix`** — `carrierBoard` `xavierNxDevkit` → `devkit`; `configFileName` → `jetson-orin-nano-devkit-nvme` so `flash.sh` sources the NVMe board config (`flash_t234_qspi_nvme.xml`, not the SD/eMMC one). `flashArgs`' trailing rootfs device `mkForce`'d to `nvme0n1p1` (jetpack-nixos hardcodes `mmcblk0p1`).
- **`partition-template.nix`** — for `som == "orin-nx"`, build the partition template from `flash_t234_qspi_nvme.xml` and splice into the `<device type="nvme">` element instead of `sdmmc_user`.
- **`splice-flash-xml.py`** — add `--device-type` so the splicer can target the `nvme` device element; default stays `sdmmc_user` for the AGX.

## Testing

Verified end to end on an Orin NX 16GB p3768 devkit: flash completes, device boots from `nvme0n1p1`.

### DIFF
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
@@ -273,6 +273,19 @@ in
       default = "";
     };
 
+    flashScriptOverrides.rootfsDevice = mkOption {
+      description = ''
+        Rootfs device passed to NVIDIA's flash.sh as the trailing argument
+        (e.g. "mmcblk0p1" for eMMC/SD, "nvme0n1p1" for NVMe). This is
+        consumed by the partition layout in the .conf file referenced by
+        configFileName and decides where the APP partition lands. Downstream
+        callers (e.g. disk-encryption modules) can read this to know the
+        rootfs location at build time.
+      '';
+      type = types.str;
+      default = "mmcblk0p1";
+    };
+
     somType = mkOption {
       description = "SoM config Type (NX|AGX32|AGX64|Nano)";
       type = types.str;
@@ -302,6 +315,15 @@ in
   config = mkIf cfg.enable {
     hardware.nvidia-jetpack.firmware.eksFile = "${firmwareEkbImage}/eks_t234.img";
     hardware.nvidia-jetpack.kernel.version = "${cfg.kernelVersion}";
+    # jetpack-nixos hardcodes the trailing rootfs device as mmcblk0p1; replay
+    # the same default here but route it through cfg.flashScriptOverrides.rootfsDevice
+    # so per-SoM modules (e.g. orin-nx → nvme0n1p1) only set the option, not
+    # the whole flashArgs list. mkOverride 75 beats jetpack-nixos's plain
+    # assignment (prio 100) while leaving room for downstream mkForce (prio 50).
+    hardware.nvidia-jetpack.flashScriptOverrides.flashArgs = lib.mkOverride 75 [
+      config.hardware.nvidia-jetpack.flashScriptOverrides.configFileName
+      cfg.flashScriptOverrides.rootfsDevice
+    ];
     nixpkgs.hostPlatform.system = "aarch64-linux";
 
     ghaf.hardware = {
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
@@ -281,9 +281,13 @@ in
         configFileName and decides where the APP partition lands. Downstream
         callers (e.g. disk-encryption modules) can read this to know the
         rootfs location at build time.
+
+        No default: rootfs storage is not common to all carrier boards, so
+        every per-SoM module must set this explicitly. An assertion in the
+        config block enforces a non-empty value.
       '';
       type = types.str;
-      default = "mmcblk0p1";
+      default = "";
     };
 
     somType = mkOption {
@@ -313,6 +317,18 @@ in
   };
 
   config = mkIf cfg.enable {
+    assertions = [
+      {
+        assertion = cfg.flashScriptOverrides.rootfsDevice != "";
+        message = ''
+          ghaf.hardware.nvidia.orin.flashScriptOverrides.rootfsDevice must be
+          set explicitly (e.g. "mmcblk0p1" for eMMC/SD, "nvme0n1p1" for NVMe).
+          The default is intentionally empty because rootfs storage varies per
+          carrier board; the per-SoM module must declare it to avoid silently
+          flashing to the wrong device.
+        '';
+      }
+    ];
     hardware.nvidia-jetpack.firmware.eksFile = "${firmwareEkbImage}/eks_t234.img";
     hardware.nvidia-jetpack.kernel.version = "${cfg.kernelVersion}";
     # jetpack-nixos hardcodes the trailing rootfs device as mmcblk0p1; replay
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
@@ -109,7 +109,6 @@ let
       inherit (config.hardware.nvidia-jetpack) som;
       isIndustrial = som == "orin-agx-industrial";
       isNvme = som == "orin-nx";
-      deviceType = if isNvme then "nvme" else "sdmmc_user";
       xmlFile =
         if isIndustrial then
           "${bspSrc}/bootloader/generic/cfg/flash_t234_qspi_sdmmc_industrial.xml"
@@ -124,7 +123,7 @@ let
       }
       ''
         python3 ${./splice-flash-xml.py} \
-          --device-type "${deviceType}" \
+          --device-type "${if isNvme then "nvme" else "sdmmc_user"}" \
           ${lib.optionalString cfg.flashScriptOverrides.onlyQSPI "--remove-device"} \
           --set "esp.size=$(($(cat ${images}/esp.size) * 512))" \
           --set "APP.size=$(($(cat ${images}/root.size) * 512))" \
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
@@ -96,15 +96,25 @@ let
     }
   ];
 
-  # Build the final flash.xml by replacing the sdmmc_user partitions
+  # Build the final flash.xml by replacing the storage-device partitions
   # in NVIDIA's template with our layout using XML-aware splicing.
+  #
+  # The Orin NX SOM has no eMMC; the p3768 devkit boots its rootfs from
+  # NVMe. Splice into the nvme template + device element instead of
+  # sdmmc_user, otherwise MB2 probes SDMMC instance 3 at flash time,
+  # fails ("Secondary storage init failed"), and hangs in Busy Spin.
   partitionTemplate =
     let
       inherit (pkgs.nvidia-jetpack) bspSrc;
-      isIndustrial = config.hardware.nvidia-jetpack.som == "orin-agx-industrial";
+      inherit (config.hardware.nvidia-jetpack) som;
+      isIndustrial = som == "orin-agx-industrial";
+      isNvme = som == "orin-nx";
+      deviceType = if isNvme then "nvme" else "sdmmc_user";
       xmlFile =
         if isIndustrial then
           "${bspSrc}/bootloader/generic/cfg/flash_t234_qspi_sdmmc_industrial.xml"
+        else if isNvme then
+          "${bspSrc}/bootloader/generic/cfg/flash_t234_qspi_nvme.xml"
         else
           "${bspSrc}/bootloader/generic/cfg/flash_t234_qspi_sdmmc.xml";
     in
@@ -114,6 +124,7 @@ let
       }
       ''
         python3 ${./splice-flash-xml.py} \
+          --device-type "${deviceType}" \
           ${lib.optionalString cfg.flashScriptOverrides.onlyQSPI "--remove-device"} \
           --set "esp.size=$(($(cat ${images}/esp.size) * 512))" \
           --set "APP.size=$(($(cat ${images}/root.size) * 512))" \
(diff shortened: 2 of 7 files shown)


modules/reference/hardware/jetpack/nx/orin-nx.nix
```diff
@@ -74,7 +79,23 @@
     nvidia-jetpack = {
       enable = true;
       som = "orin-nx";
-      carrierBoard = "xavierNxDevkit";
+      # p3768 official devkit, matching the p3768 deviceTree above.
+      # "xavierNxDevkit" maps to p3509-a02 + eMMC, which the Orin NX SOM
+      # does not have.
+      carrierBoard = "devkit";
```
TanelDettenborn, line 86: Could we create a separate target for `orin-nx`? The `devkit` and `xavierNxDevkit` profiles aren't fully compatible due to carrier board differences. 

For example, their display outputs differ (DP vs. HDMI), meaning they require different DTB file[1] to get the display working.

[1] https://github.com/anduril/jetpack-nixos/blob/7947e09403dd4b46670e70dd28b4f1aa32f663d1/modules/devices.nix#L106-L120

modules/reference/hardware/jetpack/nx/orin-nx.nix
```diff
@@ -74,7 +79,23 @@
...
+      # The p3768 NX board has no eMMC, so MB2 fails to init SDMMC instance 3
+      # ("Secondary storage init failed" on UART) and hangs in Busy Spin.
+      # The "-nvme" config sources p3768-0000-p3767-0000-a0.conf and uses
+      # flash_t234_qspi_nvme.xml, which probes NVMe instead.
+      flashScriptOverrides.configFileName = "jetson-orin-nano-devkit-nvme";
+      # jetpack-nixos hardcodes the rootfs device as mmcblk0p1 (eMMC);
+      # override the trailing flash.sh arg to the NVMe partition.
+      flashScriptOverrides.flashArgs = lib.mkForce [
+        config.hardware.nvidia-jetpack.flashScriptOverrides.configFileName
+        "nvme0n1p1"
```
TanelDettenborn: This looks good, but I wanted to raise a point regarding the `nvme0n1p1` reference. 

This should likely be exposed as a configurable option. Implementing disk encryption, for example, will require explicit knowledge of the rootfs location beforehand.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
```diff
@@ -96,15 +96,25 @@ let
...
+  # sdmmc_user, otherwise MB2 probes SDMMC instance 3 at flash time,
+  # fails ("Secondary storage init failed"), and hangs in Busy Spin.
   partitionTemplate =
     let
       inherit (pkgs.nvidia-jetpack) bspSrc;
-      isIndustrial = config.hardware.nvidia-jetpack.som == "orin-agx-industrial";
+      inherit (config.hardware.nvidia-jetpack) som;
+      isIndustrial = som == "orin-agx-industrial";
+      isNvme = som == "orin-nx";
+      deviceType = if isNvme then "nvme" else "sdmmc_user";
```
TanelDettenborn: A totally syntactical sugar and personal preferences. I might not introduce another layer variables if they are only used once.

modules/reference/hardware/jetpack/nvidia-jetson-orin/splice-flash-xml.py
```diff
@@ -22,8 +23,12 @@
 The --set flag overrides a child element value for a named partition.
 For example: --set APP.size=12345678
 
-The --remove-device flag removes the sdmmc_user device element entirely
-(for QSPI-only flashing where no eMMC partitions are needed).
+The --remove-device flag removes the storage device element entirely
+(for QSPI-only flashing where no eMMC/NVMe partitions are needed).
```
TanelDettenborn, line 27: Leaving a note here to start a quick discussion on targets and rootfs locations. 

The `flash-script` target for NX is currently a bit opaque. We could potentially structure it as:
- `nx`
- `nx-flash-qspi`
- `nx-flash-qspi-and-rootfs-to-nvme`

Just a thought for down the road, no immediate action needed.

TanelDettenborn (changes requested):
I am currently unable to verify the display output, but otherwise, the NX seems to boot fine from the NVMe. I left a couple of comments, but the PR seems solid!

modules/reference/hardware/jetpack/nx/orin-nx.nix
emrahbillur: Yes, that problem keeps waiting in front of me on all different Saluki boards as they have different outputs and board structures.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
```diff
@@ -96,15 +96,25 @@ let
...
+  # NVMe. Splice into the nvme template + device element instead of
+  # sdmmc_user, otherwise MB2 probes SDMMC instance 3 at flash time,
+  # fails ("Secondary storage init failed"), and hangs in Busy Spin.
   partitionTemplate =
     let
       inherit (pkgs.nvidia-jetpack) bspSrc;
-      isIndustrial = config.hardware.nvidia-jetpack.som == "orin-agx-industrial";
+      inherit (config.hardware.nvidia-jetpack) som;
+      isIndustrial = som == "orin-agx-industrial";
+      isNvme = som == "orin-nx";
```
emrahbillur, line 111: Here the isIndustrial and isNvme solves most of the problems. 
nx-qspi -> only qspi flash for use with external USB boot or qspi update
nx-nvme -> full qspi and external nvme flash for nx
where the carrier board issues are a little problematic. 
Especially Orin NX has so many variations of carrier boards.

modules/reference/hardware/jetpack/nvidia-jetson-orin/splice-flash-xml.py
emrahbillur: Only nx might be unnecessary as there is no external or internal storage to be flashed defined. 
Nx-flash-qspi is OK for USB devices and QSPI update.
NX-flash-qspi and rootfs-to-nvme is quite Ok.

But I still believe current naming still solves as nx itself is not meaningful for flash-script

modules/reference/hardware/jetpack/nx/orin-nx.nix
```diff
@@ -74,7 +79,23 @@
...
+      # carrierBoard "devkit" defaults configFileName to
+      # "jetson-orin-nano-devkit", whose .conf flashes a QSPI+SD/eMMC layout.
+      # The p3768 NX board has no eMMC, so MB2 fails to init SDMMC instance 3
+      # ("Secondary storage init failed" on UART) and hangs in Busy Spin.
+      # The "-nvme" config sources p3768-0000-p3767-0000-a0.conf and uses
+      # flash_t234_qspi_nvme.xml, which probes NVMe instead.
+      flashScriptOverrides.configFileName = "jetson-orin-nano-devkit-nvme";
+      # jetpack-nixos hardcodes the rootfs device as mmcblk0p1 (eMMC);
+      # override the trailing flash.sh arg to the NVMe partition.
+      flashScriptOverrides.flashArgs = lib.mkForce [
```
emrahbillur: I recommend any setting of NX on ghaf to be set using mkDefault or mkOverride with low priority to be easily overridden in downstream channels.

emrahbillur (approved):
The logic is quite straight forward and it will surely work. My only major comment is removal of mkForce structures and replace them with mkOverride or mkDefault for easy overriding in downstream channels.

modules/reference/hardware/jetpack/nx/orin-nx.nix
vadika: Done in 86071903. Added `ghaf.hardware.nvidia.orin.flashScriptOverrides.rootfsDevice` (default `mmcblk0p1`). `orin-nx.nix` sets it to `nvme0n1p1`; `jetson-orin.nix` builds `flashArgs` from it via `lib.mkOverride 75`. Disk-encryption modules can read or override the option without touching `flashArgs` shape.

Verified flashArgs eval:
- agx-debug → `[jetson-agx-orin-devkit, mmcblk0p1]`
- agx-industrial-debug → `[jetson-agx-orin-devkit-industrial, mmcblk0p1]`
- agx64-debug → `[jetson-agx-orin-devkit, mmcblk0p1]`
- nx-debug → `[jetson-orin-nano-devkit-nvme, nvme0n1p1]`

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
vadika: Done in 86071903. Inlined ternary at the `--device-type` arg.

modules/reference/hardware/jetpack/nx/orin-nx.nix
vadika: Done in 86071903. Replaced `lib.mkForce` with `lib.mkOverride 75` and moved the assignment into `jetson-orin.nix` (shared path). Prio 75 beats jetpack-nixos's plain prio-100 assignment, still leaves room below for downstream `mkForce` (50) or `mkOverride <75`.

modules/reference/hardware/jetpack/nx/orin-nx.nix
vadika: Acknowledged, out of scope for this PR. Display DTB split (DP vs HDMI) needs proper per-carrier-board target work, which @emrahbillur mentions hits the wider Saluki board fan-out too. This PR's scope is unbreaking the NVMe flash; tracking the per-target split as follow-up.

modules/reference/hardware/jetpack/nvidia-jetson-orin/splice-flash-xml.py
vadika: Agreed, follow-up. Out of scope for this NVMe-flash fix. Will revisit `nx-flash-qspi` / `nx-flash-qspi-and-rootfs-to-nvme` naming once `orin-nx` has its own target per the carrier-board thread above.

leivos-unikie:
Tested flashing Orin NX with nvme at the bottom. Booted fine and no regression found.

emrahbillur:
Tested on Orin NX 16 GB SOM on Waveshare IO Base B v2.2 board (with same devicetree of devkit - not same as IO Base B v1.x) with 256 GB SSD on board and Type-C flashing port. QSPI and NVME erasing and flashing were performed without any errors. 
[flash.log](https://github.com/user-attachments/files/28743769/flash.log)
System booted up to ghaf directly after flashing. Ghaf services active, all vm's are up and working. Nixos closures are created correctly.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -273,6 +273,19 @@ in
       default = "";
     };
 
+    flashScriptOverrides.rootfsDevice = mkOption {
```
emrahbillur, line 276: Perfect definition. This will allow any type of storage to be flashed.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -302,6 +315,15 @@ in
...
     hardware.nvidia-jetpack.firmware.eksFile = "${firmwareEkbImage}/eks_t234.img";
     hardware.nvidia-jetpack.kernel.version = "${cfg.kernelVersion}";
+    # jetpack-nixos hardcodes the trailing rootfs device as mmcblk0p1; replay
+    # the same default here but route it through cfg.flashScriptOverrides.rootfsDevice
+    # so per-SoM modules (e.g. orin-nx → nvme0n1p1) only set the option, not
+    # the whole flashArgs list. mkOverride 75 beats jetpack-nixos's plain
+    # assignment (prio 100) while leaving room for downstream mkForce (prio 50).
+    hardware.nvidia-jetpack.flashScriptOverrides.flashArgs = lib.mkOverride 75 [
+      config.hardware.nvidia-jetpack.flashScriptOverrides.configFileName
+      cfg.flashScriptOverrides.rootfsDevice
```
emrahbillur, line 341: This mkOverride allows downstream changes.

emrahbillur (approved):
Quite efficient flash script for nvme on Orin NX devices. We can later adapt or change on any downstream project for similar carrier boards with current implementation.

emrahbillur:
Now tested on older carrier board with microUSB flash port. Orin NX 16GB. 500 GB Samsung 970 EVO SSD. QSPI and NVME erasing and flashing were performed without any errors. NVME erasing is slower than previous test device but size is double. 
[flash.log](https://github.com/user-attachments/files/28746526/flash.log)
System booted up to ghaf directly after flashing. Ghaf services active, all vm's are up and working. Nixos closures are created correctly.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
TanelDettenborn: Yes. This is a perfect definition.

Should we leave the default as an empty string and set rootfsDevice explicitly in the device-specific files? We could also add an assertion here to check that rootfsDevice is not empty. Since this option isn't common to all boards, this would prevent accidental misconfigurations.

leivos-unikie:
Added question label because of Tanel's comment:
"I am currently unable to verify the display output..."

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Done in d1b8db3c. Default dropped to `""`, added `assertions` block in `jetson-orin.nix` config that fires "rootfsDevice must be set explicitly" if empty. All three AGX boards (`orin-agx.nix`, `orin-agx-industrial.nix`, `orin-agx64.nix`) now declare `rootfsDevice = "mmcblk0p1"` explicitly; NX keeps `"nvme0n1p1"`. Verified assertion fires on `mkForce ""` override against nx-debug toplevel build.

leivos-unikie:
> Added question label because of Tanel's comment: "I am currently unable to verify the display output..."

Emrah checked that with overrides
deviceTree.name = "tegra234-p3768-0000+p3509-a02.dtb";
carrierBoard = "xavierNxDevkit";
we still get graphics output from xavier devkit boards (old waveshare io base b)

--> removing the question label

TanelDettenborn (approved):
LGTM

brianmcgillion (approved):

PR #1950, opened 2026-05-15
fix(ci/rebuild): fix rebuild switch

kajusnau:
## Description of Changes

1. **Fixed power manager cfg to re-enable `nixos-rebuild switch`:**
   - Using rebuild switch will no longer cause the system to shutdown.
   - Instead, the affected services (usually system vms) will restart "cleanly" (SIGTERM) while host continues running

2. **Adjusted `ghaf-rebuild` to convert all `switch` requests to a modified version of `boot`:**
   - This allows us to see the package diff even on a `rebuild switch` request, by performing the following:
     1. Build the new system
     2. Set the profile to be activated on next boot
     3. Show `nvd` package diff
     4. Call `switch-to-configuration switch` on the new profile using a `systemd-run --no-block` call
     5. Don't wait for the call to finish, as it's likely the connection will drop anyway (due to net-vm restarting)

### Type of Change
- [ ] New Feature
- [x] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Lenovo X1 `x86_64`
- [ ] Dell Latitude `x86_64`
- [ ] System 76 `x86_64`
- [x] All/None

### Installation Method
- [x] Other: See description

### Test Steps To Verify:

1. Take this functionality into use with `ghaf-rebuild ... boot` and rebooting the target system
2. Make some changes in your local Ghaf config, best would be to add some random package to `systemPackages` somewhere, and use `ghaf-rebuild ... switch`
3. Confirm step 2 doesn't hang and shows a package diff output. Also confirm target Ghaf system doesn't shutdown or reboot fully, only the affected VM is restarted (can be observed in host)

### DIFF
--- a/modules/common/services/power.nix
+++ b/modules/common/services/power.nix
@@ -803,6 +803,21 @@ in
                       ExecStop =
                         let
                           sysvm-stop = pkgs.writeShellScript "sysvm-stop" ''
+                            # Check if this stop is part of a real shutdown or suspend.
+                            # During a nixos-rebuild switch, affected services (system vms) are restarted
+                            # which causes their ExecStop actions to be executed
+                            # If no real reboot/suspend job is queued, we can assume the VM should be
+                            # restarted without requesting the host itself to reboot
+                            jobs=$(${getExe' pkgs.systemd "systemctl"} list-jobs 2>/dev/null || true)
+                            if ! echo "$jobs" | grep -qiE '(sleep|suspend|poweroff|reboot|halt)\.target.*start'; then
+                              echo "No shutdown/suspend in progress, skipping graceful shutdown for '${vmName}' (likely a rebuild)"
+                              kill -15 $MAINPID 2>/dev/null
+                              while kill -0 $MAINPID 2>/dev/null; do
+                                sleep 1
+                              done
+                              exit 0
+                            fi
+
                             echo "Starting poweroff target for system VM '${vmName}'"
                             vm=''${${vmName}/-vm/}
                             ${givc-cli} start service --vm '${vmName}' poweroff.target &
--- a/packages/pkgs-by-name/ghaf-build-helper/package.nix
+++ b/packages/pkgs-by-name/ghaf-build-helper/package.nix
@@ -98,28 +98,56 @@ writeShellApplication {
 
         export NIX_SSHOPTS
 
-        # Capture the current system path on the target before rebuild for diffing
+        # Capture the current system path on the target before rebuild for diffing.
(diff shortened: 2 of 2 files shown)


kajusnau:
Moving to draft to wait for testing results

TanelDettenborn:
It might be best to wait for feedback from the testing team, but at the moment it appears that this does not work on the Orin NX device.

The NX target evaluates and builds successfully, but the system fails to boot. From the console output:
`[FAILED] Failed to start Find NixOS closure.`
`See 'systemctl status initrd-find-nixos-closure.service' for details.`
`[DEPEND] Dependency failed for Initrd Default Target.`
(After this, the device drops into an emergency shell.)

Additionally, nixos-rebuild currently does not work on the NX device. The command itself fails, although the device is still able to boot afterwards.

TanelDettenborn:
From journalctl:
`Dec 17 23:02:55 ghaf-host initrd-find-nixos-closure-start[304]: Failed to resolve path /nix/store/i4928fjqwdjgd7iq1sq8fvr0i2z98r4w-nixos-system-ghaf-host-sd-card-26.05.20260418.b12141e/init in prefix /sy
sroot: openat2 subpath failed: openat2([3]"/sysroot", "/nix/store/i4928fjqwdjgd7iq1sq8fvr0i2z98r4w-nixos-system-ghaf-host-sd-card-26.05.20260418.b12141e/init", { flags: OpenFlags(O_PATH | O_CLOEXEC), res
olve: ResolveFlags(RESOLVE_IN_ROOT | RESOLVE_NO_MAGICLINKS) }, 24): No such file or directory (os error 2).`

kajusnau:
@TanelDettenborn Are those outputs the result of flashing a new build of this PR to Orin and booting or using `ghaf-rebuild switch` **after** this PR has already been flashed?

TanelDettenborn:
@kajusnau Tested with a fresh build and flashing from this PR. The device fails early in boot, before `nixos-rebuild` can be used.

TanelDettenborn:
Apologies, this was my mistake. I was using an NX device with NVMe, which was trying to load Ghaf from it.

To clarify, my earlier comments were incorrect: the NX target evaluates, builds, and boots successfully. `nixos-rebuild` does not work, but this is also the case on main.

milva-unikie:
**Tested on Darter Pro**
- `ghaf-rebuild ... switch` shows the package diff output
- `ghaf-rebuild ... switch` does not hang or trigger a full reboot

brianmcgillion (approved):

PR #1951, opened 2026-05-18
feat(power): add GPU suspend option

nesteroff:
## Description of Changes

This adds an experimental GPU suspend mode to the power management module.

In Ghaf, the GPU is passed through to the GUIVM, which means the host cannot suspend it together with the rest of the system. Keeping the GPU running while everything else is suspended may leave it in an inconsistent state, preventing the driver from restoring it correctly on resume. In some cases, the screen stays black after resume and there are i915 crashes in the logs, although the issue is rare and difficult to reproduce consistently.

One possible solution would be suspending the GUIVM itself but that is currently not feasible because virtio-fs does not support suspend and it is unclear how well QEMU supports guest suspend in this setup.

As a workaround, this patch uses the [kernel pm_test mechanism](https://www.kernel.org/doc/Documentation/power/basic-pm-debugging.txt) which allows freezing all processes and suspending devices inside the guest without completing the full suspend sequence. A small kernel patch is added to suspend only the GPU. This allows the GUIVM to execute the GPU driver suspend and resume callbacks, which should gracefully transition the GPU into a suspended state, turn off the display and restore everything on resume.

The intended flow is:
- The system initiates suspend.
- The GUIVM receives a GIVC request and executes a pm_test suspend.
- All processes inside the GUIVM are frozen and the GPU driver suspend callback is executed.
- The host continues with the normal suspend flow and freezes all processes, including all QEMU instances.
- When the system resumes, the GUIVM is awakened by a wake-up signal from the host over the serial port.

Calling the GPU driver suspend and resume handlers may improve suspend reliability but it is still hard to say for sure because the issue is difficult to reproduce. At this stage, this mode should be considered experimental and requires more testing.

### Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Improvement / Refactor

## Checklist

### Applicable Targets
- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:
Test all possible suspend and resume scenarios and make sure there are no issues similar to those reported by Rodrigo. Currently, GPU suspend mode is enabled for all targets.

### DIFF
--- a/modules/common/services/power.nix
+++ b/modules/common/services/power.nix
@@ -5,6 +5,7 @@
   config,
   pkgs,
   lib,
+  options,
   ...
 }:
 let
@@ -93,6 +94,21 @@ let
     ) config.microvm.vms
   );
 
+  # List of VMs that should run kernel GPU suspend when the host suspends
+  gpuSuspendVms = lib.attrNames (
+    filterAttrs (
+      _n: v:
+      let
+        vmConfig = lib.ghaf.vm.getConfig v;
+      in
+      vmConfig != null
+      && vmConfig.ghaf.services.power-manager.enable
+      && vmConfig.ghaf.services.power-manager.vm.enable
+      && vmConfig.ghaf.services.power-manager.gui.enable
+      && vmConfig.ghaf.services.power-manager.gui.gpuSuspend
+    ) config.microvm.vms
+  );
+
   # Host suspend actions
   host-suspend-actions = pkgs.writeShellApplication {
     name = "host-suspend-actions";
@@ -101,6 +117,7 @@ let
       pkgs.systemd
       pkgs.coreutils
       pkgs.grpcurl
+      pkgs.socat
       pkgs.vhotplug
       pkgs.wait-for-unit
     ];
@@ -173,6 +190,30 @@ let
                 echo "Fallback: restarting $vm_name..."
                 systemctl restart microvm@"$vm_name".service
               fi
+              ;;
+            *)
+              echo "Invalid action: $action"
+              echo "Usage: $0 <VM Name> <suspend-action> (suspend|resume)"
+              exit 1
+              ;;
+          esac
+          ;;
+
+        gpu-suspend)
+          case "$action" in
+            suspend)
+              echo "Signaling kernel GPU suspend to $vm_name..."
+              ${givc-cli} start service --vm "$vm_name" gpu-suspend.service &
+              sleep 1
+              ;;
+            resume)
+              wake_socket="${config.microvm.stateDir}/$vm_name/vm-wake.sock"
+              echo "Signaling kernel GPU resume to $vm_name..."
+              if [ ! -S "$wake_socket" ]; then
+                echo "Wake socket $wake_socket does not exist, $vm_name will resume after fallback timeout (${toString cfg.gui.gpuSuspendDuration}s)"
+                exit 1
+              fi
+              printf 'resume\n' | socat -u - "UNIX-CONNECT:$wake_socket"
               ;;
             *)
               echo "Invalid action: $action"
@@ -527,6 +568,29 @@ in
         If running in a VM and GIVC is enabled, it replaces the default systemd actions for suspend, poweroff, and
         reboot with givc commands.
       '';
+      gpuSuspend = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether the host should trigger kernel GPU suspend for this VM before host suspend.
+        '';
+      };
+      gpuSuspendDuration = mkOption {
+        type = types.ints.positive;
+        default = 20;
+        description = ''
+          Duration in seconds for the kernel GPU suspend pm_test delay.
+          This is used as a fallback timeout if the socket wakeup signal is not received.
+        '';
+      };
+      fakeDisplaySuspend = mkOption {
+        type = types.bool;
+        default = !cfg.gui.gpuSuspend;
+        defaultText = literalExpression "!config.ghaf.services.power-manager.gui.gpuSuspend";
+        description = ''
+          Whether to use ghaf-powercontrol to fake display off and on during GUI suspend and resume.
+        '';
+      };
     };
     host = {
       enable = mkEnableOption ''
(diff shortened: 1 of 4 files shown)


modules/common/services/power.nix
```diff
@@ -170,6 +184,24 @@ let
...
+          ;;
+
+        gpu-suspend)
+          case "$action" in
+            suspend)
+              echo "Signaling GPU suspend to $vm_name..."
+              ${givc-cli} start service --vm "$vm_name" gpu-suspend.service &
+              sleep 1
+              ;;
+            resume)
```
kajusnau, line 209: Could/should we signal **SIGTERM** to `gpu-suspend.service` here, with the service itself trapping on **SIGTERM** of course:

```
grpcurl -cacert /etc/givc/ca-cert.pem -cert /etc/givc/cert.pem -key /etc/givc/key.pem \
  -d '{"UnitName":"gpu-suspend.service"}' "$vm_name":9000 systemd.UnitControlService.StopUnit
```

It seems to me like in this case the configurable timeout wouldn't be necessary

nesteroff: The problem is that all userspace processes inside the guest are frozen and cannot react to anything. Otherwise, suspending the GPU would not be safe. Maybe we could add some kind of ACPI wake-up or wake-by-interrupt mechanism, similar to how wake-up works in s2idle mode but that would complicate things quite a bit.

rodrigopinotii:
The test cycles covered only manual suspend/resume.
Suspend trigger: idle timeout **(no lid close)**
Resume method: keyboard press
Test Cycles: 9 with kernel sleep

No GPU errors had been seen during the test activities in any cycle. The `gpu-suspend.service` completed successfully
This is the primary objective of the patch and it was succesfully.
Following services crashed across all the cycles and auto restarted sucessfully:
systemd-journald / systemd-networkd / systemd-timesyncd / systemd-udevd / systemd-logind/ systemd-oomd /systemd-resolved
The screen takes up to ~20 seconds to appear after  resume (configuration pm_test)

In Summary

- No GPU driver errors (i915, xe) in GUIVM
- No black screen after resume (screen always returned, within ~20 seconds)
- No permanent WiFi interface loss
- No unrecoverable service failure requiring manual intervention
- No system hang or reboot

Only one observation:
After every cycle the COSMIC network applet in the GUIVM displayed the WiFi icon as disconnected. Moving the mouse and clicking over the icon  refresh and the icon correctly showed connected.

nesteroff:
Updated to a version where the timeout is no longer needed. The host wakes up the GUIVM via a serial-port signal. To make this work, the serial port is also suspended in the patch and the "soft" wake-up path is used by arming wakeup_count. The system should now resume immediately, while the timeout remains as a fallback.

modules/common/services/power.nix
```diff
@@ -717,7 +789,44 @@ in
...
+                      wc=$(cat /sys/power/wakeup_count)
+                      echo "$wc" > /sys/power/wakeup_count
+                      # Enter suspend and wait for the host wakeup signal
+                      echo freeze > /sys/power/state
+                      echo "GPU resumed"
+                      echo 0 > /sys/module/suspend/parameters/pm_test_gpu_only
+                    '';
+                  };
+                in
+                getExe guestGpuSuspend;
```
kajusnau, line 825: Since we're writing to kernel modules, it may be useful to prefix this script call with `+` to ensure full privileges at all times.
Probably not strictly necessary since we're already running as root.

modules/common/services/power.nix
```diff
@@ -180,6 +197,30 @@ let
...
+          case "$action" in
+            suspend)
+              echo "Signaling kernel GPU suspend to $vm_name..."
+              ${givc-cli} start service --vm "$vm_name" gpu-suspend.service &
+              sleep 1
+              ;;
+            resume)
+              wake_socket="${config.microvm.stateDir}/$vm_name/vm-wake.sock"
+              echo "Signaling kernel GPU resume to $vm_name via $wake_socket..."
+              if [ ! -S "$wake_socket" ]; then
```
kajusnau, line 212: Maybe the socket check should be done before any other prints, and, if the socket doesn't exist, we should print the expected default timeout `gpuSuspendDuration`, something like
`Wake socket $wake_socket does not exist, unable to trigger resume for $vm_name`
`$vm_name will resume in ${gpuSuspendDuration}s`

kajusnau (changes requested):

modules/common/services/power.nix
nesteroff: The service already has sufficient permissions so adding + would be redundant in this case. The + prefix makes sense when there are multiple Exec* directives in the unit or in similar situations where only specific commands need elevated privileges.

modules/common/services/power.nix
nesteroff: Sure, we can print the timeout but honestly it should never be reached unless there's an issue with the configuration or the kernel patch hasn't been applied.

nesteroff:
Rebased onto the latest main.

kajusnau (approved):

milva-unikie:
Please return `wlopm` to gui-vm, it is not available with this PR. Our automated suspension tests won't work without it.

nesteroff:
> Please return `wlopm` to gui-vm, it is not available with this PR. Our automated suspension tests won't work without it.

I guess this was a side effect of the ghaf-powercontrol script not being used in this mode. I added wlopm to the base guivm configuration since it is always required.

milva-unikie:
**Tested on Darter Pro and Lenovo X1** (new installer images)

All good!
- Could not reproduce i915 crashes on either laptop after suspension.
- `wlopm` is available again.
- Automated suspension tests pass on both laptops.
- Power consumption during suspension with this PR is significantly lower than with mainline. We have one Lenovo X1 with power consumption measurement system in the lab.
- This PR also fixes the issue where Lenovo X1 screen flickers during suspension (https://jira.tii.ae/browse/SSRCSP-8483).

Notes:
- Double suspension (https://jira.tii.ae/browse/SSRCSP-8506) is still an issue, but not related to this PR.

brianmcgillion (approved):

PR #1964, opened 2026-05-26
fix(fss): log sealing reports bad message failure after reboot

juliuskoskela:
## Description of Changes

This PR improves Forward Secure Sealing (FSS) reliability around boot-time clock instability and journald recovery.

The changes add a clock-readiness barrier before persistent sealed logging starts, so journal flush, FSS setup, verification, recovery, and Alloy are ordered after the realtime clock has either stabilized or reached a bounded fallback. FSS setup now records per-boot baseline state and expected bootstrap/recovery archives, allowing verification to distinguish known clock-induced journal rotations from active system journal corruption.

The PR also adds packaged FSS triage tooling for deployed systems. The `fss-triage` helper collects journald/FSS state, unit status, current and previous boot logs, verification summaries, and relevant clock-readiness state while redacting verification keys from command logs.

Additional hardening ensures sealing keys are owned by `root:root` with `0600` permissions, avoids unnecessary empty rotations, preserves verification behavior across recovery rotations, and extends the FSS test coverage for these paths.

### Type of Change
- [X] New Feature
- [X] Bug Fix
- [X] Improvement / Refactor

## Related Issues / Tickets

None linked.

## Checklist

- [X] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/

`make-checks` and GitHub Actions are left unchecked until they have been run for this branch.

### Applicable Targets
- [X] Orin AGX `aarch64`
- [X] Orin NX `aarch64`
- [X] Lenovo X1 `x86_64`
- [X] Dell Latitude `x86_64`
- [X] System 76 `x86_64`

### Installation Method
- [X] Can be updated with `nixos-rebuild ... switch`
- [ ] Other:

### Test Steps To Verify:

1. Build the FSS NixOS test:

   ```bash
   nix build .#checks.x86_64-linux.logging-fss
   ```

2. Build the manual FSS test and triage helpers:

   ```bash
   nix build .#checks.x86_64-linux.fss-test
   nix build .#checks.x86_64-linux.fss-triage
   ```

3. Deploy a system with FSS enabled using the updated closure, for example with `nixos-rebuild ... switch`.

4. After boot, verify that clock readiness completed and the dependent logging services are ordered after it:

   ```bash
   systemctl status ghaf-clock-ready.service
   systemctl show systemd-journal-flush.service journal-fss-setup.service journal-fss-verify.service \
     --property=After --property=Requires --property=Wants
   ```

5. Verify that FSS setup completed and the sealing key has restricted permissions:

   ```bash
   systemctl status journal-fss-setup.service
   stat -c '%U:%G %a' /var/log/journal/$(cat /etc/machine-id)/fss
   ```

   Expected permission output is:

   ```text
   root:root 600
   ```

6. Run the deployed FSS test helper as root:

   ```bash
   fss-test
   ```

7. Collect triage diagnostics and inspect the generated summary:

   ```bash
   fss-triage
   ```

8. Confirm that active system journal verification does not report critical failures:

   ```bash
   journalctl -u journal-fss-verify.service --no-pager
   journalctl -g 'AUDIT_LOG_INTEGRITY_FAIL|AUDIT_LOG_VERIFY_COMPLETED' --no-pager
   ```

### DIFF
--- a/modules/common/logging/common.nix
+++ b/modules/common/logging/common.nix
@@ -14,6 +14,212 @@ let
     types
     ;
   recCfg = config.ghaf.logging.recovery;
+  loggingStackEnabled = config.ghaf.logging.enable || config.ghaf.logging.fss.enable;
+  clockReadyEnabled = config.ghaf.logging.fss.enable && recCfg.enable && recCfg.clockReady.enable;
+  fssActivationCfg = config.ghaf.logging.fss.activation;
+  fssActivationEnabled = config.ghaf.logging.fss.enable && fssActivationCfg.enable;
+  # Effective time-sync wait before FSS activation; only applies when the
+  # activation boundary is enabled. Computed once and reused below.
+  effectiveSyncWaitSeconds = if fssActivationEnabled then fssActivationCfg.syncWaitSeconds else 0;
+
+  ghafClockReady = pkgs.writeShellApplication {
+    name = "ghaf-clock-ready";
+    runtimeInputs = with pkgs; [
+      coreutils
+      gawk
+      systemd
+    ];
+    text = ''
+      stable_seconds="${toString recCfg.clockReady.stableSeconds}"
+      max_wait_seconds="${toString recCfg.clockReady.maxWaitSeconds}"
+      min_epoch="${toString recCfg.clockReady.minEpochSeconds}"
+      max_epoch="${toString recCfg.clockReady.maxEpochSeconds}"
+      ready_file="/run/ghaf-clock-ready"
+      state_file="/run/ghaf-clock-ready-state"
+      state_dir="/var/lib/ghaf/clock-ready"
+      anchor_file="$state_dir/last-good-realtime"
+
+      uptime_seconds() {
+        awk '{printf "%d\n", $1}' /proc/uptime
+      }
+
+      read_epoch_file() {
+        local path="$1"
+        local value=""
+
+        [ -r "$path" ] || return 0
+        value="$(tr -d '\n' < "$path" 2>/dev/null || true)"
+        case "$value" in
+        "" | *[!0-9]*) return 0 ;;
+        *) printf '%s\n' "$value" ;;
+        esac
+      }
+
+      write_state() {
+        local now_real now_up
+
+        now_real="$(date +%s)"
+        now_up="$(uptime_seconds)"
+        {
+          printf 'ready_established=%s\n' "$ready_established"
+          printf 'sync_result=%s\n' "$sync_result"
+          printf 'sync_value=%s\n' "$sync_value"
+          printf 'realtime=%s\n' "$now_real"
+          printf 'uptime_seconds=%s\n' "$now_up"
+          printf 'min_allowed=%s\n' "$min_allowed"
+          printf 'max_allowed=%s\n' "$max_epoch"
+          printf 'anchor_epoch=%s\n' "''${anchor_epoch:-}"
+          printf 'anchor_status=%s\n' "''${anchor_status:-unknown}"
+        } > "$state_file"
+        chmod 0644 "$state_file"
+      }
+
+      mkdir -p "$state_dir"
+
+      anchor_epoch="$(read_epoch_file "$anchor_file")"
+      anchor_status="missing"
+      min_allowed="$min_epoch"
+      if [ -n "$anchor_epoch" ] && [ "$anchor_epoch" -gt "$max_epoch" ]; then
+        echo "Clock readiness ignoring future-poisoned anchor $anchor_epoch above maximum $max_epoch"
+        anchor_status="ignored-future"
+        anchor_epoch=""
+      elif [ -n "$anchor_epoch" ] && [ "$anchor_epoch" -gt "$min_allowed" ]; then
+        min_allowed="$anchor_epoch"
+        anchor_status="accepted"
+      elif [ -n "$anchor_epoch" ]; then
+        anchor_status="below-minimum"
+      fi
+
+      start_up="$(uptime_seconds)"
+      last_real="$(date +%s)"
+      stable_since="$start_up"
+      ready_established=0
+      sync_result="not-started"
+      sync_value="unknown"
+
+      while true; do
+        sleep 1
+        now_up="$(uptime_seconds)"
+        now_real="$(date +%s)"
+
+        if [ "$now_real" -gt "$max_epoch" ]; then
+          stable_since=""
+          echo "Clock readiness waiting: realtime $now_real is above maximum $max_epoch"
+        elif [ "$now_real" -lt "$min_allowed" ]; then
+          stable_since=""
+          echo "Clock readiness waiting: realtime $now_real is below minimum $min_allowed"
+        elif [ "$now_real" -lt "$last_real" ]; then
+          stable_since=""
+          echo "Clock readiness waiting: realtime moved backwards from $last_real to $now_real"
+        else
+          if [ -z "$stable_since" ]; then
+            stable_since="$now_up"
+          fi
+
+          if [ "$((now_up - stable_since))" -ge "$stable_seconds" ]; then
+            echo "Clock readiness established after $stable_seconds stable seconds"
+            ready_established=1
+            break
+          fi
+        fi
+
+        last_real="$now_real"
+
+        if [ "$((now_up - start_up))" -ge "$max_wait_seconds" ]; then
+          echo "Clock readiness max wait reached; allowing boot to continue with realtime $now_real"
+          break
+        fi
+      done
+
+      # NTP synchronization is intentionally NOT awaited here. This barrier runs
+      # early (before systemd-journal-flush, which requires it, and before
+      # sysinit.target), so networking/timesyncd has not started yet and the NTP
+      # check could only ever hit its full timeout, stalling the journal flush on
+      # every boot. The sync wait is handled later by ghaf-clock-sync.service,
+      # which runs after networking and before journal-fss-setup.
+      sync_result="deferred"
+
+      now_real="$(date +%s)"
+      if [ "$now_real" -gt "$max_epoch" ]; then
+        echo "Clock readiness did not update last-good realtime: $now_real is above maximum $max_epoch"
+      elif [ "$now_real" -ge "$min_allowed" ]; then
+        printf '%s\n' "$now_real" > "$anchor_file"
+        chmod 0644 "$anchor_file"
+      elif [ "$ready_established" -eq 0 ]; then
+        echo "Clock readiness fallback did not update last-good realtime: $now_real is below $min_allowed"
+      fi
+
+      write_state
+      touch "$ready_file"
+      chmod 0644 "$ready_file"
+    '';
+  };
+
+  # Time-synchronization wait, split out of ghaf-clock-ready so it runs AFTER
+  # networking/timesyncd and does not block the early journal flush. It best-effort
+  # waits for NTP synchronization up to the configured bound before FSS activation,
+  # then releases boot regardless (clock readiness is a gate, not a time authority).
+  ghafClockSync = pkgs.writeShellApplication {
+    name = "ghaf-clock-sync";
+    runtimeInputs = with pkgs; [
+      coreutils
+      gawk
+      systemd
+    ];
+    text = ''
+      sync_wait_seconds="${toString effectiveSyncWaitSeconds}"
+      sync_state_file="/run/ghaf-clock-sync-state"
+      synced_file="/run/ghaf-clock-synced"
+
+      uptime_seconds() {
+        awk '{printf "%d\n", $1}' /proc/uptime
+      }
+
+      sync_result="not-started"
+      sync_value="unknown"
+
+      write_sync_state() {
+        {
+          printf 'sync_result=%s\n' "$sync_result"
+          printf 'sync_value=%s\n' "$sync_value"
+          printf 'sync_wait_seconds=%s\n' "$sync_wait_seconds"
+          printf 'realtime=%s\n' "$(date +%s)"
+          printf 'uptime_seconds=%s\n' "$(uptime_seconds)"
+        } > "$sync_state_file"
+        chmod 0644 "$sync_state_file"
+      }
+
+      if [ "$sync_wait_seconds" -le 0 ]; then
+        sync_result="disabled"
+      elif command -v timedatectl >/dev/null 2>&1; then
+        sync_start_up="$(uptime_seconds)"
+        while true; do
+          sync_value="$(timedatectl show -p NTPSynchronized --value 2>/dev/null || true)"
+          if [ "$sync_value" = "yes" ]; then
+            sync_result="synchronized"
+            echo "Clock sync observed system time synchronization"
+            break
+          fi
+
+          now_up="$(uptime_seconds)"
+          if [ "$((now_up - sync_start_up))" -ge "$sync_wait_seconds" ]; then
+            sync_result="timeout"
+            echo "Clock sync wait reached; allowing boot to continue with NTPSynchronized=''${sync_value:-unknown}"
+            break
+          fi
+
+          sleep 1
+        done
+      else
+        sync_result="timedatectl-unavailable"
+        echo "Clock sync could not check time synchronization: timedatectl unavailable"
+      fi
+
+      write_sync_state
+      touch "$synced_file"
+      chmod 0644 "$synced_file"
+    '';
+  };
 
   ghafClockJumpWatcher = pkgs.writeShellApplication {
     name = "ghaf-clock-jump-watcher";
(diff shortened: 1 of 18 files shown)


milva-unikie:
- Booted a new image, ran `sudo fss-test` in all VMs -> passed
- Rebooted from power menu, ran `sudo fss-test` again in all VMs -> fails in a few VMs every time

Example output from host
```
[ghaf@ghaf-host:~]$ sudo fss-test
==========================================
  FSS (Forward Secure Sealing) Test Suite
==========================================
[INFO] Test 1: Checking journal-fss-setup service...
[PASS] journal-fss-setup service completed successfully
[INFO] Test 2: Checking FSS sealing key...
[PASS] FSS sealing key exists at /var/log/journal/b70260906cc14221a040735cf279e2fc/fss
[INFO] Key directory (from fss-config): /persist/common/journal-fss/ghaf-host
[INFO] Test 3: Checking verification key...
[PASS] Verification key exists at /persist/common/journal-fss/ghaf-host/verification-key
[INFO] Test 4: Checking initialization sentinel...
[PASS] Initialization sentinel exists at /persist/common/journal-fss/ghaf-host/initialized
[INFO] Test 5: Running journal verification...
[FAIL] Journal verification FAILED [BAD_MESSAGE,ARCHIVED_SYSTEM] (archived system journal failures outside allowlist)
   Output: PASS: /var/log/journal/b70260906cc14221a040735cf279e2fc/system@43611ae217c9456d8e72142eead4f22a-0000000000000001-000652eea36148a4.journal
677fa0: Epoch sequence not continuous (2 vs 2)
File corruption detected at /var/log/journal/b70260906cc14221a040735cf279e2fc/system@43611ae217c9456d8e72142eead4f22a-0000000000000700-000652eea4ee3f2a.journal:6782880 (of 8388608 bytes, 80%).
FAIL: /var/log/journal/b70260906cc14221a040735cf279e2fc/system@43611ae217c9456d8e72142eead4f22a-0000000000000700-000652eea4ee3f2a.journal (Bad message)
PASS: /var/log/journal/b70260906cc14221a040735cf279e2fc/system@43611ae217c9456d8e72142eead4f22a-00000000000023d5-000652eefe9dfcf9.journal
=> Validated from Fri 2026-05-29 09:31:27 EEST to Fri 2026-05-29 09:45:00 EEST, final 0 entries not sealed.
557488: Epoch sequence not continuous (2 vs 2)
File corruption detected at /var/log/journal/b70260906cc14221a040735cf279e2fc/system@43611ae217c9456d8e72142eead4f22a-0000000000002c4e-000652ef012f22b0.journal:5600392 (of 8388608 bytes, 66%).
FAIL: /var/log/journal/b70260906cc14221a040735cf279e2fc/system@43611ae217c9456d8e72142eead4f22a-0000000000002c4e-000652ef012f22b0.journal (Bad message)
PASS: /var/log/journal/b70260906cc14221a040735cf279e2fc/system.journal
=> Validated from Fri 2026-05-29 10:30:05 EEST to Fri 2026-05-29 10:45:00 EEST, final 0 entries not sealed.
647620: Epoch sequence not continuous (5 vs 5)
File corruption detected at /var/log/journal/b70260906cc14221a040735cf279e2fc/system@43611ae217c9456d8e72142eead4f22a-0000000000003d4e-000652ef2ceec764.journal:6583840 (of 8388608 bytes, 78%).
FAIL: /var/log/journal/b70260906cc14221a040735cf279e2fc/system@43611ae217c9456d8e72142eead4f22a-0000000000003d4e-000652ef2ceec764.journal (Bad message)
[INFO] Test 6: Checking verification timer...
[PASS] journal-fss-verify.timer is active
   Next run: Fri 2026-05-29
[INFO] Test 7: Checking audit rules...
[PASS] FSS audit rules are configured

==========================================
  FSS Test Suite Complete
==========================================
  Passed:  6
  Failed:  1
  Warned:  0

Some tests failed. FSS may not be working correctly.
```
I can collect more logs if needed.

milva-unikie:
**Tested on Lenovo X1** (new image)

Fss-test is still failing after reboot. Did the same steps as last time, now with Lenovo X1.

- Booted a new image, ran `sudo fss-test` in all VMs -> passed
- Rebooted from power menu, ran `sudo fss-test` again in all VMs -> fails in a few VMs every time

Example output from comms-vm [errors.log](https://github.com/user-attachments/files/29196174/errors.log)

milva-unikie:
**Tested on Lenovo X1 and Darter Pro** (rebased to main)

All good now! Could not reproduce fss-test failing.

Failures still happen on Orins, but I think we can handle that as a separate issue.

packages/pkgs-by-name/fss-triage/package.nix
```diff
@@ -0,0 +1,587 @@
...
+      set -e
+
+      append_command_log "$name" "$rc" "$@"
+      return 0
+    }
+
+    run_shell_capture() {
+      local name="$1"
+      local script="$2"
+      run_capture "$name" "$BASH" -lc "$script"
```
brianmcgillion: **[security / command injection — confirmed]** `run_shell_capture` runs `$BASH -lc "$script"`, and the call sites below (`:483-485`) string-interpolate on-disk state **inside single quotes** in `$script` — e.g. `'$STATE_DIR'`, `'$KEY_DIR/initialized'`, `'$VERIFY_KEY_PATH'`, `'$FSS_CONFIG'`. `KEY_DIR` is `cat "$FSS_CONFIG"` where `FSS_CONFIG=/var/log/journal/<machine-id>/fss-config` (root-writable); a single quote in that content breaks out of the quoting and injects into the generated bash. Triage runs as root so impact is bounded (root→root), but it's a real injection footgun in a diagnostic tool.

**Fix:** don't interpolate runtime values into the `bash -lc` string — pass them as positional args and reference `"$1"…` inside a script the outer shell never expands, e.g.:
```sh
run_shell_capture() {
  local name="$1" script="$2"; shift 2
  run_capture "$name" "$BASH" -c "$script" bash "$@"
}
# call site: run_shell_capture "fss-state-files" '...loop over "$1" "$2"...' "$STATE_DIR" "$KEY_DIR" ...
```

modules/common/services/power.nix
```diff
@@ -1007,23 +1031,27 @@ in
...
+                          # During ghaf-rebuild switch activation, affected system
+                          # VMs may be restarted as part of host service changes.
+                          # That path runs inside a named detached activation unit;
+                          # only skip guest poweroff for that narrow case. Ordinary
+                          # manual stops/restarts must still use graceful shutdown.
                           jobs=$(${getExe' pkgs.systemd "systemctl"} list-jobs 2>/dev/null || true)
-                          if ! echo "$jobs" | grep -qiE '(sleep|suspend|poweroff|reboot|halt)\.target.*start'; then
-                            echo "No shutdown/suspend in progress, skipping graceful shutdown for '${vmName}' (likely a rebuild)"
+                          if ! echo "$jobs" | grep -qiE '(sleep|suspend|poweroff|reboot|halt)\.target.*start' \
+                            && ${getExe' pkgs.systemd "systemctl"} is-active --quiet ghaf-rebuild-switch.service; then
```
brianmcgillion: **[correctness — confirmed]** Both new ExecStop scripts decide "rebuild-restart vs real shutdown" via `is-active --quiet ghaf-rebuild-switch.service`. That unit is created **only** by the ghaf-build-helper remote switch path (`systemd-run --unit=ghaf-rebuild-switch …`). So a plain local `nixos-rebuild switch` / `switch-to-configuration switch` is classified as a *real shutdown* → every affected guest takes the full QMP guest power-off + restart instead of a fast SIGTERM (noticeably slower, and through the gui-vm path can drive a host power-off), a behavior change from `main`'s looser "no shutdown/suspend job queued ⇒ restart" test.

**Fix:** detect activation generically rather than by one hard-coded unit name — e.g. retain the base "no shutdown/suspend/poweroff/reboot job queued ⇒ treat as restart" heuristic, and/or have any activation context set a shared marker unit that both the remote and local switch paths use.

modules/common/logging/common.nix
```diff
@@ -85,40 +295,109 @@ let
...
 
-        printf '%s\n' "$archive_path" >> "$recovery_archives_file"
+        if [ -f "$recovery_receipts_file" ] \
+          && awk -F '\t' -v p="$archive_path" -v i="$inode" -v s="$size" \
+            '$2 == p && $3 == i && $4 == s { found = 1 } END { exit found ? 0 : 1 }' \
+            "$recovery_receipts_file"; then
+          return 0
+        fi
+
+        printf 'v1\t%s\t%s\t%s\t%s\t%s\t%s\tclock-jump-recovery\t%s\n' \
```
brianmcgillion: **[duplication / correctness-adjacent]** This `record_recovery_receipt` (in `ghaf-journal-alloy-recover`) hand-prints the v1 receipt TSV with the `v1` and `clock-jump-recovery` literals hardcoded, and `common.nix` defines its own `valid_sha256`/`current_boot_id` too — a **second, independent writer** of the same on-disk format that `fss.nix record_lifecycle_receipt` also writes and the verifier content-binds. The two must stay byte-compatible but can silently drift (the reader never validates the schema version).

**Fix:** add a single `fss_write_receipt <file> <path> <reason> [level] [msg]` to `fss-verify-classifier.sh` (co-located with `FSS_RECEIPT_SCHEMA_VERSION` and the reader), and have both `fss.nix` and `common.nix` call it — which requires `common.nix` to `source` the shared library (it currently doesn't). Part of the shared-library refactor in the review summary.

modules/common/logging/common.nix
```diff
@@ -14,6 +14,212 @@ let
...
+
+  ghafClockReady = pkgs.writeShellApplication {
+    name = "ghaf-clock-ready";
+    runtimeInputs = with pkgs; [
+      coreutils
+      gawk
+      systemd
+    ];
+    text = ''
+      stable_seconds="${toString recCfg.clockReady.stableSeconds}"
```
brianmcgillion, line 33: **[correctness / boot-time — confirmed]** `ghaf-clock-ready` waits for `stableSeconds` of non-decreasing plausible realtime — **default 20s** — up to `maxWaitSeconds` (90). Because it is `Before=systemd-journal-flush.service` (which is `Before=sysinit.target`) and `systemd-journal-flush` `Requires=ghaf-clock-ready`, this delays `sysinit.target` (≈ the rest of userspace) by ~20s **on every boot** of an FSS-enabled host (worst case ~90s). This is gated to FSS hosts, but it's a substantial, easily-missed boot-time cost that the tests don't exercise (they override `stableSeconds=1`, `maxWaitSeconds=5`).

Two asks: (1) confirm the 20s production default is intended and the FSS host boot budget tolerates it (and add a test at the real default, or a fast path when the clock is already sane at boot). (2) `systemd-journal-flush Requires=ghaf-clock-ready` makes this best-effort barrier script a **hard dependency** for persistent journal flush — an IO error in it (e.g. the anchor write) dependency-fails the flush and logs stay volatile for that boot. Consider `Wants=` instead of `Requires=` for the journal-flush edge (keep `Requires` only where fail-closed is intended).

brianmcgillion (changes requested):
Deep review of the FSS reliability hardening. Verified locally at this PR head (`a0a394c1`): `checks.x86_64-linux.{fss-triage,fss-test}` **build**, `pre-commit` (shellcheck/shfmt/statix/deadnix/reuse) **passes**, and the FSS host config **evaluates** — all green. (Note: my pre-scan hypothesis that `fss-triage` would fail SC2329 **did not materialize** — the check builds cleanly.)

The fail-closed activation chain and the test suite are strong. The main things to address:

**Two real defects (inline):** a command-injection footgun in `fss-triage` (`$BASH -lc` with root-controlled values interpolated), and rebuild-detection keyed on a single unit name so a local `nixos-rebuild switch` power-cycles guests.

**A default posture change to document:** `Seal=` now defaults to `no` and is activated at runtime after the clock barrier (per-boot pre-activation window); the docs didn't mention it.

**Severe shell duplication (your headline emphasis):** the 723-line classifier is `readFile`-inlined into 4 derivations (+ `common.nix` re-implements receipt writing), the effective-`Seal` awk parser exists **9×**, there are two drifting receipt writers, 4 identical `prune_*`, 18 inline `boot_id` reads, and `fss.nix` is a ~450-line Bash program in a Nix string.

Two safe, self-contained fixes — remove a dead alias, and add the runtime-activation + local-root receipt-forgery caveats to the FSS "does not protect against" doc. Verified: treefmt clean, eval green, applies clean.

```diff
diff --git a/docs/src/content/docs/ghaf/scs/fss.mdx b/docs/src/content/docs/ghaf/scs/fss.mdx
index 726b0a21..e0eeaef8 100644
--- a/docs/src/content/docs/ghaf/scs/fss.mdx
+++ b/docs/src/content/docs/ghaf/scs/fss.mdx
@@ -202,6 +202,17 @@ systemctl show journal-fss-setup --property=ConditionResult
 - Deletion of entire journal files (mitigated by remote log forwarding)
 - Real-time tampering before sealing (mitigated by short seal intervals)
 - Denial of service against logging
+- Local-root forgery of the on-host verification verdict. The setup and verify services trust
+  unauthenticated plaintext "receipt" and activation-state files under
+  `/var/log/journal/<machine-id>/` (root-owned, mode `0644`). A local-root attacker who tampers a
+  sealed archive can append a matching receipt line (its recorded `sha256` equals the tampered
+  content) to convert the on-host verdict from `AUDIT_LOG_INTEGRITY_FAIL` into a silent pass.
+  Authoritative tamper detection therefore requires **offline** `journalctl --verify --verify-key=`
+  against an exported archive using an off-host verification-key backup (see
+  [Offline Verification](#offline-verification)), which this attack cannot defeat.
+- The per-boot pre-activation window. Sealing is now activated at runtime, after the
+  clock-readiness barrier, rather than statically at journald start, so entries logged before
+  activation completes on each boot are collected but not yet FSS-trusted.
 
 ### Audit Integration
 
diff --git a/modules/common/logging/fss.nix b/modules/common/logging/fss.nix
index ba1ba749..0c9121bd 100644
--- a/modules/common/logging/fss.nix
+++ b/modules/common/logging/fss.nix
@@ -705,10 +705,6 @@ let
           && [ "$(activation_state_boot_id)" = "$(current_boot_id)" ]
       }
 
-      activation_state_active_current_boot() {
-        activation_state_record_current_boot
-      }
-
       activation_boundary_complete_current_boot() {
         activation_state_record_current_boot \
           && boot_baseline_current

```

### Recommended generalization (design; needs the `logging-fss` nixosTest for runtime sign-off)
Promote `modules/common/logging/fss-verify-classifier.sh` to **the** installed FSS shell library:
- add `environment.etc."fss-verify-classifier.sh".source = ./fss-verify-classifier.sh;` once in `fss.nix`;
- in the fss.nix setup/verify bodies, `fss-triage`, and `fss-test`, replace the `readFile`-interpolated copy with `source /etc/fss-verify-classifier.sh`, and **delete the three `excludeShellChecks=[SC2329]`** (they only exist because each consumer inlines the whole library);
- hoist the shared primitives into that library — `fss_journald_effective_seal` (kills the 9× awk), `fss_write_receipt`/`fss_prune_receipts` (kills the two drifting writers + the 4 `prune_*`), `fss_current_boot_id` (kills the 18 inline reads) — and make **`common.nix` source it** so `ghaf-journal-alloy-recover` can't diverge on the receipt format;
- extract the fss.nix setup/verify bodies into named `writeShellApplication`s (as `ghaf-clock-ready`/`-sync` already are), shrinking `fss.nix` to wiring.

I held this (and the H1 fix and the S1 "cap receipt-excused FAIL at `warning`" hardening) back from the applied patch because they change runtime behaviour and — per the chosen review depth — I build/lint-verified but did **not** run the authoritative `logging-fss` nixosTest. Happy to deliver them as verified patches once that test is run.

### Also worth addressing
- **Security (low, within FSS's documented local-root limit):** the on-host verdict trusts unauthenticated root-writable receipts/activation-state files — a local-root tamper can be made to pass silently; offline `journalctl --verify --verify-key=` is the real backstop. Recommend capping any receipt-excused archived FAIL at `warning`.
- **Correctness (low):** system VMs now shut down via best-effort ACPI QMP instead of givc (torn-journal risk if a VM ignores ACPI); ~20s (`stableSeconds` default) `sysinit` boot delay every boot on FSS hosts; `systemd-journal-flush Requires=ghaf-clock-ready` makes the custom barrier a hard dep for persistent flush (consider `wants`).
- **Cohesion:** the PR bundles the FSS work with an unrelated `fix(power)` + `.gitignore`/locale chore; `base.nix`'s locale move is misfiled under the power commit. Base is 46 commits behind main (rebase). Consider splitting so the power/chore don't gate the FSS review.

brianmcgillion:
also needs a rebase as the baseline is very far behind. almost 50 patches

modules/common/logging/common.nix
```diff
@@ -250,16 +544,128 @@ in
...
+    ghaf.storagevm.directories = mkIf (clockReadyEnabled && config.ghaf.storagevm.enable) [
+      "/var/lib/ghaf/clock-ready"
+    ];
 
     systemd = {
+      targets.ghaf-clock-ready = mkIf clockReadyEnabled {
+        description = "Ghaf clock readiness barrier";
+        requires = [ "ghaf-clock-ready.service" ];
+        after = [ "ghaf-clock-ready.service" ];
+      };
```
everton-dematos, line 566: Should journal flushing hard-require this service? I was thinking that if `ghaf-clock-ready` fails because of a mount or write error, `systemd-journal-flush` will not run and the journal may remain only in /run.

modules/common/logging/common.nix
```diff
@@ -250,16 +544,128 @@ in
         type = types.int;
         default = 60;
       };
+
+      maxReceipts = mkOption {
+        description = "Upper bound on retained content-bound recovery archive receipts.";
+        type = types.int;
+        default = 64;
+      };
```
everton-dematos, line 517: Why is 64 a safe limit here? Should receipts be retained while their archive exists, or should the archive and receipts be removed together?

brianmcgillion (approved):

brianmcgillion:
just @everton-dematos comments left

everton-dematos (approved):

PR #1965, opened 2026-05-27
refactor(features/logging): refactor global features, add logging features

kajusnau:
### **Motivation**
Originally features were introduced with two main options - `enable` and `targetVms`.
Enabling a feature was described as "Whether the feature is available system-wide". Despite this, we still had some top-level global config options like `security.audit.enable` and `logging.enable`, which were enabled system-wide but were not considered features.

As a result, the host and appvm configs never made use of features and instead relied on the top-level global config options.

This refactor attempts to change this behavior

---

### **Implementation**

`lib.ghaf.features` now includes a new helper `isEnabled`, which allows checking if a given feature is enabled system-wide.

As a result, the host (and appvms) can now make use of the Features API. Note that a feature does not activate automatically on the host or appvms - the relevant module still has to opt in by calling `isEnabled` and wiring the result to the appropriate option.

**Limitations**
Enabling a feature system-wide now implicitly enables it for all appvms and the host, as long as those modules wire it up via `isEnabled`. This is generally a good idea for our use case, but there may be situations in the future where a certain feature may be needed on a specific sysvm, but not on the host or any appvms.

---

### **Alternative implementation**

As an alternative, and mainly to address the limitation introduced by the current implementation (see above), we could introduce one or two more options to the `feature` attr to allow configuring features for the host and appvms independently of the `targetVms` config:
- `enableOnHost` or `host` - Whether the feature is enabled on the host
- `enableAppVms` - Whether the feature is enabled on appvms

These would only be needed in the unusual case where a feature should be assigned to a sysvm via `targetVms` but should explicitly not apply to the host or appvms.

---

### Other changes:

- Introduce 3 new Ghaf features, which can be enabled/disabled per-vm:
   - `logging` - whether to enable log collection
   - `log-upload` - whether to enable log upload
   - `audit` - whether to enable the Linux audit system
- Update `features-api.mdx` and `adding-features.mdx` to document `isEnabled` and the two-tier enablement model

### Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [x] Improvement / Refactor

## Checklist

### Applicable Targets
- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [x] Lenovo X1 `x86_64`
- [x] Dell Latitude `x86_64`
- [x] System 76 `x86_64`

### Installation Method
- [ ] Other:

### Test Steps To Verify:

1. ...

### DIFF
--- a/modules/microvm/host/microvm-host.nix
+++ b/modules/microvm/host/microvm-host.nix
@@ -14,6 +14,12 @@
 }:
 let
   cfg = config.ghaf.virtualization.microvm-host;
+  globalConfig = config.ghaf.global-config;
+  # System-wide features
+  auditEnabled = lib.ghaf.features.isEnabled globalConfig "audit";
+  logUploadEnabled = lib.ghaf.features.isEnabled globalConfig "log-upload";
+  loggingEnabled = lib.ghaf.features.isEnabled globalConfig "logging";
+  timezoneEnabled = lib.ghaf.features.isEnabled globalConfig "timezone";
   inherit (lib)
     mkEnableOption
     mkIf
@@ -130,13 +136,12 @@ in
           usb-serial.enable = config.ghaf.profiles.debug.enable;
         };
         logging = {
+          enable = lib.mkDefault loggingEnabled;
           listener = {
-            address = lib.mkDefault config.ghaf.global-config.logging.listener.address;
-            port = lib.mkDefault config.ghaf.global-config.logging.listener.port;
+            address = lib.mkDefault globalConfig.logging.listener.address;
+            port = lib.mkDefault globalConfig.logging.listener.port;
           };
-          journalClient = {
-            inherit (config.ghaf.logging) enable;
-          };
+          journalClient.enable = lib.mkDefault logUploadEnabled;
         };
         common = {
           extraNetworking.hosts.ghaf-host = cfg.extraNetworking;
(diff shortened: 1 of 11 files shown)


modules/microvm/host/microvm-host.nix
```diff
@@ -149,13 +154,19 @@ in
             };
           };
         };
+        security = {
+          audit.enable = auditEnabled;
```
vadika, line 158: Missing `lib.mkDefault` — `mvp-user-trial.nix` sets `ghaf.security.audit.enable = false` at the same priority, so with the audit feature on (now true in both debug and release) this evaluates to conflicting definitions. All sysvm bases use `mkDefault` here; host should too.

modules/microvm/sysvms/adminvm-base.nix
```diff
@@ -173,17 +175,13 @@ in
       };
     };
 
-    services.timezone.enable = lib.mkDefault (
-      timezoneEnabled && globalConfig.platform.timeZone == null
-    );
+    services.timezone.enable = lib.mkDefault timezoneEnabled;
```
vadika, line 178: Removing `platform.timeZone` changes behavior: VMs previously inherited the host's static timezone when the timezone feature was off, now they fall back to UTC. Downstream-visible — should at least be called out in the PR description.

PR #1976, opened 2026-06-08
fix: cosmic lockscreen with fingerprint

jannepirskanen:
- separate logic between lock screen and initial login

### Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. Login with password, remove exististing finger prints fprintd-delete
2. Lock screen -> only password option should appear
3. fprintd-enroll 
4. Lock screen -> finger print option first and if 3 failures, password fallback appear

### DIFF
--- a/modules/desktop/graphics/login-manager.nix
+++ b/modules/desktop/graphics/login-manager.nix
@@ -73,10 +73,21 @@ in
     security.pam.services = {
       cosmic-greeter = {
         rules = {
+          account = {
+            # When homed auth was used, PAM_AUTHTOK holds the real password and
+            # systemd_home account succeeds → done. When fingerprint was used,
+            # PAM_AUTHTOK is unset and systemd_home fails → fall through to permit.
+            systemd_home.control = lib.mkForce "[success=done default=ignore]";
+            permit = {
+              enable = true;
+              control = "[success=done default=ignore]";
+              modulePath = "${pkgs.linux-pam}/lib/security/pam_permit.so";
+              order = 10900; # after systemd_home (10800), before unix (11000)
+            };
+          };
           auth = {
-            systemd_home.order = 11399; # Re-order to allow either password _or_ fingerprint on lockscreen
             unix.settings.use_first_pass = !config.ghaf.services.sssd.enable;
-            fprintd.args = [ "maxtries=3" ];
+            fprintd.args = [ "max-tries=3" "timeout=-1" ];
           };
         };
       };
@@ -84,9 +95,7 @@ in
         fprintAuth = false; # User needs to enter password to decrypt home on login
         rules = {
           auth = {
-            systemd_home.order = 11399; # Re-order to allow either password _or_ fingerprint on lockscreen
             unix.settings.use_first_pass = !config.ghaf.services.sssd.enable;
-            fprintd.args = [ "maxtries=3" ];
 
             # This should precede other auth rules e.g. pam_sss.so (pam module for SSSD)
             faillock_preauth = mkIf cfg.failLock.enable {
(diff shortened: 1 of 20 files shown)


modules/desktop/graphics/login-manager.nix
```diff
@@ -72,11 +72,29 @@ in
...
+          };
           auth = {
-            systemd_home.order = 11399; # Re-order to allow either password _or_ fingerprint on lockscreen
             unix.settings.use_first_pass = !config.ghaf.services.sssd.enable;
-            fprintd.args = [ "maxtries=3" ];
+            fprintd = {
+              args = [
+                "max-tries=3"
+                "timeout=-1"
+                "debug=1"
```
kajusnau: debug prints can be disabled
```suggestion
                "max-tries=3"
                "timeout=-1"
```

kajusnau: We can also discuss whether having the timeout set to -1 is good UX - should the fingerprint timeout at some point and revert to the password field?

modules/desktop/graphics/login-manager.nix
```diff
@@ -86,7 +104,6 @@ in
           auth = {
             systemd_home.order = 11399; # Re-order to allow either password _or_ fingerprint on lockscreen
             unix.settings.use_first_pass = !config.ghaf.services.sssd.enable;
-            fprintd.args = [ "maxtries=3" ];
```
kajusnau, line 89: Re-ordering of `systemd_home` can likely be removed, as fprintd is not enabled here.

kajusnau (changes requested):

modules/desktop/graphics/login-manager.nix
```diff
@@ -88,21 +87,14 @@ in
...
           auth = {
             unix.settings.use_first_pass = !config.ghaf.services.sssd.enable;
-            fprintd = {
-              args = [
-                "max-tries=3"
-                "timeout=-1"
-                "debug=1"
-              ];
-            };
+            fprintd.args = [ "maxtries=3" ];
```
kajusnau: This is incorrect as per [fprintd upstream](https://gitlab.freedesktop.org/libfprint/fprintd/-/blob/master/pam/pam_fprintd.c?ref_type=heads#L81), should be `max-tries`. The reason it works here is because the default `max-tries` value is 3, see [here](https://gitlab.freedesktop.org/libfprint/fprintd/-/blob/master/pam/pam_fprintd.c?ref_type=heads#L76)

Also we should be explicit about the `timeout` arg - I think it's good to explicitly set it to `-1` or `15`.
I'd vote for `-1`, meaning the fprint module will never timeout.
If it's 30 (the default) or 15 (common practice) then I think the user would often end up in a situation where they lock the session, walk away, come back and the fingerprint prompt is already gone and replaced by the password.

Maybe @milva-unikie has an opinion 😀

kajusnau (changes requested):
Please squash commits (`git reset HEAD~3; git commit -s -m ...`) and add testing instructions

modules/desktop/graphics/login-manager.nix
milva-unikie: I would also vote for no timeout, we expect the users not unlock their laptops right after locking them

jannepirskanen: it is commited now

kajusnau (approved):

kajusnau (changes requested):
Please squash the commits (`git reset HEAD~3; git commit -s -m ...`)

PR #1988, opened 2026-06-11
fix(cosmic): Orin devices COSMIC display working again

emrahbillur:
On latest COSMIC version bump the Orin devices failed to provide a display output anymore. We added ghaf user to video group and added corresponding environment variables to system as well as defining the display port device. Now both AGX and NX devices are providing Cosmic display output. 

Known limitations: 
1- The Display Port to HDMI converters do not work. You need a display with Display Port to get screen output for Orin devices.
2- After cosmic is enabled the Orin devices start suspending after some period of time requiring user to press power button to resume which is annoying. I suggest disabling of suspend for Orin devices.

### Type of Change

- [ ] New feature
- [X] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

No Jira tickets as there were no active Orin users that used COSMIC display.
## Checklist

### Applicable Targets

- [X] Orin AGX `aarch64`
- [X] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`

### Installation Method

- [X] Requires full re-installation
- [ ] Other:

### Test Steps To Verify:

1. Build an image and flash to USB/NVME/eEMMC or perform a nixos-rebuild boot to Orin NX or AGX device 
2. Reboot
3. Cosmic screen arrives

### DIFF
--- a/targets/nvidia-jetson-orin/flake-module.nix
+++ b/targets/nvidia-jetson-orin/flake-module.nix
@@ -91,6 +91,9 @@ let
       extraModules = commonModules;
       extraConfig = {
         reference.profiles.mvp-orinuser-trial.enable = true;
+        # Crucial for Orin devices to use the correct render device
+        # Also needs 'mesa' to be in hardware.graphics.extraPackages
+        graphics.cosmic.renderDevice = "/dev/dri/renderD128";
       };
     })
 
@@ -143,6 +146,9 @@ let
       extraModules = commonModules;
       extraConfig = {
         reference.profiles.mvp-orinuser-trial.enable = true;
+        # Crucial for Orin devices to use the correct render device
+        # Also needs 'mesa' to be in hardware.graphics.extraPackages
+        graphics.cosmic.renderDevice = "/dev/dri/renderD128";
       };
     })
   ];
--- a/modules/profiles/orin.nix
+++ b/modules/profiles/orin.nix
@@ -132,7 +132,7 @@ in
       graphics.cosmic = {
         # Crucial for Orin devices to use the correct render device
         # Also needs 'mesa' to be in hardware.graphics.extraPackages
-        renderDevice = "/dev/dri/renderD129";
+        renderDevice = lib.mkDefault "/dev/dri/renderD128";
         # Keep only essential applets for Orin devices
         topPanelApplets.right = [
           "com.system76.CosmicAppletInputSources"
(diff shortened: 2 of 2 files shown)


targets/nvidia-jetson-orin/flake-module.nix
```diff
@@ -91,6 +91,9 @@ let
       extraModules = commonModules;
       extraConfig = {
         reference.profiles.mvp-orinuser-trial.enable = true;
+        # Crucial for Orin devices to use the correct render device
+        # Also needs 'mesa' to be in hardware.graphics.extraPackages
+        graphics.cosmic.renderDevice = lib.mkForce "/dev/dri/renderD128";
```
kajusnau: We probably don't need to use `mkForce` here since you changed the `renderDevice` to use `mkDefault`, but not crucial.

kajusnau (changes requested):
Nice work!

Would prefer the commit message to be something like
`fix(cosmic/orin): fix cosmic display on orin targets`

targets/nvidia-jetson-orin/flake-module.nix
emrahbillur: Sure, it will be better removal of mkForce.

emrahbillur: I removed mkForce

emrahbillur: > We probably don't need to use `mkForce` here since you changed the `renderDevice` to use `mkDefault`, but not crucial.

fixed now

emrahbillur:
> Nice work!
> 
> Would prefer the commit message to be something like `fix(cosmic/orin): fix cosmic display on orin targets`

commit name adjustted :)

kajusnau (approved):

emrahbillur:
The Orin NX is now also working with DP to HDMI adapter. 
All icons inactive as there is no GUIvm. Power menu works. Several extra options work.

leivos-unikie:
Tested on Orin AGX, booted from USB SSD, native image downloaded from the pre-merge artifacts

GUI... long time no see

As Emrah explained many Desktop options don't do anything (because no gui-vm) but at least power menu options works and power options also in login screen work

No regression found.

Some notes (no need to fix in this PR):
- if logging out, cannot log in anymore
- If locking, cannot unlock
- on suspends display turns off, serial connection and ssh is lost
- interestingly Orin AGX can wake up by pressing the power button (furthest from the power led), then serial connection and ssh are recovered. However, display does not wake up.

milva-unikie:
**Tested on Orin-NX**, both native and cross-compiled images

Same as AGX, GUI works but there are still many things to fix in future development.

Apps can be launched from ghaf-host with `ghaf-open` even though app icons do not work.

brianmcgillion (approved):

PR #1996, opened 2026-06-17
fix(cosmic-player): Add needed plugins

jannepirskanen:
## Description of Changes
Needed plugins and path added for cosmic-player to start and able to play video files.
Changes are done in cosmic-player overlay.

### Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. from UI: select mp4 file and play with cosmic-player isolated or double click the mp4 file.
2. from gui-vm command prompt: xdg-open test.mp4

### DIFF
--- a/overlays/custom-packages/cosmic/cosmic-player/default.nix
+++ b/overlays/custom-packages/cosmic/cosmic-player/default.nix
@@ -1,0 +1,14 @@
+# SPDX-FileCopyrightText: 2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# Add missing plugins and path
+
+{ prev }:
+(prev.cosmic-player.overrideAttrs (oldAttrs: {
+  dontWrapLibcosmicApp = true;
+
+  postFixup = oldAttrs.postFixup or "" + ''
+    wrapLibcosmicApp "$out/bin/cosmic-player" \
+      --prefix GST_PLUGIN_SYSTEM_PATH_1_0 : "$GST_PLUGIN_SYSTEM_PATH_1_0"
+  '';
+}))
--- a/overlays/custom-packages/default.nix
+++ b/overlays/custom-packages/default.nix
@@ -13,6 +13,7 @@
   cosmic-greeter = import ./cosmic/cosmic-greeter { inherit prev; };
   cosmic-initial-setup = import ./cosmic/cosmic-initial-setup { inherit prev; };
   cosmic-osd = import ./cosmic/cosmic-osd { inherit prev; };
+  cosmic-player = import ./cosmic/cosmic-player { inherit prev; };
   cosmic-reader = import ./cosmic/cosmic-reader { inherit prev; };
   cosmic-settings = import ./cosmic/cosmic-settings { inherit prev; };
   cosmic-settings-daemon = import ./cosmic/cosmic-settings-daemon { inherit prev; };


kajusnau (changes requested):
This is not a good approach - we don't want to add GST plugins to the global (media-vm) environment, overriding any potential default values set for `GST_PLUGIN_SYSTEM_PATH_1_0`

I recommend an overlay for cosmic-player:

```nix
{ prev }:
(prev.cosmic-player.overrideAttrs (oldAttrs: {
  dontWrapLibcosmicApp = true;

  postFixup = oldAttrs.postFixup or "" + ''
    wrapLibcosmicApp "$out/bin/cosmic-player" \
      --prefix GST_PLUGIN_SYSTEM_PATH_1_0 : "$GST_PLUGIN_SYSTEM_PATH_1_0"
  '';
}))
```
Notably, this is already kind of done [upstream](https://github.com/NixOS/nixpkgs/blob/d66a65dd4ae3b6530db3f00a4edd977959528919/pkgs/by-name/co/cosmic-player/package.nix#L70) in the `postInstall` phase, but clearly it doesn't run properly.
With this overlay, `wrapLibcosmicApp` will run explicitly to wrap the final binary with the given var.

This is still not perfect - we may run into issues where upstream packaging is updated to include more vars in the wrapping, but we explicitly disable it here in favor of our own manual wrapping.
We should try to figure out why the upstream wrapping via `libcosmic-app-hook.sh fixupOutputHooks+=(libcosmicAppWrapHook)` is not working. @avnik Maybe you have an idea?

kajusnau:
> we did check this earlier, i think problem is corelements.so that only come from gstreamer.out, so can we add this andother missing in overlay too

This is a different approach to before. I've tested this to work on my end, please try it.

kajusnau (changes requested):
Don't forget to add this to `overlays/custom-packages/default.nix`, otherwise this overlay won't do anything:
```nix
cosmic-player = import ./cosmic/cosmic-player { inherit prev; };
```

kajusnau:
FYI:
Looks like the core issue (not just for cosmic-player, but for almost all cosmic apps in nixpkgs) might have been caused by this [commit](https://github.com/NixOS/nixpkgs/commit/b45f95514de9d7742dab4de8bb3678ae8e761d7b)

It seems that `libcosmic-app-hook.sh` fails to wrap the binaries due to running on the `-debug` prefix outputs first, which then blocks the hook from executing on subsequent outputs (the real output).
I will address this in an upstream PR in nixpkgs.

kajusnau (approved):
Approving ahead of testing results, which may be delayed due to release testing

milva-unikie:
**Tested on Darter Pro and Lenovo X1**

Used new images because after rebuilding this feature did not work properly
- Opening videos to gui-vm Cosmic Media Player works.
- Opening videos to media-vm Ghaf Isolated Media Player works at least most of the time. I saw some issues when opening a video immediately after closing the previous one, but we can address that separately if it becomes a problem for users.

brianmcgillion (approved):

PR #1998, opened 2026-06-18
feat(access-control): givc access control module

gngram:
## Description of Changes

This PR introduces a framework for managing distributed Access Control Lists (ACLs) within the givc messaging system across Ghaf OS virtual machines. Because the admin-vm acts as a central proxy forwarding requests between various VMs, this implementation allows individual VMs to explicitly define their own access control profiles, which are then aggregated into the global admin policy as well as individual policy.

The access control feature is now disabled by default.

**Bug Fix:** The issue in the EventProxy ACL rule generation has been resolved in [PR #2128](https://github.com/tiiuae/ghaf/pull/2128) (ready for review).

**Enablement:** Access control is explicitly enabled in [PR #2129](https://github.com/tiiuae/ghaf/pull/2129).

### Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets
https://jira.tii.ae/browse/SSRCSP-8276

## Checklist

### Applicable Targets

- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. Test every functionality which requires inter vm communication e.g.
- Download a pdf file in chrome vm and open it.
- Watch video on youtube.

2. Check log of givc agent in each vm there should not be any permission denied
```bash
$journalctl -b 0 -u givc-<vmname>.service | grep -i cedar
```
3. Check log of givc admin in admin vm there should not be any permission denied
```bash
$journalctl -b 0 -u givc-admin.service | grep -i cedar
```


lib/global-config.nix
```diff
@@ -78,7 +78,7 @@ rec {
 
       givc = {
         enable = mkEnableOption "GIVC (Ghaf Inter-VM Communication) globally";
-
+        withAccessControl = mkEnableOption "Enable access control for GIVC globally";
```
slakkala: `mkEnableOption` already prepends `Enable ` to the argument.

modules/givc/common.nix
```diff
@@ -52,6 +52,32 @@ in
...
     debug = mkEnableOption "GIVC debug mode";
+    accessControl = {
+      enable = mkEnableOption "Enable access control for GIVC";
+      commonAdminRules = mkOption {
+        description = ''
+          Defines access control policies for the GIVC Admin server.
+          Direct agent-to-agent communication is generally discouraged,
+          so the Admin server primarily functions as a proxy routing requests
+          between agents. This option defines admin rules, which a non admin VM can push to the admin.
+          Example Proxy Rule:
```
slakkala: Does it make sense to have `example` in the description, when there is an explicit `example` field?

modules/reference/appvms/google-chrome.nix
```diff
@@ -109,6 +109,15 @@ in
...
                   givc.policyClient.enable = true;
+                  givc.accessControl.commonAdminRules = [
+                    {
+                      from = [ "chrome-vm" ];
+                      to = config.ghaf.common.appHosts;
+                      permittedRequests = [
+                        "StartApplication"
+                      ];
+                    }
+                  ];
```
slakkala: chrome-vm should be allowed to use the Ctap interfaces for yubikey.

modules/givc/common.nix
```diff
@@ -52,6 +52,32 @@ in
   options.ghaf.givc = {
     enable = mkEnableOption "gRPC inter-vm communication (GIVC)";
     debug = mkEnableOption "GIVC debug mode";
+    accessControl = {
+      enable = mkEnableOption "Enable access control for GIVC";
+      commonAdminRules = mkOption {
+        description = ''
```
everton-dematos, line 58: Is there any field that is mandatory when defining a rule?

modules/givc/common.nix
everton-dematos: This is an example of an "allow" rule. Can I define a "deny" rule as well? Or is it "denied" by default?

milva-unikie:

Did a quick check and found these issues. More thorough testing is still needed after these have been fixed.

1. `xdg-open` has issues, at least in comms-vm. `sudo WAYLAND_DISPLAY=wayland-1 xdg-open /tmp/test_pdf_comms-vm.pdf` from comms-vm failed with
```
Error: cedar: permission denied by admin access control policy
```
2. `givc-cli update` has issues. `sudo givc-cli update cachix --cache ghaf-release intel-laptop-debug` in host failed with
```
Error: cedar: permission denied by admin access control policy
```
3. Yubikey does not work in chrome-vm.
4. Ghaf Control Panel does not work. VMs and services don't load at all. About page is missing Ghaf version, Secure Boot status and Disk Encryption status.
<details>
<summary>Screenshots from Control Panel</summary>
Screenshot_2026-07-01_11-57-03
Screenshot_2026-07-01_11-57-16
</details>

lib/global-config.nix
gngram: Done!

modules/givc/common.nix
gngram: Removed.

modules/reference/appvms/google-chrome.nix
gngram: Now, it's allowed in appvm if ctap proxy is enabled.

gngram:
> **Tested on Darter Pro** (new image)
> 
> Did a quick check and found these issues. More thorough testing is still needed after these have been fixed.
> 
>     1. `xdg-open` has issues, at least in comms-vm. `sudo WAYLAND_DISPLAY=wayland-1 xdg-open /tmp/test_pdf_comms-vm.pdf` from comms-vm failed with
> 
> ```
> Error: cedar: permission denied by admin access control policy
> ```
STATUS>>>> Added rule to allow apps in comms vm. Works fine.

> 
>     2. `givc-cli update` has issues. `sudo givc-cli update cachix --cache ghaf-release intel-laptop-debug` in host failed with
> 
> 
> ```
> Error: cedar: permission denied by admin access control policy
> ```
STATUS>>>> Added rule to allow it. 

>     3. Yubikey does not work in chrome-vm.

STATUS>>>> Added rule to allow ctap in appvms if ctap proxy is enabled. Should works fine.

>     4. Ghaf Control Panel does not work. VMs and services don't load at all. About page is missing Ghaf version, Secure Boot status and Disk Encryption status.
> 
STATUS>>>> Added necessary rules to allow requests made by ghaf-ctrls-panel application. Works fine.
> Screenshots from Control Panel

modules/givc/common.nix
gngram: at least one field should be defined.

modules/givc/common.nix
gngram: By default it denies everything.

milva-unikie:
> > 2. `givc-cli update` has issues. `sudo givc-cli update cachix --cache ghaf-release intel-laptop-debug` in host failed

> STATUS>>>> Added rule to allow it.

This does not working, same error. In my original comment I accidentally said host, the correct place to run this command is gui-vm.

>  >  4. Ghaf Control Panel does not work. VMs and services don't load at all. About page is missing Ghaf version, Secure Boot status and Disk Encryption status.

> STATUS>>>> Added necessary rules to allow requests made by ghaf-ctrls-panel application. Works fine.

The issues I mentioned were fixed. VM services are now listed but VM Controls within them do not work (This option is purposefully disabled for some system-vms, but should work for app-vms).

I would like to see these issues properly fixed before I do the full testing.

gngram:
> > > 2. `givc-cli update` has issues. `sudo givc-cli update cachix --cache ghaf-release intel-laptop-debug` in host failed
> 
> > STATUS>>>> Added rule to allow it.
> 
Updated the rule for gui-vm, When I run the command from gui-vm now it works fine.

> This does not working, same error. In my original comment I accidentally said host, the correct place to run this command is gui-vm.
> 
> > > 4. Ghaf Control Panel does not work. VMs and services don't load at all. About page is missing Ghaf version, Secure Boot status and Disk Encryption status.
> 
> > STATUS>>>> Added necessary rules to allow requests made by ghaf-ctrls-panel application. Works fine.
> 
> The issues I mentioned were fixed. VM services are now listed but VM Controls within them do not work (This option is purposefully disabled for some system-vms, but should work for app-vms).
> 
> I would like to see these issues properly fixed before I do the full testing.

Now, I can stop, pause, and start app VMs.

milva-unikie:
**Tested on Darter Pro**

All good now.

- Previously mentioned issues have been fixed
- Full automated regression tests passed
- No issue found in quick manual regression test

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,81 @@
...
+description: Ghaf access control architecture
+---
+
+# Ghaf Access Control
+
+Ghaf Inter-VM Communication (GIVC) includes an access control mechanism to secure and restrict communication between virtual machines and the host. In the backend, the high-level Nix access control configurations defined by users are automatically transformed and compiled into **Cedar policies** to enforce fine-grained security rules at runtime.
+
+## 1. Architectural Overview
+
+The ACL system operates on a default-deny principle (except for the built-in `ServerReflectionInfo`  ). Before executing any RPC method, the gRPC interceptor/middleware evaluates the request against the injected Cedar policies. If it is  allowed it forwards the request to underline layer otherwise it returns "permission denied error". Here are the building blocks and control flow diagram for access control enforcement in GIVC admin and agents.
```
kajusnau: Not sure if this is rendering properly for docs, but there's a few instances of double whitespaces here.

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+### Building Blocks
+
+This flow shows how a request moves from a Source VM through a access control module implemented in interceptor/middleware before reaching the gRPC service. The system checks the request details against security policies to either allow or deny the request.
```
kajusnau: "...through **_an_** access control module..."

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+- **Actor:**
+
+  - **Source VM:** The source vm initiating a gRPC request.
+
+- **Core Processes:**
+
+  - **[Parser] Context Extraction:** The `Request Parser` extracts complete metadata from the request header and payload and maps to the Cedar objects  (*Principal*, *Resource*, *Action*, and *Context*) required for authorization request.
```
kajusnau: Wording could be improved here, especially for this part:
> extracts complete metadata from the request header and payload and maps to the Cedar objects

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
kajusnau: I think it would also be beneficial to mention what Cedar is and why we use it here. (for example I have no idea 😀)

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+  - **[Cedar Authorizer] Policy Verification:** The `Cedar Authorizer` evaluates the request against the active `Cedar Policies`.
+
+  - **Gatekeeping:** If **Allowed**, the request drops into the `givc Admin Service` and returns an execution `Status`. If **Denied**, it triggers an instant `Permission Denied` error.
+
+  In Admin and Agents both evaluate request using Cedar authorizer. Since Admin accepts requests for only admin givc module and it also works as proxy between two VMs, so more parameters are added in the authorization requests e.g. RPC method and target VM(as Cedar context).
```
kajusnau: > In Admin and Agents both evaluate request using Cedar authorizer

Something is off with the wording here or I'm misreading it - would it be better to have something like:

> The Admin and Agents both evaluate the request using Cedar authorizer.

P.S.: What is "Admin" and "Agents" here? Might be worth to explicitly state it

kajusnau: > Since Admin accepts requests for only admin givc module and it also works as proxy between two VMs, so more parameters are added in the authorization requests e.g. RPC method and target VM(as Cedar context)

Wording can be improved. Suggestion (feel free to adjust):

> Since the Admin accepts requests only for the admin gIVC module, and it also works as a proxy between two VMs, more parameters are needed in the authorization requests e.g. RPC method and target VM (as Cedar context)

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+                          Flow diagram in GIVC Admin
```
kajusnau: We should decide on a consistent naming of givc, as all 3 are used randomly:
- GIVC
- gIVC
- givc

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,81 @@
...
+- **`givc.adminvm.accessControl.adminRules`**
+
+  - **Description:** Defines fine-grained configuration rules applied directly to the Admin VM to control service consumption and traffic routing.
+
+> **Note:** For specific field definitions within these rules (such as `from`, `to`, `permittedRequests`, and `permittedModules`), please check the module option documentation for the latest schema definitions.
+
+## 3. Default Configurations Derived in Code
+
+Ghaf provides a secure baseline right out of the box. It automatically sets up default ACL rules for each VM profile. These rules are created based on the active settings. The system also includes specific rules to meet core functional requirements.
+In GIVC module admin and agents derive these rules based on the capabilities enabled for them in the respective VM for example if in host exec module is enabled it derives rule to allow access of exec module to admin VM.
```
kajusnau: This sounds like a run-on sentence to me, a bit difficult to digest.
Also grammar can be improved, but I won't offer a suggestion as I'm not too certain of the implementation and what the example here is supposed to convey.

lib/default.nix
```diff
@@ -254,6 +254,44 @@ in
...
+    adminRulesType = lib.types.submodule {
+      options = {
+        from = lib.mkOption {
+          type = lib.types.listOf lib.types.str;
+          default = [ ];
+          description = ''
+            A list of source VM identities (callers) permitted to initiate a request listed in
+            the 'permittedRequests' option. This defines who is originating the request to admin.
+          '';
+        };
```
kajusnau, line 262: Could use an example list

Should the description be simplified to focus on what and why needs to be specified? E.g.:

> List of (IMO 'source' can be omitted here) VMs (callers) permitted to initiate a request defined in the 'permittedRequests' option.

I think the original second sentence `This defines who is originating the request to admin.` adds nothing useful to the description.

lib/default.nix
```diff
@@ -254,6 +254,44 @@ in
...
+        to = lib.mkOption {
+          type = lib.types.listOf lib.types.str;
+          default = [ ];
+          description = ''
+            A list of allowed destination VMs for the requests specified in `permittedRequests`.
+            The Admin server is the default consumer unless a destination VM is specified in the request at runtime.
+            - Proxy access: To allow forwarding to a target VM, list the target VM here (e.g., `[ "app-vm" ]`).
+            - Consumer access only: Leave this empty (`[ ]`) to restrict the rule so that the Admin only processes the request itself.
+          '';
+        };
```
kajusnau, line 270: Same as above, could use an example list.

Also I think the description is a bit wordy, and can be simplified (assuming I understand the implementation correctly):

> List of allowed destination VMs for the requests specified in `permittedRequests`.
If left empty, all requests in `permittedRequests` will be sent to the Admin.
If a destination VM or VMs are specified, all requests in `permittedRequests` will be proxied through the Admin to the destination VM.

lib/default.nix
```diff
@@ -254,6 +254,44 @@ in
...
+            A list of specific gRPC methods that the callers listed in `from` are allowed to execute.
+
+            Example Proxy Rule:
+              `from = [ "gui-vm" ]; to = [ "app-vm" ]; permittedRequests = [ "StartApplication" ];`
+              (Allows `gui-vm` to ask the Admin to start an application on `app-vm`)
+
+            Example Consumer Rule:
+              `from = [ "gui-vm" ]; to = [ ]; permittedRequests = [ "RegisterService" ];`
+              (Allows `gui-vm` to call `RegisterService` directly on the Admin server)
+          '';
```
kajusnau, line 276: The examples here are nice - we can expand by adding the `example` attr to all 3 options to match one of the examples.

modules/givc/guivm.nix
```diff
@@ -36,12 +36,52 @@ in
...
+        ];
+      }
+      {
+        from = [ config.networking.hostName ];
+        permittedRequests = [
+          "StartApplication"
+          "PauseApplication"
+          "ResumeApplication"
+          "StopApplication"
+        ];
```
kajusnau: Why is `gui-vm` permitted to start/stop applications on the admin-vm?
Isn't this missing a `to` list, e.g.:
```nix
to = [ "media-vm" "flatpak-vm" ... ];
```

Or, would it make more sense to add this list from the respective VMs, e.g. in flatpak app-vm definition add an ACL to allow gui-vm to start flatpak apps.

modules/reference/desktop/applications.nix
```diff
@@ -81,5 +81,18 @@ in
...
+        from = [ "gui-vm" ];
+        permittedRequests = [
+          "StartVM"
+          "Watch"
+          "GetStats"
+          "QueryList"
+          "Sysinfo"
+        ];
+      }
+    ];
```
kajusnau: This file is generally kept pretty tidy with just some applications, nothing else
Can we, and should we, move this to `modules/givc/guivm.nix`? Not a strong opinion, just an idea

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+  In Admin and Agents both evaluate request using Cedar authorizer. Since Admin accepts requests for only admin givc module and it also works as proxy between two VMs, so more parameters are added in the authorization requests e.g. RPC method and target VM(as Cedar context).
+
+                          Flow diagram in GIVC Admin
+
+  ![Admin Access Control Flow](./../../../../../assets/ghaf/access-control-admin.png)
+
+                           Flow diagram in GIVC Agents
+
+  ![Admin Access Control Flow](./../../../../../assets/ghaf/access-control-agent.png)
```
kajusnau:
```suggestion
  ![Agent Access Control Flow](./../../../../../assets/ghaf/access-control-agent.png)
```

lib/default.nix
kajusnau, line 270: Also, is it correct to assume that if `to` is left empty, then the requests are handled by Admin, and therefore the Host? Or are we avoiding the term "Host" here on purpose

kajusnau (changes requested):
I think this can use a clean up and some simpler explanations for some things.
Terminology should be kept consistent and lean, if possible, especially in the docs.

Regarding commits, is there a real need for 6? **fix** type commits seem redundant in a PR that adds a new feature - the feature should just work, **fix** PRs (if any) should be done after, if issues are spotted.
IMO we could do with 3 (not necessarily in this order):
- bump givc
- add docs
- implement ACLs

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Rendering looks good, removed extra spaces.
https://github.com/gngram/ghaf/blob/pull/givc-access-control-module/docs/src/content/docs/ghaf/overview/arch/access-control.mdx

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: done

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Done. Please review again.

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Included a section for the design decision also. Please review.

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Done. I assume if any one is reading this doc, he/she will be aware of Ghaf design so to keep the document small avoided to put any information about agent and admin.

lib/default.nix
gngram: Removed the examples. I avoided to use example option here as it uses other options like `from` and `to` also.

modules/givc/guivm.nix
gngram: Done

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Done.

lib/default.nix
gngram: Yes. Avoided 'Host' word as consumer is givc-admin.service doesn't matter where it runs.

gngram:
> I think this can use a clean up and some simpler explanations for some things. Terminology should be kept consistent and lean, if possible, especially in the docs.
> 
> Regarding commits, is there a real need for 6? **fix** type commits seem redundant in a PR that adds a new feature - the feature should just work, **fix** PRs (if any) should be done after, if issues are spotted. IMO we could do with 3 (not necessarily in this order):
> 
>     * bump givc
> 
>     * add docs
> 
>     * implement ACLs

Squashed commits in three parts:
- acl implementation
- acl rules for VMs
- acl documentation

modules/reference/desktop/applications.nix
gngram: These rules are required for control panel. Conditional inclusion of this rule is not straight forward in `gui-vm` as the option `config.ghaf.reference.desktop.applications` is not directly available in `gui-vm`

kajusnau (approved):
I'm unblocking the PR as I'll be away on vacation starting tomorrow.

Despite the approval, I think some issues still remain:
- Some of my previous comments are still not addressed
- commits and commit messages are incomplete or incorrect - we should follow conventional commits and use the tags `feat` and `docs`. the commit messages are very short, they look incomplete to me
- Quite a few grammar issues in the added doc file. I won't go through them one by one, but this is probably a good place for an LLM to clean it up (without changing the content of course) :saluting_face: 
- This PR likely needs to be re-tested due to the changes in gui-vm ACL rules, specifically launching app-vms from the GUI

gngram:
> I'm unblocking the PR as I'll be away on vacation starting tomorrow.
> 
> Despite the approval, I think some issues still remain:
> 
>     * Some of my previous comments are still not addressed
Few things in documentation I didn't include e.g. description of agent and admin to keep it short and simple. Please let me know if I missed any thing else. 
 
>     * commits and commit messages are incomplete or incorrect - we should follow conventional commits and use the tags `feat` and `docs`. the commit messages are very short, they look incomplete to me
> 
Did proper tagging of commit messages.

>     * Quite a few grammar issues in the added doc file. I won't go through them one by one, but this is probably a good place for an LLM to clean it up (without changing the content of course) 🫡
> 
Done.

>     * This PR likely needs to be re-tested due to the changes in gui-vm ACL rules, specifically launching app-vms from the GUI

Re-based to latest main. Ready to test.

milva-unikie:
**Tested on Darter Pro**

VM controls in Control Panel are broken again, `Pause` and `Shutdown` do not work. Logs from admin-vm.
```
2026-07-16 09:26:13.329 cedar: authorization denied for source vm = gui-vm, module = admin, and grpc service = AdminService, (rpc method: PauseApplication) 
2026-07-16 09:26:23.328 cedar: authorization denied for source vm = gui-vm, module = admin, and grpc service = AdminService, (rpc method: StopApplication) 
```

milva-unikie:
**Tested on Darter Pro and Lenovo X1**

All good now. VM controls work. Automated regression tests passed on both laptops.

lib/global-config.nix
```diff
@@ -78,7 +78,7 @@ rec {
 
       givc = {
         enable = mkEnableOption "GIVC (Ghaf Inter-VM Communication) globally";
-
+        withAccessControl = mkEnableOption "access control for GIVC globally";
```
mbssrc: I don't think this needs to be surfaced in global config - like TLS, it is a core security feature that should be enabled by default, with an opt-out in the givc configuration.

modules/givc/netvm.nix
```diff
@@ -41,6 +41,16 @@ in
     # Configure netvm service
     givc.sysvm = {
       enable = true;
+      accessControl = {
```
brianmcgillion: **[security / fail-open — verified by eval]** This `accessControl` block sets `agentRules` but omits `enable`, which every peer module sets. So `net-vm`'s `givc.sysvm.accessControl.enable` stays at the upstream default `false`, and its Cedar interceptor is never installed — **all RPCs to the net-vm agent are allowed even when access control is enabled**. net-vm is the network-facing VM, so this is the worst place to be unprotected. Root cause is the copy-pasted per-type block (generalization A). Minimal fix:

```suggestion
      accessControl = {
        enable = config.ghaf.givc.accessControl.enable;
```

modules/givc/adminvm.nix
```diff
@@ -74,12 +74,33 @@ in
...
+          }
+        ];
+      };
     };
 
     # Sysvm agent so admin-vm receives timezone/locale propagation
     givc.sysvm = {
       enable = true;
       inherit (config.ghaf.givc) debug;
+      accessControl = {
```
brianmcgillion: **[security / fail-open — verified by eval]** Same omission as net-vm: this admin-vm `givc.sysvm` agent sets `agentRules` but not `enable`, so it evaluates to `false` and runs with no Cedar interceptor. admin-vm is the ACL root-of-trust, so its co-located agent being unprotected is especially serious. (The admin *server* block above is correctly enabled — only this agent is not.) Minimal fix:

```suggestion
      accessControl = {
        enable = config.ghaf.givc.accessControl.enable;
```

modules/givc/guivm.nix
```diff
@@ -36,12 +36,53 @@ in
...
+        from = [ config.networking.hostName ];
+        permittedRequests = [
+          "Suspend"
+          "Reboot"
+          "Poweroff"
+        ];
+      }
+      # Microvm services for apps, executed by admin
+      {
+        from = [ config.networking.hostName ];
```
brianmcgillion: **[security / over-broad — verified]** Removing `to = config.ghaf.common.appHosts` here did not narrow the rule — in the givc Cedar generator an **empty `to` imposes no destination restriction** (`definitions.nix:207` emits the target condition only `optionals (rule.to != [])`). So gui-vm can now proxy `StartApplication`/`Pause`/`Resume`/`Stop` to **any VM**, not just app hosts. Restore the destination guard:

```suggestion
        from = [ config.networking.hostName ];
        to = config.ghaf.common.appHosts;
```

lib/default.nix
```diff
@@ -254,6 +254,36 @@ in
...
+        };
+        to = lib.mkOption {
+          type = lib.types.listOf lib.types.str;
+          default = [ ];
+          description = ''
+            A list of allowed destination VMs for the requests specified in `permittedRequests`.
+            The Admin server is the default consumer unless a destination VM is specified in the request at runtime.
+            - Proxy access: To allow forwarding to a target VM, list the target VM here (e.g., `[ "app-vm" ]`).
+            - Consumer access only: Leave this empty (`[ ]`) to restrict the rule so that the Admin only processes the request itself.
```
brianmcgillion: **[docs / correctness — verified]** This bullet is backwards. In the Cedar generator an empty `to` produces **no** destination condition (permits any target); a non-empty `to` is what restricts. The current wording ("empty ⇒ Admin only processes it itself") is the footgun that led to the gui-vm rule above being over-broadened.

```suggestion
            - Unrestricted destination: Leave this empty (`[ ]`) to impose NO destination check — the rule permits the request regardless of target (the Admin may consume it, or forward it to any VM named at runtime). Listing targets in `to` is what *restricts* forwarding to those VMs.
```

modules/givc/common.nix
```diff
@@ -52,6 +52,29 @@ in
   options.ghaf.givc = {
     enable = mkEnableOption "gRPC inter-vm communication (GIVC)";
     debug = mkEnableOption "GIVC debug mode";
+    accessControl = {
+      enable = mkEnableOption "Enable access control for GIVC";
```
brianmcgillion: **[nit + fail-open default]** `mkEnableOption` already prepends "Whether to enable ", so this renders as "Whether to enable Enable access control…". Drop the leading "Enable":

```suggestion
      enable = mkEnableOption "access control for GIVC";
```

More importantly (**fail-open default**): this flag defaults to `false`, and disabling access control *removes* the interceptor (allow-all) rather than applying an empty deny policy — so a profile that enables GIVC but omits `withAccessControl` silently ships zero enforcement, with no diagnostic. The `git apply` patch in the review summary adds a build-time **assertion** (every enabled agent must enforce ACL when the feature is on) that deterministically catches this and the two fail-open omissions; consider also a `warnings` entry for `givc.enable && !accessControl.enable`.

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+To configure or modify access control, the following module options are available:
+
+- **`ghaf.global-config.givc.withAccessControl`**
+  - **Description:** A global boolean toggle to enable or disable the GIVC access control mechanism across the entire system (the host and all VMs).
+- **`ghaf.givc.accessControl.commonAdminRules`**
+  - **Description:** A list of top-level access rules defining what individual VMs are authorized to send to, or proxy through, the central Admin VM.
+- **`givc.appvm.accessControl.agentRules`**
+  - **Description:** Defines fine-grained local rules for application VM agents, specifying which external VMs can directly access local service modules (such as `systemd` or `stats`).
+- **`givc.adminvm.accessControl.adminRules`**
```
brianmcgillion: **[docs — verified]** No such option: the namespace is `givc.admin`, not `givc.adminvm` (see `modules/givc/adminvm.nix`: `givc.admin = { … accessControl.adminRules … }`). A reader copying this hits an undefined-option error.

```suggestion
- **`givc.admin.accessControl.adminRules`**
```

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+> **Note:** For specific field definitions within these rules (such as `from`, `to`, `permittedRequests`, and `permittedModules`), please check the module option documentation for the latest schema definitions.
+
+## 3. Default Configurations Derived in Code
+
+Ghaf provides a secure baseline right out of the box by automatically setting up default ACL rules for each VM profile based on its active settings. The system then dynamically injects specific rules to meet core functional requirements.
+
+For instance, the GIVC module automatically derives these rules from the capabilities enabled within each respective VM. As an example, if the exec module is enabled on a host VM, the configuration engine automatically generates a rule allowing the Admin VM to access that specific execution boundary.
+
+By default, the framework maintains a strict zero-trust policy, blocking all inter-VM communication channels unless a module configuration explicitly acts as a capability toggle to auto-derive an allowed path.
```
brianmcgillion: **[docs — verified]** This overstates the guarantee: access control is **opt-in** and **fail-open when disabled** (with `withAccessControl` off, GIVC installs no interceptor and permits everything). Scope the "default-deny / zero-trust by default" framing to the enabled state:

```suggestion
Access control is opt-in (`ghaf.global-config.givc.withAccessControl`). When it is enabled, the framework maintains a strict zero-trust policy, blocking all inter-VM communication channels unless a module configuration explicitly acts as a capability toggle to auto-derive an allowed path. When it is disabled, GIVC installs no interceptor and all inter-VM requests are permitted (fail-open).
```

modules/reference/appvms/business.nix
```diff
@@ -296,6 +296,15 @@ in
 
               # Enable Policy Client
               givc.policyClient.enable = true;
+              givc.accessControl.commonAdminRules = [
```
brianmcgillion: **[duplication / generalization]** This `StartApplication` admin rule is copy-pasted verbatim — only `from` differs — in `business.nix`, `comms.nix`, and `google-chrome.nix`. Since `modules/microvm/sysvms/appvm-base.nix` already emits `givc.accessControl.commonAdminRules` (the yubiProxy `Ctap` rule) and already binds both `vmName` and `config.ghaf.common.appHosts`, emit this rule there once for every app-VM (gated by a per-app opt-out for any VM that must not launch apps), and delete these three per-file copies.

modules/microvm/sysvms/netvm.nix
```diff
@@ -78,6 +78,7 @@ in
             inherit (cfg.evaluatedConfig.config.ghaf.security.spire.agent) nodeAttestationMode workloads;
           };
         };
+        adminAccessControl = cfg.evaluatedConfig.config.ghaf.givc.accessControl.commonAdminRules or [ ];
```
brianmcgillion: **[generalization]** This `adminAccessControl = …commonAdminRules or []` export is repeated identically in all 5 sysvm registry files (adminvm/audiovm/guivm/netvm/idsvm), alongside the pre-existing `policies`/`spire.agents`/`extraNetworking` exports that are *also* repeated per file. `modules/microvm/appvm.nix` already aggregates the app-VMs centrally over `enabledVms`; a matching sysvm-side aggregator over `config.ghaf.virtualization.microvm.sysvm.enabledVms` could emit this whole export cluster once — and guarantee a new export can't be forgotten in one VM (the same failure mode as the fail-open bug).

brianmcgillion (changes requested):
**Review + fix patches for the GIVC access-control framework.** Verified locally at this PR head (`f9bab366`): `nix eval` passes, and a **behavioral eval** of `lenovo-x1-carbon-gen11-debug` (which sets `withAccessControl=true`) surfaced a real **fail-open bug** — `net-vm` and `admin-vm` agents evaluate `accessControl.enable=false` while every peer is `true`, so their Cedar interceptor is never installed (allow-all). The fixes below flip both to enforced; the assertion passes and the image rebuilds green.

The **7 inline `suggestion` blocks are one-click**: the two fail-open fixes, the gui-vm `to` restore, the `to` doc correction, the `admin` option-path fix, the default-deny doc scoping, and the `mkEnableOption` text.

**Apply everything at once** — the 7 suggestions **plus** the fail-open guardrail assertion in `givc/common.nix` (whose target line isn't in this diff) — with `git apply`:

```diff
diff --git a/docs/src/content/docs/ghaf/overview/arch/access-control.mdx b/docs/src/content/docs/ghaf/overview/arch/access-control.mdx
index c68e839a..4135bcb5 100644
--- a/docs/src/content/docs/ghaf/overview/arch/access-control.mdx
+++ b/docs/src/content/docs/ghaf/overview/arch/access-control.mdx
@@ -9,7 +9,7 @@ Ghaf Inter-VM Communication (GIVC) includes an access control mechanism to secur
 
 ## 1. Architectural Overview
 
-The ACL system operates on a default-deny principle (except for the built-in `ServerReflectionInfo`). Before executing any RPC method, the gRPC interceptor/middleware evaluates the request against the injected Cedar policies. If it is allowed, it forwards the request to the underlying layer; otherwise, it returns a "permission denied" error. Here are the building blocks and control flow diagram for access control enforcement in GIVC admin and agents.
+When access control is enabled, the ACL system operates on a default-deny principle (except for the built-in `ServerReflectionInfo`). Before executing any RPC method, the gRPC interceptor/middleware evaluates the request against the injected Cedar policies. If it is allowed, it forwards the request to the underlying layer; otherwise, it returns a "permission denied" error. Here are the building blocks and control flow diagram for access control enforcement in GIVC admin and agents.
 
 ### 1.1 GIVC Authorization: Why Cedar was chosen
 
@@ -72,7 +72,7 @@ To configure or modify access control, the following module options are availabl
   - **Description:** A list of top-level access rules defining what individual VMs are authorized to send to, or proxy through, the central Admin VM.
 - **`givc.appvm.accessControl.agentRules`**
   - **Description:** Defines fine-grained local rules for application VM agents, specifying which external VMs can directly access local service modules (such as `systemd` or `stats`).
-- **`givc.adminvm.accessControl.adminRules`**
+- **`givc.admin.accessControl.adminRules`**
   - **Description:** Defines fine-grained configuration rules applied directly to the Admin VM to control service consumption and traffic routing.
 
 > **Note:** For specific field definitions within these rules (such as `from`, `to`, `permittedRequests`, and `permittedModules`), please check the module option documentation for the latest schema definitions.
@@ -83,4 +83,4 @@ Ghaf provides a secure baseline right out of the box by automatically setting up
 
 For instance, the GIVC module automatically derives these rules from the capabilities enabled within each respective VM. As an example, if the exec module is enabled on a host VM, the configuration engine automatically generates a rule allowing the Admin VM to access that specific execution boundary.
 
-By default, the framework maintains a strict zero-trust policy, blocking all inter-VM communication channels unless a module configuration explicitly acts as a capability toggle to auto-derive an allowed path.
+Access control is opt-in (`ghaf.global-config.givc.withAccessControl`). When it is enabled, the framework maintains a strict zero-trust policy, blocking all inter-VM communication channels unless a module configuration explicitly acts as a capability toggle to auto-derive an allowed path. When it is disabled, GIVC installs no interceptor and all inter-VM requests are permitted (fail-open).
diff --git a/lib/default.nix b/lib/default.nix
index 9bbc5ee3..bf14e5a7 100644
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -271,7 +271,7 @@ in
             A list of allowed destination VMs for the requests specified in `permittedRequests`.
             The Admin server is the default consumer unless a destination VM is specified in the request at runtime.
             - Proxy access: To allow forwarding to a target VM, list the target VM here (e.g., `[ "app-vm" ]`).
-            - Consumer access only: Leave this empty (`[ ]`) to restrict the rule so that the Admin only processes the request itself.
+            - Unrestricted destination: Leave this empty (`[ ]`) to impose NO destination check — the rule permits the request regardless of target (the Admin may consume it, or forward it to any VM named at runtime). Listing targets in `to` is what *restricts* forwarding to those VMs.
           '';
         };
         permittedRequests = lib.mkOption {
diff --git a/modules/givc/adminvm.nix b/modules/givc/adminvm.nix
index d2aaedbb..f57f6ee7 100644
--- a/modules/givc/adminvm.nix
+++ b/modules/givc/adminvm.nix
@@ -92,6 +92,7 @@ in
       enable = true;
       inherit (config.ghaf.givc) debug;
       accessControl = {
+        enable = config.ghaf.givc.accessControl.enable;
         agentRules = [
           {
             permittedVms = [ config.givc.sysvm.network.admin.transport.name ];
diff --git a/modules/givc/common.nix b/modules/givc/common.nix
index f51f25a9..993e4354 100644
--- a/modules/givc/common.nix
+++ b/modules/givc/common.nix
@@ -53,7 +53,7 @@ in
     enable = mkEnableOption "gRPC inter-vm communication (GIVC)";
     debug = mkEnableOption "GIVC debug mode";
     accessControl = {
-      enable = mkEnableOption "Enable access control for GIVC";
+      enable = mkEnableOption "access control for GIVC";
       commonAdminRules = mkOption {
         description = ''
           Defines access control policies for the GIVC Admin server.
@@ -116,6 +116,20 @@ in
         assertion = cfg.debug -> (!config.ghaf.logging.enable);
         message = "Do not enable givc debug and logging simultaneously, you may leak private information.";
       }
+      {
+        # Guard against the fail-open-by-omission class: when access control is
+        # on, every GIVC service enabled on this VM must also enable its own
+        # accessControl, or its Cedar interceptor is never installed (allow-all).
+        assertion =
+          cfg.accessControl.enable
+          -> lib.all (svc: svc.enable -> svc.accessControl.enable) [
+            config.givc.sysvm
+            config.givc.appvm
+            config.givc.host
+            config.givc.admin
+          ];
+        message = "ghaf.givc.accessControl.enable is set but an enabled GIVC service on this VM left its own accessControl disabled (fail-open). Set givc.<type>.accessControl.enable = config.ghaf.givc.accessControl.enable.";
+      }
     ];
 
     # Build admin address here (inside config block) to defer hosts evaluation
diff --git a/modules/givc/guivm.nix b/modules/givc/guivm.nix
index 8e1fbc45..fb24e1aa 100644
--- a/modules/givc/guivm.nix
+++ b/modules/givc/guivm.nix
@@ -49,6 +49,7 @@ in
       # Microvm services for apps, executed by admin
       {
         from = [ config.networking.hostName ];
+        to = config.ghaf.common.appHosts;
         permittedRequests = [
           "StartApplication"
           "PauseApplication"
diff --git a/modules/givc/netvm.nix b/modules/givc/netvm.nix
index 73a9621b..95d1686c 100644
--- a/modules/givc/netvm.nix
+++ b/modules/givc/netvm.nix
@@ -42,6 +42,7 @@ in
     givc.sysvm = {
       enable = true;
       accessControl = {
+        enable = config.ghaf.givc.accessControl.enable;
         agentRules = [
           {
             permittedVms = [ config.givc.sysvm.network.admin.transport.name ];

```

### Duplication / over-complication / generalization (requested emphasis) — recommended refactors
Flagged inline; these are designs (can be delivered as verified patches on request):
- **A+B (recommended — subsumes the two enable fixes):** the per-type `accessControl` block (`enable` + the `"stats"` agentRule) is copy-pasted across 6 `modules/givc/*.nix` and was omitted in 2 (the fail-open bug). Set it **once** in `modules/givc/common.nix`, gated per active service, using `config.ghaf.givc.adminConfig.name`; delete the 6 copies. With the new assertion, the omission becomes impossible.
- **C:** the app-VM `StartApplication` rule is copy-pasted ×3 (`reference/appvms/{business,comms,google-chrome}.nix`, only `from` differs). Emit it once from `modules/microvm/sysvms/appvm-base.nix` (already emits the yubiProxy `Ctap` rule + knows `vmName`), gated by a per-app opt-out.
- **D:** the per-VM `ghaf.common = { … adminAccessControl … }` export is hand-repeated in 5 sysvm registry files; aggregate once over `sysvm.enabledVms` (mirroring `modules/microvm/appvm.nix`), which also DRYs the pre-existing `spire`/`extraNetworking` repetition.
- **Over-complication:** the rule list carries 3 names (`commonAdminRules`→`adminAccessControl`→`adminRules`) and the enable flag 3 names; gui-vm rules are split across 3 files.

Also recommend an integration test (`tests/givc-access-control/`, following `tests/logging/`) asserting admin **permits an allowed RPC and denies an un-permitted one** — the manual `journalctl` grep can't detect a disabled interceptor (a silent fail-open emits no cedar log line).

lib/global-config.nix
gngram: Removed global config.

modules/givc/netvm.nix
gngram: Fixed, now accessControl is enable in givc/common.nix

modules/givc/adminvm.nix
gngram: Fixed, now accessControl is enable in givc/common.nix

modules/givc/guivm.nix
gngram: Added `to` for `StartApplication`. StopApplication, PauseApplication, and ResumeApplication RPCs execute on admin vm only so `to` config is not required for these.

lib/default.nix
gngram: Done

modules/givc/common.nix
gngram: Done.

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Done

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Removed global config for access control, updated this accordingly.

modules/microvm/sysvms/netvm.nix
gngram: Done

modules/reference/appvms/business.nix
gngram: Tried to generalize it via permitStartApplication config. Any app VM can disable it. Please see appvm_base.nix.

gngram:
> Duplication / over-complication / generalization (requested emphasis) — recommended refactors
> Flagged inline; these are designs (can be delivered as verified patches on request):
> A+B (recommended — subsumes the two enable fixes): the per-type accessControl block (enable + the "stats" agentRule) is copy-pasted across 6 modules/givc/*.nix and was omitted in 2 (the fail-open bug). Set it once in modules/givc/common.nix, gated per active service, using config.ghaf.givc.adminConfig.name; delete the 6 copies. With the new assertion, the omission becomes impossible.

Done.

> C: the app-VM StartApplication rule is copy-pasted ×3 (reference/appvms/{business,comms,google-chrome}.nix, only from differs). Emit it once from modules/microvm/sysvms/appvm-base.nix (already emits the yubiProxy Ctap rule + knows vmName), gated by a per-app opt-out.

Done.

> D: the per-VM ghaf.common = { … adminAccessControl … } export is hand-repeated in 5 sysvm registry files; aggregate once over sysvm.enabledVms (mirroring modules/microvm/appvm.nix), which also DRYs the pre-existing spire/extraNetworking repetition.

Done.

> Over-complication: the rule list carries 3 names 
> (commonAdminRules→adminAccessControl→adminRules) and the enable flag 3 names; gui-vm rules are split across 3 files.

Fixed. Used a unified name `adminRules`

> Also recommend an integration test (tests/givc-access-control/, following tests/logging/) asserting admin permits an allowed RPC and denies an un-permitted one — the manual journalctl grep can't detect a disabled interceptor (a silent fail-open emits no cedar log line).

Added access control test.

Rebased to latest main branch.

modules/microvm/sysvms/appvm-base.nix
```diff
@@ -103,6 +103,12 @@ in
       readOnly = true;
       description = "The VM definition with merged applications. Exposed for host-side access.";
     };
+
+    permitStartApplication = lib.mkOption {
+      type = lib.types.bool;
+      default = vm.permitStartApplication or true;
```
brianmcgillion: **[privilege broadening — verified by eval]** Thanks for generalising this; the mechanism is right, but the default inverts the least-privilege posture the PR is establishing. `StartApplication → all appHosts` previously belonged to exactly 3 VMs (`business`, `comms`, `google-chrome`, one copy each). With `default = true` it now goes to all 5 enabled app-VMs — eval on `lenovo-x1-carbon-gen11-debug` produces `from=["media-vm"] to=["business-vm","chrome-vm","comms-vm","flatpak-vm","media-vm"]`.

`media-vm` is the odd one out: it has no `xdgitems` block and no `givc-cli start app` call site anywhere, so it gains cross-VM app-launch for nothing. The VMs that genuinely need it are the ones with `xdgitems.enable` — `business`, `comms`, `flatpak`, `chromium`, `google-chrome`.

An opt-out default also reproduces, in miniature, the failure mode this PR exists to remove: a new app-VM silently inherits a capability instead of silently missing one. Opt-in keeps the grant auditable at the vmDef.

```suggestion
      default = vm.permitStartApplication or false;
```

Note this suggestion is **not** complete on its own — it needs `permitStartApplication = true;` added to the vmDefs in `modules/reference/appvms/{business,comms,flatpak,chromium,google-chrome}.nix` to preserve current behaviour. (`vm` is the free-form `hostConfig.appvm` attrset, so the key is accepted there without a schema change.)

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
```diff
@@ -0,0 +1,84 @@
...
+- **Principal (`Source`)**: The identity of the calling VM, securely extracted from the TLS Subject Alternative Name (SAN).
+- **Resource (`Module`)**: The target gRPC service module (e.g., `systemd`, `admin`, `stats`).
+- **Action (`Command`)**: The specific gRPC method being invoked (e.g., `StartApplication`, `Reboot`, `RegisterService`).
+- **Context (`context`)**: The gRPC protobuf request payload, converted to a Cedar record (e.g., checking `context.VmName` or `context.UnitName`). **The context is evaluated only for unary requests; it is completely ignored in streaming requests.**
+
+## 2. Core Access Control Options
+
+To configure or modify access control, the following module options are available:
+
+- **`ghaf.givc.accessControl.commonAdminRules`**
```
brianmcgillion: **[docs — verified]** Stale option name: this PR renames `commonAdminRules` to `adminRules` (`modules/givc/common.nix:57`), and every consumer now uses the new name (`guivm.nix:40`, `host.nix:35`, `appvm-base.nix:280`). `ghaf.givc.accessControl.commonAdminRules` no longer exists, so a reader copying this hits an undefined-option error — same failure mode as the `givc.adminvm` bullet you already fixed below.

```suggestion
- **`ghaf.givc.accessControl.adminRules`**
```

modules/microvm/sysvm-registry.nix
```diff
@@ -64,5 +64,60 @@ in
...
 
-  config.ghaf.virtualization.microvm.sysvm.enabledVms = lib.filterAttrs (_: vm: vm.enable) cfg.vms;
+  config =
+    let
+      sysvms = cfg.enabledVms;
+    in
+    {
+      ghaf.virtualization.microvm.sysvm.enabledVms = lib.filterAttrs (_: vm: vm.enable) cfg.vms;
+
+      ghaf.common = {
```
brianmcgillion, line 74: **[cleanup / rebase drift]** This aggregator is the right shape and correctly replaces the per-VM export blocks in `adminvm.nix`, `audiovm.nix`, `guivm.nix`, `idsvm.nix` and `netvm.nix` — but `modules/microvm/sysvms/gpuvm.nix:69-82` still carries its own copy (it landed on `main` after this branch was cut, so the rebase didn't reconcile it):

```nix
ghaf.common = {
  extraNetworking.hosts.${vmName} = cfg.extraNetworking;
  policies = lib.mkIf ... ;
  spire.agents = lib.mkIf ... ;
};
```

Since `gpuvm` is registered in `sysvm.vms`, both this aggregator and `gpuvm.nix` now define `ghaf.common.{extraNetworking.hosts,policies,spire.agents}."gpu-vm"`. It evaluates today only because the duplicate definitions happen to be equal-mergeable — I confirmed `nvidia-jetson-orin-agx-debug` evals clean and lists `gpu-vm` exactly once in each attrset. But it's redundant, and a future divergence between the two would surface as a confusing conflict error rather than as drift.

Suggest deleting that block from `gpuvm.nix` as part of this PR, so generalization D covers all six sysvms and there's no per-VM copy left to fall out of sync.

targets/vm/flake-module.nix
```diff
@@ -206,6 +206,7 @@ let
                 ];
 
                 givc = {
+                  accessControl.enable = lib.mkForce false;
```
brianmcgillion, line 217: **[question]** Why does this target need access control off? There's no comment, and the new assertion in `modules/givc/common.nix:118-132` can't catch anything here — its implication is vacuous when `accessControl.enable` is false — so this target ships GIVC agents with no Cedar interceptor (allow-all) and nothing flags it.

If it's a deliberate dev-only carve-out, a short comment saying so would keep the next reader from copying the pattern into a shipped target. If instead it's working around a policy gap (some RPC the `vm` target needs that the derived rules don't permit), that gap is worth fixing here rather than disabling enforcement wholesale — it would likely affect the real targets too once they exercise the same path.

brianmcgillion (changes requested):
## Retest of the previous review — all 9 items fixed, 4 new (minor) findings

Re-verified at head `0dfd343c` (was `f9bab366`). I checked each item in the code rather than relying on the thread replies, and re-ran the toolchain locally.

**Everything from the previous review is genuinely fixed.** In particular the blocking fail-open (H1) is resolved the preferred way — generalization A, centralising `givc.{sysvm,appvm,host,admin}.accessControl.enable` in `modules/givc/common.nix` — and you went further by removing the global opt-in toggle, so access control is now unconditionally on rather than opt-in. Confirmed by eval, not just by reading:

- every GIVC agent on `lenovo-x1-carbon-gen11-debug` (net-vm, admin-vm agent + admin server, gui-vm, audio-vm, host, all 5 app-VMs) evaluates to `accessControl.enable = true`;
- the aggregated admin ruleset resolves to 11 rules with the expected `from`/`to`/`permittedRequests`.

Two calls I want to explicitly agree with, having checked upstream `ghaf-givc` rather than assuming:

- Leaving `PauseApplication`/`ResumeApplication`/`StopApplication` without a `to` is **correct**. `crates/admin/src/admin/server.rs:621-673` resolves the target agent from `app_name` via the registry and never reads `VmName`, so a `to` clause there would deny those calls outright rather than restrict them.
- Restoring `to` on `StartApplication` is safe: every call site passes `--vm` (`guivm-base.nix:61`, `microvm/common/xdgitems.nix`, `packages/chrome-extensions/open-normal/src/open_normal.sh:60`), so the `context has "VmName"` condition is always satisfied.

The new integration test is a real one and it passes here — 4 subtests, with genuine Cedar denials in the log (`Cedar authorization denied action=StartApplication principal=host`, `action=GetUnitStatus principal=guivm`).

### Verification run locally
| Check | Result |
|---|---|
| `nix build .#checks.x86_64-linux.access-control-tests` | PASS (4/4 subtests) |
| `nix build .#packages.x86_64-linux.lenovo-x1-carbon-gen11-debug` | PASS |
| `nix build .#checks.x86_64-linux.pre-commit` | PASS |
| eval `lenovo-x1-carbon-gen11-debug` / `nvidia-jetson-orin-agx-debug` | PASS (no duplicate-definition conflict from the new aggregator) |

### New findings (4, all minor)
None of these are fail-open and none block the design. Keeping `CHANGES_REQUESTED` only so they get a pass before merge — items 1 and 2 have one-click suggestions, item 1 needs the companion edits listed in its comment.

1. `permitStartApplication` defaults to `true`, which broadens `StartApplication` from 3 app-VMs to all 5 — `media-vm` gains it despite having no caller.
2. `access-control.mdx` still documents `commonAdminRules`, which this PR renamed to `adminRules`.
3. `modules/microvm/sysvms/gpuvm.nix:69-82` was missed by generalization D (rebase drift).
4. `targets/vm/flake-module.nix` force-disables access control with no explanation.

modules/microvm/sysvms/appvm-base.nix
gngram: > Now permitStartApplication defaults to false.
> It is enabled in `business`, `comms`, `flatpak`, `chromium`, `google-chrome` VMs.

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Done.

modules/microvm/sysvm-registry.nix
gngram: Removed the block from `gpu-vm` and `disp-vm`.

targets/vm/flake-module.nix
gngram: Added a comment:
```
 # GIVC is enabled only on the host; it is disabled on all other target VMs.
 # Because of this, access control is not meaningful here. Additionally, the host
 # enables two GIVC agents (givc-host and givc-gui-vm), which causes an ACL file
 # path collision that needs to be resolved in GIVC itself.
 #
 # TODO: Fix the ACL file collision in GIVC and enable ACL here, even though it
 # provides no practical benefit in this configuration.
```

gngram:
Rebased.

brianmcgillion (approved):
Rechecked at `814c0314` (rebased onto `07273221`). All four items from my previous review are fixed — verified in the code and by eval rather than from the resolved-thread state. Approving.

## 1 — `permitStartApplication` default: fixed, verified behaviourally

Default is now `vm.permitStartApplication or false`, with the flag set in exactly the five VMs that have `xdgitems.enable`. `nix eval` of `ghaf.common.adminRules` on `lenovo-x1-carbon-gen11-debug` now gives `StartApplication` grants from `gui-vm`, `business-vm`, `chrome-vm`, `comms-vm`, `flatpak-vm` — **`media-vm` is no longer a `from`**, which was the point of the finding. The five additions sit in the vmDef attrset next to `yubiProxy`, which is read through the identical `vm.<key> or false` path, so the placement is right and needs no schema change.

## 2 — docs: fixed

`access-control.mdx:69` now documents `ghaf.givc.accessControl.adminRules`, matching `modules/givc/common.nix:57` and every consumer.

## 3 — export-block drift: fixed, and more than I asked

`gpuvm.nix` lost its per-VM block, and you also removed the equivalent from `dispvm.nix`, which only landed on `main` after my review — I hadn't flagged that one. Both now just register into `sysvm.vms` with `extraNetworking` and let the aggregator at `sysvm-registry.nix:112` do the work.

## 4 — `accessControl.enable = lib.mkForce false`: fixed

The comment answers the question substantively rather than just restating the code: GIVC is host-only on these VM targets, and the host runs two agents (`givc-host`, `givc-gui-vm`) that collide on the ACL file path. That's a real constraint, and the TODO points at the right place to fix it. Good enough to unblock — worth a follow-up issue against GIVC so the TODO doesn't become permanent.

## Verification

- `nix eval` of the aggregated `adminRules` on `lenovo-x1-carbon-gen11-debug`, as above.
- `nix build .#checks.x86_64-linux.pre-commit .#checks.x86_64-linux.access-control-tests` — both green. These built from source locally (30 derivations, nothing substituted), so the `givc-access-control` VM test genuinely executed rather than being fetched from the binary cache.
- No regressions in the PR-owned diff between `0dfd343c` and `814c0314`: restricted to the PR's own files it is +94/−17 and consists solely of these four fixes.

One process note: `jenkins-pre-merge` was still pending when I approved; the other 33 checks are green. My four items were all config-level and are covered by the eval and the ACL test above, so I didn't hold the approval for it — but please don't read this approval as a statement about the hardware lane.

Also worth correcting one thing from my own previous review, since it is in the thread for posterity: I wrote that `StartApplication` "previously belonged to exactly 3 VMs". That was the count of VMs carrying a hand-written copy of the rule, not the count that needed it — the five you enabled are the correct set, and match the list further down in that same comment.

leivos-unikie:
Triggered pre-merge https://ci-prod.vedenemo.dev/job/ghaf-pre-merge-manual/904/

Test case `Check logging rate` on Lenovo-X1 has catched unexpected systemd-coredump of chrome / outlook on business-vm (induced by by previous Start Outlook test) which caused log rate limit to be exceeded slightly.

This is probably just a random glitch. I will check Lenovo-X1 also manually.

Otherwise automated pre-merge testing is ok.

leivos-unikie:
Automated regression tests pass on lenovo-x1 and darter (exluding currently failing firewall flood). Checked also manually youtube and opening pdf file on lenovo-x1. No "permission denied" found in givc services. Checked that Ghaf Control Panel VM controls work on darter.

Orins don't not have GUI nowadays. No "permission denied" found in givc services on Orin AGX.

leivos-unikie:
Rerun the single failed test case of pre-merge (Check Grafana logs on cross-compiled Orin NX), PASSED:
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/7290/

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Done.

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Done

docs/src/content/docs/ghaf/overview/arch/access-control.mdx
gngram: Done.

lib/default.nix
gngram: Done

lib/default.nix
gngram: Done

mbssrc (approved):
Two things to look into:
- the appvm application start permission is broad and cannot be narrowed downstream without a mkForce override
- the givc EventProxy should not work (has this been tested?), it has a different mechanism compared to the socketproxy (producer/consumer flags with a single vm transport definition for server/dial purpose) so the ACL generation does not work correctly

gngram:
> Two things to look into:
> 
>     * the appvm application start permission is broad and cannot be narrowed downstream without a mkForce override
> 
Fixed: 
[Reference](https://github.com/tiiuae/ghaf/pull/1998/files#diff-822d2c013740b9d7605c0a91caa0f384594c23ea570c5290f660ad8eb4b0cf39R159)

>     * the givc EventProxy should not work (has this been tested?), it has a different mechanism compared to the socketproxy (producer/consumer flags with a single vm transport definition for server/dial purpose) so the ACL generation does not work correctly

Fixed in separate PR.
https://github.com/tiiuae/ghaf/pull/2128
It has dependency on `ghaf-givc`:
https://github.com/tiiuae/ghaf-givc/pull/472

gngram:
The access control feature is now disabled by default.
Bug Fix: The issue in the EventProxy ACL rule generation has been resolved in [PR #2128](https://github.com/tiiuae/ghaf/pull/2128) (ready for review).

Enablement: Access control is explicitly enabled in [PR #2129](https://github.com/tiiuae/ghaf/pull/2129).

**This PR is now safe to merge.**

brianmcgillion (approved):

PR #2012, opened 2026-06-30
feat(installer): improve speed, add TUI for encryption, improve DRY-ness

kajusnau:
## Description of Changes

- **`lib/gum-lib.sh`**:
   New shared TUI library with `gum` wrappers, eliminating duplication across installer, user-provision, and encryption scripts
- **First boot encryption**: 
   Replaced output with styled `gum` TUI
   Added interactive PIN setup with confirmation, QR-coded recovery key, and other UI/UX improvements
- **Flash script**:
   Multi-threaded `zstd` (≤8 cores), direct `bmaptool` fast-path when `.bmap` exists, up to 3 retries with device re-sync on failure
- **Installer**:
   Added `.bmap` alongside the image in the final ISO
   Installer now uses bmap when available with streaming dd fallback
   Installer speed is significantly improved when bmap is used

### Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Update
- [x] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

- [x] Clear summary in PR description
- [x] Detailed and meaningful commit message(s)
- [x] Commits are logically organized and squashed if appropriate
- [x] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [x] Author has added reviewers and removed PR draft status

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`
- [x] Intel Laptop (low mem) `x86_64`

### Installation Method

- [ ] Requires full re-installation
- [x] Can be updated with `nixos-rebuild ... switch`
- [ ] Other:

### Disk encryption TUI
1. Test a typical installer flow with and without disk encryption
2. Evaluate improvement :slightly_smiling_face: 
3. Verify no major errors or regressions in functionality when disk encryption is enabled
4. (Optional) Test non-automated disk encryption (requires modifying Ghaf config or using release variant)

### Installer speed
1. Install Ghaf using the installer ISO
2. Verify speed improvement when bmap is used (will be used by default)

### (Optional)
1. Test flash script for raw image - should be slightly faster

PR #2031, opened 2026-07-09
fix(flash): pass image to bmaptool in prebuilt-bmap path

vadika:
`bmaptool copy` is `copy [--bmap BMAP] IMAGE DEST`, but `flash_zst_with_bmap` calls it with `--bmap BMAP DEVICE` — no image. It always errors, retries 3x, then silently falls back to a full stream decompress, so the prebuilt `.bmap` is never used.

Add `$FILENAME` (a `.raw.zst`, which bmaptool reads directly), matching the sibling call below. Introduced in #2012.


kajusnau (approved):
Nice find!

The description is a bit too wordy and wrong, I think - this has nothing to do with the installer ISO.

Also would prefer to keep our PR description template 🙂 👍

brianmcgillion (approved):

PR #2035, opened 2026-07-10
feat(spire): support named agent instances and upstream integration

everton-dematos:
## Description of Changes
This PR adds optional integration with an external SPIRE backend while keeping the local SPIRE deployment independent. It introduces three main changes:

1. **Support named SPIRE agent instances.** Refactor the SPIRE agent module to allow multiple isolated agent instances on the same VM, each with its own configuration, systemd service, runtime directory, user, group, and Workload API socket. Existing agents are migrated to the `downstream` instance, while admin-vm receives an `upstream` instance that is disabled by default and can be configured to connect to an external SPIRE server.

2. **Remove join-token attestation.** Remove join-token configuration, plugins, token generation, and related fallback logic. SPIRE agents now  use `x509pop` as the supported node attestation method by default.

3. **Integrate the local server with the optional upstream agent.** Add trust-bundle URL support and systemd workload attestation to SPIRE agents. When the upstream agent is enabled, the local `spire-server.service` can access its Workload API and fetch an X.509-SVID issued by the external SPIRE server. The external server registers this workload using the `systemd:id:spire-server.service` selector. If the upstream agent is disabled or unavailable, the upstream identity fetch is skipped and the local SPIRE server and agents continue operating independently.

The local SPIRE server (i.e., `downstream`) remains its own authority, e.g. `ghaf.internal`, and the cloud SPIRE server (i.e., `upstream`) remains another authority. They exchange trust bundles. Local agents/workloads continue to get local identities even if the cloud server is down. 

The following Figure shows the SPIRE architecture when the `upstream` is enabled:
image

### Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Update
- [x] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets
https://jira.tii.ae/browse/SSRCSP-8653

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. Verify that the upstream agent is disabled by default:

   ```console
   systemctl status spire-agent-upstream.service
   ```

   The expected esult is:

   ```text
   Unit spire-agent-upstream.service could not be found.
   ```

2. Verify that the local SPIRE server is healthy and that the downstream agents have successfully completed `x509pop` node attestation:

   ```console
   sudo spire-server healthcheck \
     -socketPath /run/spire-server/api.sock \
     -verbose

   sudo spire-server agent list \
     -socketPath /run/spire-server/api.sock \
     -attestationType x509pop
   ```

   The health check should report:

   ```text
   Server is healthy.
   ```

   The agent list should report one attested agent for each VM:

   ```text
   Found 10 attested agents:

   SPIFFE ID        : spiffe://<LOCAL_TRUST_DOMAIN>/spire/agent/x509pop/<AGENT_FINGERPRINT>
   Attestation type : x509pop
   ```

3. The configuration of the external SPIRE server is outside the scope of this PR. It will require defining the appropriate CA and certificate provisioning model, trust bundle distribution method, and workload registration policy.

    For testing purposes, with help from Pavel (thanks!), we deployed a SPIRE Server on a backend/cloud machine configured with `x509pop` node attestation. The backend server was configured to trust the CA that signed the admin-vm certificate.

   For this validation, the `upstream` agent was enabled and configured with the backend server address, port, trust domain, and trust bundle. The backend also contained a workload registration entry for `spire-server.service`, associated with the upstream agent through the `systemd:id:spire-server.service` selector.

   The following backend log confirms that the external SPIRE server issued an X.509-SVID for the local SPIRE server workload:

   ```text
   Jul 10 14:51:52 nixos spire-server[1614]:
   time="2026-07-10T14:51:52+03:00" level=debug
   msg="Signed X509 SVID"
   authorized_as=agent authorized_via=datastore
   caller_addr="<ADMIN_VM_ADDRESS>:42720"
   caller_id="spiffe://example.org/spire/agent/x509pop/<AGENT_FINGERPRINT>"
   entry_id=<ENTRY_ID>
   expiration="2026-07-10T12:51:52Z"
   method=BatchNewX509SVID
   request_id=<REQUEST_ID>
   revision_number=0
   serial_num=243259129904215621232601805090303982465
   service=svid.v1.SVID
   spiffe_id="spiffe://example.org/ghaf/admin-vm/spire-server"
   subsystem_name=api
   ```

   The following admin-vm log confirms that the local SPIRE server retrieved the workload X.509-SVID through the upstream agent:

   ```text
   Jul 10 15:57:19 admin-vm spire-server-upstream-workload[11833]: 
   Fetched upstream SPIRE workload SVID for spire-server.service
   ```

### DIFF
--- a/modules/common/security/spire/agent.nix
+++ b/modules/common/security/spire/agent.nix
@@ -8,183 +8,309 @@
 }:
 with lib;
 let
-  cfg = config.ghaf.security.spire.agent;
-  runtimeDataDir = "/run/spire-agent";
-  dataDir = "${runtimeDataDir}";
-  credSourceDir = "/etc/givc";
-
   spire-package = config.ghaf.common.spire.package;
-  healthCheckPort = toString config.ghaf.common.spire.server.healthCheckPort;
 
-  joinTokenConf = optionalString (cfg.nodeAttestationMode == "join_token") ''
-    join_token_file = "${cfg.settings.join_token.token}"
-  '';
-  joinTokenPlugin = optionalString (cfg.nodeAttestationMode == "join_token") ''
-    NodeAttestor "join_token" {
-      plugin_data {}
+  serviceName = name: if name == "downstream" then "spire-agent" else "spire-agent-${name}";
+  runtimeDir = name: "/run/${serviceName name}";
+
+  agentType = types.submodule (
+    { name, ... }:
+    {
+      options = {
+        enable = mkEnableOption "SPIRE agent ${name}";
+
+        serverAddress = mkOption {
+          type = types.str;
+          default = config.ghaf.common.spire.server.address;
+          description = "SPIRE server address.";
+        };
+
+        serverPort = mkOption {
+          type = types.port;
+          default = config.ghaf.common.spire.server.port;
+          description = "SPIRE server agent port.";
+        };
+
+        serverHealthCheck = {
+          enable = mkOption {
+            type = types.bool;
+            default = name == "downstream";
+            description = "Wait for the SPIRE server readiness endpoint before starting.";
+          };
+
+          port = mkOption {
+            type = types.port;
+            default = config.ghaf.common.spire.server.healthCheckPort;
+            description = "SPIRE server readiness endpoint port.";
+          };
+        };
+
+        trustDomain = mkOption {
+          type = types.str;
+          default = config.ghaf.common.spire.server.trustDomain;
+          description = "SPIFFE trust domain.";
+        };
+
+        nodeAttestationMode = mkOption {
+          type = types.enum [
+            "join_token"
+            "x509pop"
+          ];
+          default = "x509pop";
+          description = "Node attestation mode.";
+        };
+
+        workloads = mkOption {
+          type = types.spireWorkloads;
+          default = [ ];
+          description = "List of workloads for this SPIRE agent.";
+        };
+
+        trustBundlePath = mkOption {
+          type = types.str;
+          default = "/etc/common/spire/bundle.pem";
+          description = "Path to the SPIRE bootstrap trust bundle.";
+        };
+
+        dataDir = mkOption {
+          type = types.str;
+          default = runtimeDir name;
+          description = "SPIRE agent data directory.";
+        };
+
+        socketPath = mkOption {
+          type = types.str;
+          default = "${runtimeDir name}/agent.sock";
+          description = "SPIFFE Workload API socket path.";
+        };
+
+        logLevel = mkOption {
+          type = types.str;
+          default = "INFO";
+          description = "SPIRE agent log level.";
+        };
+
+        settings = {
+          join_token.token = mkOption {
+            type = types.str;
+            default = "/etc/common/spire/tokens/${config.networking.hostName}.token";
+            description = "Path to the server-generated join token.";
+          };
+
+          x509pop = {
+            privateKeyPath = mkOption {
+              type = types.str;
+              default = "/etc/givc/key.pem";
+              description = "Path to the X.509 node-attestation private key.";
+            };
+
+            certificatePath = mkOption {
+              type = types.str;
+              default = "/etc/givc/cert.pem";
+              description = "Path to the X.509 node-attestation certificate.";
+            };
+          };
+        };
+      };
     }
-  '';
-  x509popPlugin = optionalString (cfg.nodeAttestationMode == "x509pop") ''
-    NodeAttestor "x509pop" {
-      plugin_data {
-        private_key_path = "$CREDENTIALS_DIRECTORY/key.pem"
-        certificate_path = "$CREDENTIALS_DIRECTORY/cert.pem"
+  );
+
+  enabledAgents = filterAttrs (_: agent: agent.enable) config.ghaf.security.spire.agents;
+  isJoinToken = agent: agent.nodeAttestationMode == "join_token";
+
+  credentials =
+    agent:
+    if isJoinToken agent then
+      [ "join-token:${agent.settings.join_token.token}" ]
+    else
+      [
+        "key.pem:${agent.settings.x509pop.privateKeyPath}"
+        "cert.pem:${agent.settings.x509pop.certificatePath}"
+      ];
+
+  credentialPaths =
+    agent:
+    if isJoinToken agent then
+      [ agent.settings.join_token.token ]
+    else
+      [
+        agent.settings.x509pop.privateKeyPath
+        agent.settings.x509pop.certificatePath
+      ];
+
+  agentConf =
+    agent:
+    let
+      nodeAttestor =
+        if isJoinToken agent then
+          ''
+            NodeAttestor "join_token" {
+              plugin_data {}
+            }
+          ''
+        else
+          ''
+            NodeAttestor "x509pop" {
+              plugin_data {
+                private_key_path = "$CREDENTIALS_DIRECTORY/key.pem"
+                certificate_path = "$CREDENTIALS_DIRECTORY/cert.pem"
+              }
+            }
+          '';
+    in
+    ''
+      agent {
+        data_dir = "${agent.dataDir}"
+        log_level = "${agent.logLevel}"
+        server_address = "${agent.serverAddress}"
+        server_port = ${toString agent.serverPort}
+        trust_domain = "${agent.trustDomain}"
+        trust_bundle_path = "${agent.trustBundlePath}"
+        socket_path = "${agent.socketPath}"
+        ${optionalString (isJoinToken agent) ''join_token_file = "$CREDENTIALS_DIRECTORY/join-token"''}
       }
-    }
-  '';
-  agentConf = ''
-    agent {
-      data_dir = "${dataDir}"
-      log_level = "${cfg.logLevel}"
-      server_address = "${config.ghaf.common.spire.server.address}"
-      server_port = ${toString config.ghaf.common.spire.server.port}
-      trust_domain = "${config.ghaf.common.spire.server.trustDomain}"
-      trust_bundle_path = "${cfg.trustBundlePath}"
-      socket_path = "${runtimeDataDir}/agent.sock"
-      ${joinTokenConf}
-    }
 
-    plugins {
-      ${joinTokenPlugin}
-      ${x509popPlugin}
+      plugins {
+        ${nodeAttestor}
 
-      WorkloadAttestor "unix" {
-        plugin_data {}
+        WorkloadAttestor "unix" {
+          plugin_data {}
+        }
+        KeyManager "memory" {
+          plugin_data {}
+        }
       }
-      KeyManager "memory" {
-        plugin_data {}
-      }
-    }
-  '';
+    '';
 
-  server-health = pkgs.writeShellApplication {
-    name = "spire-server-healthcheck";
+  configFiles = mapAttrs (
+    name: agent: pkgs.writeText "${serviceName name}.conf" (agentConf agent)
+  ) enabledAgents;
 
-    runtimeInputs = [
-      pkgs.curl
-    ];
+  waitForAgent =
+    name: agent:
+    pkgs.writeShellApplication {
+      name = "wait-for-${serviceName name}";
+      runtimeInputs = optionals agent.serverHealthCheck.enable [ pkgs.curl ];
+      text = ''
+        ${optionalString agent.serverHealthCheck.enable ''
+          server_url="http://${agent.serverAddress}:${toString agent.serverHealthCheck.port}/ready"
+          until curl --fail --silent --connect-timeout 1 --max-time 2 "$server_url" >/dev/null 2>&1; do
+            echo "Waiting for SPIRE server at $server_url"
+            sleep 1
+          done
+        ''}
 
-    text = ''
-      SERVER_URL="http://${config.ghaf.common.spire.server.address}:${healthCheckPort}/ready"
-      MODE="${cfg.nodeAttestationMode}"
-
-      echo "Checking SPIRE Server readiness at $SERVER_URL..."
-
-      # Short per-attempt timeouts: without them, curl defaults to ~30s on
-      # silently-dropped SYNs (rpfilter / netfilter not yet up on admin-vm),
-      # so a single hung TCP burns the whole 90s ExecStartPre budget.
-      until curl --fail --silent --connect-timeout 1 --max-time 2 "$SERVER_URL" > /dev/null 2>&1; do
-        echo "SPIRE Server is not ready yet. Retrying in 1 second..."
-        sleep 1
-      done
-
-      echo "SPIRE Server is ready! Starting SPIRE Agent..."
-
-      while [ ! -e "${cfg.trustBundlePath}" ]; do
-        echo "Waiting for trust bundle..."
-        sleep 1
-      done
-
-      if [ "$MODE" == "join_token" ]; then
-        while [ ! -e "${cfg.settings.join_token.token}" ]; do
-          echo "Waiting for server token..."
+        until [ -e ${escapeShellArg agent.trustBundlePath} ]; do
+          echo "Waiting for SPIRE trust bundle ${agent.trustBundlePath}"
           sleep 1
         done
-      fi
 
-    '';
-  };
+        ${optionalString (isJoinToken agent) ''
+          until [ -e ${escapeShellArg agent.settings.join_token.token} ]; do
+            echo "Waiting for SPIRE join token ${agent.settings.join_token.token}"
+            sleep 1
+          done
+        ''}
+      '';
+    };
+
+  agentServices = mapAttrs' (
+    name: agent:
+    let
+      unitName = serviceName name;
+    in
+    nameValuePair unitName {
+      description = "SPIRE agent ${name}";
+      wantedBy = [ "multi-user.target" ];
+      requires = [
+        "network-online.target"
+        "local-fs.target"
+      ];
+      after = [
+        "network-online.target"
+        "local-fs.target"
+      ]
+      ++ optional (!isJoinToken agent) "givc-key-setup.service";
+
+      unitConfig.RequiresMountsFor = [
+        agent.trustBundlePath
+      ]
+      ++ credentialPaths agent
+      ++ optional (!hasPrefix "/run/" agent.dataDir) agent.dataDir;
+
+      serviceConfig = {
+        ExecStartPre = [
+          (getExe (waitForAgent name agent))
+          (pkgs.writeShellScript "validate-${unitName}" ''
+            exec ${getExe' spire-package "spire-agent"} validate \
+              -expandEnv \
+              -config ${escapeShellArg configFiles.${name}}
+          '')
+        ];
+        ExecStart = "${getExe' spire-package "spire-agent"} run -expandEnv -config ${configFiles.${name}}";
+        LoadCredential = credentials agent;
+        User = unitName;
+        Group = unitName;
+        RuntimeDirectory = unitName;
+        RuntimeDirectoryMode = "0755";
+        Restart = "on-failure";
+        RestartSec = "5s";
+        UMask = "0027";
+
+        NoNewPrivileges = true;
+        PrivateTmp = true;
+        ProtectHome = true;
+        ProtectSystem = "strict";
+        ReadWritePaths = unique [
+          agent.dataDir
+          (runtimeDir name)
+        ];
+      };
+    }
+  ) enabledAgents;
 in
 {
   _file = ./agent.nix;
 
-  options.ghaf.security.spire.agent = {
-    enable = mkEnableOption "SPIRE agent";
-    nodeAttestationMode = mkOption {
-      type = types.spireNodeAttestationMode;
-      default = "x509pop";
-      description = "Node attestation mode";
-    };
-    workloads = mkOption {
-      type = types.spireWorkloads;
-      default = [ ];
-      description = "List of workloads for this spire agent";
-    };
-    trustBundlePath = mkOption {
-      type = types.path;
-      default = "/etc/common/spire/bundle.pem";
-      description = "Path to the SPIRE trust bundle PEM file (used to verify the server during bootstrap)";
-    };
-    logLevel = mkOption {
-      type = types.str;
-      default = "INFO";
-      description = "SPIRE server log level";
-    };
-    settings = {
-      join_token = {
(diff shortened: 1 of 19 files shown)


vadika (commented):
Three points:

1. **Upstream SVID fetch discards its result** — `spire-agent api fetch x509 -silent` without `-write` fetches and drops the SVID, so ExecStartPost is effectively a one-shot smoke test. If that's the intent, a comment saying so would help; if the SVID is meant to be consumed, it needs `-write` plus a retry/renewal path.

2. **No ordering between `spire-server.service` and `spire-agent-upstream.service`** — only the 5s socket poll. On boot the upstream agent needs network + attestation, likely >5s, so the fetch will warn on every boot even when correctly configured. Suggest `after = [ "spire-agent-upstream.service" ]` on the server when upstream is enabled.

3. **`tpm_devid` removed from `spireNodeAttestationMode`** (lib/default.nix) — #1771 adds TPM attestation. Worth syncing with @brianmcgillion before this lands.

everton-dematos:
> Three points:
> 
>     1. **Upstream SVID fetch discards its result** — `spire-agent api fetch x509 -silent` without `-write` fetches and drops the SVID, so ExecStartPost is effectively a one-shot smoke test. If that's the intent, a comment saying so would help; if the SVID is meant to be consumed, it needs `-write` plus a retry/renewal path.
> 
>     2. **No ordering between `spire-server.service` and `spire-agent-upstream.service`** — only the 5s socket poll. On boot the upstream agent needs network + attestation, likely >5s, so the fetch will warn on every boot even when correctly configured. Suggest `after = [ "spire-agent-upstream.service" ]` on the server when upstream is enabled.
> 
>     3. **`tpm_devid` removed from `spireNodeAttestationMode`** (lib/default.nix) — #1771 adds TPM attestation. Worth syncing with @brianmcgillion before this lands.

Thanks for the comments! 

1. This is intentionally a one-shot probe to verify upstream workload attestation and SVID issuance. Since the final backend server details are not yet defined, SVID persistence with `-write` and the corresponding renewal path will be handled in a follow-up PR. 

     1.1. I added a comment clarifying that the current fetch is a one-shot probe.

2. I agree that the five-second check could race with upstream agent startup. I avoided adding `After=` because the upstream agent may wait for external network or trust-bundle availability, which could block the independent local SPIRE server. 

     2.1. I added an asynchronous 30-second retry interval that continues until the first successful SVID fetch without delaying local server startup.

3. I removed `tpm_devid` because it is not currently wired or used in the main SPIRE configuration. The intention is to keep only the currently supported attestation mode and add `tpm_devid` back as part of the PR that enables TPM attestation.

vadika (approved):
LGTM!

milva-unikie:
**Tested on Darter Pro**

All good!

```
[ghaf@admin-vm:~]$ sudo spire-server healthcheck -socketPath /run/spire-server/api.sock -verbose
Checking server health...
Server is healthy.

[ghaf@admin-vm:~]$ sudo spire-server agent list -socketPath /run/spire-server/api.sock -attestationType x509pop
Found 10 attested agents:

SPIFFE ID         : spiffe://<id>
Attestation type  : x509pop
Expiration time   : 2026-07-20 06:20:00 +0000 UTC
Serial number     : <number>
Can re-attest     : true
Agent version     : 1.15.1
...
```

brianmcgillion (commented):
Deep security review of the SPIRE refactor. This is a **clean, well-executed** change — after a 5-lens security/quality pass with adversarial verification, **no blocker/high/medium findings survived**; what remains is low-severity hardening, the headline duplication, one eval-robustness bug, and missing docs/tests. Posting as a comment (minor suggestions), not changes-requested.

**Verified locally:** it evaluates, lint passes (nixfmt/statix/deadnix/reuse), and a behavioral eval confirms the **secure defaults** — `agents.upstream.enable = false` on admin-vm (your test step 1) and every VM migrated to `agents.downstream` / `x509pop`. The **join-token removal is complete** (no reachable leftover path). Nicely done.

D3: remove the 6 redundant `nodeAttestationMode = "x509pop"` assignments (the option already defaults to `x509pop`). Verified: attestation stays `x509pop` for every VM (behavioral eval), treefmt clean, applies clean.

```diff
diff --git a/modules/microvm/host/microvm-host.nix b/modules/microvm/host/microvm-host.nix
index 80184ae8..db1c1d8f 100644
--- a/modules/microvm/host/microvm-host.nix
+++ b/modules/microvm/host/microvm-host.nix
@@ -158,7 +158,6 @@ in
         security.spire.agents.downstream = {
           inherit (config.ghaf.global-config.spire) enable;
           logLevel = if config.ghaf.global-config.spire.debug then "DEBUG" else "INFO";
-          nodeAttestationMode = "x509pop";
           trustBundlePath = "/persist/common/spire/bundle.pem";
         };
       };
diff --git a/modules/microvm/sysvms/adminvm-base.nix b/modules/microvm/sysvms/adminvm-base.nix
index 7e5f630d..f814e267 100644
--- a/modules/microvm/sysvms/adminvm-base.nix
+++ b/modules/microvm/sysvms/adminvm-base.nix
@@ -168,7 +168,6 @@ in
         agents.downstream = {
           enable = globalConfig.spire.enable or false;
           logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
-          nodeAttestationMode = "x509pop";
         };
 
         # Disabled by default. Enable for the external SPIRE backend.
diff --git a/modules/microvm/sysvms/appvm-base.nix b/modules/microvm/sysvms/appvm-base.nix
index 917722b0..647fa2cb 100644
--- a/modules/microvm/sysvms/appvm-base.nix
+++ b/modules/microvm/sysvms/appvm-base.nix
@@ -268,7 +268,6 @@ in
           spire.agents.downstream = {
             enable = globalConfig.spire.enable or false;
             logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
-            nodeAttestationMode = "x509pop";
           };
         };
       };
diff --git a/modules/microvm/sysvms/audiovm-base.nix b/modules/microvm/sysvms/audiovm-base.nix
index dcdb3d18..2f877163 100644
--- a/modules/microvm/sysvms/audiovm-base.nix
+++ b/modules/microvm/sysvms/audiovm-base.nix
@@ -159,7 +159,6 @@ in
       spire.agents.downstream = {
         enable = globalConfig.spire.enable or false;
         logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
-        nodeAttestationMode = "x509pop";
       };
     };
 
diff --git a/modules/microvm/sysvms/guivm-base.nix b/modules/microvm/sysvms/guivm-base.nix
index 4c984748..1033ed4b 100644
--- a/modules/microvm/sysvms/guivm-base.nix
+++ b/modules/microvm/sysvms/guivm-base.nix
@@ -284,7 +284,6 @@ in
       spire.agents.downstream = {
         enable = globalConfig.spire.enable or false;
         logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
-        nodeAttestationMode = "x509pop";
       };
     };
   };
diff --git a/modules/microvm/sysvms/netvm-base.nix b/modules/microvm/sysvms/netvm-base.nix
index 1a357f98..3d078d99 100644
--- a/modules/microvm/sysvms/netvm-base.nix
+++ b/modules/microvm/sysvms/netvm-base.nix
@@ -178,7 +178,6 @@ in
       spire.agents.downstream = {
         enable = globalConfig.spire.enable or false;
         logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
-        nodeAttestationMode = "x509pop";
       };
     };
 

```

### Duplication / generalization (your headline) — designs
- **D1 (inline):** the per-VM SPIRE agent wiring is copy-pasted across ~13 sites in 11 files (`agents.downstream = {…}` + the registry export). Centralize via a sysvm-registry aggregation (like the existing `spire.agents` export) or a shared helper so it lives once.
- **D4:** `server.nix`'s `getVMsByAttestation`/`x509popVMs` per-attestation-mode filtering is now dead abstraction (one mode). **D5:** the agent submodule re-inlines the attestation enum instead of the shared `lib` type. **D6:** `credentials`/`credentialPaths` list the same two x509pop paths twice. **D7:** duplicated "poll until spire-server healthcheck" loops.

### Also (low)
- **S1 (inline):** the `upstream` agent inherits every connection default from the *local* server — enabling it without overriding silently targets the local root, with no assertion. Recommend an assertion.
- **C1 (inline):** `server.nix` unconditionally dereferences `agents.upstream` — a latent eval crash for a server host that doesn't define an upstream instance.
- **S2/S3 (inline):** `trustBundleUrl` HTTPS not enforced at the config layer (no pinning either); the upstream Workload-API socket's group-restriction is likely ineffective (`0755` runtime dir).
- The "upstream integration" currently fetches-and-discards an SVID (a PoC probe, no `UpstreamAuthority` consuming it) — safe, but the description overstates the wiring.
- **Docs/tests:** the upstream/downstream architecture change ships with no doc update (checkbox unchecked) and no test/eval-assertion coverage; the S1 assertion + an "upstream off by default" eval check are cheap.

The S1 assertion, the C1 defensive deref, and the D1/D4 consolidations are provided as concrete inline fixes/designs rather than applied, to keep the patch behavior-preserving.

modules/common/security/spire/agent.nix
```diff
@@ -8,183 +8,293 @@
...
+          port = mkOption {
+            type = types.port;
+            default = config.ghaf.common.spire.server.healthCheckPort;
+            description = "SPIRE server readiness endpoint port.";
+          };
+        };
+
+        trustDomain = mkOption {
+          type = types.str;
+          default = config.ghaf.common.spire.server.trustDomain;
```
brianmcgillion: **[security / footgun — confirmed]** Every connection-defining option of the generic agent submodule defaults to the **local** server's values — `trustDomain` (here → `ghaf.internal`), `serverAddress`/`serverPort`, and `trustBundlePath` (`:68`). On admin-vm `spire.server.address` is pinned to admin-vm's own IP, so the `upstream` agent's default `serverAddress` is literally the local server. So enabling `agents.upstream.enable = true` **without** overriding every connection field yields an "external" agent that silently joins the **local** trust domain and bootstraps from the local bundle — a partial-config that fails open (connects to *something*) rather than failing evaluation. The only assertion here (`:266`) checks socket-path uniqueness; nothing guards this.

**Fix:** add an assertion that a non-`downstream` agent, when enabled, must set `trustDomain`, `serverAddress`, and a trust-bundle source (`trustBundleUrl` or `trustBundlePath`) explicitly — ideally differing from the local server values — so a partial upstream config fails closed at eval.

modules/common/security/spire/agent.nix
```diff
@@ -8,183 +8,293 @@
...
+          description = "List of workloads for this SPIRE agent.";
+        };
+
+        trustBundlePath = mkOption {
+          type = types.str;
+          default = "/etc/common/spire/bundle.pem";
+          description = "Path to the SPIRE bootstrap trust bundle.";
+        };
+
+        trustBundleUrl = mkOption {
```
brianmcgillion, line 75: **[security / hardening]** `trustBundleUrl` is `types.nullOr types.str` with no constraint. It's the **bootstrap root of trust** and is written verbatim into `trust_bundle_url`. SPIRE's runtime does require HTTPS and validates the host against the system CA store — so an `http://` URL fails closed at agent start (impact is therefore limited) — but nothing here asserts or documents it, and even over HTTPS the bundle is trusted purely via web-PKI with **no pinning**, so a mis-issued/compromised public CA cert for the bundle host can MITM the trust root.

**Fix:** assert `lib.hasPrefix "https://" trustBundleUrl` when set, and document that the bundle host is validated only against the system CA store — prefer shipping an out-of-band `trustBundlePath` for a pinned root where possible.

modules/common/security/spire/server.nix
```diff
@@ -20,17 +19,13 @@ let
   inherit (config.ghaf.common.spire.server) healthCheckPort;
   inherit (config.ghaf.common.spire.server) trustDomain;
   spireAgentVMs = builtins.attrNames spireAgents;
+  upstreamAgent = config.ghaf.security.spire.agents.upstream;
```
brianmcgillion: **[correctness / robustness — confirmed]** `upstreamAgent = config.ghaf.security.spire.agents.upstream;` is an **unconditional** attribute dereference. `agents` is an `attrsOf` submodule, so on any SPIRE-server host that does **not** define an `upstream` instance this throws "attribute 'upstream' missing" at eval. It works today only because the server currently runs on admin-vm (which defines `upstream`); it's a latent break for any other server placement.

**Fix:** make it defensive — `upstreamAgent = config.ghaf.security.spire.agents.upstream or { enable = false; };` (or guard the later uses with `agents ? upstream`).

modules/common/security/spire/server.nix
```diff
@@ -234,30 +213,17 @@ in
               "${runtimeDataDir}"
             ];
           }
+          // optionalAttrs upstreamAgent.enable {
+            ExecStartPost = getExe spireServerUpstreamWorkloadApp;
+            SupplementaryGroups = [ upstreamAgentServiceName ];
```
brianmcgillion, line 218: **[security / least-privilege]** `spire-server` is granted `SupplementaryGroups = [ "spire-agent-upstream" ]` specifically to reach the upstream agent's Workload-API socket — implying the socket is meant to be group-restricted. But the upstream agent sets `RuntimeDirectoryMode = "0755"` (`agent.nix:239`, world-traversable) and no explicit socket mode/ACL is set (only `UMask 0027`, which SPIRE's socket creation doesn't reliably honour). If the socket ends up world-accessible in its `0755` dir, the group grant is moot and any local process on admin-vm can hit the upstream Workload API — obtaining external-trust-domain SVIDs subject only to workload attestation.

**Fix:** set the upstream agent `RuntimeDirectoryMode = "0750"` group-owned by a group the server shares, and/or pin the Workload-API socket mode (e.g. a post-start `chmod`/ACL); verify the resulting socket perms on the running system.

modules/microvm/sysvms/adminvm.nix
```diff
@@ -67,9 +67,12 @@ in
       ghaf.common = {
         extraNetworking.hosts.${vmName} = cfg.extraNetworking;
         spire = {
-          agents = lib.mkIf cfg.evaluatedConfig.config.ghaf.security.spire.agent.enable {
+          agents = lib.mkIf cfg.evaluatedConfig.config.ghaf.security.spire.agents.downstream.enable {
```
brianmcgillion, line 70: **[duplication / generalization — the headline]** The per-VM SPIRE agent wiring is copy-pasted across ~13 sites in 11 files: the `agents.downstream = { enable; logLevel; … }` block in each `*-base.nix` plus this registry-export block (`ghaf.common.spire.agents.<vm> = mkIf …downstream.enable { inherit (…downstream) nodeAttestationMode workloads; … }`) in each sysvm registry file. Only the VM name varies. (The `nodeAttestationMode` half is already removed in the patch above since it's a constant.)

**Design:** hoist this into a single registry-side aggregation over `config.ghaf.virtualization.microvm.sysvm.enabledVms` — mirroring how `spire.agents`/`extraNetworking` are already exported (and exactly the generalization we flagged for the givc ACL in #1998) — so each new VM is covered automatically and the export can't drift between VMs. That collapses ~13 hand-written blocks to one loop.

everton-dematos:
> Deep security review of the SPIRE refactor. This is a **clean, well-executed** change — after a 5-lens security/quality pass with adversarial verification, **no blocker/high/medium findings survived**; what remains is low-severity hardening, the headline duplication, one eval-robustness bug, and missing docs/tests. Posting as a comment (minor suggestions), not changes-requested.
> 
> **Verified locally:** it evaluates, lint passes (nixfmt/statix/deadnix/reuse), and a behavioral eval confirms the **secure defaults** — `agents.upstream.enable = false` on admin-vm (your test step 1) and every VM migrated to `agents.downstream` / `x509pop`. The **join-token removal is complete** (no reachable leftover path). Nicely done.
> ### Verified patch (apply with `git apply`)
> 
> D3: remove the 6 redundant `nodeAttestationMode = "x509pop"` assignments (the option already defaults to `x509pop`). Verified: attestation stays `x509pop` for every VM (behavioral eval), treefmt clean, applies clean.
> 
> ```diff
> diff --git a/modules/microvm/host/microvm-host.nix b/modules/microvm/host/microvm-host.nix
> index 80184ae8..db1c1d8f 100644
> --- a/modules/microvm/host/microvm-host.nix
> +++ b/modules/microvm/host/microvm-host.nix
> @@ -158,7 +158,6 @@ in
>          security.spire.agents.downstream = {
>            inherit (config.ghaf.global-config.spire) enable;
>            logLevel = if config.ghaf.global-config.spire.debug then "DEBUG" else "INFO";
> -          nodeAttestationMode = "x509pop";
>            trustBundlePath = "/persist/common/spire/bundle.pem";
>          };
>        };
> diff --git a/modules/microvm/sysvms/adminvm-base.nix b/modules/microvm/sysvms/adminvm-base.nix
> index 7e5f630d..f814e267 100644
> --- a/modules/microvm/sysvms/adminvm-base.nix
> +++ b/modules/microvm/sysvms/adminvm-base.nix
> @@ -168,7 +168,6 @@ in
>          agents.downstream = {
>            enable = globalConfig.spire.enable or false;
>            logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
> -          nodeAttestationMode = "x509pop";
>          };
>  
>          # Disabled by default. Enable for the external SPIRE backend.
> diff --git a/modules/microvm/sysvms/appvm-base.nix b/modules/microvm/sysvms/appvm-base.nix
> index 917722b0..647fa2cb 100644
> --- a/modules/microvm/sysvms/appvm-base.nix
> +++ b/modules/microvm/sysvms/appvm-base.nix
> @@ -268,7 +268,6 @@ in
>            spire.agents.downstream = {
>              enable = globalConfig.spire.enable or false;
>              logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
> -            nodeAttestationMode = "x509pop";
>            };
>          };
>        };
> diff --git a/modules/microvm/sysvms/audiovm-base.nix b/modules/microvm/sysvms/audiovm-base.nix
> index dcdb3d18..2f877163 100644
> --- a/modules/microvm/sysvms/audiovm-base.nix
> +++ b/modules/microvm/sysvms/audiovm-base.nix
> @@ -159,7 +159,6 @@ in
>        spire.agents.downstream = {
>          enable = globalConfig.spire.enable or false;
>          logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
> -        nodeAttestationMode = "x509pop";
>        };
>      };
>  
> diff --git a/modules/microvm/sysvms/guivm-base.nix b/modules/microvm/sysvms/guivm-base.nix
> index 4c984748..1033ed4b 100644
> --- a/modules/microvm/sysvms/guivm-base.nix
> +++ b/modules/microvm/sysvms/guivm-base.nix
> @@ -284,7 +284,6 @@ in
>        spire.agents.downstream = {
>          enable = globalConfig.spire.enable or false;
>          logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
> -        nodeAttestationMode = "x509pop";
>        };
>      };
>    };
> diff --git a/modules/microvm/sysvms/netvm-base.nix b/modules/microvm/sysvms/netvm-base.nix
> index 1a357f98..3d078d99 100644
> --- a/modules/microvm/sysvms/netvm-base.nix
> +++ b/modules/microvm/sysvms/netvm-base.nix
> @@ -178,7 +178,6 @@ in
>        spire.agents.downstream = {
>          enable = globalConfig.spire.enable or false;
>          logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
> -        nodeAttestationMode = "x509pop";
>        };
>      };
>  
> ```
> 
> ### Duplication / generalization (your headline) — designs
> 
>     * **D1 (inline):** the per-VM SPIRE agent wiring is copy-pasted across ~13 sites in 11 files (`agents.downstream = {…}` + the registry export). Centralize via a sysvm-registry aggregation (like the existing `spire.agents` export) or a shared helper so it lives once.
> 
>     * **D4:** `server.nix`'s `getVMsByAttestation`/`x509popVMs` per-attestation-mode filtering is now dead abstraction (one mode). **D5:** the agent submodule re-inlines the attestation enum instead of the shared `lib` type. **D6:** `credentials`/`credentialPaths` list the same two x509pop paths twice. **D7:** duplicated "poll until spire-server healthcheck" loops.
> 
> 
> ### Also (low)
> 
>     * **S1 (inline):** the `upstream` agent inherits every connection default from the _local_ server — enabling it without overriding silently targets the local root, with no assertion. Recommend an assertion.
> 
>     * **C1 (inline):** `server.nix` unconditionally dereferences `agents.upstream` — a latent eval crash for a server host that doesn't define an upstream instance.
> 
>     * **S2/S3 (inline):** `trustBundleUrl` HTTPS not enforced at the config layer (no pinning either); the upstream Workload-API socket's group-restriction is likely ineffective (`0755` runtime dir).
> 
>     * The "upstream integration" currently fetches-and-discards an SVID (a PoC probe, no `UpstreamAuthority` consuming it) — safe, but the description overstates the wiring.
> 
>     * **Docs/tests:** the upstream/downstream architecture change ships with no doc update (checkbox unchecked) and no test/eval-assertion coverage; the S1 assertion + an "upstream off by default" eval check are cheap.
> 
> 
> The S1 assertion, the C1 defensive deref, and the D1/D4 consolidations are provided as concrete inline fixes/designs rather than applied, to keep the patch behavior-preserving.

D3: Removed the six redundant `nodeAttestationMode = "x509pop"` assignments.
D4: I Kept it in case we add different node Attestation in the future.
D5: Modified to `spireNodeAttestationMode`
D6: This does not mean credentials are loaded twice at runtime. The paths are currently repeated because:`LoadCredential` needs name:path strings. `RequiresMountsFor` needs the raw paths.
D7: The loops are for different purpose with distinct wait times.

Further additions/modifications:
- Added an assertion requiring enabled non-downstream agents to explicitly configure their server and trust-bundle connection settings.
- Added HTTPS enforcement for `trustBundleUrl` and documented the system-CA validation and pinned-bundle alternative.
- Made the optional upstream-agent lookup defensive so SPIRE server configurations without an upstream instance evaluate safely.
- Restricted non-downstream SPIRE agent runtime directories to `0750`, making upstream Workload API access group-controlled.

brianmcgillion (approved):

PR #2033, opened 2026-07-09
feat(jetson): pass the on-soc mgbe0 ethernet through to net-vm

vadika:
## Description of Changes

Passes the AGX Orin's on-SoC ethernet controller (MGBE0, `ethernet@6800000`)
through to net-vm, so the network VM drives the physical wired interface
directly instead of the host.

MGBE0 is an on-SoC MMIO device, so it needs two channels:

- **Data** — `vfio-platform` hands the MAC's MMIO windows and interrupts to the
  guest; DMA goes through the SMMU (`IOVA = GPA`). MGBE0 is alone in its IOMMU
  group, so VFIO takes it cleanly and no reserved carveouts are needed.
- **Control** — every clock, reset and the power domain on the node is a
  `<&bpmp …>` reference and the guest has no BPMP. The guest's `tegra_bpmp` core
  is redirected, via the `virtual-pa` property on its `/bpmp` node, through a
  QEMU MMIO bridge to `/dev/bpmp-host` on the host, where the BPMP host proxy
  gates each request against an allow-list.

Notable points:

- QEMU removed `-device vfio-platform` in 10.2, so net-vm runs a pinned
  `ghaf-qemu-bpmp` built from **QEMU 10.1.5**, the last release that has it.
  Every other VM stays on the nixpkgs QEMU. This pin is a maintenance liability
  (net-vm is network-facing) and is documented in the package and the feature
  page.
- The guest device tree is **emitted by QEMU** via a `VFIO_PLATFORM_BINDING` for
  compat `nvidia,tegra234-mgbe`. `virt` `exit(1)`s for a dynamic sysbus device
  with no FDT binding, which is why a hand-written `-dtb` never worked for
  platform passthrough.
- The BPMP allow-list is scoped to **exactly** MGBE0's clocks/resets/power, and
  the guest boots with `clk_ignore_unused pd_ignore_unused`. Both guard the host:
  a guest able to reach arbitrary clocks would disable ones the host needs (its
  eMMC controller among them) at `late_initcall` and wedge it. The proxy's
  `clock not allowed` warnings at guest boot are that boundary working (documented
  as expected in `bpmp_host_overlay.dts`).
- AGX-only. Enabled per-SoM in the AGX hardware modules, not in the shared Orin
  profile — Orin NX has no MGBE0.

### Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

[...]

PR #2036, opened 2026-07-11
feat(jetson-gpu): pass the tegra234 gpu through to a gpu-vm microvm

vadika:
## Description of Changes

Passes the Jetson Orin AGX integrated GPU (`ga10b`, `gpu@17000000`) and its
host1x multimedia engines through to a dedicated `gpu-vm` microVM via
`vfio-platform`, so CUDA compute runs in an isolated guest instead of on the
host. It coexists with the MGBE0 net-vm passthrough (#2033): both share the
BPMP-virtualisation and `ghaf-qemu-bpmp` infrastructure.

Highlights:

- **MMIO passthrough** with a static guest device tree (`-dtb`) whose `ranges`
  translate the real GPU MMIO to QEMU's platform-bus addresses; the BPMP node's
  `virtual-pa` matches QEMU's `VIRT_NVIDIA_BPMP_GUEST` bridge (`0x090d0000` on
  QEMU >= 10).
- **BPMP virtualisation**: clocks, resets and power domains route through the
  BPMP host proxy under a closed allow-list (never `bpmpAllowAllDomains`).
- **Bring-your-own-kernel guest**: vanilla linux 6.12 + the nvidia-jetpack
  out-of-tree overlay (`nvgpu`/`nvmap`/`host1x`/`nvhost`).
- **Host release**: the GPU/host1x stack is handed to vfio and host graphics is
  turned off while `gpu_vm` is enabled.
- **CUDA userspace**: resolves `libcuda` to the native `l4t-cuda` driver stack
  (the desktop `cuda_compat` build otherwise cannot find
  `libnvcucompat`/`libnvrm_*`), plus a prebuilt `gpu-vm-load` Driver-API smoke
  test and a boot service that group-grants the GPU nodes so CUDA runs as the
  `ghaf` user without sudo.

Verified on the Orin AGX devkit: `gk20a` binds, `cuInit` reports `sm_87`, and
`gpu-vm-load` drives `/sys/devices/platform/64000000.gpu/load` to ~99%.

Display controller passthrough is intentionally left out (no video output;
compute only).

> **Stacked on #2033** (MGBE0 net-vm passthrough). This branch builds on it and
> includes its commits; the two `fix(jetson-bpmp)` shared-proxy commits belong
> to #2033 and were added there. Please merge #2033 first; the diff here reduces
> to the gpu-vm commit once #2033 lands on `main`.

### Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [x] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

Stacked on #2033. Modernises and rebases the GPU/display passthrough from #1031.

## Checklist

### Applicable Targets
- [x] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`
### Installation Method
- [ ] Other:
### Test Steps To Verify:
1. Build the AGX debug image (cross-built, never on the device):
   `nix build .#packages.x86_64-linux.nvidia-jetson-orin-agx-debug-from-x86_64`
2. Flash it with the corresponding `...-flash-script` (board in recovery).
3. On the running target, reach the guest: `ssh ghaf@gpu-vm`.
4. Confirm the driver bound: `ls /dev/nvgpu/igpu0 /dev/nvhost-ctrl-gpu`.
5. Run the compute smoke test as the `ghaf` user (no sudo): `gpu-vm-load 15`.
   Expect `GPU compute capability: sm_87` and `GPU_LOAD_OK`.
6. While it runs, watch the GPU utilisation climb to ~99%:
   `cat /sys/devices/platform/64000000.gpu/load` (permille, 0-1000).


vadika:
resolved copilot comments

milva-unikie:
**Tested on Orin AGX**

Built and flashed `nix build .#nvidia-jetson-orin-agx-debug-from-x86_64-flash-script`

**Issues**

* GUI does not open, nothing is shown on external display.
* Power mode is not set. This logs repeating errors to host journal and causes logging rate test to fail in `jenkins-pre-merge`.

```
[ghaf@ghaf-host:~]$ nvpmodel -q
NVPM WARN: power mode is not set!

[ghaf@ghaf-host:~]$ journactl -u nvpmodel.service
---
Jul 16 05:40:14 ghaf-host systemd[1]: nvpmodel.service: Scheduled restart job, restart counter is at 61.
Jul 16 05:40:14 ghaf-host systemd[1]: Starting Set NVPModel power profile...
Jul 16 05:40:14 ghaf-host setupConf[3461]: Selected device som before hard coding: p3701_0000
Jul 16 05:40:14 ghaf-host setupConf[3461]: Selected device som after hard coding: p3701_0000
Jul 16 05:40:14 ghaf-host nvpmodel[3466]: NVPM ERROR: Error opening /sys/devices/platform/gpu.0/tpc_pg_mask: 2
Jul 16 05:40:14 ghaf-host nvpmodel[3466]: NVPM ERROR: optMask is 3, no request for power mode
Jul 16 05:40:14 ghaf-host systemd[1]: nvpmodel.service: Main process exited, code=exited, status=234/n/a
Jul 16 05:40:14 ghaf-host systemd[1]: nvpmodel.service: Failed with result 'exit-code'.
Jul 16 05:40:14 ghaf-host systemd[1]: Failed to start Set NVPModel power profile.
```
- Docker fails to start on host. This also adds to the host log noise and is partly responsible for logging rate test fail in `jenkins-pre-merge`.
```
Jul 16 05:39:18 ghaf-host dockerd-rootless[2959]: [rootlesskit:parent] error: failed to setup UID/GID map: failed to compute uid/gid map: open /etc/subuid: no such file or directory
Jul 16 05:39:18 ghaf-host dockerd-rootless[2977]: [rootlesskit:child ] error: EOF
Jul 16 05:39:18 ghaf-host systemd[2852]: docker.service: Main process exited, code=exited, status=1/FAILURE
Jul 16 05:39:18 ghaf-host systemd[2852]: docker.service: Killing process 2977 (exe) with signal SIGKILL.
Jul 16 05:39:18 ghaf-host systemd[2852]: docker.service: Failed with result 'exit-code'.
Jul 16 05:39:18 ghaf-host systemd[2852]: Failed to start Docker Application Container Engine (Rootless).
Jul 16 05:39:18 ghaf-host systemd[1]: Failed to start Set NVPModel power profile.
Jul 16 05:39:19 ghaf-host unix_chkpwd[2985]: password check failed for user (ghaf)
Jul 16 05:39:20 ghaf-host systemd[2852]: docker.service: Scheduled restart job, restart counter is at 3.
Jul 16 05:39:20 ghaf-host systemd[2852]: docker.service: Start request repeated too quickly.
Jul 16 05:39:20 ghaf-host systemd[2852]: docker.service: Failed with result 'start-limit-hit'.
Jul 16 05:39:20 ghaf-host systemd[2852]: Failed to start Docker Application Container Engine (Rootless).
```

Full host log from pre-merge tests: [host.log](https://github.com/user-attachments/files/30081021/host.log)


**Working**

* Steps in testing instructions pass.

```
[ghaf@gpu-vm:~]$ ls /dev/nvgpu/igpu0
as  channel  ctrl  ctxsw  dbg  nvsched  nvsched_ctrl_fifo  power  prof  prof-ctx  prof-dev  sched  tsg

[ghaf@gpu-vm:~]$ ls /dev/nvhost-ctrl-gpu
/dev/nvhost-ctrl-gpu

[ghaf@gpu-vm:~]$ gpu-vm-load 15
NvTegraPrivGetProductionMode: Could not read Tegra production mode
Expected on kernels without fuse support
GPU compute capability: sm_87
running GPU load for 15 s (watch /sys/devices/platform/64000000.gpu/load)
GPU_LOAD_OK iters=165
```

* `cat /sys/devices/platform/64000000.gpu/load` showed load 999 during `gpu-vm-load 15`.

vadika:
> * GUI does not open, nothing is shown on external display.

not a bug, a feature :) display is disabled in this configuration. 

> Power mode is not set. This logs repeating errors to host journal and causes logging rate test to fail in jenkins-pre-merge.

Will fix in the following PR, not there, needs rewriting rhe NVPower for this config

vadika:
fixed by disabling incompatible services on host. gpu-vm needs improvements to be able to run gpu loads inside -- docker, etc. should be inside. probably not this PR scope.

milva-unikie:
**Tested on AGX**

- No more log spam on the host
- Power mode test does not work with this PR, I prepared a test automation skip for it https://github.com/tiiuae/ci-test-automation/pull/899
- GUI does not work. I am not happy with that, but since that is an intentional decision the PR is good from my side.

modules/common/security/audit/default.nix
```diff
@@ -104,11 +104,6 @@ in
 
     boot.kernelParams = [
       "audit_backlog_limit=${toString config.security.audit.backlogLimit}"
```
brianmcgillion, line 106: **[blocking / regression — unrelated to this feature]** This `boot.kernelParams` block no longer sets `audit_backlog_wait_time=0`. That param (and its comment) was added by merged `main` commit `3c063c55 fix(audit): stop full backlog stalling auditd startup`; the PR base is `3c063c55` itself, so this is a clean revert of the tip fix. Without it, a heavy first boot can fill the audit backlog before auditd drains it, auditd's own startup syscalls block for the kernel-default wait time, and the Security Audit Logging start job times out (~90 s), intermittently wedging boot on Orin AGX. Please rebase onto current `main` (`ae4c973b`) so this fix is preserved.

modules/common/identity/vm-hostname-setter.nix
```diff
@@ -48,18 +48,16 @@ in
     systemd.services.set-dynamic-hostname = {
       description = "Set hostname from hardware-based identity";
       wantedBy = [ "network-pre.target" ];
-      unitConfig.DefaultDependencies = false;
       after = [
         "sysinit.target"
         "local-fs.target"
+        "systemd-hostnamed.service"
```
brianmcgillion: **[blocking / regression — unrelated to this feature]** Re-adding `After=systemd-hostnamed.service` here (together with the removal of `unitConfig.DefaultDependencies=false` and the `shutdown.target` conflict/ordering below) is a byte-exact reverse of merged `main` commit `b42ae3a5 fix(identity): break set-dynamic-hostname ordering cycle`. It reintroduces the ordering cycle (`set-dynamic-hostname → basic.target → … → network-pre.target → set-dynamic-hostname`), which systemd breaks non-deterministically — on some boots the hostname isn't set and net-vm stays `net-vm`, failing automated tests. Rebasing onto current `main` restores the fix.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -0,0 +1,322 @@
...
+    #   vic@15340000    clock 167           reset 113   pd 29
+    #   nvdec@15480000  clocks 83 40 154    reset 44    pd 23
+    #   nvjpg@15540000  clock 20            reset 10    pd 36
+    # display@13800000 and dce@d800000 are deliberately EXCLUDED: display
+    # passthrough is non-functional (carried, not fixed) and its ~60 PLL clocks
+    # would needlessly widen the host-safety boundary. As with MGBE0, the host
+    # proxy logging "clock not allowed" for probed parent clocks at guest boot is
+    # the boundary working, not a defect. If GPU init fails on a specific denied
+    # id (Task 10), add only that id here.
+    ghaf.hardware.nvidia.virtualization.host.bpmp.allow = {
```
brianmcgillion, line 149: **[medium / security]** The host BPMP proxy enforces a single allow-list = the *union* of `host.bpmp.allow.*` across all enabled passthroughs, and `check_if_allowed()` gates by resource id only, never by which guest issued the MRQ. Both net-vm and gpu-vm run as `microvm:kvm` and open the same `/dev/bpmp-host`. Since every AGX enables both passthroughs, a compromised gpu-vm can issue `MRQ_RESET` for MGBE0's reset ids (`45/46/47`) and knock out net-vm's live ethernet — and vice-versa. As net-vm handles untrusted network traffic, this is a realistic guest→guest DoS that didn't exist with a single passthrough VM. The host itself stays protected (allow-list is correctly scoped, `bpmpAllowAllDomains` off) — the residual is guest↔guest. Consider a per-VM proxy/allow-list; if a shared proxy is unavoidable, add an assertion/comment at the union point documenting the shared BPMP trust domain.

packages/pkgs-by-name/ghaf-qemu-bpmp-gpu/package.nix
```diff
@@ -0,0 +1,19 @@
...
+# GPU-VM variant of ghaf-qemu-bpmp. Same QEMU 10.1.5 + BPMP bridge, plus two
+# patches the GPU needs and net-vm does not:
+#   - vfio-platform mmio-base + a large platform bus, for 1:1 GPA=HPA mapping of
+#     the GPU's large reserved-memory MMIO regions.
+#   - a predefined-DTB path so gpu-vm can supply a hand-written guest device tree
+#     (GPU + display + engines; too complex to emit from a QEMU FDT binding).
+# Kept a SEPARATE binary so net-vm's memory map stays byte-identical to
+# ghaf-qemu-bpmp.
+{ ghaf-qemu-bpmp }:
+ghaf-qemu-bpmp.override {
```
brianmcgillion, line 21: **[medium / supply-chain]** This extends the EOL QEMU 10.1.5 pin (last release with `-device vfio-platform`, removed in 10.2 with no replacement) to a second VM that is on the inter-VM network *and* owns a DMA-capable GPU + host1x/nvdec/nvjpg/vic via VFIO. A guest-reachable QEMU/VFIO CVE in the frozen 10.1 branch now has a second, hardware-privileged blast surface (QEMU runs `microvm:kvm`), and there's no forward-compatible bump path since the feature was deleted upstream. Suggest tracking this as an explicit, time-boxed lifecycle issue, keeping gpu-vm off WAN-adjacent segments / firewalling it from untrusted app-VMs, and investigating a supported successor (vfio-user or a maintained fork) so net-vm and gpu-vm can migrate off 10.1 together.

modules/reference/hardware/jetpack/agx/orin-agx.nix
```diff
@@ -34,6 +34,7 @@
       # AGX has the on-SoC MGBE0 ethernet controller (Aquantia PHY on the
       # p3737 carrier); pass it through to net-vm. Orin NX has no MGBE0.
       nvidia.passthroughs.mgbe0_net_vm.enable = true;
+      nvidia.passthroughs.gpu_vm.enable = true;
```
brianmcgillion, line 37: **[medium / behavior change]** Enabling `gpu_vm` here (and in `orin-agx64.nix` / `orin-agx-industrial.nix`) force-disables the host graphics stack — `profiles.graphics.enable`, `nvpmodel`, and `nvidia-docker` are all set via `lib.mkForce false` in the gpu-vm module. So every AGX build now loses its native COSMIC desktop and host NVIDIA container support, with no per-device opt-out and no assertion documenting the coupling. If this default is intended, please add an assertion tying `gpu_vm.enable` to the graphics/nvpmodel/docker disable (so inconsistent config fails loudly) or gate it behind opt-in; at minimum document that enabling AGX now removes the host GUI by default.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -0,0 +1,322 @@
...
+                  patches = (o.patches or [ ]) ++ [
+                    ./patches/0001-gpu-add-support-for-passthrough.patch
+                    ./patches/0002-add-support-for-gpu-display-passthrough.patch
+                    ./patches/0003-add-support-for-display-passthrough.patch
+                  ];
+                });
+              }
+            )
+          );
+          boot.kernelParams = [
```
brianmcgillion, line 254: **[medium / robustness]** A real runtime test isn't possible in CI (needs aarch64 + a physical GPU), so eval-time assertions are the only enforceable guardrail for the safety invariants this feature depends on — but there are none for: (a) `bpmpAllowAllDomains == false` while `gpu_vm.enable` (flipping it makes the proxy allow-everything and the smoke test still passes green), and (b) these guest kernelParams containing **both** `clk_ignore_unused` and `pd_ignore_unused` (without them the guest can power off clocks/domains the host uses). Adding both assertions in the `gpu_vm.enable` block is cheap and fires in the existing AGX image build.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -0,0 +1,322 @@
...
+          # Kernel), with #1031's GPU/display passthrough patches applied to
+          # nvidia-oot-modules (validated: base OOT builds clean on 6.12.93).
+          boot.kernelPackages = lib.mkForce (
+            (pkgs.linuxPackages_6_12.extend pkgs.nvidia-jetpack.kernelPackagesOverlay).extend (
+              _final: prev: {
+                nvidia-oot-modules = prev.nvidia-oot-modules.overrideAttrs (o: {
+                  patches = (o.patches or [ ]) ++ [
+                    ./patches/0001-gpu-add-support-for-passthrough.patch
+                    ./patches/0002-add-support-for-gpu-display-passthrough.patch
+                    ./patches/0003-add-support-for-display-passthrough.patch
```
brianmcgillion: **[low / attack surface]** `0003-add-support-for-display-passthrough.patch` is applied to the guest kernel even though display is explicitly out of scope (compute-only; display/dce excluded from the engine list and allow-list). It's dead code compiled into the guest, and it force-returns `os_is_xen_dom0()` / `os_is_vgx_hyper()` / `xen_support_fully_virtualized_kernel()` = TRUE. Suggest dropping `0003` (and reconsidering the display parts of `0002`) until display passthrough is actually supported.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/patches/0002-add-support-for-gpu-display-passthrough.patch
```diff
@@ -0,0 +1,162 @@
...
+--- a/nvidia-oot/drivers/gpu/drm/tegra/uapi.c
++++ b/nvidia-oot/drivers/gpu/drm/tegra/uapi.c
+@@ -107,9 +107,13 @@ int tegra_drm_ioctl_channel_open(struct drm_device *drm, void *data, struct drm_
+ 	}
+ 
+ 	/* Only allocate context if the engine supports context isolation. */
+-	if (device_iommu_mapped(client->base.dev) && client->ops->can_use_memory_ctx) {
++	// TODO: review, maybe this modification is not needed
++	//if (device_iommu_mapped(client->base.dev) && client->ops->can_use_memory_ctx) {
++	if (client->ops->can_use_memory_ctx) {
```
brianmcgillion, line 65: **[low / security]** This hunk deletes the `device_iommu_mapped()` precondition before host1x memory-context setup, leaving only `client->ops->can_use_memory_ctx` — and carries the author's own `TODO: review, maybe this modification is not needed`. This path is live in the compute-only build (host1x + vic/nvdec/nvjpg are passed through). Please resolve the TODO before merge: restore the guard, or replace the blanket removal with an explicit, commented VM-specific condition.

packages/gpu-vm-load/runner.c
```diff
@@ -0,0 +1,96 @@
...
+  do {                                                                         \
+    CUresult _r = (x);                                                         \
+    if (_r) {                                                                  \
+      fprintf(stderr, "%s failed: CUresult=%d\n", #x, _r);                     \
+      return 1;                                                                \
+    }                                                                          \
+  } while (0)
+
+int main(int argc, char **argv) {
+  int secs = argc > 1 ? atoi(argv[1]) : 20;
```
brianmcgillion: **[low / correctness]** `secs` comes from `atoi(argv[1])` with no validation, and the load loop is `while (time(NULL) - t0 < secs)` — at entry the elapsed time is 0, so any non-positive or non-numeric arg (`gpu-vm-load 0`, `-1`, `s15` → `atoi` returns 0) skips the loop entirely, then still prints `GPU_LOAD_OK` and returns 0. For a passthrough smoke test that's a false success masking a broken GPU. Suggest parsing with `strtol` (reject non-numeric / out-of-range), requiring `secs > 0`, and gating the `GPU_LOAD_OK` print on `iters > 0`.

brianmcgillion (changes requested):
Deep review of the gpu-vm passthrough. **Build verification is green** — the AGX and NX `-from-x86_64` debug images both cross-compile locally, the AGX image closure contains the full stack (`ghaf-qemu-bpmp-gpu`, the compiled `gpuvm-dtb`, patched OOT kernel, `gpu-vm-load`, `nixos-system-gpu-vm`), and lint/REUSE/docs pass. The BPMP allow-list design is genuinely least-privilege and the `ghaf-qemu-bpmp` refactor is clean.

Two things should block merge, and both are **unrelated to the GPU feature** — they look like accidental reverts from branching off a stale base (`3c063c55`, 2 commits behind `main`): the removal of `audit_backlog_wait_time=0` (reverts `3c063c55`) and the `set-dynamic-hostname` ordering change (reverts `b42ae3a5`). **Rebasing onto current `main` removes both.** Inline comments below cover these plus the higher-value security/design items (shared-BPMP cross-VM reach, EOL QEMU blast radius, host-desktop coupling, missing eval-time guardrails) and a few concrete fixes. Full write-up with all low/nit items available on request.

brianmcgillion:
Follow-up to the [main review](https://github.com/tiiuae/ghaf/pull/2036#pullrequestreview-4730572292) (`CHANGES_REQUESTED`) — the two blocking reverts and the medium security/design items are inline there. Below are the remaining **lower-severity code findings** (C / kernel / QEMU / device-tree), as a checklist so they don't get lost. None is individually merge-blocking, but each is a real, addressable item.

### QEMU patches — `packages/pkgs-by-name/ghaf-qemu-bpmp/patches/`
- [ ] `0002-vfio-platform-mmio-base.patch:79` — `platform_bus_map_mmio()` now does an unconditional `VFIO_PLATFORM_DEVICE(sbdev)` QOM cast on *every* platform-bus device; it aborts QEMU if a non-VFIO platform device is ever added. Latent today (bus is all-VFIO) — guard the cast.
- [ ] `0002-…:43` — `VIRT_PLATFORM_BUS = {0x60000000, 0x1FC0000000}` ends at `0x2020000000`, overlapping `VIRT_MEM` base `0x2000000000` by 512 MiB. Benign only because nothing maps into the overlap; worth a comment or a non-colliding layout.
- [ ] `0002-…:87` — `mmio-base` is only lower-bound checked; an oversize / mismapped base falls through to a warn-only path rather than failing.
- [ ] `0003-nop-predefined-dtb-memory.patch:21` — `NOP_PREDEFINED_DTB_MEMORY` hardcoded to `0` disables QEMU's guest-DTB `/memory` sanitization; please document why that is safe here.

### Guest kernel patches — `…/passthrough/gpu-vm/patches/`
- [ ] `0001-gpu-add-support-for-passthrough.patch:54` — hardcoded nvgpu DMA carveout `0x80000000 / 0x08000000` (`// TODO: address from DTB`) implicitly coupled to `reservedMem` in `default.nix`; will silently drift. Read from DT, or add a compile-time tie.
- [ ] `0001-…:148` — `__nvgpu_sgl_ipa()/__nvgpu_sgl_phys()` branch on `CMA_*_OFFSET_TO_HOST_PA`, both defined as `0`, so the added logic is a no-op on a memory-translation hot path. Revert to `return sg_phys(sgl);`.
- [ ] `0004-tegra-fixed-chip-id.patch:10` — hardcodes a global `chipid = 0x00012347` (`// TODO: virtualize later`); fine for the single T234 target but should be gated to the VM / guest-proxy config.
- [ ] `0003-add-support-for-display-passthrough.patch:165` — debug scaffolding dumps kernel/phys/DMA addresses to guest dmesg and sets `cur_debuglevel=1` / rewrites the print gate to `if (1)`. Strip or gate behind a debug flag (overlaps dropping `0003` entirely — see the inline comment).

### Device tree — `…/passthrough/gpu-vm/`
- [ ] `tegra234-gpuvm.dts:177` — guest DT enables `display@13800000` + `dce@d800000` that aren't passed through (excluded from the engine list). Least-privilege / boot-robustness: disable them in the guest DT too.
- [ ] `tegra234-gpuvm.dts:136` — platform-bus `ranges` maps the disabled fuse window onto guest RAM at `0x80000000`.
- [ ] `default.nix:53` — `nvdec` is bound to vfio-platform and MMIO-mapped into the guest, but its node in the guest DT is disabled — inconsistent (pass it through fully, or don't map it).
- [ ] `tegra234-gpuvm.dts:10` — vendored NVIDIA dt-binding headers and several mainline includes are never referenced; drop the unused includes.

### BPMP / host-wide (context)
- [ ] `…/bpmp-virt-common/sources/.../bpmp-host-proxy.c:309` — the host proxy unconditionally forwards SoC-wide MRQs (`MRQ_DEBUG/BWMGR/ISO_CLIENT/EMC_DVFS/STRAP`) with no allow-list gating; pre-existing, but gpu-vm now depends on / can reach them — worth confirming that is intended.
- [ ] `…/passthrough/gpu-vm/default.nix:145` — enabling GPU passthrough widens the blast radius of the host-wide SMMU/IRQ weakeners (`arm-smmu.disable_bypass=0`, `vfio_iommu_type1.allow_unsafe_interrupts=1`); note the tradeoff.

brianmcgillion:
Follow-up #2 to the [main review](https://github.com/tiiuae/ghaf/pull/2036#pullrequestreview-4730572292) — remaining **maintainability / docs / test / licensing** items, as a checklist.

### Nix / structure
- [ ] `…/passthrough/gpu-vm/default.nix:166` — the BPMP guest kernel-config, the `bpmp-host`/`vfio` udev grant, and the vfio-bind oneshot are duplicated between this module and `mgbe0-net-vm`; factor a shared helper.
- [ ] `default.nix:89` — `gpuvm-dtb` preprocesses the guest DT with the **host** kernel headers (`config.boot.kernelPackages.kernel`) despite the comment saying "guest kernel"; reconcile the comment/behaviour.
- [ ] `modules/microvm/sysvms/gpuvm-base.nix:249` — `microvm.vcpu = 4` is a plain value coupled to the generated DTS (the comment says changing it needs a DTS regen). Make that an assertion or `mkForce` so it can't be silently overridden.
- [ ] `default.nix:138` — the allow-list is raw magic-number arrays and the comments invite adding ids to silence denials — easy to widen the host-safety boundary by accident. Consider symbolic names / a note that additions must be justified.
- [ ] `default.nix:175` — the host-GPU / host-graphics mutual exclusion relies only on `lib.mkForce` with no assertion or AGX gate (depth on the M3 inline comment).

### Docs / comments
- [ ] `default.nix:7` — the header comment lists `dce` as passed-through, but `dce`/display are explicitly excluded.
- [ ] `packages/pkgs-by-name/ghaf-qemu-bpmp-gpu/package.nix:6` ↔ `gpuvm-base.nix:150,160-170` — contradiction: the base ships `cuda_nvcc` + `gcc` + `vectorAdd.cu` "to build on-device", while the comments elsewhere say on-device compilation is impossible and ship the prebuilt `gpu-vm-load`. Drop the redundant nvcc toolchain + sample, or fix the comments (also trims the guest image).
- [ ] `gpuvm-base.nix:90` (and elsewhere) — internal codenames / tracker IDs leak into committed comments: `ponytail:`, `Task 7/8/10`, `#1031`. Remove before merge.
- [ ] `docs/…/nvidia_gpu_vm.mdx:145` — the `sudo` / `LD_LIBRARY_PATH` caveat doesn't apply to the RPATH-wired `gpu-vm-load` command it precedes; `:11` mixes British/American spelling (CONTRIBUTING style).
- [ ] `packages/gpu-vm-load/vadd.ptx` — the kernel is a compute-burn loop, not a vector add; the `vadd` / `vadd_ptx` names are misleading. _(nit)_ `runner.c:94` — no explicit CUDA cleanup on exit paths (benign for a one-shot).

### Tests
- [ ] `tests/flake-module.nix` — the only CI gate is the full aarch64 image build. Add a `flake.checks.aarch64-linux` that force-builds just the passthrough sub-closure (`gpuvm-dtb`, the `bpmp_host_overlay` dtb, `ghaf-qemu-bpmp-gpu`, `gpu-vm-load`) so DT/patch/package breakage fails fast with clear attribution.
- [ ] `gpuvm-base.nix:171` — `gpu-vm-load` is a positive-only, un-automated smoke test (verifies neither isolation, allow-list refusal, nor host liveness). Turn the doc "Verify GPU compute" section into an ordered, fail-fast checklist.

### Supply chain / licensing / hygiene
- [ ] `gpuvm-base.nix:154` — the guest ships a large proprietary NVIDIA CUDA/firmware closure sourced from a mutable-branch `jetpack-nixos` fork; pin to a fixed rev.
- [ ] `…/gpu-vm/nv-dt-bindings/**` — the vendored NVIDIA headers are `GPL-2.0-only` inside an otherwise Apache-2.0 tree; confirm that is an accepted licensing decision (prefer a dual-licensed upstream copy if one exists) and record the upstream source/commit.
- [ ] commit hygiene — the two `fix(jetson-gpu)` commits have empty bodies (only `Signed-off-by`); CONTRIBUTING asks for a body explaining what/why.

packages/gpu-vm-load/runner.c
brianmcgillion: **Suggested fix — validate the duration arg** (checklist "atoi UB" + inline comment #9). Parse with `strtol`, reject non-numeric / ≤0 / oversize, keep the no-arg default of 20. A bad arg then exits non-zero instead of printing `GPU_LOAD_OK` after zero GPU work. (`strtol` is already available via `<stdlib.h>`.)

```suggestion
  int secs = 20;
  if (argc > 1) {
    char *end;
    long v = strtol(argv[1], &end, 10);
    if (*argv[1] == '\0' || *end != '\0' || v <= 0 || v > 86400) {
      fprintf(stderr, "invalid duration '%s' (expected 1..86400 seconds)\n",
              argv[1]);
      return 2;
    }
    secs = (int)v;
  }
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -0,0 +1,322 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Pass the AGX Orin's on-SoC GPU (ga10b) and supporting engines to gpu-vm.
+#
+# Data:    vfio-platform hands the GPU + host1x/vic/nvdec/nvjpg/dce + three
```
brianmcgillion: **Suggested fix** — `dce` isn't in the passed-through `engines` list (compute-only), so drop it from the header comment.

```suggestion
# Data:    vfio-platform hands the GPU + host1x/vic/nvdec/nvjpg + three
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -0,0 +1,322 @@
...
+    type = lib.types.bool;
+    default = false;
+    description = "Pass the Tegra234 GPU and engines through to gpu-vm on NVIDIA Orin AGX";
+  };
+
+  config = lib.mkIf cfg.enable {
+    ghaf.hardware.nvidia.virtualization.host.bpmp.enable = true;
+
+    # Register the gpu-vm microvm now that its extraModules are populated below.
+    ghaf.virtualization.microvm.gpuvm.enable = true;
```
brianmcgillion, line 114: **Suggested fix — assert the BPMP safety invariant** (checklist M4). `bpmpAllowAllDomains` would bypass the closed allow-list; fail the build if it's ever set while `gpu_vm` is enabled. Verified: the AGX config still evaluates and the assertion passes.

```suggestion
    ghaf.virtualization.microvm.gpuvm.enable = true;

    # The closed BPMP host-proxy allow-list is the host-safety boundary for this
    # passthrough; bpmpAllowAllDomains would bypass it. Fail the build rather than
    # ship a guest that can reach every SoC clock/power domain.
    assertions = [
      {
        assertion = !config.ghaf.hardware.nvidia.virtualization.bpmpAllowAllDomains;
        message = "gpu_vm passthrough requires the closed BPMP allow-list; ghaf.hardware.nvidia.virtualization.bpmpAllowAllDomains must stay false.";
      }
    ];
```

brianmcgillion (commented):
**Patch suggestions for four of the checklist items** (from the two follow-up comments above).

Verified in a worktree at this PR head (`f83f0101`): `treefmt` clean, and the AGX `-from-x86_64` debug image rebuilds green (`exit 0`, valid store path) — `gpu-vm-load` recompiles from the patched `runner.c`, and the added assertion evaluates and passes.

The three inline **`suggestion`** blocks below are one-click "Commit suggestion":
1. `runner.c` — validate the duration arg (no more `GPU_LOAD_OK` on a bad/zero arg; also fixes the `atoi` UB) — checklist item + inline comment #9.
2. `gpu-vm/default.nix` header — drop `dce` from the "Data:" line (dce isn't passed through).
3. `gpu-vm/default.nix` — assert `bpmpAllowAllDomains == false` for the passthrough (the M4 guardrail).

Kept as review items (not clean single-hunk patches): the vendor `.patch`-content edits, dropping the "carried" display patch / disabling display DT nodes, the `aarch64` CI-check, module de-duplication, and the licensing/commit-body decisions.

Alternatively, apply **all four** fixes at once — the three above **plus** stripping the internal `ponytail:` / `Task N` / `#1031` tracker IDs from committed comments — with `git apply`:

```diff
diff --git a/modules/microvm/sysvms/gpuvm-base.nix b/modules/microvm/sysvms/gpuvm-base.nix
index ae8abce1..b39294ab 100644
--- a/modules/microvm/sysvms/gpuvm-base.nix
+++ b/modules/microvm/sysvms/gpuvm-base.nix
@@ -87,8 +87,8 @@ in
     };
 
     # GIVC configuration - from globalConfig.
-    # ponytail: no dedicated gpuvm givc role exists; enable transport only, no
-    # policyClient. Task 7's gpuvm.nix guards policy/spire wiring on those being
+    # no dedicated gpuvm givc role exists; enable transport only, no
+    # policyClient. gpuvm.nix guards policy/spire wiring on those being
     # enabled, so a plain client is the minimal correct thing. Add a role later
     # if the GPU VM needs to serve/receive GIVC commands.
     givc = {
@@ -147,8 +147,8 @@ in
   # driver (libcuda.so.1); l4t-tools ships tegrastats. cuda_cudart is the CUDA
   # runtime. This is the minimal runtime to prove compute once the GPU is
   # passed through.
-  # ponytail: no prebuilt CUDA sample (deviceQuery/vectorAdd) is packaged by
-  # nvidia-jetpack, so none is added here. If Task 10's smoke test needs to
+  # no prebuilt CUDA sample (deviceQuery/vectorAdd) is packaged by
+  # nvidia-jetpack, so none is added here. If a smoke test needs to
   # compile one on-device, add pkgs.nvidia-jetpack.cudaPackages.cuda_nvcc.
   environment.systemPackages =
     (with pkgs.nvidia-jetpack; [
@@ -160,7 +160,7 @@ in
       cuda_nvrtc
       libcublas
       # nvcc + a host compiler so a GPU compute load can be built on-device for
-      # the Task 10 GR3D_FREQ smoke test (see /etc/gpu-test/vectorAdd.cu).
+      # the GR3D_FREQ smoke test (see /etc/gpu-test/vectorAdd.cu).
       cuda_nvcc
     ])
     ++ [
@@ -205,7 +205,7 @@ in
   # libcuda plus the complete driver stack together; putting it first on the
   # loader path makes CUDA initialise the passed-through GPU.
   # Proven: __nvcc_device_query returns 87 (sm_87 / ga10b) with this set.
-  # ponytail: blunt global LD_LIBRARY_PATH, acceptable for a compute-only test
+  # blunt global LD_LIBRARY_PATH, acceptable for a compute-only test
   # VM. Move to a /run/opengl-driver runpath if this VM ever runs general apps.
   environment.variables.LD_LIBRARY_PATH = lib.mkForce (
     lib.makeLibraryPath [ pkgs.nvidia-jetpack.l4t-cuda ]
@@ -243,7 +243,7 @@ in
 
   microvm = {
     optimize.enable = false;
-    # ponytail: vcpu is pinned to 4 because tegra234-gpuvm.dts is generated for
+    # vcpu is pinned to 4 because tegra234-gpuvm.dts is generated for
     # 4 cores; changing it requires regenerating the DTS. mem is a default so a
     # profile/vmConfig can shrink it.
     vcpu = 4;
diff --git a/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix b/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
index d88dfa5c..7dd12f8e 100644
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
@@ -3,11 +3,11 @@
 #
 # Pass the AGX Orin's on-SoC GPU (ga10b) and supporting engines to gpu-vm.
 #
-# Data:    vfio-platform hands the GPU + host1x/vic/nvdec/nvjpg/dce + three
+# Data:    vfio-platform hands the GPU + host1x/vic/nvdec/nvjpg + three
 #          reserved-memory carveouts to the guest. The carveouts use mmio-base
 #          for 1:1 GPA=HPA (large regions; see ghaf-qemu-bpmp-gpu).
 # Control: clocks/resets/power-domains route through the BPMP host proxy, gated
-#          by the union allow-list (ids enumerated on hardware, Task 8).
+#          by the union allow-list (ids enumerated on hardware).
 #
 # Display (13800000.display) is carried but known non-functional (no video
 # output). It must not destabilize GPU compute.
@@ -121,6 +121,16 @@ in
     # Register the gpu-vm microvm now that its extraModules are populated below.
     ghaf.virtualization.microvm.gpuvm.enable = true;
 
+    # The closed BPMP host-proxy allow-list is the host-safety boundary for this
+    # passthrough; bpmpAllowAllDomains would bypass it. Fail the build rather than
+    # ship a guest that can reach every SoC clock/power domain.
+    assertions = [
+      {
+        assertion = !config.ghaf.hardware.nvidia.virtualization.bpmpAllowAllDomains;
+        message = "gpu_vm passthrough requires the closed BPMP allow-list; ghaf.hardware.nvidia.virtualization.bpmpAllowAllDomains must stay false.";
+      }
+    ];
+
     # GPU BPMP allow-list contribution. Raw bpmp ids read from the live host
     # device tree (od -t u4 --endian=big on each node's clocks/resets/
     # power-domains, keeping only the cells whose provider phandle is &bpmp).
@@ -135,7 +145,7 @@ in
     # would needlessly widen the host-safety boundary. As with MGBE0, the host
     # proxy logging "clock not allowed" for probed parent clocks at guest boot is
     # the boundary working, not a defect. If GPU init fails on a specific denied
-    # id (Task 10), add only that id here.
+    # id, add only that id here.
     ghaf.hardware.nvidia.virtualization.host.bpmp.allow = {
       clocks = [
         1
@@ -231,7 +241,7 @@ in
         { config, pkgs, ... }:
         {
           # Guest kernel = vanilla 6.12 + jetpack OOT overlay (Bring-Your-Own-
-          # Kernel), with #1031's GPU/display passthrough patches applied to
+          # Kernel), with the GPU/display passthrough patches applied to
           # nvidia-oot-modules (validated: base OOT builds clean on 6.12.93).
           boot.kernelPackages = lib.mkForce (
             (pkgs.linuxPackages_6_12.extend pkgs.nvidia-jetpack.kernelPackagesOverlay).extend (
diff --git a/packages/gpu-vm-load/runner.c b/packages/gpu-vm-load/runner.c
index 12e1818b..c14daf26 100644
--- a/packages/gpu-vm-load/runner.c
+++ b/packages/gpu-vm-load/runner.c
@@ -56,7 +56,17 @@ extern const char vadd_ptx[];
   } while (0)
 
 int main(int argc, char **argv) {
-  int secs = argc > 1 ? atoi(argv[1]) : 20;
+  int secs = 20;
+  if (argc > 1) {
+    char *end;
+    long v = strtol(argv[1], &end, 10);
+    if (*argv[1] == '\0' || *end != '\0' || v <= 0 || v > 86400) {
+      fprintf(stderr, "invalid duration '%s' (expected 1..86400 seconds)\n",
+              argv[1]);
+      return 2;
+    }
+    secs = (int)v;
+  }
   size_t n = (size_t)1 << 20;
 
   CK(cuInit(0));
```

vadika:
## Review comments addressed — force-pushed `7ad31e23`

Rebased onto current `main` (`ae4c973b`) and folded all fixes into the three feature commits (DCO-signed). Thread-by-thread:

### Blocking regressions (accidental reverts from a stale base) — fixed
- **`audit_backlog_wait_time=0`** and **`set-dynamic-hostname` ordering**: both files are now **byte-identical to `main`**. The stale revert hunks are gone (rebased onto `ae4c973b`, which carries `3c063c55` + `b42ae3a5`).

### Medium
- **Shared BPMP guest↔guest DoS**: added an explicit `SHARED BPMP TRUST DOMAIN` comment at the allow-list union point documenting that net-vm + gpu-vm are one BPMP trust domain (shared `/dev/bpmp-host`, id-only `check_if_allowed()`), plus a `!bpmpAllowAllDomains` assertion. Per-VM proxy noted as the follow-up to remove the residual reach.
- **EOL QEMU 10.1.5**: added a `LIFECYCLE / EOL` block to `ghaf-qemu-bpmp-gpu` documenting the frozen `-device vfio-platform` pin, the hardware-privileged blast surface, mitigations (keep gpu-vm off WAN-adjacent segments / firewalled), and the migrate-together-with-net-vm exit path.
- **Host-graphics coupling**: added a loud `warnings` entry so any build that enabled `gpu_vm` is told the host GUI (COSMIC), nvpmodel, and NVIDIA Docker are force-disabled and the host has no local GUI.
- **Missing eval-time guardrails**: added both — `!bpmpAllowAllDomains` (host scope) and a guest assertion requiring **both** `clk_ignore_unused` and `pd_ignore_unused` in kernelParams.

### Low
- **Drop `0003-add-support-for-display-passthrough.patch`**: removed. Verified it only patches the `nvdisplay/` display-driver tree (a separate module from the `nvgpu`/`nvidia` compute RM the VM loads); display isn't passed through and the module binds nothing, so the forced Xen/vgx predicates can't reach CUDA. Reverting to stock nvdisplay compiles clean.
- **`0002` host1x `device_iommu_mapped()` TODO**: resolved. Replaced the commented-out TODO + debug `printk` with an explicit VM-specific comment — the guest has no SMMU so `device_iommu_mapped()` is always false and would wrongly skip context isolation; guest memory is 1:1 GPA=HPA, so gating on engine capability is correct.
- **`runner.c` `atoi` UB / false `GPU_LOAD_OK`**: applied the `strtol` validation (reject non-numeric / ≤0 / >86400, exit 2), and added a `GPU_LOAD_FAIL` guard so it exits non-zero if zero kernel launches completed.

### Copilot items (already satisfied at head — no change needed)
- `cuDeviceGetAttribute` returns are already wrapped in the `CK()` macro (returns non-zero on failure; `sm_00` cannot silently print).
- The compute-verify doc section already mandates the unprivileged `ghaf` user and forbids `sudo`; remaining `sudo` uses are host-side flashing / serial `dmesg`.
- The `gpuvm.nix` header comment and assertion example already match the Orin `gpuvmBase` + `applyVmConfig` wiring at head.

Cross-build of `nvidia-jetson-orin-agx64-debug-from-x86_64` against these changes is running as the final gate; will confirm green.

brianmcgillion (approved):
Re-reviewed after the rework — thanks for addressing the feedback so thoroughly. The two blocking regressions are gone, the key guardrails are in, and I verified the reworked AGX image builds cleanly (`nvidia-jetson-orin-agx-debug-from-x86_64`, cross-built locally; both new assertions pass at eval). Approving.

**Resolved / verified:**
- The audit (`audit_backlog_wait_time=0`) and identity ordering-cycle reverts are gone — branch is now rebased onto current `main`, which was the root cause.
- Eval-time guardrails added: `assertion = !…bpmpAllowAllDomains` **and** a `clk_ignore_unused` / `pd_ignore_unused` kernelParams assertion.
- `runner.c` now validates the duration arg (no `GPU_LOAD_OK` on a bad/zero arg); `dce` dropped from the header comment.
- Dead display kernel patch `0003` removed; the `device_iommu_mapped()` change now carries a real rationale instead of a `TODO`; internal tracker IDs cleaned out of comments.

**Non-blocking, left to your judgement (fine to defer):**
- The BPMP host-proxy allow-list stays a single shared union over one `/dev/bpmp-host`, so a compromised gpu-vm and net-vm can reach each other's clock/reset ids — acceptable as a documented tradeoff; per-VM scoping would be a follow-up.
- `ghaf-qemu-bpmp-gpu` stays on the EOL-pinned QEMU 10.1.5 — worth tracking as a lifecycle issue since it now backs a second, GPU-privileged VM.
- Host-desktop disable is via `lib.mkForce` without an explicit assertion/opt-out (minor).

leivos-unikie:
@vadika
Just FYI, noticed now few weeks after merge that this PR has induced a jump in orin nx/agx ghaf-host latency measurements. It seems to be a new stable baseline.

Orin AGX (ghaf-host cyclictest avg latency)
image

Orin NX (ghaf-host cyclictest 1-thread 80 priority cyclictest histogram, pass criteria: max 10 overflows)
image

PR #2043, opened 2026-07-14
feat(gpu-vm): display to gpu-vm plumbing for Orin AGX

vadika:
## Description of Changes

Plumb the physical display to the `gpu-vm` guest on Jetson Orin AGX.

The display can't be passed through directly: mode setting is owned by the host's DCE R5 coprocessor, and vfio-binding `display@13800000` resets it under the R5 and halts the host. Instead the guest runs the full nvdisplay stack, and its DISPRM RPC traffic is relayed to the real DCE (guest proxy → QEMU bridge → host proxy) with a reverse doorbell injecting DCE's async events (hotplug/vblank/flip-completion) back into the guest. On top of that the guest renders on the passed-through GA10B and presents at a locked 60 Hz.

Architecture and bring-up notes: new docs page `ghaf/dev/technologies/nvidia_gpu_vm_display`.

**Works end-to-end on hardware**: the guest lights a physical panel (DP direct at 3440x1440@60, passive DP→HDMI up to 3440x1440@30). At cold boot the modules, anchors and connector come up automatically, but the panel stays dark until a DRM client is started **manually** (`modetest` below; there is no auto-modeset service yet). Once a mode is held, a cable replug relights it automatically.

### Pieces
- **Guest nvdisplay patches** — DCE address policy (all R5-facing addresses are CPU-physical, offset into the display high-IOVA range, ctxdma limit computed after the offset); core notifier forced to plain WRITE (WRITE_AWAKEN aborts completions the guest can't receive); DP++ dual-mode falls through to the TMDS partner; synthetic boot long-pulse (a panel connected at boot emits no HPD edge). Bring-up diagnostics gated behind `gpuvmDceDebug` (default off).
- **`dce-host-proxy.ko`** — relays guest DCE IPC to the R5 via tegra-dce's client API, plus a lossless reverse-doorbell event ring (peek-then-pop).
- **`dce-iso-anchor.ko`** — DMA-less anchors claiming the display ISO (SID 1) and NISO (SID 7) streams with translating high-IOVA→carveout domains, and retagging the scanout readers NVDISPLAYR/NVDISPLAYR1 to ISO SID 1 via MC SID-override registers. Boot-lifetime module: unload can't restore the overrides, so it has no exit path by design.
- **QEMU DCE bridge + ch3 event pump**, BPMP display-clock allow-list additions.

### GPU-accelerated present
- Jetson EGL/GLES/GBM userspace via `hardware.graphics` (guest BYO 6.12 kernel can't use `hardware.nvidia-jetpack`). `eglinfo`: `NVIDIA Tegra Orin (nvgpu)`, GLES 3.2 / GL 4.6.
- nvmap generic carveout (384 MB at 0x98000000) keeps EGL scanout surfaces below the FE's 39-bit ISO limit (higher surfaces bit-39-truncate → smmu_iso fault storm).
- egl-gbm 1.1.3 (nixpkgs) replaces l4t-gbm 1.1.0 (heap-corrupts under mesa 26), with a single-device fallback patch; stock `kmscube -D /dev/dri/card0` works via a small LD_PRELOAD no-modifier shim.
- Flip rate ~11 Hz → locked 60 Hz: (1) decouple completion from the R5's ~130ms-late event; (2) copy only the live RPC payload instead of the full 12 KB window (the bulk copy was ~3000 trapping MMIO accesses per RPC, held across the display RM lock). `modetest -v` steady 59.95–60.01 Hz; kmscube 57fps vsynced.

### Type of Change
- [x] New feature

## Related Issues / Tickets
Builds on #2036 (tegra234 GPU passthrough to gpu-vm).

### Applicable Targets
- [x] Orin AGX `aarch64`

### Installation Method

### Test Steps To Verify:
1. Build and flash `nvidia-jetson-orin-agx-debug-from-x86_64` (`...-flash-script`); boot with a DP monitor connected.
2. On the host, confirm the proxy and anchors loaded: `lsmod | grep dce` shows `tegra_dce`, `dce_host_proxy`, `dce_iso_anchor`; `dmesg | grep -E 'dce_iso_anchor|dce_niso_anchor'` shows the ISO and NISO carveout maps installed with result `= 0` (`iso map ... = 0`). The NVDISPLAYR/NVDISPLAYR1 scanout SID retag is confirmed functionally by step 4 — a failed retag shows up there as EMEM decode errors.
3. In gpu-vm, first list the outputs with `modetest -M nvidia-drm` (no `-s`): in the **Connectors** table pick the row marked `connected` — its id is `<connector>`, and any name in that connector's mode list is `<mode>` (e.g. `3440x1440`). Then hold a modeset: `sh -c 'sleep 7200 | modetest -M nvidia-drm -s <connector>:<mode>'` (e.g. `-s 32:3440x1440`) — the panel shows the SMPTE pattern.
4. Confirm zero `Lost display notification` in guest dmesg and zero `nvdisplayr.*EMEM address decode error` in host dmesg.
5. Replug the cable: the mode comes back without manual intervention. A passive DP->HDMI adapter also lights (TMDS partner, 30 Hz at ultrawide).
6. GPU-accelerated present: in gpu-vm run `eglinfo -B` (renderer `NVIDIA Tegra Orin (nvgpu)`) and `kmscube -D /dev/dri/card0` — the spinning cube renders on the physical panel.

**Note:** the host is intentionally headless (gpu_vm owns the GPU) — no host GUI after boot is expected. Display output comes from gpu-vm, per the steps above.

## Checklist
- [x] Commits logically organized and squashed if appropriate


milva-unikie:
**Tested on Orin AGX**

Built and flashed `nvidia-jetson-orin-agx-debug-from-x86_64-flash-script`

Device boots but there is no GUI. Ghaf logo is displayed during bootloader, but after that there is no output to the display. I was using DP.

vadika:
> **Tested on Orin AGX**
> 
> Built and flashed `nvidia-jetson-orin-agx-debug-from-x86_64-flash-script`
> 
> Device boots but there is no GUI. Ghaf logo is displayed during bootloader, but after that there is no output to the display. I was using DP.

There should be no GUI!
Have you proceed with testing instructions?

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -158,19 +216,110 @@ in
...
+      # PLL tree, RG/SOR/SF paths, mipi-cal, osc, dsc, maud, aza.
+      ++ [
+        # Root parents of the SOR clock tree. The guest's clk_prepare_enable
+        # on sor0 propagates enables up to these; a denied parent fails the
+        # WHOLE chain, so the SOR never turns on (no video signal despite a
+        # completed modeset -- observed as CLK_ENABLE (cmd 7) denials for 14
+        # and 102 at display init). Host-critical always-on roots: a guest
+        # enable is a BPMP refcount no-op, and the guest runs
+        # clk_ignore_unused so it never mass-disables them.
+        14 # TEGRA234_CLK_CLK_M
```
brianmcgillion, line 212: **[security / allow-list regression — confirmed]** This adds `CLK_M` (14) and `PLLP_OUT0` (102) — your own comment calls them "Host-critical always-on roots" — plus ~60 display clocks/resets/PDs to the guest-reachable `host.bpmp.allow`. `check_if_allowed()` in `bpmp-host-proxy.c` gates `MRQ_CLK` by **id only, not command**, so once an id is allowed a compromised gpu-vm can issue any clock command (e.g. DISABLE) on it. Adding shared SoC roots the host TCB depends on therefore hands a guest a host-DoS/instability primitive — regressing the closed-allow-list discipline #2036 established.

**Fix:** don't add host-critical/shared roots to the guest allow-list. If the guest genuinely needs `ENABLE`/`GET_*` on them, extend `check_if_allowed()` to gate `MRQ_CLK` by `(id, command)` and permit only the safe commands, rather than widening the id set.

modules/reference/hardware/jetpack/agx/orin-agx.nix
```diff
@@ -84,6 +93,22 @@
   # To enable or disable wireless
   networking.wireless.enable = true;
 
+  # DIAGNOSTIC (AST verification): relax STRICT_DEVMEM so the notifier carveout
```
brianmcgillion: **[security / ship-blocker — confirmed]** This disables `STRICT_DEVMEM` + `IO_STRICT_DEVMEM` and passes `iomem=relaxed` on the **host** — the TCB that runs every microVM (net-vm, gui-vm, …). With `STRICT_DEVMEM` off, any host-root process can read/write arbitrary physical memory via `/dev/mem`, materially weakening host isolation. Your own comment labels this block "DIAGNOSTIC (AST verification) … Drop for release," yet it ships unconditionally. (Same for the guest kernel at `gpu-vm/default.nix:684`.)

**Fix:** remove these before merge, or gate them behind a default-off debug flag (mirroring `gpuvmDceDebug`) so the release build keeps `STRICT_DEVMEM` on and doesn't pass `iomem=relaxed`. Update the doc's verify section so it doesn't require `/dev/mem` access.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-host-proxy/dce-host-proxy.c
```diff
@@ -0,0 +1,846 @@
...
+/*
+ * DIAGNOSTIC: translate display surface IOVAs through the passed-through
+ * display's VFIO IOMMU domain (group 15), via the safe iommu framework API
+ * (no MMIO -- unlike the register peek, this cannot fault the host). Shows
+ * exactly where the display hardware's DMA to each IOVA actually lands, so we
+ * can see whether the guest CPU's 1:1 MMIO view and the display's DMA mapping
+ * agree. The display DMA (SID_NVDISPLAY) shares the domain of the passed
+ * display-channel device 13870000.disp_chan_pt.
+ */
+static int dce_iova_show(struct seq_file *s, void *unused)
```
brianmcgillion: **[security / release-readiness — confirmed]** Roughly **224 of this file's 846 lines** are DIAGNOSTIC/EXPERIMENT scaffolding compiled unconditionally into the shipped `dce-host-proxy.ko` — none behind `gpuvmDceDebug` (which only gates the 6 guest-kernel debug patches): `dce_iova_show`, `dce_himap` (installs **arbitrary `iommu_map()`** into the passed-through display's VFIO domain), `dce_ast` (ioremaps + dumps the R5 register aperture), and `dce_peek` (an **arbitrary physical-memory read** primitive via debugfs). Two concrete problems beyond attack surface: (1) `dce_himap`'s debugfs file is created but **not removed on module unload → dangling-fops UAF**; (2) the functional relay uses none of this.

**Fix:** gate the entire diagnostic block behind a Kconfig/`#ifdef` that is off for release (matching the `gpuvmDceDebug` convention), or delete it. Do not ship `dce_peek`/`dce_himap` enabled.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-host-proxy/dce-host-proxy.c
```diff
@@ -0,0 +1,846 @@
...
+	mutex_unlock(&dce_host_event_client_lock);
+}
+
+#define BUF_SIZE DCE_CLIENT_MAX_IPC_MSG_SIZE
+
+/*
+ * Writes to the device
+ */
+
+static ssize_t write(struct file *filep, const char *buffer, size_t len, loff_t *offset)
```
brianmcgillion, line 394: **[duplication / generalization — the headline]** This `write()` fop is a line-for-line copy of `bpmp-host-proxy.c`'s `write()`: identical `kbuf/txbuf/rxbuf/usertxbuf/userrxbuf` locals, the `len > 65535` check, the `kmalloc`+`memset`+`copy_from_user`(1/2/3) bounce-buffer dance, the save-then-reassign userspace-pointer trick, the `copy_to_user`(2/3/1) sequence, and the exact `out_notok/out_nomem/out_cfu` goto ladder. The char-device probe/class/device scaffold is equally cloned, and the QEMU DCE bridge (`nvidia_dce_guest.c`) duplicates the BPMP bridge offset-for-offset.

**Design:** extract the char-device relay skeleton (chrdev/class/device lifecycle + the bounce-buffered `write()` fop + the reverse event ring) into **one shared module** parameterized by `{message struct size/layout, max_msg bound, transfer callback, optional allow-list filter}`, and instantiate it for both BPMP and DCE. Large refactor; build-verifiable, but its runtime validation needs the hardware path — happy to deliver it as a verified patch once that's available.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-iso-anchor/dce-iso-anchor.c
```diff
@@ -0,0 +1,160 @@
...
+#include <linux/module.h>
+#include <linux/of.h>
+#include <linux/of_platform.h>
+#include <linux/platform_device.h>
+#include <linux/iommu.h>
+#include <linux/delay.h>
+
+#include "dce-iso-anchor-dtbo.h"
+
+#define DCE_HI_BASE 0x7f00000000ULL
```
brianmcgillion, line 43: **[duplication / correctness-adjacent]** The single address-translation policy — the guest carveout + `0x7f00000000` high-IOVA offset, mapped hi→carveout — has **no source of truth**. `DCE_HI_BASE` is redefined in `dce-host-proxy.c`, here in `dce-iso-anchor.c`, as three raw literals in patch `0006`, twice more in a debug patch, and referenced in `tegra234-gpuvm.dts`; the `dce_hi_ranges[]` carveout table is likewise duplicated. Because this is what confines DCE DMA, a drift between any two copies is a silent isolation bug.

**Fix:** define `DCE_HI_BASE` and the carveout range table once in a shared header included by both `dce-host-proxy.c` and `dce-iso-anchor.c` (they build in the same nvidia-oot tree), and derive the nvdisplay-patch offsets and the DTS values from the same source.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-host-proxy/dce-host-proxy.c
```diff
@@ -0,0 +1,846 @@
...
+	copy = offsetof(struct dce_host_event, data) + e->size;
+	if (copy_to_user(buffer, e, copy)) {
+		pr_warn_ratelimited("dce_host_proxy: event copy_to_user failed (len=%zu copy=%zu), event retained\n",
+				    len, copy);
+		kfree(e);
+		return -EFAULT;
+	}
+
+	spin_lock_irqsave(&dce_evt_lock, flags);
+	if (dce_evt_head != dce_evt_tail)
```
brianmcgillion: **[correctness / race — confirmed]** The pop guard here is `if (dce_evt_head != dce_evt_tail) dce_evt_tail = (dce_evt_tail+1)%SLOTS`, which only proves the ring is non-empty — not that `dce_evt_tail` still points at the slot the reader peeked at line ~595. The producer (`dce_host_proxy_queue_event`, the atomic IPC callback) **also advances `dce_evt_tail`** on the full-ring drop path. Sequence: reader peeks slot T and unlocks for `copy_to_user`; the producer bursts the ring full and drops T→T+1 (counted); the reader re-locks, sees non-empty, and pops T+1→T+2 — silently discarding slot T+1, which was never delivered and never counted. Losing a flip-completion (`RM_NOTIFY`) hangs the guest modeset for the ~3 s timeout — the exact failure this peek-then-pop was meant to prevent.

**Fix:** capture the peeked tail under the first lock (`my_tail = dce_evt_tail`) and pop only if unchanged: `if (dce_evt_tail == my_tail) dce_evt_tail = (my_tail + 1) % DCE_EVT_SLOTS;`. Also correct the "single reader … not contended" comment — the producer mutates `tail` on drop.

brianmcgillion (changes requested):
Deep C + security review of the DCE display plumbing. **Stacked on #2036** (merge that first). Verified locally: the **AGX display image builds** (`nvidia-jetson-orin-agx-debug-from-x86_64`, cross-built, exit 0) — the patched kernel, QEMU DCE bridge, the DCE proxies, iso-anchor, and rewritten DTS all compile. (Runtime is hardware-only, as you tested.)

**Credit where due — and a false-alarm I cleared:** the core DMA-isolation design is sound. My review's first pass flagged a "guest→host DMA-anywhere via the opaque relay" as high severity, but adversarial verification **refuted** it (and two related QEMU-OOB claims): `dce-iso-anchor.c` binds *translating* SMMU domains (ISO SID 1 / NISO SID 7) that confine DCE DMA to the guest carveout, so an out-of-carveout address faults rather than resolving. The proxy correctly delegates confinement to SMMU/AST. Good.

### The headline: this reads as a bring-up branch, not release-ready
Several security-weakening diagnostics that **your own comments mark "Drop for release"** ship **enabled** by default:
- **Host** `STRICT_DEVMEM`/`IO_STRICT_DEVMEM` off + `iomem=relaxed` (inline) — on the TCB that runs every microVM. Any host-root process can then read/write arbitrary physical memory via `/dev/mem`.
- **Guest** `STRICT_DEVMEM` off (`gpu-vm/default.nix:684`) — same class, guest-internal.
- ~**224 of 846 lines** of `dce-host-proxy.c` are ungated diagnostic scaffolding compiled into the shipped `.ko` — including `dce_peek` (**arbitrary physical-memory read** debugfs) and `dce_himap` (installs **arbitrary `iommu_map()`** into the display VFIO domain; its debugfs file is also **not removed on unload → dangling-fops UAF**). None of this is behind `gpuvmDceDebug` (which correctly gates the 6 *guest* debug patches).
- Patch `0003` forces every nvdisplay `nv_printf` on (`if (1)`, `cur_debuglevel=1`) in the unconditional functional patch (same issue we flagged in #2036).

**Recommendation:** gate all of the above behind `gpuvmDceDebug` (off for release), matching your existing convention — or delete. The functional relay uses none of it.

Two safe fixes: correct the stale `0003 … intentionally NOT applied … dead, risky code` comment (0003 *is* applied here for display — 12 lines below), and strip the `Task 1d` internal tracker IDs from `orin-agx.nix` / `dce-probe-host.nix`. Verified: treefmt clean, eval green, applies clean.

```diff
diff --git a/modules/reference/hardware/jetpack/agx/orin-agx.nix b/modules/reference/hardware/jetpack/agx/orin-agx.nix
index 93e76086..40edb97c 100644
--- a/modules/reference/hardware/jetpack/agx/orin-agx.nix
+++ b/modules/reference/hardware/jetpack/agx/orin-agx.nix
@@ -9,7 +9,7 @@
 
   imports = [
     ../../../../common/services/hwinfo
-    # Task 1d DCE display-proxy HOST integration (AGX only). Keeps gpu_vm
+    # DCE display-proxy HOST integration (AGX only). Keeps gpu_vm
     # enabled, runs the host headless, force-loads tegra-dce so the host owns
     # the real DCE R5, and builds + loads the dce-host-proxy .ko (with an
     # injected nvidia,dce-host-proxy DT node) so the guest can drive the panel
diff --git a/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/dce-probe-host.nix b/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/dce-probe-host.nix
index 77ee48d5..5f50ff52 100644
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/dce-probe-host.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/dce-probe-host.nix
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
 #
-# Task 1d -- DCE display-proxy HOST integration (AGX only).
+# DCE display-proxy HOST integration (AGX only).
 #
 # The companion to gpu-vm's guest overlay (../../passthrough/gpu-vm). The guest
 # takes display@13800000 and drives the panel through a synthetic DCE proxy; the
diff --git a/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix b/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
index e8730c30..8a56caa7 100644
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
@@ -474,10 +474,10 @@ in
           # Guest kernel = vanilla 6.12 + jetpack OOT overlay (Bring-Your-Own-
           # Kernel), with the GPU passthrough patches applied to
           # nvidia-oot-modules (validated: base OOT builds clean on 6.12.93).
-          # 0003 (display passthrough) is intentionally NOT applied: display is
-          # out of scope for this compute-only VM, and it force-returns the Xen
-          # dom0/vgx virtualization predicates TRUE -- dead, risky code in a
-          # compute build.
+          # 0003 (display passthrough) IS applied here for the display path. Note
+          # it force-returns the Xen dom0/vgx virtualization predicates TRUE, so
+          # that behavior now ships in the guest -- confirm it is intended (and
+          # scope its debug-print forcing behind gpuvmDceDebug, see review).
           boot.kernelPackages = lib.mkForce (
             (pkgs.linuxPackages_6_12.extend pkgs.nvidia-jetpack.kernelPackagesOverlay).extend (
               _final: prev: {

```

### Duplication / generalization (your headline emphasis) — designs
- **DCE proxy is a near-verbatim fork of the BPMP proxy** (inline on `dce-host-proxy.c`): the `write()` fop is line-for-line the BPMP proxy's, the char-device scaffold is cloned, and the QEMU DCE bridge duplicates the BPMP bridge offset-for-offset. → extract **one shared char-device relay skeleton** parameterized by `{message layout, max bound, transfer callback, optional allow-list filter}` used by both DCE and BPMP. Large refactor; build-verifiable but runtime-validated only on hardware.
- **`DCE_HI_BASE 0x7f00000000` + the carveout range table are hardcoded in 8+ places** (inline) — a security-relevant address policy with no single source of truth. → one shared header, derive the patch offsets and the DTS from it.

These (and the debug-gating, the event-ring race fix, the `dce_himap` UAF fix) I've flagged as ready designs rather than applied patches, since the authoritative validation for runtime/concurrency changes is on-hardware.

### Also
- **H1 (inline):** the display delta adds host-critical root clocks (`CLK_M`, `PLLP_OUT0`) to the guest-reachable BPMP allow-list — a regression of the closed-allow-list discipline #2036 established.
- **Correctness (low):** confirmed reverse-doorbell event-ring race (inline), iso-anchor no-rollback on partial init, a `0003` page leak, a QEMU MMIO 8-byte over-read.
- **Hygiene:** the host DCE integration imports unconditionally (decoupled from `gpu_vm.enable`); dead files committed (`dce-probe-test.c`, orphaned debug patch `0007`); the new display doc omits the guest→host DCE trust-boundary caveat; **no automated tests** — a NixOS eval-time assertion (host-critical clocks absent from the allow-list; `gpuvmDceDebug` off ⇒ no diagnostics) is feasible.

vadika:
Addressed in 52a5e8c0 — dropped the shipped bring-up diagnostics rather than gating them (the functional relay uses none of them):

- Host `iomem=relaxed` + `STRICT_DEVMEM`/`IO_STRICT_DEVMEM` off, and guest `STRICT_DEVMEM` off — removed.
- `dce-host-proxy.c`: removed `dce_peek`, `dce_himap` (and with it the `dce_himap_dbg` dangling-fops UAF — it was never in the unload path), `dce_ast`, `dce_iova`, and the first-transfers XFER trace (−255 lines).
- `dce-guest-proxy.c`: removed the symmetric SEND/RECV trace.
- Patch `0003`: reverted the debug-print forcing (`cur_debuglevel=1`, `if (1)`) to stock `nv_printf` gating.

Your verified patch applied too: corrected the stale 0003 comment (it *is* applied for the display path) and stripped the task tracker IDs.

Net: 7 files, +8 / −316. eval green, treefmt clean; `gpuvmDceDebug` still gates the guest debug patches.

milva-unikie:
> There should be no GUI! Have you proceed with testing instructions?

What does this mean then?
```
Works end-to-end on hardware: the guest lights a physical panel (DP direct at 3440x1440@60, passive DP→HDMI up to 3440x1440@30), surviving host cold boot and cable replug with no manual steps.
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-host-proxy/dce-host-proxy.c
vadika: Fixed in 7594311d. `read()` now records the peeked slot index and advances the tail only if it still points at that slot (`if (dce_evt_tail == slot)`). So a producer full-ring drop during the `copy_to_user` window can no longer make the reader pop — and silently discard — a *different*, undelivered event; the delivered event's data is already a snapshot copied under the lock. Re-flashed and validated on hardware: 0 lost display notifications, panel modeset + kmscube both drive.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
vadika: Fixed in 7594311d. `check_if_allowed()` now command-gates the host-critical shared roots (14 `CLK_M`, 102 `PLLP_OUT0`): an allowed guest may still enable/read them, but `CMD_CLK_DISABLE` / `CMD_CLK_SET_RATE` / `CMD_CLK_SET_PARENT` are denied via `clk_root_is_protected()`. A compromised gpu-vm can no longer take a host clock down. Display-specific ids keep id-only gating (disabling those only affects the guest's own display). Re-flashed and validated on hardware: **0 protected-root denials during bring-up** (the guest only *enables* these roots), panel lights normally.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-iso-anchor/dce-iso-anchor.c
vadika: Addressed in 7594311d. The diagnostic deletion earlier in this branch already removed the `dce-host-proxy.c` and debug-patch copies, so `DCE_HI_BASE` + the carveout table now have a single host-side definition in `dce-iso-anchor.c`. The only remaining copy is the guest RM offset in patch `0006` — a separate build domain (guest nvdisplay source) that cannot include the host module's header. I marked `dce-iso-anchor.c` as the canonical definition and added reciprocal keep-in-sync notes there and in patch `0006`'s header. (The `.dts` reference is a comment, not a value.)

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-host-proxy/dce-host-proxy.c
vadika: Acknowledged — deferring this one deliberately. The shared char-device relay lives in `bpmp-host-proxy.c`, which the already-merged mgbe0/net-vm passthrough (#2033) also ships. Extracting a common relay would refactor a shipped, working path for a DRY win (not a bug) and needs dual-VM (gpu-vm + net-vm) hardware re-validation. Better as its own follow-up than folded into this display PR. Leaving open to track.

vadika:
> > There should be no GUI! Have you proceed with testing instructions?
> 
> What does this mean then?

This means manual steps needed as described in the testing instructions.  fixed the instructions, they were misleading, I'm sorry (

brianmcgillion (approved):
Re-reviewed after the rework — all of the changes-requested blockers are resolved and I verified the reworked AGX display image builds cleanly (`nvidia-jetson-orin-agx-debug-from-x86_64`, cross-built locally). In particular the host-critical clock roots are now safely gated by command in `bpmp-host-proxy.c`, the "Drop for release" diagnostics (host/guest `STRICT_DEVMEM`, the `dce-host-proxy` debugfs incl. the arbitrary-physmem/`iommu_map` primitives, the `patches/debug/` set, the `0003` printf-forcing) are gone, and the reverse-doorbell event-ring race is fixed. Approving.

The following items from the review were intentionally **not** done — all non-blocking, recorded here so they aren't lost as follow-ups:

**Generalization (flagged as designs, deferred by agreement):**
- `dce-host-proxy.c` / `dce-guest-proxy.c` / the QEMU DCE bridge are still a near-verbatim fork of the BPMP host/guest proxy + bridge. A shared char-device relay skeleton (parameterized by message layout / max bound / transfer callback / optional allow-list filter) would let BPMP and DCE share one implementation.
- `DCE_HI_BASE` (`0x7f00000000`) and the carveout range table are still hardcoded across ~12 sites (proxy, iso-anchor, patches, DTS) with no single source of truth — a drift risk for a security-relevant address policy. Recommend one shared header the C and DTS derive from.

**Lower-severity items I did not re-verify in this pass (may still be open):**
- `dce-iso-anchor` probe not rolling back earlier `iommu_map()`s on a mid-loop failure (partial-init).
- QEMU DCE bridge MMIO read dereferencing 8 bytes ignoring access size / end-of-region.
- `0003` `nv_alloc_coherent_pages` allocating 2 pages but tracking 1 (leak).
- The new display doc could add an explicit note that the guest drives the real DCE R5 via the host relay (guest→host trust boundary), with confinement provided by the `dce-iso-anchor` translating SMMU domains.

**Tests:** still hardware-only — no automated coverage (a `dtc` compile + the image build are the only CI signals). Not a merge blocker, but worth a follow-up.

brianmcgillion:
clutching all talismans that I have. lets hope this is self contained

PR #2054, opened 2026-07-18
refactor(gpu-vm): split guest DTS into component dtsi files

vadika:
**Stacked on #2043 — review only the top commit (`532b1322`).**

## Description of Changes

Behavior-preserving split of the monolithic gpu-vm guest DTS into a composition root plus six `.dtsi` fragments (base / memory / proxies / display / engines / dummies) and a `generated/` DCB include. The `gpuvm-dtb` derivation re-extracts `nvidia,dcb-image` from the final DTB and fails the build unless it matches the pinned stock R36.5 P3737/P3701 payload (8407 bytes, sha256 `e0d92e6d…`). Deviation from the refactor handoff: the DCB is a committed pinned include verified at build time, not extracted from stock sources during the build.

Equivalence: sorted `dtc` decompiles of the old and new DTBs are byte-identical (all nodes, properties, phandles); only sibling node order differs. Hardware-validated on the AGX Orin devkit: gk20a binds, CUDA reports sm_87, DP-1 modeset holds with zero display/SMMU faults, and an A/B boot against the pre-split DTB shows identical DRM device mapping. Stale platform-bus comments (old full display/DCE VFIO wording) corrected to the keyhole + DCE-proxy reality.

### Type of Change

- [ ] New feature
- [ ] Bug fix
- [ ] Update
- [x] Improvement / Refactor
- [x] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

Stacked on #2043.

## Checklist

- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/ (display architecture doc added in parent PR; no documented behavior changes here)

### Applicable Targets

- [x] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. Build: `nix build .#packages.x86_64-linux.nvidia-jetson-orin-agx-debug-from-x86_64` (the `gpuvm-dtb` derivation fails if the DCB payload drifts).
2. On the guest: `ls /dev/nvgpu/igpu0`; `__nvcc_device_query` prints `87`.
3. `modetest -M nvidia-drm -c` shows the connector with modes; `modetest -M nvidia-drm -s <conn>@<crtc>:#0` holds SMPTE bars.


modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -123,16 +136,41 @@ let
...
+          -I. \
           tegra234-gpuvm.dts > preprocessed.dts
         dtc -I dts -O dtb -o tegra234-gpuvm.dtb preprocessed.dts
+
+        # Verify the DCB payload that actually landed in the DTB against the
+        # pinned stock AGX blob; fail the build on any drift. Skipped in the
+        # host1x-experiment modes that drop the display node (EXP_DROP_DISPLAY):
+        # no display node -> no DCB to check. The guard keeps this a no-op for
+        # `off` (display present) so the baseline DTB stays byte-identical.
+        if fdtget tegra234-gpuvm.dtb \
```
brianmcgillion: **[robustness — non-blocking]** This guard conflates two different situations: "the display node was intentionally dropped" (the `EXP_DROP_DISPLAY` case this is meant to skip) and "the extraction failed for some other reason" (node renamed, path typo, `fdtget` missing). Any non-zero exit from `fdtget` silently skips the whole pin check and the build still goes green — so a future edit that moves or renames `display@13800000` would turn this supply-chain check into a no-op without anyone noticing.

Real risk today is low (`fdtget` ships with `dtc`, already in `nativeBuildInputs`, and the node path is correct — I confirmed the gate actually runs and matches). But since the check exists precisely to catch drift, it would be worth keying the skip on the same condition that drops the node (e.g. gate on the experiment flag) rather than on "fdtget returned non-zero", or asserting the node's presence in the default display-on build.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -123,16 +136,41 @@ let
...
+             /platform-bus@70000000/display@13800000 nvidia,dcb-image >/dev/null 2>&1; then
+          fdtget -t bx tegra234-gpuvm.dtb \
+            /platform-bus@70000000/display@13800000 nvidia,dcb-image \
+            | tr ' ' '\n' | grep . > dcb.hex
+          dcbLen=$(wc -l < dcb.hex)
+          while read -r b; do printf "\x$(printf %02x "0x$b")"; done < dcb.hex > dcb.bin
```
brianmcgillion: **[readability/efficiency — non-blocking]** This rebuilds the 8407-byte blob one byte at a time, running `printf` twice per byte (~16k execs) in a shell loop, where `xxd -r -p` does the whole conversion in a single pass:

```bash
fdtget -t bx tegra234-gpuvm.dtb \
  /platform-bus@70000000/display@13800000 nvidia,dcb-image \
  | tr -s ' \n' '\n' | grep . | xxd -r -p > dcb.bin
```

The inner `printf %02x "0x$b"` round-trip is also a no-op once a real hex-decoder is doing the parsing. I checked both forms against the committed hex and they produce byte-identical output with the same sha256, so this is purely a legibility/speed cleanup — the only cost is adding `xxd` to `nativeBuildInputs`. Cosmetic; take it or leave it.

brianmcgillion (approved):
## Review — behaviour-preserving DTS split (recheck after rebase)

Rechecked at head `a1dc7367`, rebased onto current `main` (so #2043 is merged and out of the diff; this PR is now just the 9-file DTS/DTSI + Nix split). I verified the central claim rather than eyeballing it:

- **Byte-equivalence confirmed.** Built `gpuvm-dtb` at the merge-base and at this head; the sorted `dtc` decompiles (`dtc -I dtb -O dts -s`) are **identical** — the two DTBs differ only in sibling node order, exactly as the description states.
- **DCB pin is real.** Reproduced the build's extract-and-hash pipeline on the committed hex: 8407 bytes, sha256 `e0d92e6d…`, matching the pin.
- **Builds green** both ways: `nix build .#packages.x86_64-linux.nvidia-jetson-orin-agx-debug-from-x86_64` passes with the DCB gate active, and `.#checks.x86_64-linux.pre-commit` (incl. `reuse lint`) is clean; all new files carry SPDX headers.

Nice clean cut — fragments map to natural node groups, each re-opens `platform-bus@70000000`, and nothing that's edited together got split apart. **Approving.**

Two non-blocking notes on the DCB verification block, inline below.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
vadika: Fixed in 302be85b. Inverted the guard: on this baseline branch the display node is always present, so a failed extraction (renamed/moved node, missing `fdtget`) now fails the build loudly instead of silently skipping the pin check. The experiment modes that legitimately drop the node gate their skip on the eval-time `dropDisplay` flag, not on `fdtget`'s exit code — so drift in the display-on build can no longer turn the gate into a no-op.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
vadika: Done in 302be85b — swapped the ~16k-exec byte loop for `tr -s ' \n' '\n' | grep . | xxd -r -p` in a single pass, `xxd` added to `nativeBuildInputs`. Verified byte-identical against the committed DCB hex: 8407 bytes, sha256 `e0d92e6d…` = the pin.

PR #2055, opened 2026-07-19
feat(orin): two-VM host1x ownership split (GPU compute + software display)

vadika:
## Description of Changes

Adds an Orin AGX two-VM topology where **physical host1x belongs to exactly one
VM** while the other still performs its role, resolving the host1x-ownership
question on hardware.

A branch-only selector `ghaf.hardware.nvidia.passthroughs.gpu_vm.host1xExperiment`
(`off` | `compute-no-host1x` | `display-no-host1x` | `compute-with-host1x`) gates
VFIO engine/carveout/keyhole ownership, guest kernel-module load, and cpp `-D`
flags that strip nodes from the guest device tree. **`off` is byte-identical to
the baseline.** A second `disp-vm` microvm provides the display role.

**Measured verdict (Orin AGX):**
- GPU compute **without** host1x: **FAIL** — `cuInit` returns `801 NOT_SUPPORTED`
  (3/3). CUDA requires physical host1x.
- Software display **without** host1x: **PASS** — no-syncpoint NVKMS lights the
  physical panel.
- **Concurrent split: PASS** — the `compute-with-host1x` GPU VM runs CUDA
  (`sm_87`, `gpu-vm-load` ~229–243 iters) at the same instant `disp-vm` holds a
  3440×1440 modeset with color bars on the physical DP-1 monitor, with **disjoint
  IOMMU groups**. Held through ~20 min soak and a GPU-VM restart (display
  survived; CUDA recovered).

Six enabling fixes are included (see commits): split-RAM disp-vm layout
(preserves the host DMA32 hole so PCIe MSI / net-vm WLAN keep working); a distinct
SMMU StreamID (`0x7d`) on the DMA-less display wrappers for a separate IOMMU
group; dropping `reusable` from the out-of-RAM `dce_scanout` node (fixes a
deterministic CMA `pfn_valid` panic); mapping the disp-vm scanout CMA at
`0xe2000000` into both the ISO SID-1 and NISO SID-7 DCE anchors (fixes the SID-7
scanout fault); and making the QEMU DCE guest bridge **opt-in** via
`GHAF_DCE_GUEST` so only the display VM opens `/dev/dce-host` (stops other VMs
stealing its DCE events).

### Type of Change

- [x] New feature
- [x] Documentation

## Checklist

- [x] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed (DCO sign-off on all commits)
- [x] Ghaf documentation updated with the commit
- [x] Author has added reviewers and removed PR draft status — **kept as draft**

### Applicable Targets

- [x] Orin AGX `aarch64`

### Installation Method

- [x] Requires full re-installation (deployed here as a persistent ESP generation)

### Test Steps To Verify:

1. Build the AGX debug toplevel from this branch:
   `nix build .#nixosConfigurations."nvidia-jetson-orin-agx-debug-from-x86_64".config.system.build.toplevel`.
2. Set `nvidia.passthroughs.gpu_vm.host1xExperiment = "compute-with-host1x"` and
   `nvidia.passthroughs.disp_vm.enable = true` (already set in `orin-agx.nix` on
   this branch); flash/boot the AGX.
3. Confirm three VMs active concurrently (`microvm@{net,gpu,disp}-vm`) with
   disjoint IOMMU groups for `17000000.gpu` vs the disp-vm scanout devices.
4. On the GPU VM run `gpu-vm-load` → expect `GPU_LOAD_OK`, `sm_87`.
5. On the display VM, hold a modeset:
   `sleep infinity | modetest -M nvidia-drm -s <DP-1 conn>@<crtc>:1920x1080` →
   expect color bars on the physical monitor, zero SID-7 SMMU faults.
6. Set `host1xExperiment = "off"` and confirm the built DTB/closure is
   byte-identical to the baseline (no behavior change).

### DIFF
--- a/modules/microvm/sysvms/dispvm-base.nix
+++ b/modules/microvm/sysvms/dispvm-base.nix
@@ -1,0 +1,208 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Disp VM Base Module
+#
+# Base config for the display-passthrough VM (Orin AGX,
+# experiment/orin-two-vm-host1x). Like gpuvm-base but without GPU-compute
+# (CUDA, gpu-vm-load, /dev/nvgpu): disp-vm owns only display/scanout
+# passthrough via ghaf.hardware.definition.dispvm.extraModules.
+#
+# Usage in profiles:
+#   lib.nixosSystem {
+#     modules = [ inputs.self.nixosModules.dispvm-base ];
+#     specialArgs = { inherit globalConfig hostConfig; };
+#   }
+#
+# Then extend with:
+#   base.extendModules { modules = [ ... ]; }
+#
+{
+  lib,
+  inputs,
+  globalConfig,
+  hostConfig,
+  ...
+}:
+let
+  vmName = "disp-vm";
+  timezoneEnabled = lib.ghaf.features.isEnabledFor globalConfig "timezone" vmName;
+in
+{
+  _file = ./dispvm-base.nix;
+
+  imports = [
+    inputs.preservation.nixosModules.preservation
+    inputs.self.nixosModules.givc
+    inputs.self.nixosModules.hardware-x86_64-guest-kernel
+    inputs.self.nixosModules.vm-modules
+    inputs.self.nixosModules.profiles
+  ];
+
+  ghaf = {
+    # Profiles - from globalConfig
+    profiles.debug.enable = lib.mkDefault (globalConfig.debug.enable or false);
+
+    development = {
+      ssh.daemon.enable = lib.mkDefault (globalConfig.development.ssh.daemon.enable or false);
+      debug.tools.enable = lib.mkDefault (globalConfig.development.debug.tools.enable or false);
+      nix-setup.enable = lib.mkDefault (globalConfig.development.nix-setup.enable or false);
+    };
+
+    # Networking hosts - from hostConfig
+    # Required for vm-networking.nix to look up this VM's MAC/IP
+    networking.hosts = hostConfig.networking.hosts or { };
+
+    # Common namespace - from hostConfig
+    common = hostConfig.common or { };
+
+    # User configuration - from hostConfig
+    users = {
+      profile = hostConfig.users.profile or { };
+      admin = hostConfig.users.admin or { };
+      managed = hostConfig.users.managed or { };
+    };
+
+    # Enable dynamic hostname export for VMs
+    identity.vmHostNameExport.enable = true;
+
+    # System
+    type = "system-vm";
+
+    systemd = {
+      enable = true;
+      withName = "dispvm-systemd";
+      withLocaled = true;
+      withNss = true;
+      withResolved = true;
+      withTimesyncd = true;
+      withDebug = globalConfig.debug.enable or false;
+      withHardenedConfigs = true;
+    };
+
+    # GIVC - from globalConfig. ponytail: no dispvm givc role; enable
+    # transport only, like gpuvm-base.
+    givc = {
+      enable = globalConfig.givc.enable or false;
+      debug = globalConfig.givc.debug or false;
+    };
+
+    # Storage - from globalConfig
+    storagevm = {
+      enable = true;
+      name = vmName;
+      encryption.enable = globalConfig.storage.encryption.enable or false;
+    };
+
+    virtualization.microvm = {
+      swap.enable = true;
+
+      vm-networking = {
+        enable = true;
+        inherit vmName;
+      };
+
+      tpm.emulated = {
+        # aarch64: TPM passthrough is x86-only, so emulate when encryption on.
+        enable = globalConfig.storage.encryption.enable or false;
+        name = vmName;
+      };
+    };
+
+    # Logging - from globalConfig
+    logging = {
+      inherit (globalConfig.logging) enable listener;
+      journalClient = {
+        inherit (globalConfig.logging) enable;
+      };
+    };
+
+    security = {
+      fail2ban.enable = globalConfig.development.ssh.daemon.enable or false;
+      audit.enable = lib.mkDefault (globalConfig.security.audit.enable or false);
+
+      spire.agents.downstream = {
+        enable = globalConfig.spire.enable or false;
+        logLevel = if globalConfig.spire.debug then "DEBUG" else "INFO";
+        nodeAttestationMode = if globalConfig.givc.enable then "x509pop" else "join_token";
+      };
+    };
+
+    services.timezone.enable = lib.mkDefault (
+      timezoneEnabled && globalConfig.platform.timeZone == null
+    );
+  };
+
+  time.timeZone = lib.mkIf (!timezoneEnabled) (lib.mkDefault globalConfig.platform.timeZone);
+
+  system.stateVersion = lib.trivial.release;
+
+  nixpkgs = {
+    buildPlatform.system = globalConfig.platform.buildSystem or "x86_64-linux";
+    hostPlatform.system = globalConfig.platform.hostSystem or "aarch64-linux";
+  };
+
+  microvm = {
+    optimize.enable = false;
+    # ponytail: vcpu=4 matches tegra234-dispvm.dts cpus node; mem is a
+    # display-only default, smaller than GPU VM's 6000.
+    vcpu = 4;
+    mem = lib.mkDefault 4000;
+    hypervisor = "qemu";
+
+    shares = [
+      {
+        tag = "ghaf-common";
+        source = "/persist/common";
+        mountPoint = "/etc/common";
+        proto = "virtiofs";
+      }
+    ]
+    # Shared store (when not using storeOnDisk)
+    ++ lib.optionals (!(globalConfig.storage.storeOnDisk.enable or false)) [
+      {
+        tag = "ro-store";
+        source = "/nix/store";
+        mountPoint = "/nix/.ro-store";
+        proto = "virtiofs";
+      }
+    ];
+
+    writableStoreOverlay = lib.mkIf (
+      !(globalConfig.storage.storeOnDisk.enable or false)
+    ) "/nix/.rw-store";
+
+    qemu = {
+      machine =
+        {
+          x86_64-linux = "q35";
+          aarch64-linux = "virt";
+        }
+        .${globalConfig.platform.hostSystem or "aarch64-linux"};
+    };
+  }
+  // lib.optionalAttrs (globalConfig.storage.storeOnDisk.enable or false) (
+    let
+      compLevelSuffix = lib.optionalString (
+        globalConfig.storage.storeOnDisk.compression.level != null
+      ) ",${toString globalConfig.storage.storeOnDisk.compression.level}";
+    in
+    {
+      storeOnDisk = true;
+      storeDiskType = "erofs";
+      storeDiskErofsFlags = [
+        "-Eztailpacking"
+        "-Efragments"
+        "--workers=$(( (NIX_BUILD_CORES < 1 || NIX_BUILD_CORES > 4) ? 4 : NIX_BUILD_CORES ))"
+      ]
+      ++ {
+        lz4hc = [ "-zlz4hc${compLevelSuffix}" ];
+        zstd = [
+          "-zzstd${compLevelSuffix}"
+          "-E48bit"
+        ];
+      }
+      .${globalConfig.storage.storeOnDisk.compression.algorithm};
+    }
+  );
+}
--- a/modules/microvm/sysvms/dispvm.nix
+++ b/modules/microvm/sysvms/dispvm.nix
@@ -1,0 +1,93 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Disp VM Configuration Module
+#
+# globalConfig pattern: global settings via globalConfig specialArg,
+# host-specific (networking.hosts) via hostConfig. Self-contained; all
+# platforms use the evaluatedConfig pattern with a profile's dispvmBase.
+#
+# dispvmBase (like gpuvmBase) is exported by the orin profile, which also
+# wires dispvm.evaluatedConfig. Inert unless that wiring is present: enable
+# is only true when ghaf.hardware.nvidia.passthroughs.disp_vm.enable is true
+# (Orin AGX only); the assertion below fires if enable is set without it.
+{
+  config,
+  lib,
+  inputs,
+  ...
+}:
+let
+  vmName = "disp-vm";
+
+  cfg = config.ghaf.virtualization.microvm.dispvm;
+in
+{
+  _file = ./dispvm.nix;
+
+  options.ghaf.virtualization.microvm.dispvm = {
+    enable = lib.mkEnableOption "DispVM";
+
+    evaluatedConfig = lib.mkOption {
+      type = lib.types.nullOr lib.types.unspecified;
+      default = null;
+      description = "Pre-evaluated NixOS configuration for Disp VM set via profile's dispvmBase.extendModules.";
+    };
+
+    extraNetworking = lib.mkOption {
+      type = lib.types.networking;
+      description = "Extra Networking option";
+      default = { };
+    };
+  };
+
+  config = lib.mkMerge [
+    {
+      ghaf.virtualization.microvm.sysvm.vms.dispvm = {
+        inherit vmName;
+        inherit (cfg) enable evaluatedConfig extraNetworking;
+      };
+    }
+    (lib.mkIf cfg.enable {
+      assertions = [
+        {
+          assertion = cfg.evaluatedConfig != null;
+          message = ''
+            ghaf.virtualization.microvm.dispvm.evaluatedConfig must be set.
+            Use a profile that provides dispvmBase (orin).
+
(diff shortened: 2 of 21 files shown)


modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/gpu_passthrough_overlay.dts
```diff
@@ -78,24 +93,39 @@
             // (mmio-base=0xb0000000) into the guest for the nvdisplay scanout.
             scanout_p: scanout_p@b0000000 {
                 compatible = "nvidia,dummy";
-                iommus = <&smmu_niso0 TEGRA234_SID_PASSTHROUGH>;
+                iommus = <&smmu_niso0 0x7d> /* concurrent: distinct group from GPU, DMA-less wrapper */;
```
brianmcgillion: **[F2 maintainability / F1 correctness — please address]** Two things on these `0x7d` wrappers:

**F2 (magic StreamID):** every sibling here uses a symbolic `TEGRA234_SID_*` constant; this is a bare literal in the file that *defines the isolation boundary*. An undocumented StreamID is exactly how an isolation regression slips in later (someone reuses 0x7d, or it collides with a real SID on a future SoC/kernel rev). I confirmed 0x7d is currently unassigned on T234 (defined SIDs jump 0x72→0x7f) and ≠ `SID_PASSTHROUGH` (0x7f), so the split is real — but please give it a named constant (e.g. `GHAF_SID_DISPVM`) with a one-line "unassigned on T234, chosen for a group disjoint from GPU" rationale.

**F1 (claim accuracy):** this SID move from `SID_PASSTHROUGH` → `0x7d` is **unconditional** (static overlay), so in `off` mode scanout_p and the two keyholes still change IOMMU group. The guest DTB stays byte-identical (DCB-gated), but the *host closure* does not — so the "off is byte-identical / no behaviour change" wording in the PR/commit is only true of the guest DTB. It's functionally benign (DMA-less wrappers, gpu-vm still owns both groups, 0x7d has no other tenant), but please narrow the claim so a future reader doesn't rely on "off == baseline" for the host side.

modules/microvm/sysvms/dispvm.nix
```diff
@@ -0,0 +1,93 @@
...
+let
+  vmName = "disp-vm";
+
+  cfg = config.ghaf.virtualization.microvm.dispvm;
+in
+{
+  _file = ./dispvm.nix;
+
+  options.ghaf.virtualization.microvm.dispvm = {
+    enable = lib.mkEnableOption "DispVM";
```
brianmcgillion, line 29: **[N1 maintainability — MED] `dispvm.nix` is a near-verbatim copy of `gpuvm.nix`.** 93 vs 92 lines; after `s/disp/gpu/` the only differences are comments and the `enable`/`description` strings — the actual logic (the `evaluatedConfig` export, the `ghaf.common` aggregation, the `microvm.vms` wiring, the assertion) is identical. Two full copies of one sysvm wrapper means a fix to one silently diverges from the other — the same drift failure mode the sysvm-registry aggregation was introduced to kill in #1998. Please factor these into one parametrized wrapper (e.g. `mkPassthroughSysvm { name; base; }`) rather than shipping the second copy.

modules/microvm/sysvms/dispvm-base.nix
```diff
@@ -0,0 +1,208 @@
...
+  globalConfig,
+  hostConfig,
+  ...
+}:
+let
+  vmName = "disp-vm";
+  timezoneEnabled = lib.ghaf.features.isEnabledFor globalConfig "timezone" vmName;
+in
+{
+  _file = ./dispvm-base.nix;
```
brianmcgillion, line 32: **[N2 maintainability — LOW/MED] `dispvm-base.nix` is ~50% duplicated from `gpuvm-base.nix`** (208 vs 277 lines; ~111 lines differ after normalization). There's genuine divergence here (disp-vm correctly drops the GPU carveouts/VRAM), so this is lower priority than N1 — but the shared scaffolding (imports, storagevm, givc, spire, networking) is copy-pasted and will drift. Worth extracting a `tegra-passthrough-vm-base` that both extend, ideally in the same pass as N1.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -22,66 +21,107 @@
...
 
-  # Reserved-memory carveouts take an explicit mmio-base for 1:1 GPA=HPA;
-  # engines use the default mapping.
-  reservedMem = [
-    {
+  # Host1x-ownership experiment (experiment/orin-two-vm-host1x). Both no-host1x
+  # directions drop physical host1x + 64MiB syncpoint shim + media (one unit).
+  # compute-no-host1x also drops display/DCE; display-no-host1x also drops
+  # GA10B/nvgpu and forces NVKMS no-syncpt mode.
+  exp = cfg.host1xExperiment;
```
brianmcgillion, line 33: **[N3 over-complication — MED] Selector branch density.** Five derived booleans (`computeNoHost1x`/`computeWithHost1x`/`displayOnly`/`dropHost1x`/`dropDisplay`) thread through the VFIO device lists, the cpp `-D` flags, the `reservedMem` carveouts, the patch list, the kernel config, and the DCE env across this 763-line file — and the load-bearing memory-map coupling (`EXP_SHRINK_BANK1` ↔ scanout carveout ↔ bank1 tail) is enforced only by prose comments. Consider collapsing to a single `modes.${exp} = { devices; defines; patches; dropScanout; … }` table so each mode's full effect is readable in one place and illegal combinations are unrepresentable. This would also remove much of the disp-vm scaffolding that N1/N2 currently duplicate.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -394,7 +453,14 @@ in
...
       };
     };
-    systemd.services."microvm@gpu-vm".after = [ "bindGpuVm.service" ];
+    systemd.services."microvm@gpu-vm" = {
+      after = [ "bindGpuVm.service" ];
+      # Exclusive DCE ownership (opt-in): the QEMU DCE bridge is created only when
+      # GHAF_DCE_GUEST=1 (ghaf-qemu-bpmp patch 0002). A display-capable GPU-VM opts
+      # in; a display-less one leaves it unset -> never opens /dev/dce-host, so it
+      # can't steal disp-vm's DCE events.
+      environment = lib.mkIf (!dropDisplay) { GHAF_DCE_GUEST = "1"; };
```
brianmcgillion, line 469: **[F3 correctness — please confirm] Exactly one VM must set `GHAF_DCE_GUEST`.** This gate exists so only one VM opens `/dev/dce-host` and drains the single host DCE event ring (per the patch 0002 rationale — two openers race and steal each other's events). It's set on the `!dropDisplay` arm. In the shipped `compute-with-host1x` + `disp_vm` config, please confirm that resolves to exactly disp-vm — a display-capable GPU-VM *and* disp-vm both hitting `!dropDisplay` would re-introduce the ring race this patch is meant to prevent. A short assertion ("at most one enabled VM sets GHAF_DCE_GUEST") would make it self-enforcing.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-iso-anchor/dce-iso-anchor.c
```diff
@@ -46,6 +46,11 @@ static const struct { u64 pa; u64 size; } dce_hi_ranges[] = {
 	{ 0x60000000, 0x04000000 },  /* vm_hs */
 	{ 0x80000000, 0x30000000 },  /* vm_cma */
 	{ 0xb0000000, 0x08000000 },  /* scanout */
+	{ 0xe2000000, 0x04000000 },  /* disp-vm CMA (split-RAM low bank): NVKMS
```
brianmcgillion, line 49: **[nit — confirmed correct, one robustness note]** This range is well-formed — I verified it matches disp-vm's `linux,cma` (0xe2000000..0xe6000000) exactly and stops at the DMA32/WLAN hole, so it doesn't widen DCE DMA beyond disp-vm's own RAM. Good.

Pre-existing robustness note (not a blocker): the `iommu_map` loop below returns on the first failure without unmapping the ranges already mapped this probe — a partial-map left behind on error. Host-init only (these are static host ranges, not guest-controlled), and this PR just adds one more entry, but an `iommu_unmap` rollback on the error path would make a failed probe leave a clean domain.

brianmcgillion (changes requested):
## Review — two-VM host1x split (full-depth, incl. cold C review of the DCE/BPMP surface)

Reviewed at `56556631` (rebased on main; #2043/#2054 merged, so this is the clean top of the stack). The architecture is sound and I verified the security-critical part rather than trusting the hardware run:

- **IOMMU isolation holds in the DT, not just on the devkit.** gpu-vm devices use `SID_PASSTHROUGH` (0x7f), disp-vm wrappers use `0x7d`; 0x7f ≠ 0x7d ⇒ genuinely disjoint SMMU groups, and 0x7d is unassigned in `tegra234-mc.h` (defined SIDs jump 0x72→0x7f) ⇒ no collision.
- **The `dce-iso-anchor.c` range `{0xe2000000,0x04000000}` is correctly bounded** — matches disp-vm's `linux,cma` exactly, stays inside disp-vm's own LOW bank, and stops precisely at the DMA32/WLAN hole (0xe6000000). The #2043 "guest→host DMA-anywhere" refutation still holds (translating SMMU domains, identity map within disp-vm's own RAM).
- **Builds green** (`nvidia-jetson-orin-agx-debug-from-x86_64`) and pre-commit/REUSE/SPDX pass.

Requesting changes so the feature lands clean — this is the last PR in the set, so let's fold the cleanups in here rather than trailing them. Inline comments cover the med/low items (SID naming, the unconditional-overlay wording, the `dispvm.nix`/`dispvm-base.nix` duplication, selector density, the DCE-owner check). The one item above nit-level is a **pre-existing** but guest-triggerable memory-safety bug in the QEMU bridges — details below; since it's in this feature's own DCE/BPMP surface and same authorship, it belongs in this PR.

### C1 [HIGH — guest→hypervisor memory safety] QEMU MMIO handlers read/write past the region buffer

Both device models have an out-of-bounds access reachable from an unprivileged guest via ordinary MMIO:

`packages/pkgs-by-name/ghaf-qemu-bpmp/sources/hw/misc/nvidia_dce_guest.c`
- `nvidia_dce_guest_read` (~L169-178): `if (addr >= MEM_SIZE) return 0xDEADBEEF; return *(uint64_t*)&s->mem[addr];` — the handler dereferences a **`uint64_t` regardless of the access `size`** and only rejects `addr >= MEM_SIZE`. A guest read at `addr ∈ [MEM_SIZE-7, MEM_SIZE-1]` reads up to **7 bytes past `s->mem[]`** and returns them to the guest → disclosure of the adjacent `NvidiaDceGuestState` fields (`host_device_fd`, thread state, `evt_seq`).
- `nvidia_dce_guest_write` default case (~L218): `memcpy(&s->mem[addr], &data, size)` with the same `addr < MEM_SIZE`-only guard → **OOB write** a few bytes past `s->mem[]` into those adjacent fields → guest-driven QEMU state corruption.

`packages/pkgs-by-name/ghaf-qemu-bpmp/sources/hw/misc/nvidia_bpmp_guest.c` has the identical pattern (read ~L60-63, write default ~L118).

Fix in this PR: bound `addr + size <= MEM_SIZE` in both handlers, honor the access `size` in `read` instead of the hardcoded `uint64_t` deref, and set `.impl.min/max_access_size` + `.valid` on the `MemoryRegionOps` so QEMU clamps for you. These sources are edited via `0002-nvidia-dce-guest-hooks.patch`, which this PR already touches.

### C2 [LOW — defense-in-depth] host-proxy short-`len` OOB read

`bpmp-host-proxy.c` `write()` (~L462) and `dce-host-proxy.c` `write()` (~L422) `kmalloc(len)` then dereference `kbuf->tx.size`/`rx.size`/`mrq`/`iface` without checking `len >= sizeof(*kbuf)`. Not currently reachable (the only writer is the QEMU bridge, which always writes `sizeof(struct)`, and `/dev/{bpmp,dce}-host` is root-only), but a one-line `if (len < sizeof(*kbuf)) return -EINVAL;` closes it. Optional.

### Also
- PR body checklist self-contradicts ("removed PR draft status — **kept as draft**") and the API says not-draft — please clarify the intended state.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/gpu_passthrough_overlay.dts
vadika: Both done in d2104fc6.

**F2:** replaced the bare `0x7d` at all sites with a named `GHAF_SID_DISPVM` constant carrying the rationale (unassigned on T234 — SIDs jump 0x72→0x7f — and ≠ `TEGRA234_SID_PASSTHROUGH` 0x7f, so gpu-vm and disp-vm never share a stream).

**F1:** narrowed the claim — the commit now states the guest DTB stays byte-identical in `off`, but the disp-vm SID wrappers are an *unconditional* host-closure change (DMA-less, gpu-vm still owns both groups), so the host closure is explicitly not identical.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
vadika: Done in d2104fc6 — added an assertion in gpu-vm that fails the build when a display-capable gpu-vm (`!dropDisplay`) and an enabled `disp_vm` would both set `GHAF_DCE_GUEST`. In the shipped `compute-with-host1x`+disp_vm config gpu-vm has `dropDisplay = true`, so only disp-vm sets it — the assertion makes that self-enforcing rather than prose.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/sources/drivers/platform/tegra/dce-iso-anchor/dce-iso-anchor.c
vadika: Added the rollback in d2104fc6 — on `iommu_map` failure the probe now unmaps only the ranges it mapped this pass (tracked in a small `mapped[]`), leaving any already-present maps intact, so a failed probe leaves a clean domain.

modules/microvm/sysvms/dispvm.nix
vadika: Agreed this is real duplication. Deferring the `mkPassthroughSysvm` extraction (N1) to a dedicated follow-up rather than folding a cross-cutting sysvm-wrapper refactor into this feature PR — it touches gpuvm.nix as well and is cleaner reviewed on its own. Tracking it with N2/N3.

modules/microvm/sysvms/dispvm-base.nix
vadika: Same call as N1 — the `tegra-passthrough-vm-base` extraction is deferred to the same follow-up. Noted that the divergence (disp-vm dropping the GPU carveouts/VRAM) is intended.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
vadika: Deferring the `modes.${exp}` table (N3) to the same follow-up as N1/N2 — collapsing the five selectors + the EXP_SHRINK_BANK1 memory-map coupling into one table is a substantial refactor of this file and is safer as its own reviewed change than mixed into the security/naming fixes here.

vadika:
Thanks for the deep review. Addressed in d2104fc6 (pushed shortly after an aarch64 build validates on the builder):

**C1 [HIGH — MMIO OOB]** — fixed in both bridges. `nvidia_dce_guest.c` / `nvidia_bpmp_guest.c`: reads/writes now bound the *full* access width (`addr > MEM_SIZE - size`) instead of only `addr >= MEM_SIZE`, `read` honors the access `size` via `memcpy` instead of a fixed `uint64_t` deref, and both `MemoryRegionOps` gained `.valid`/`.impl` `min/max_access_size` (1..8, aligned) so QEMU clamps malformed accesses before the handlers run. Closes the tail-of-`mem[]` disclosure and the OOB write into the adjacent state fields. `ghaf-qemu-bpmp` rebuilt clean.

**C2 [LOW]** — added `if (len < sizeof(*kbuf)) … -EINVAL;` before the header deref in both `bpmp-host-proxy.c` and `dce-host-proxy.c` `write()` paths.

**F1/F2/F3 + iommu nit** — replied inline; all in d2104fc6.

**N1/N2/N3** — the three dedup/selector refactors are real but cross-cutting (N1/N2 touch gpuvm.nix too; N3 restructures the whole selector surface). Deferring them to a dedicated follow-up rather than mixing a large refactor into this feature + security PR — replied on each thread.

brianmcgillion (approved):

PR #2056, opened 2026-07-20
fix(ci): reduce build eval worker pressure

henrirosten:
Cancel stale workflow runs at the workflow level instead of per matrix target so older runs from the same PR or branch do not overlap with newer commits.

Keep matrix jobs parallel, but run nix-fast-build with one eval worker and an explicit 4096 MiB eval memory cap per job. This avoids multiplying nix-eval-jobs memory across x86 targets on the shared builder.

### Type of Change

- [ ] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. ...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,12 @@ on:
     branches:
       - main
 
+concurrency:
+  # Cancel stale runs from the same PR or branch while allowing the active run's
+  # matrix jobs to run concurrently.
+  group: ${{ github.workflow }}.${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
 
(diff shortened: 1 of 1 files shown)


brianmcgillion (approved):

brianmcgillion:
only question is if there is enough memory if limited to 4GB

henrirosten:
@brianmcgillion: 
> _only question is if there is enough memory if limited to 4GB_

The limit was already effectively 4 GiB already before this change, since nix-eval-jobs defaults `--max-memory-size` to 4 GiB per worker. This PR just makes that explicit.

The important change is `--eval-workers 1`. `--eval-max-memory-size` is passed to nix-eval-jobs as a per-worker evaluator threshold. When a worker reaches it, nix-eval-jobs restarts that worker. It is not a hard RSS/cgroup limit for the whole eval process, so total observed memory can still exceed 4 GiB.

Before this PR, each matrix job could use 4 eval workers, so evaluator memory pressure was roughly multiplied by 4 per job. This change keeps matrix parallelism, but avoids multiplying eval workers inside each matrix job.

This change was tested here: https://github.com/tiiuae/ghaf/actions/runs/29744652892

PR #2057, opened 2026-07-20
fix(qemu-bpmp): restore vfio-platform irqfd fast path on gicv3

vadika:
## Description of Changes

MGBE0 passthrough into net-vm topped out ~76 Mbit/s on a 1 Gbps link. Two independent causes, both in QEMU:

- **No irqfd on GICv3.** `arm_gic_kvm.c` registers the `qemu_irq->GSI` map `kvm_irqchip_add_irqfd_notifier()` needs; the GICv3 model never did, so vfio-platform fell back to userspace injection — ~12 MMIO exits per IRQ. `0004` mirrors the GICv2 loop.
- **No DMA burst config.** Mainline stmmac skips `stmmac_init_dma_engine` without an `snps,axi-config` child; the guest node had none. `add_tegra_mgbe_fdt_node()` now synthesizes a guest-local one. `snps,blen` is 7 cells — mainline reads it with `AXI_BLEN=7` and silently drops the whole property (`-EOVERFLOW`) on fewer.

Once irqfd binds, an assertion in the ~17 s gap between vfio arming the line (during device connect) and the guest requesting the SPI wedges it permanently. `0005`/`0006` recover it via a **default-off, MGBE0-only `startup-rearm`** property, bounded to `startup-rearm-window` ms (default 30 s), not a perpetual all-device timer.

Also: base `ghaf-qemu` moved to QEMU 11.0.1 (which still lacks vfio-platform, so the 10.1.5 pin stays); its ACPI laptop patches no longer apply to 10.1.5 and are filtered out.

## Testing

AGX Orin, net-vm TX, `dd | nc` 1 GB over MGBE0 (`tx_bytes` confirms the path), qemu the only variable:

| | throughput | IRQs / GB |
|---|---:|---:|
| before (no irqfd) | 9.5 MB/s (**76 Mbit/s**) | 183,639 |
| this PR | 118 MB/s (**944 Mbit/s**) | 37,889 |

**12.4× throughput, near 1 GbE line rate.** Lifecycle checks all recover with no wedge/storm: clean boot, delayed `ip link up` past the 30 s window, 20× down/up, and guest reboot (resamplefd handles reboot recovery). Guest DT verified live: `snps,axi-config` resolves, 7-cell `snps,blen`, `rd/wr_osr_lmt=63`, driver init clean. `nix fmt`, `reuse lint`, pre-commit, AGX+NX images, and `ghaf-qemu-bpmp-gpu` (fast path off) all build.

## Scope

net-vm only. `ghaf-qemu-bpmp-gpu` opts out via `withIrqfdFastPath = false`; the GPU/display VMs' irqfd behavior is unevaluated and load-bearing — a follow-up, not this PR.

### DIFF
--- a/packages/pkgs-by-name/ghaf-qemu-bpmp/package.nix
+++ b/packages/pkgs-by-name/ghaf-qemu-bpmp/package.nix
@@ -12,6 +12,10 @@
 # is the last release that has it. Only the VM receiving such a device uses this
 # QEMU (set via microvm.qemu.package); everything else stays on pkgs.ghaf-qemu.
 #
+# 11.0.1 still lacks vfio-platform (checked 2026-07-21), so the pin holds. When
+# base ghaf-qemu bumps QEMU its patches may stop applying here -- fix the
+# prev.patches filter below, don't drop the pin.
+#
 # The pin is a security liability: net-vm is network-facing and won't get QEMU
 # fixes past 10.1.x. Revisit if upstream restores platform passthrough.
 #
@@ -19,11 +23,13 @@
 # and rename the binary. pkgs-by-name auto-calls with {}, so both default to the
 # net-vm build -- pkgs.ghaf-qemu-bpmp is byte-identical to before.
 {
+  lib,
   ghaf-qemu,
   fetchurl,
-  lib,
   extraPatches ? [ ],
   variantName ? "",
+  # validated on MGBE0 only; the GPU variant opts out
+  withIrqfdFastPath ? true,
   ...
 }:
 ghaf-qemu.overrideAttrs (
@@ -40,11 +46,25 @@ ghaf-qemu.overrideAttrs (
     # series (extended-GPE / Battery / AC-adapter / lid-button) no longer applies
     # to the pinned 10.1.5 and is irrelevant to the Orin aarch64 microvm. Drop it
     # so the package builds; keep everything else.
+    #
+    # 0004-0006 fix vfio-platform's irqfd fast path on GICv3 (the GICv3 model never
+    # registered the qemu_irq->GSI map, so every IRQ trapped to userspace). One unit,
+    # do not split: 0004 alone kills the NIC. With 0001's AXI config: 76 -> 944 Mbit/s.
+    # net-vm only; the GPU variant sets withIrqfdFastPath = false.
+    # The hw-acpi- series is x86-only in base ghaf-qemu (isx86_64-guarded), so on
+    # the aarch64 Orin guest prev.patches carries none and this filter is a no-op;
+    # it only bites if base ever makes them cross-platform. No count assert: the
+    # expected number is 0 here and 4 on x86, so a fixed check would be wrong.
     patches =
       (lib.filter (p: !(lib.hasInfix "hw-acpi-" (baseNameOf p))) (prev.patches or [ ]))
       ++ [
         ./patches/0001-nvidia-bpmp-guest-hooks.patch
         ./patches/0002-nvidia-dce-guest-hooks.patch
+      ]
+      ++ lib.optionals withIrqfdFastPath [
+        ./patches/0004-arm-gicv3-kvm-register-qemuirq-gsi.patch
+        ./patches/0005-vfio-platform-unmask-after-resamplefd.patch
+        ./patches/0006-vfio-platform-rearm-automasked-lines.patch
       ]
       ++ extraPatches;
 
--- a/packages/pkgs-by-name/ghaf-qemu-bpmp/patches/0005-vfio-platform-unmask-after-resamplefd.patch
+++ b/packages/pkgs-by-name/ghaf-qemu-bpmp/patches/0005-vfio-platform-unmask-after-resamplefd.patch
@@ -1,0 +1,56 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Ghaf contributors <ghaf@tii.ae>
+Date: Mon, 20 Jul 2026 17:00:00 +0300
+Subject: [PATCH] hw/vfio/platform: unmask the line after arming the resamplefd
+
+vfio enables the IRQ inside vfio_start_irqfd_injection() (from
+sysbus_connect_irq() during device connect), before the guest installs a
+handler for the SPI. An assertion in that window is auto-masked, and on the
+irqfd path only the resamplefd re-arms the line - which needs a guest EOI for
+an interrupt it never got, so it stays masked forever. (Measured on an AGX
+Orin: host vfio IRQ fires once, guest SPI stays 0, NIC receives nothing.) The
+userspace path self-heals via vfio_platform_eoi() on any MMIO.
+
+Unmask once after the resamplefd is armed, opt-in via the "startup-rearm"
+property so no other vfio-platform device is affected. This patch carries the
+struct field + property so it is independently buildable and bisectable; the
+bounded retry loop (next patch) builds on top of it.
+---
+--- a/hw/vfio/platform.c
++++ b/hw/vfio/platform.c
+@@ -403,6 +403,15 @@
+         if (vfio_set_resample_eventfd(intp) < 0) {
+             goto fail_vfio;
+         }
++
++        /*
++         * The line is armed during device connect, before the guest installs
++         * its SPI handler; a stale assertion then wedges it (only the resamplefd
++         * re-arms on irqfd, and it needs a guest EOI). Opt-in unmask once here.
++         */
++        if (vdev->startup_rearm) {
++            vfio_device_irq_unmask(&vdev->vbasedev, intp->pin);
++        }
+         trace_vfio_platform_start_level_irqfd_injection(intp->pin,
+                                     event_notifier_get_fd(intp->interrupt),
+                                     event_notifier_get_fd(intp->unmask));
+@@ -653,6 +662,7 @@
+     DEFINE_PROP_UINT32("mmap-timeout-ms", VFIOPlatformDevice,
+                        mmap_timeout, 1100),
+     DEFINE_PROP_BOOL("x-irqfd", VFIOPlatformDevice, irqfd_allowed, true),
++    DEFINE_PROP_BOOL("startup-rearm", VFIOPlatformDevice, startup_rearm, false),
+ #ifdef CONFIG_IOMMUFD
+     DEFINE_PROP_LINK("iommufd", VFIOPlatformDevice, vbasedev.iommufd,
+                      TYPE_IOMMUFD_BACKEND, IOMMUFDBackend *),
+--- a/include/hw/vfio/vfio-platform.h
++++ b/include/hw/vfio/vfio-platform.h
+@@ -60,6 +60,7 @@
+     unsigned int num_compat; /* number of compatible values */
+     uint32_t mmap_timeout; /* delay to re-enable mmaps after interrupt */
+     QEMUTimer *mmap_timer; /* allows fast-path resume after IRQ hit */
++    bool startup_rearm; /* opt-in: re-arm early automasked lines at boot */
+     QemuMutex intp_mutex; /* protect the intp_list IRQ state */
+     bool irqfd_allowed; /* debug option to force irqfd on/off */
+ };
+--
+2.49.0
--- a/packages/pkgs-by-name/ghaf-qemu-bpmp/patches/0006-vfio-platform-rearm-automasked-lines.patch
+++ b/packages/pkgs-by-name/ghaf-qemu-bpmp/patches/0006-vfio-platform-rearm-automasked-lines.patch
@@ -1,0 +1,133 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Ghaf contributors <ghaf@tii.ae>
+Date: Mon, 20 Jul 2026 18:00:00 +0300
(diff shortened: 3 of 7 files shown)


enesoztrk:
Related Tickets could be https://jira.tii.ae/browse/SSRCSP-8720

vadika:
> (**944 Mbit/s**)
@enesoztrk found another problem , now have near 1Gbps thoughput!

enesoztrk:
I have tested it! Good work!

packages/pkgs-by-name/ghaf-qemu-bpmp/patches/0006-vfio-platform-rearm-automasked-lines.patch
```diff
@@ -0,0 +1,119 @@
...
++    int64_t now = qemu_clock_get_ms(QEMU_CLOCK_VIRTUAL);
++    VFIOINTp *intp;
++
++    if (vdev->startup_rearm_window && now >= vdev->startup_rearm_deadline) {
++        return; /* window elapsed: let the resamplefd contract own the line */
++    }
++
++    WITH_QEMU_LOCK_GUARD(&vdev->intp_mutex) {
++        QLIST_FOREACH(intp, &vdev->intp_list, next) {
++            if (intp->kvm_accel && vfio_irq_is_automasked(intp)) {
```
brianmcgillion, line 66: **[C1 — correctness precision / doc] The gate is a static type flag, not "currently masked".**

`vfio_irq_is_automasked()` (`hw/vfio/platform.c:46`) is `return intp->flags & VFIO_IRQ_INFO_AUTOMASKED` — a **static property set once at init** from `info.flags`, so it is true for *every* level SPI, permanently. It says "this line is the auto-masked kind", not "this line is currently masked". The check also never consults `intp->state` (`VFIO_IRQ_INACTIVE` / `PENDING` / `ACTIVE`).

So this loop re-arms **every** `kvm_accel` level line once a second for the whole window — including a line that is legitimately masked awaiting a real guest EOI (`VFIO_IRQ_ACTIVE`). If such a line is still physically asserted, `vfio_device_irq_unmask()` re-triggers it, i.e. a spurious duplicate injection.

In this deployment the impact is genuinely small — `startup-rearm=on` is set only on MGBE0, a single-IRQ device that is quiescent during bring-up, and the rate is capped at 1 Hz for 30 s — which matches your "no wedge/storm" lifecycle results. The issue is that the doc comment above ("re-arm lines **automasked before the guest was listening**") describes a precision the code doesn't have: it re-arms all level lines regardless of whether they were wedged at startup or are actively cycling.

Please either narrow the comment to say plainly that it re-arms all level lines for the duration of the window, or — since this is general vfio-platform code that a future multi-IRQ device could enable — gate on something that actually distinguishes the wedged case (e.g. a per-`intp` "has not yet seen its first EOI" flag, or skipping `intp->state == VFIO_IRQ_ACTIVE`) so an actively-cycling line is never re-armed mid-flight.

packages/pkgs-by-name/ghaf-qemu-bpmp/patches/0005-vfio-platform-unmask-after-resamplefd.patch
```diff
@@ -0,0 +1,36 @@
...
+         if (vfio_set_resample_eventfd(intp) < 0) {
+             goto fail_vfio;
+         }
++
++        /*
++         * The line is armed during device connect, before the guest installs
++         * its SPI handler; a stale assertion then wedges it (only the resamplefd
++         * re-arms on irqfd, and it needs a guest EOI). Opt-in unmask once here.
++         */
++        if (vdev->startup_rearm) {
```
brianmcgillion, line 31: **[C2 — patch hygiene / bisectability] 0005 is not independently buildable.**

This hunk reads `vdev->startup_rearm`, but that struct field (`include/hw/vfio/vfio-platform.h`) and its `DEFINE_PROP_BOOL("startup-rearm", ...)` are both added by **0006**. The series is applied as a unit (`lib.optionals withIrqfdFastPath [ 0004 0005 0006 ]`), so the final tree compiles and the build is green — but 0005 on its own references an undeclared member.

That breaks `git bisect` across the series and would be an immediate review comment if these ever go upstream. Cheapest fix: move the `startup_rearm` field + property definition into 0005 (leaving the window/deadline/timer machinery in 0006), or simply merge 0005 and 0006 — they're arguably one logical change ("bounded opt-in re-arm"), which would also address the split noted in the review body.

packages/pkgs-by-name/ghaf-qemu-bpmp/package.nix
```diff
@@ -40,12 +46,22 @@ ghaf-qemu.overrideAttrs (
     # series (extended-GPE / Battery / AC-adapter / lid-button) no longer applies
     # to the pinned 10.1.5 and is irrelevant to the Orin aarch64 microvm. Drop it
     # so the package builds; keep everything else.
+    #
+    # 0004-0006 fix vfio-platform's irqfd fast path on GICv3 (the GICv3 model never
+    # registered the qemu_irq->GSI map, so every IRQ trapped to userspace). One unit,
+    # do not split: 0004 alone kills the NIC. With 0001's AXI config: 76 -> 944 Mbit/s.
+    # net-vm only; the GPU variant sets withIrqfdFastPath = false.
     patches =
       (lib.filter (p: !(lib.hasInfix "hw-acpi-" (baseNameOf p))) (prev.patches or [ ]))
```
brianmcgillion, line 59: **[N1 — Nix, minor] The ACPI drop is a basename *infix* match.**

`lib.hasInfix "hw-acpi-" (baseNameOf p)` will drop any patch whose filename happens to contain `hw-acpi-`, and conversely silently drops nothing if base `ghaf-qemu` renames the series. You've documented the fragility directly above ("fix the `prev.patches` filter below, don't drop the pin"), which is the right call and most of what's needed here.

Worth considering while you're in this file: an exact-basename allowlist of the patches being removed, or asserting the filter actually removed the expected count, so a future base-QEMU bump fails loudly at eval instead of silently changing which patches apply to the pinned 10.1.5. Non-blocking — the comment already carries the warning.

brianmcgillion (changes requested):
## Review — vfio-platform irqfd fast path on GICv3

Reviewed at `d7286446`. Since the C here lives inside `.patch` files and the QEMU sources aren't in-tree, I materialized QEMU 10.1.5, applied 0004/0005/0006, and reviewed the resulting `.c` with full surrounding context rather than reading the diffs in isolation.

This is a well-scoped, unusually well-documented fix and the design is sound. **Builds verified green on all three relevant targets:**

| Target | Result |
|---|---|
| `nvidia-jetson-orin-agx-debug-from-x86_64` (net-vm, fast path on) | PASS |
| `ghaf-qemu-bpmp` (0004-0006 applied) | PASS — series applies + compiles |
| `ghaf-qemu-bpmp-gpu` (`withIrqfdFastPath = false`) | PASS — opt-out genuinely excludes the patches |

**Verified correct against the real source** (not taken on trust):
- **0004's loop is right.** `for (i = 0; i < s->num_irq - GIC_INTERNAL; i++) kvm_irqchip_set_qemuirq_gsi(kvm_state, qdev_get_gpio_in(dev, i), i)` matches the existing GICv3 idiom in the same file (`arm_gicv3_kvm.c:928`) and the GICv2 `num_irq - GIC_INTERNAL` SPI convention — no off-by-one, and GSI↔SPI numbering is consistent between registration and the vfio lookup.
- **The DoS surface is properly bounded.** The watchdog's `if (window && now >= deadline) return;` exits *without* `timer_mod`, so the timer genuinely stops rescheduling once the window elapses (`window = 0` short-circuits to perpetual, as documented). Opt-in, MGBE0-only, GPU variant opts out, `QEMU_CLOCK_VIRTUAL` freezes on guest pause. A guest cannot escalate this into an IRQ storm — it's capped at 1 Hz.
- **`unmask_timer` lifecycle is fine.** It's only `timer_del`'d on the `irq_err:` path — but so is the pre-existing `mmap_timer`, and neither is freed in a normal unrealize (there is no `dc->unrealize`). The new timer faithfully mirrors the established pattern; not a new leak, and these sysbus devices aren't hot-unplugged.
- **0001's `snps,axi-config` synthesis is correct** — 7-cell `snps,blen`, `rd/wr_osr_lmt = 63`, guest-local phandle, and the `-EOVERFLOW`-on-fewer-cells rationale checks out.

Marking **changes requested** so the items below land before merge. In fairness they are all **low severity and none is a correctness blocker** on its own — the substantive gate on this PR remains the `Needs Testing` AGX-Orin runtime validation, which I can't reproduce without the hardware. Details inline; no duplication or over-generalization found (the three patches are genuinely distinct concerns, and `withIrqfdFastPath` follows the existing `variantName`/`extraPatches` override idiom).

### Hygiene (no action needed unless something looks off)
- SPDX/REUSE on the three new `.patch` files is covered by the existing globs — `reuse lint` / pre-commit pass.
- The `lib` argument reorder is cosmetic.
- Commit message is detailed and its claims (IRQ counts, "0004 alone kills the NIC") are consistent with the code paths I traced.

vadika:
@brianmcgillion thanks for the deep review — all three items addressed and force-pushed (squashed to a single signed-off commit `d6ed9f08`).

- **C1 (re-arm precision / doc):** you're right that the loop can't distinguish a startup-wedged line from an active one — and on the irqfd path `intp->state` isn't even maintained (the eventfd goes straight to KVM, bypassing `vfio_intp_interrupt`), so a state gate would be theatre. Took the honest option: reworded the `vfio_platform_unmask_watchdog` doc comment and the commit message to state plainly that it re-arms *every* `kvm_accel` level line for the window, with an explicit note that a future multi-IRQ opt-in must accept that. No behavior change (harmless here — MGBE0 is single-IRQ and quiescent during bring-up, unmask of an unmasked line is a no-op).

- **C2 (bisectability):** moved the `startup_rearm` struct field + `startup-rearm` property out of 0006 and into 0005, so 0005 now carries everything it references and is independently buildable. 0006 keeps only the window/deadline/timer machinery. Verified against a materialized QEMU 10.1.5 tree: the 0005-only tree has zero references to `startup_rearm_window` / `startup_rearm_deadline` / `unmask_timer`, and the full 0004→0005→0006 series still applies clean.

- **N1 (ACPI filter fragility):** added an eval-time `lib.assertMsg` that exactly 4 `hw-acpi-` patches are dropped from `prev.patches`. A future base-QEMU bump that renames or removes them now fails loudly at eval instead of silently changing which patches apply to the 10.1.5 pin.

The substantive gate remains the AGX-Orin runtime validation, which is unchanged by these edits (patch hygiene + docs only).

packages/pkgs-by-name/ghaf-qemu-bpmp/package.nix
brianmcgillion: **Blocker — and it's my fault: this assert breaks the Orin target.**

Reverified at `2ec11d4f`. C1 and C2 are properly fixed, and the `flake-module.nix` revert from the previous head is correctly gone. But the assert added for this comment makes the **AGX cross image fail to evaluate**:

```
$ nix eval .#packages.x86_64-linux.nvidia-jetson-orin-agx-debug-from-x86_64.drvPath
error: ghaf-qemu-bpmp: expected to drop 4 hw-acpi- patches from base ghaf-qemu,
dropped 0. Base QEMU patch set changed -- re-check this
filter (see comment above), don't drop the 10.1.5 pin.
```

Root cause — the `hw-acpi-*` patches are x86-laptop-only, so the count is platform-dependent:

| base `ghaf-qemu` | `hw-acpi-*` patches | `dropped` | result |
|---|---:|---:|---|
| `x86_64-linux` | 4 (extended-GPE, Battery, AC-adapter, lid-button) | 4 | assert passes |
| `aarch64-linux` (Orin) | **0 — none shipped** | 0 | **assert fires, eval dies** |

`ghaf-qemu-bpmp` and `-gpu` still evaluate fine on `x86_64-linux`, which is why this isn't obvious locally — but the primary target of this PR doesn't evaluate.

This one is on me: I suggested "an exact-basename allowlist, or asserting the filter removed the expected count" without noting the count differs per platform. You implemented exactly what I asked for.

If you want to keep an assert, the platform-independent form is to assert the actual invariant — that no ACPI patch *survived* the filter — rather than a magic count. I evaluated both platforms and it holds (`anyAcpiSurvives = false` for x86_64 and aarch64):

```nix
assert lib.assertMsg (!lib.any (p: lib.hasInfix "hw-acpi-" (baseNameOf p)) kept) ''
  ghaf-qemu-bpmp: an hw-acpi- patch survived the hw-acpi- filter; the base
  ghaf-qemu patch set changed -- re-check this filter, don't drop the 10.1.5 pin.
'';
```

That said, worth stating plainly: this is belt-and-braces either way. A patch that no longer applies already fails loudly in `patchPhase`, so **reverting to the comment-only version is an equally valid fix** and is simpler. Your call — either resolves the blocker.

Everything else from the review is confirmed done:
- **C1** — the reworded comment is exactly right, and your point that `intp->state` isn't maintained on the irqfd path (eventfd goes straight to KVM, bypassing `vfio_intp_interrupt`) is correct; a state gate would indeed have been theatre. Good call rejecting that half of the suggestion.
- **C2** — `startup_rearm` field + property now land in 0005; 0006 carries them as context with no double-add. 0005 is independently buildable.
- `nix build .#checks.x86_64-linux.pre-commit` passes; both `ghaf-qemu-bpmp` and `ghaf-qemu-bpmp-gpu` evaluate and build on `x86_64-linux`.

brianmcgillion (approved):

PR #2071, opened 2026-07-27
feat(orin): let EMC reach its trained maximum frequency

vadika:
## Description of Changes

On Orin, BPMP firmware boots with an internal bandwidth-manager cap on the EMC (DRAM) clock at the boot rate — 2133 MHz on AGX Orin — and every rate request is clamped to it. The trained maximum is 3199 MHz, so Ghaf devices leave roughly a third of their memory bandwidth unused under load, for every workload on the device (host, VMs, GPU).

The cap cannot be raised from stock userspace: the `nvpmodel_clk_cap` sysfs store sends `CAP_SET(clk_round_rate(S64_MAX))`, and `round_rate` is itself clamped by the current cap, so the request can never exceed it (a self-deadlock). Stock JetPack escapes this via the nvphs power-hinting service, which Ghaf does not run.

This PR adds a `ghaf.hardware.nvidia.orin.emcBoost` option backed by `emc_cap_lift.ko`, a ~100-line out-of-tree module that sends the `MRQ_BWMGR_INT`/`CMD_BWMGR_INT_CAP_SET` request once at boot. The MRQ is served by BPMP firmware, so the module is kernel-version independent (built and tested against the 6.6.129 host kernel; also validated on an r36.5 L4T 5.15 kernel). No rate pinning is done: with the cap lifted, the demand-driven EMC DVFS governor ranges freely — full bandwidth under load, low clocks (and power) at idle. The option is enabled for the AGX and AGX64 reference boards; the frequency is configurable (`emcBoost.frequencyHz`, default 3199 MHz).

**A note on why this exists at all:** there is no indication this is a deliberate restriction — the CAP_SET MRQ is not gated or authenticated in BPMP firmware, and it accepts the request from any kernel module. **It is, however, a consequence of NVIDIA's stack architecture: the only component that ever lifts the boot-time cap is the closed-source nvphs service, and the equivalent EMC scaling support has not been upstreamed to mainline. Any OS that does not run NVIDIA's closed userspace therefore silently loses ~1/3 of memory bandwidth, with `max_rate` still advertising the full 3199 MHz.** This module reimplements the single firmware request needed to remove that dependency.

### Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

None.

## Checklist

### Applicable Targets
- [x] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`
### Installation Method
### Test Steps To Verify:
1. Boot an AGX Orin target with this change; `dmesg | grep emc_cap_lift` shows `EMC cap set to 3199000000 Hz`.
2. Idle: `sudo cat /sys/kernel/debug/bpmp/debug/clk/emc/rate` reads a low rate (204/665 MHz) — demand DVFS active.
3. Load: run several parallel memory-bound streams (e.g. `for i in $(seq 12); do dd if=/dev/zero of=/dev/null bs=256M count=100 & done`) and re-read the rate — it reaches `3199000000`.
4. Without the module (baseline), step 3 never exceeds `2133000000`.

Hardware validation already performed on an AGX Orin 64GB devkit (steps 1-4, including the 2133-ceiling baseline): idle 204 MHz, loaded 3199 MHz, and the previously deadlocked `/sys/kernel/nvpmodel_clk_cap/emc` interface becomes functional after the cap lift.

### DIFF
--- a/modules/reference/hardware/jetpack/agx/orin-agx.nix
+++ b/modules/reference/hardware/jetpack/agx/orin-agx.nix
@@ -31,6 +31,8 @@
         kernelVersion = "upstream-6-6";
         somType = "agx";
         agx.enableNetvmWlanPCIPassthrough = true;
+        # Run DRAM at its trained maximum (3199 MHz vs the 2133 boot rate).
+        emcBoost.enable = true;
         carrierBoard = "devkit";
         # AGX devkit boots rootfs from eMMC.
         flashScriptOverrides = {
--- a/modules/reference/hardware/jetpack/agx/orin-agx64.nix
+++ b/modules/reference/hardware/jetpack/agx/orin-agx64.nix
@@ -22,6 +22,8 @@
         kernelVersion = "upstream-6-6";
         somType = "agx64";
         agx.enableNetvmWlanPCIPassthrough = true;
+        # Run DRAM at its trained maximum (3199 MHz vs the 2133 boot rate).
+        emcBoost.enable = true;
         carrierBoard = "devkit";
         # AGX64 devkit boots rootfs from eMMC.
         flashScriptOverrides = {
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/default.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/default.nix
@@ -10,5 +10,6 @@
     ./pci-passthrough-common.nix
     ./virtualization
     ./optee/optee.nix
+    ./emc-boost
   ];
 }
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/emc-boost/default.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/emc-boost/default.nix
@@ -1,0 +1,53 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Let the EMC (DRAM) clock reach its trained maximum instead of being
+# capped at the boot rate.
+#
+# BPMP firmware boots with an internal bandwidth-manager cap at the DRAM
+# boot rate (2133 MHz on AGX Orin), so without intervention the memory bus
+# never reaches the trained maximum (3199 MHz on AGX Orin — a ~50% memory
+# bandwidth loss under load, for every workload on the device). Stock
+# JetPack lifts the cap from the nvphs service; Ghaf does not run it, and
+# the nvpmodel sysfs cap interface cannot raise the cap (its own request
+# is rounded through a call the cap clamps). See
+# emc-boost/module/emc_cap_lift.c for the full analysis.
+#
+# The emc_cap_lift.ko module lifts the cap once at boot (a kernel module,
+# because the CAP_SET MRQ is not reachable from userspace). No rate
+# pinning is done: with the cap lifted, the demand-driven EMC DVFS
+# governor ranges freely — validated on AGX Orin hardware to reach
+# 3199 MHz under memory load and drop back to 204 MHz at idle.
+{
+  config,
+  lib,
+  ...
+}:
+let
+  cfg = config.ghaf.hardware.nvidia.orin.emcBoost;
+in
+{
+  _file = ./default.nix;
+
+  options.ghaf.hardware.nvidia.orin.emcBoost = {
+    enable = lib.mkEnableOption "running the EMC (DRAM) clock at its trained maximum";
+
+    frequencyHz = lib.mkOption {
+      type = lib.types.ints.positive;
+      default = 3199000000;
+      description = ''
+        EMC frequency cap to request from BPMP, in Hz. The default is the
+        trained maximum on AGX Orin (and Orin NX). This is a ceiling for
+        the demand-driven DVFS governor, not a fixed rate.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    boot.extraModulePackages = [
+      (config.boot.kernelPackages.callPackage ./module { })
+    ];
+    boot.kernelModules = [ "emc_cap_lift" ];
+    boot.kernelParams = [ "emc_cap_lift.cap_hz=${toString cfg.frequencyHz}" ];
+  };
+}
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/emc-boost/module/Makefile
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/emc-boost/module/Makefile
@@ -1,0 +1,22 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Out-of-tree build Makefile for the EMC cap lift module
+#
+# Override KERNEL_DIR if you want a different kernel build tree:
+#   make KERNEL_DIR=/path/to/kernel/build
+
+KERNELRELEASE ?= $(shell uname -r)
+KERNEL_DIR  ?= /lib/modules/$(KERNELRELEASE)/build
+PWD := $(shell pwd)
+
+obj-m := emc_cap_lift.o
+
+all:
+	$(MAKE) -C $(KERNEL_DIR) M=$(PWD) modules
+
+install:
+	$(MAKE) -C $(KERNEL_DIR) M=$(PWD) modules_install
+
+clean:
+	$(MAKE) -C $(KERNEL_DIR) M=$(PWD) clean
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/emc-boost/module/default.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/emc-boost/module/default.nix
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  lib,
+  stdenv,
+  kernel,
+  kernelModuleMakeFlags,
+  kmod,
+}:
+
+stdenv.mkDerivation {
+  pname = "emc-cap-lift";
+  version = "0.1.0";
+
+  src = ./.;
+
+  nativeBuildInputs = [ kmod ] ++ kernel.moduleBuildDependencies;
+
+  # External modules should not be PIC.
+  hardeningDisable = [ "pic" ];
+
+  makeFlags =
+    kernelModuleMakeFlags
+    ++ [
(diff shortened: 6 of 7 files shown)


brianmcgillion (changes requested):
Reviewed `72e5ed3b`.

## Verified correct — checked, not assumed

These were my main suspicions going in; all four cleared, so please don't spend time defending them:

- **The BPMP wire layout is right.** Against NVIDIA's downstream header (`nvidia-oot/include/soc/tegra/bpmp-abi.h`, which is present in the `l4t-oot-sources` this repo already builds): `MRQ_BWMGR_INT` is `83U` ✔, `CMD_BWMGR_INT_CAP_SET` is `3` ✔, and `struct mrq_bwmgr_int_request { uint32_t cmd; union { … cmd_bwmgr_int_cap_set_request { uint64_t rate; } … } } BPMP_ABI_PACKED` gives `cmd`@0, `rate`@4, 12 bytes — exactly your `struct bwmgr_int_cap_set_req { u32 cmd; u64 rate; } __packed`. `u64 resp` matches `mrq_bwmgr_int_response` (8-byte union) ✔.
- **`boot.kernelParams` really does reach a loadable module.** I expected this to be a silent no-op, since kernel-command-line module parameters are usually described as built-in-only. It isn't: `/proc/cmdline` parsing lives in **libkmod itself** (`libkmod/libkmod-config.c:511`, `kmod_config_parse_kcmdline()` → `kmod_config_add_options()`), so `systemd-modules-load` picks it up and not just the `modprobe` binary. `frequencyHz` is effective.
- **A guest cannot replay this.** `MRQ_BWMGR_INT` is 83; the bpmp-host-proxy allowlist permits `MRQ_BWMGR` (77) but not 83, so the cap lift stays host-only.
- **Cross-build is clean and reproducible.** `emc-cap-lift-aarch64-unknown-linux-gnu` rebuilds bit-identical under `--rebuild`; `modinfo` reports `vermagic: 6.6.129 SMP mod_unload aarch64`, `parm: cap_hz`.

Reference handling is also correct — `of_node_put()` and `put_device()` on every exit path, no leak on the `!bpmp` path, no double-put — and `msg.rx.ret` is checked. The write-up in both the NixOS module and the C header is unusually good; the analysis of *why* the sysfs path can't do this is the kind of thing that saves the next person a day.

---

## 1. [Medium] `-EPROBE_DEFER` from `module_init` is not deferral

```c
bpmp = platform_get_drvdata(pdev);
if (!bpmp) {
	put_device(&pdev->dev);
	return -EPROBE_DEFER;      /* nothing will ever retry this */
}
```

`EPROBE_DEFER` only means something when returned from a driver `probe()`. From `module_init` it is just a failed module load — nothing re-runs it, and the board boots at the 2133 MHz cap, i.e. the exact ~50% bandwidth loss this PR exists to remove, with one `pr_err` as the only trace.

To be fair on severity: this is close to unreachable in practice. `config TEGRA_BPMP` is `bool` (always built-in) and registers via `builtin_platform_driver()`, so it binds at `device_initcall` time, long before userspace loads modules. `platform_get_drvdata()` returns NULL only in the window where the DT device exists but the driver hasn't bound.

So it's latent rather than live — but the failure is silent and the fix is one line. Either return `-ENODEV` (honest: "BPMP isn't ready and we're not coming back"), or make this a real platform driver bound to the BPMP node, where deferral actually works.

## 2. [Medium] The cap is probably lost on resume, and never re-applied — please confirm before merge

The module is deliberately one-shot: *"Nothing to undo: the cap stays until the next BPMP reset."*

The 6.6.129 BPMP driver has PM ops — `tegra_bpmp_pm_ops`, `bpmp->suspended`, `tegra_bpmp_resume()`, and `tegra_bpmp_transfer()` returns `-EAGAIN` while suspended unless `TEGRA_BPMP_MESSAGE_RESET` is set. If AGX suspend power-gates BPMP and its firmware reloads, the bwmgr cap goes back to the boot rate, and nothing re-sends the CAP_SET.

That would make "the next BPMP reset" a routine suspend rather than the remote possibility the comment implies, and turn a permanent win into one that silently disappears the first time the box sleeps. Two things:

- Confirm empirically whether the cap survives SC7 on AGX.
- If it doesn't, add a PM notifier / resume hook that re-sends it, and reword the comment.

This is the finding I'd most like answered before merge, because it decides whether the feature does what it says on a running system.

## 3. [Medium] Hardcoded board-specific frequency where the goal is "no cap"

The intent is to stop BPMP capping the EMC clock, but the implementation encodes one SKU's trained maximum (`3199000000`) and then exposes it as a tunable.

`cmd_bwmgr_int_cap_set_request.rate` is a `uint64_t` cap. Requesting `U64_MAX` should let BPMP clamp to whatever that SoC actually trained to — SKU-independent, no magic number, and `frequencyHz` could disappear entirely. Note your own analysis explains the *sysfs* path fails because `clk_round_rate()` is clamped by the current cap; sending the cap directly doesn't go through `round_rate`, so that objection doesn't apply here. If BPMP does *not* clamp and a bogus value is harmful, that's worth a sentence in the comment — you already check `msg.rx.ret`, so it's easy to establish which.

Related: `frequencyHz` is documented as "the trained maximum on AGX Orin (and Orin NX)", but `emcBoost.enable` is set only on `orin-agx` and `orin-agx64` — `orin-nx.nix` doesn't enable it, and not all NX SKUs train to 3199. Either wire NX with the right value, or drop NX from the docstring. `type = ints.positive` also accepts anything with no upper bound; that stops mattering under the `U64_MAX` approach.

## 4. [Medium] The Nix packaging is a line-for-line copy of `fake-battery`

Diffed against `modules/hardware/x86_64-generic/kernel/modules/fake-battery/`:

- `default.nix` differs only in `pname`, `description`, `license`, `platforms` — and by **deleting the explanatory comments** ("Kernel exposes a list of build-time deps for external modules.", "Variable refers to the local Makefile.", "Avoid stripping which could interfere with signatures / debugging."). The comments were the part worth keeping; the boilerplate was the part worth sharing.
- `Makefile` differs only in `obj-m` and the header text — including inheriting `KERNELRELEASE ?= $(shell uname -r)`, which is dead in the Nix path (`KERNEL_DIR` is always overridden by `makeFlags`) and departs from the standard two-pass out-of-tree idiom (`ifeq ($(KERNELRELEASE),)`).

The repo now carries **four** out-of-tree kernel-module packagings: `fake-battery`, `packages/pkgs-by-name/memsocket/module.nix`, `packages/pkgs-by-name/rtl8126`, and this one. A single `lib` helper — `buildKernelModule { pname; version; src; license; platforms; }` emitting the derivation and optionally the Makefile — would absorb nearly all of it and give one place to fix the `KERNELRELEASE` wart. I'd take that as a follow-up rather than inside this PR, but the fifth copy seems like the right moment to stop and factor it out.

## 5. [Low-Medium] No size assertion or provenance for the hand-rolled ABI struct

The layout is correct today (verified above), so this is about drift, not a bug. But #2070 has just spent two review rounds establishing that hand-rolled BPMP structs in this tree need a compile-time guard, and this is the same pattern two directories away with none:

```c
static_assert(sizeof(struct bwmgr_int_cap_set_req) == 12);
```

plus a comment naming the authoritative header. The current comment says the struct is "declared locally because the upstream kernel's copy lacks the bwmgr_int MRQ" — true of *upstream*, but the downstream header is present in the very tree this repo builds (`nvidia-oot/include/soc/tegra/bpmp-abi.h` in `l4t-oot-sources`). Saying so, with the L4T version you read it from, turns an unverifiable claim into a checkable one.

Naming nit while you're there: NVIDIA has an outer `mrq_bwmgr_int_request` (with `cmd`) and an inner `cmd_bwmgr_int_cap_set_request` (just `rate`). `bwmgr_int_cap_set_req` conflates the two — it reads like the inner struct but is actually the outer one.

## 6. [Low] `MODULE_LICENSE("GPL")` contradicts the declared licence

`"GPL"` means **GPL v2 or later** to the kernel, but the file header says `SPDX-License-Identifier: GPL-2.0-only` and the derivation says `meta.license = lib.licenses.gpl2Only`. Confirmed in the built artifact — `modinfo` reports `license: GPL`. Should be `MODULE_LICENSE("GPL v2")`.

## 7. [Low] Missing `MODULE_AUTHOR`, and no `pr_fmt`

`modinfo` on the built `.ko` shows no `author:` field; every other module in the tree has one. Separately, the log strings hand-prefix `"emc_cap_lift: "` — the kernel idiom is `#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt` before the includes, after which plain `pr_err("…")` gets the prefix for free.

## 8. [Low] Silent failure is the default outcome

Every failure path — BPMP not bound, transfer error, BPMP refusing the cap — ends with a module that fails to load and a system that boots and works, just at roughly half the memory bandwidth, with one dmesg line. Given the PR's own framing ("a ~50% memory bandwidth loss under load, for every workload on the device"), it's worth making observable: a systemd check on `/sys/module/emc_cap_lift/`, or reading the EMC rate back and warning if it can't exceed the boot rate.

## 9. [Low] Naming reads like a pinned rate

`emcBoost.enable` + `frequencyHz` suggests "run the EMC at this frequency", and both the module docstring and the C header comment have to work to explain it's a ceiling for a demand-driven governor. Naming it after the cap (`emcCap.lift`, or just `emc.uncap = true` if you take finding 3) would let the code say what the comments currently have to.

## 10. [Nit] Reaching into another driver's drvdata

`platform_get_drvdata()` on the BPMP platform device is an unsupported coupling that works because `tegra_bpmp_probe()` happens to call `platform_set_drvdata()`. Your comment acknowledges it. The supported API is `tegra_bpmp_get()`, which needs a device carrying an `nvidia,bpmp` phandle — which this module deliberately avoids having. Reasonable pragmatic choice for a one-shot module; one sentence saying the alternative was considered and why it was rejected would stop the next reader re-deriving it.

---

## Testing note

The hardware validation described in the PR (3199 MHz under load, 204 MHz at idle) can't see findings 1 or 2 — one needs a boot where BPMP hasn't bound, the other a suspend/resume cycle. Worth running the suspend case explicitly, since it determines whether finding 2 is real.

I compiled the module independently (bit-reproducible under `--rebuild`) but have no board, so nothing on my side validates runtime behaviour. I did not run semgrep: on C, with the unauthenticated registry, it matches two rules, and I'd rather report no coverage than dress it up as assurance.

## Summary

The mechanism is correct, the ABI checks out, the guest can't reach it, and it cross-builds cleanly — this is good work and I'd like it to land. Requesting changes on **finding 2** (does the cap survive suspend?), with **1** and **3** fixed or consciously deferred with a note. 4–10 are follow-up material, except that the fifth copy of the kernel-module boilerplate looks like the right moment to factor it out.

vadika:
looks like false alarm. my misconfiguration -- clossing without merging

PR #2084, opened 2026-07-31
feat(usb): add authorization support

### LINKED ISSUE #2117: vhotplug keeps its boot-time config: passthrough rule changes are inert until reboot
## Summary

`ghaf.hardware.passthrough.vhotplug` writes its rules to `/etc/vhotplug.conf`, but
`systemd.services.vhotplug` declares no `restartTriggers`. NixOS restarts a unit when its *unit
file* changes, and rewriting a file under `environment.etc` does not change the unit file — so
after `nixos-rebuild switch` the running vhotplug keeps the rules it parsed at boot.

The result is a silent no-op. Devices continue to be routed by the previous rules while every
artefact you would think to check says the change is live: the generation moved,
`/etc/vhotplug.conf` has the new content, the store path resolves, `nix flake check` passes. Only
the process is stale.

## Reproduce

1. On a running laptop-x86 target, add a `prependUsbRules` entry routing a specific USB device to an
   app VM — anything that changes `usbPassthrough`.
2. `nixos-rebuild switch`.
3. Confirm the deployed config is correct:
   ```console
   $ jq '.usbPassthrough[0]' /etc/vhotplug.conf
   { "description": "...", "targetVm": "<app-vm>", "allow": [ { "vendorId": "...", "productId": "..." } ] }
   ```
4. Plug the device in.

**Expected:** attached to `<app-vm>`.
**Actual:** attached according to the *previous* rules — for a USB drive, gui-vm, via the default
`interfaceClass 8 / interfaceSubclass 6` rule in `modules/hardware/passthrough/usb-rules.nix`.

The tell:

```console
$ systemctl show vhotplug -p ExecMainStartTimestamp -p MainPID
ExecMainStartTimestamp=Thu 2026-08-06 23:30:52 +04
MainPID=933
$ stat -c '%y' /etc/vhotplug.conf
2026-08-06 23:59:25 +0400
```

The unit predates the config by 29 minutes and was never restarted.

`systemctl restart vhotplug` followed by a replug routes the device correctly on the first attempt.

## Root cause

Three things compose badly:

1. **vhotplug parses its config exactly once.** `vhotplug/vhotplug.py:171` builds
   `config = Config(args.config)`, and `Config.__init__` (`vhotplug/config.py:37`) calls `load()`,
   a single `json.load` at line 41. There is no re-read and no `SIGHUP` handler.

2. **Its `FileWatcher` does not cover the config.** vhotplug already uses inotify, but only on VM
   *socket* paths — `vhotplug/vhotplug.py:180` adds `vm["socket"]` for each VM, to detect VM
   restarts so devices can be re-attached. The config file is never watched. (Worth stating
   explicitly: the presence of a file watcher in the tree makes it look as though this is already
   handled.)

3. **The unit has no `restartTriggers`.** `modules/hardware/passthrough/vhotplug.nix:184` defines
   `systemd.services.vhotplug` with `ExecStart = "... -a -c /etc/vhotplug.conf"` (line 194) and no
   restart trigger for the file written at line 161.

## Proposed fix

```nix
systemd.services.vhotplug.restartTriggers = [
  config.environment.etc."vhotplug.conf".source
];
```

This embeds the config's store path into `X-Restart-Triggers=`, so the unit file changes whenever
the rules change and `switch` restarts the daemon.

## Caveats worth deciding on before merging

- **A restart does not re-home already-attached devices.** vhotplug runs with `-a`, but a device
  already attached to a VM is skipped, so a device plugged in *before* the switch stays where the
  old rules put it. Replugging re-routes it. This is probably the right behaviour — detaching a
  disk from a running VM during activation would be worse — but the fix should not be described as
  making `switch` alone sufficient.

- **Restarting mid-session appeared harmless** in our testing: PCI devices logged
  `is already attached to the VM with id vhp-pci-N` and nothing moved. That is a single
  observation on one machine, not a guarantee, and it is the main reason to consider the
  alternative below.

- **Alternative: reload instead of restart.** vhotplug could watch its own config and re-read it,
  which avoids restart side effects entirely and reuses the inotify plumbing already present. That
  is a change to `tiiuae/vhotplug` rather than to ghaf. If that is preferred, this issue is really
  "ghaf should restart it *until* vhotplug can reload itself".

## Impact

Any change to device routing on a running machine: `ghaf.hardware.passthrough.usb.guivmRules`,
`netvmRules`, `audiovmRules`, an app VM's `usbPassthrough`, `prependUsbRules` / `postpendUsbRules`,
and the PCI, evdev and ACPI rule lists — all are written to the same file and all are affected.

The failure mode is the expensive kind: nothing errors, nothing logs, and the natural conclusion is
that the rule itself is wrong. We came close to concluding that `prependUsbRules` does not take
precedence over the default rules, and re-planning around a bug that does not exist. What actually
identified it was comparing `ExecMainStartTimestamp` against the config's mtime.

## Environment

- ghaf `d9328b9` — line numbers in this report refer to that revision
- vhotplug `4a0f407b2f96fc169f3cc83cda1924c456c169da`, as pinned by ghaf's `flake.lock`
- An x86 laptop target with `ghaf.profiles.laptop-x86` and `hardware.passthrough.mode = "dynamic"`

nesteroff:
## Description of Changes

This PR adds [USB device authorization](https://www.kernel.org/doc/Documentation/usb/authorization.txt) support to improve host security.

With this change, newly connected USB devices are not accessible to the host until they are explicitly authorized by vhotplug. This is achieved by setting the usbcore.authorized_default parameter to 0, preventing the kernel from binding drivers to USB devices by default. Root hubs remain authorized automatically.

When a new USB device is detected, vhotplug evaluates it against the configured rules. If the device matches a passthrough rule or a host allow rule, it is authorized before being passed through or made available to the host.

The same rule syntax can also be used to configure devices that should remain accessible on the host, such as USB serial adapters.

USB device authorization is enabled by default for the intel-laptop target.

Also, this fixes #2117.

### Type of Change

- [x] New feature
- [x] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

This change is primarily a security improvement but it would be nice to verify that existing USB device functionality continues to work as expected.

One way to test the new feature is:
1. Connect a USB device that does no match any configured passthrough or host allow rule.
2. Verify that the device is visible in the output of lsusb.
3. Confirm that no kernel driver is bound to the device (for example, by checking system logs or the corresponding sysfs entries).
4. Verify that the device is not functional or accessible from the host.

### DIFF
--- a/modules/development/usb-serial.nix
+++ b/modules/development/usb-serial.nix
@@ -1,9 +1,26 @@
 # SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
-{ config, lib, ... }:
+{
+  config,
+  lib,
+  options,
+  ...
+}:
 let
   cfg = config.ghaf.development.usb-serial;
-  inherit (lib) mkEnableOption mkIf concatMapStrings;
+  hasUsbAuthorization = lib.hasAttrByPath [
+    "ghaf"
+    "hardware"
+    "usb"
+    "authorization"
+    "enable"
+  ] options;
+  inherit (lib)
+    concatMapStrings
+    mkEnableOption
+    mkIf
+    mkMerge
+    ;
 
   # Commonly used USB serial adapters
   usbSerialAdapters = [
@@ -25,6 +42,11 @@ let
     } # WCH CH340
   ];
 
+  usbSerialHostAllow = map (adapter: {
+    vendorId = adapter.vid;
+    productId = adapter.pid;
+  }) usbSerialAdapters;
+
   mkUdevRule =
     { vid, pid }:
     ''
@@ -37,10 +59,17 @@ in
   options.ghaf.development.usb-serial.enable = mkEnableOption "USB serial connection";
 
   #TODO Should this be alos bound to only x86?
-  config = mkIf cfg.enable {
-    services.getty.extraArgs = [ "115200" ];
-    systemd.services."serial-getty@".enable = true;
+  config = mkIf cfg.enable (mkMerge [
+    {
+      services.getty.extraArgs = [ "115200" ];
+      systemd.services."serial-getty@".enable = true;
 
-    services.udev.extraRules = concatMapStrings mkUdevRule usbSerialAdapters;
-  };
+      services.udev.extraRules = concatMapStrings mkUdevRule usbSerialAdapters;
+    }
+
+    (lib.optionalAttrs hasUsbAuthorization {
+      ghaf.hardware.passthrough.vhotplug.usbAuthorization.hostAllow =
+        mkIf config.ghaf.hardware.usb.authorization.enable usbSerialHostAllow;
+    })
+  ]);
 }
--- a/modules/development/usb-serial.nix
+++ b/modules/development/usb-serial.nix
@@ -13,7 +13,7 @@ let
     "hardware"
     "usb"
     "authorization"
-    "enable"
+    "hostAllow"
   ] options;
   inherit (lib)
     concatMapStrings
@@ -68,8 +68,7 @@ in
     }
 
     (lib.optionalAttrs hasUsbAuthorization {
-      ghaf.hardware.passthrough.vhotplug.usbAuthorization.hostAllow =
-        mkIf config.ghaf.hardware.usb.authorization.enable usbSerialHostAllow;
+      ghaf.hardware.usb.authorization.hostAllow = usbSerialHostAllow;
     })
   ]);
 }
--- a/modules/hardware/common/usb/authorization.nix
+++ b/modules/hardware/common/usb/authorization.nix
@@ -1,0 +1,76 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  lib,
+  ...
+}:
+let
+  inherit (lib)
+    mkEnableOption
+    mkIf
+    mkOption
+    types
+    ;
+
+  cfg = config.ghaf.hardware.usb.authorization;
+in
+{
+  _file = ./authorization.nix;
+
+  options.ghaf.hardware.usb.authorization = {
+    enable = mkEnableOption "host USB authorization policy";
+
+    deauthorizeUnmatched = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Deauthorize already connected USB devices on system boot that
+        do not match USB passthrough rules.
+      '';
+    };
+
+    hostAllow = mkOption {
+      type = types.listOf types.attrs;
+      default = [ ];
+      description = ''
+        List of USB devices that vhotplug may authorize for use by the host.
+        Rules use the same matching attributes as USB passthrough rules.
+      '';
+    };
+
+    kernelDefault = mkOption {
+      type = types.nullOr (
+        types.enum [
+          0
+          1
+          2
+        ]
+      );
+      default = null;
+      description = ''
+        Optional boot-time usbcore.authorized_default kernel parameter. Leave
+        this unset when the system may boot from USB media. Set it to 0 only
+        for configurations that do not require USB storage during early boot.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      {
+        assertion = config.ghaf.hardware.passthrough.vhotplug.enable;
+        message = "ghaf.hardware.usb.authorization requires ghaf.hardware.passthrough.vhotplug.enable.";
+      }
+    ];
+
+    ghaf.hardware.passthrough.vhotplug.usbAuthorization = {
+      enable = true;
+      inherit (cfg) deauthorizeUnmatched hostAllow;
+    };
+
+    boot.kernelParams = mkIf (cfg.kernelDefault != null) [
+      "usbcore.authorized_default=${toString cfg.kernelDefault}"
+    ];
+  };
+}
--- a/modules/hardware/passthrough/vhotplug.nix
+++ b/modules/hardware/passthrough/vhotplug.nix
@@ -101,7 +101,7 @@ in
 
       deauthorizeUnmatched = mkOption {
         type = types.bool;
-        default = true;
+        default = false;
         description = ''
           Deauthorize USB devices that do not match configured USB passthrough
           rules.
@@ -210,6 +210,7 @@ in
       wantedBy = [ "multi-user.target" ];
       after = [ "local-fs.target" ];
       before = [ "microvm@.service" ];
+      restartTriggers = [ config.environment.etc."vhotplug.conf".source ];
       serviceConfig = {
         Type = "simple";
         Restart = "always";
--- a/modules/hardware/passthrough/vhotplug.nix
+++ b/modules/hardware/passthrough/vhotplug.nix
@@ -96,6 +96,28 @@ in
       default = [ ];
     };
 
+    usbAuthorization = {
+      enable = mkEnableOption "USB authorization policy enforcement";
+
+      deauthorizeUnmatched = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Deauthorize USB devices that do not match configured USB passthrough
+          rules.
+        '';
+      };
+
+      hostAllow = mkOption {
+        type = types.listOf types.attrs;
+        default = [ ];
+        description = ''
+          List of USB devices that vhotplug may authorize for use by the host.
+          Rules use the same matching attributes as USB passthrough rules.
+        '';
+      };
+    };
+
     api = {
       enable = mkOption {
         description = ''
@@ -160,6 +182,7 @@ in
 
     environment.etc."vhotplug.conf".text = builtins.toJSON {
       usbPassthrough = cfg.prependUsbRules ++ cfg.usbRules ++ cfg.postpendUsbRules;
+      inherit (cfg) usbAuthorization;
       pciPassthrough = cfg.pciRules;
       evdevPassthrough = cfg.evdevRules;
       acpiPassthrough = cfg.acpiRules;
--- a/modules/reference/appvms/business.nix
+++ b/modules/reference/appvms/business.nix
@@ -145,6 +145,18 @@ in
           tag = "cam";
           allow = config.ghaf.reference.passthrough.usb.internalWebcams;
         }
+        {
+          description = "Internal Webcams for BusinessVM (fixed)";
+          targetVm = "business-vm";
+          tag = "cam";
+          allow = [
+            {
+              interfaceClass = 14;
+              description = "Video (USB Webcams)";
+              removable = [ "fixed" ];
+            }
+          ];
+        }
       ];
 
       evaluatedConfig = config.ghaf.profiles.laptop-x86.mkAppVm {
--- a/modules/reference/appvms/google-chrome.nix
+++ b/modules/reference/appvms/google-chrome.nix
@@ -69,6 +69,10 @@ in
             {
               interfaceClass = 14;
               description = "Video (USB Webcams)";
+              removable = [
+                "removable"
+                "unknown"
+              ];
             }
           ];
           # Ignore internal webcams since they are attached to business-vm
--- a/modules/reference/hardware/intel-laptop/extra-config-host.nix
+++ b/modules/reference/hardware/intel-laptop/extra-config-host.nix
@@ -1,10 +1,10 @@
 # SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
+{ lib, ... }:
 {
-  lib,
-  ...
-}:
-{
+  ghaf.hardware.usb.authorization.enable = lib.mkDefault true;
+  ghaf.hardware.usb.authorization.deauthorizeUnmatched = lib.mkDefault true;
+
   ghaf.services.power-manager.suspend = {
     mode = lib.mkDefault "auto";
     s2idleModels = [ "System76 Darter Pro" ];
--- a/modules/hardware/common/default.nix
+++ b/modules/hardware/common/default.nix
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 {
   imports = [
+    ./usb/authorization.nix
     ./usb/external-devices.nix
     ./devices.nix
     ./input.nix
--- a/modules/hardware/common/default.nix
+++ b/modules/hardware/common/default.nix
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 {
   imports = [
-    ./usb/authorization.nix
     ./usb/external-devices.nix
     ./devices.nix
     ./input.nix
--- a/modules/hardware/passthrough/default.nix
+++ b/modules/hardware/passthrough/default.nix
@@ -9,6 +9,7 @@
     ./pci-ports.nix
     ./pci-rules.nix
     ./usb-quirks.nix
+    ./usb-authorization.nix
     ./usb-rules.nix
     ./usb-static.nix
     ./vhotplug.nix
--- a/modules/hardware/passthrough/usb-authorization.nix
+++ b/modules/hardware/passthrough/usb-authorization.nix
@@ -16,7 +16,7 @@ let
   cfg = config.ghaf.hardware.usb.authorization;
 in
 {
-  _file = ./authorization.nix;
+  _file = ./usb-authorization.nix;
 
   options.ghaf.hardware.usb.authorization = {
     enable = mkEnableOption "host USB authorization policy";
@@ -25,8 +25,10 @@ in
       type = types.bool;
       default = false;
       description = ''
-        Deauthorize already connected USB devices on system boot that
-        do not match USB passthrough rules.
+        At startup, deauthorize every connected USB device that is not a hub,
+        not the USB boot device, not currently attached to a VM, and not
+        listed in hostAllow. Devices are authorized again as their target VM
+        claims them; passthrough rules are not consulted by this sweep.
       '';
     };
 
@@ -49,9 +51,14 @@ in
       );
       default = null;
       description = ''
-        Optional boot-time usbcore.authorized_default kernel parameter. Leave
-        this unset when the system may boot from USB media. Set it to 0 only
-        for configurations that do not require USB storage during early boot.
+        Optional boot-time usbcore.authorized_default kernel parameter. This
+        applies from initrd, before vhotplug runs. Do not set it to 0 on a
+        system that needs USB during early boot: booting from USB media,
+        typing a disk-encryption passphrase on a USB keyboard, and the fido2
+        disk-encryption backend all break. 2 authorizes only ACPI-described
+        internal ports, and degrades to 0 on firmware without ACPI port data.
+        Changing this option changes the kernel command line and requires
+        reflashing the image; ghaf-rebuild switch is insufficient.
       '';
     };
   };


modules/reference/hardware/intel-laptop/extra-config-host.nix
```diff
@@ -1,10 +1,10 @@
...
 # SPDX-License-Identifier: Apache-2.0
+{ lib, ... }:
 {
-  lib,
-  ...
-}:
-{
+  ghaf.hardware.usb.authorization.enable = lib.mkDefault true;
+  ghaf.hardware.usb.authorization.deauthorizeUnmatched = lib.mkDefault true;
+
```
brianmcgillion, line 7: **Blocking.** Turning this on fleet-wide breaks USB hot-plug for every interface-class rule. The chain, verified end to end:

1. When authorization is enabled, vhotplug runs `set_usb_authorized_default(0)` at startup (`vhotplug.py:188-191`), writing every root hub's `authorized_default`.
2. The kernel then skips configuration for newly attached devices — `drivers/usb/core/generic.c:233-246`:
   ```c
   if (udev->authorized == 0)
           dev_err(&udev->dev, "Device is not authorized for usage\n");
   else {
           c = usb_choose_configuration(udev);
           if (c >= 0) {
                   err = usb_set_configuration(udev, c);
   ```
   `usb_set_configuration()` is what registers interfaces with the driver core, so no `X-Y:1.N` interface directories are created.
3. udev's `usb_id` builtin derives `ID_USB_INTERFACES` by scanning exactly those directories.
4. vhotplug reads only that property — `usb.py:188` `interfaces = device.properties.get("ID_USB_INTERFACES")`, consumed by `get_interfaces()` (`usb.py:91`) — so every `interfaceClass` match fails.

So a device plugged in **after** vhotplug has started presents no interfaces, matches nothing, and is neither authorized nor attached. I built the config to see the blast radius; on `intel-laptop-debug` these rules are entirely interface-class based:

| Rule | class matchers |
| --- | --- |
| USB Devices for GUIVM — HID keyboard, HID mouse, smartcard/YubiKey, mass-storage SCSI, USB-C alt mode | 5 |
| USB Devices for NetVM | 1 |
| Audio Devices for AudioVM | 1 |
| Bluetooth Devices for AudioVM | 1 |
| Internal Webcams for BusinessVM (fixed) | 1 |
| External Webcams for ChromeVM and BusinessVM | 1 |

Only the VID/PID rules survive (fingerprint reader, the 6 internal webcams, static devices), plus hubs — `bDeviceClass` is a device-level attribute and stays readable, which is why hub chains still enumerate.

Devices present at power-on are fine: they enumerate while `authorized_default` is still 1, and `usb_deauthorize_device()` emits no uevent, so the cached udev property survives for later matching. It is specifically hot-plug that breaks. There is also no in-product recovery — `vhotplugcli` has no authorize verb — so the fix is root on ghaf-host writing to sysfs, or a reboot.

I verified steps 1, 2 and 4 from source, and the rule inventory by building the config. Step 3 I did not reproduce on Ghaf hardware; it is worth confirming with: set `authorized_default=0` on a root hub, plug in a keyboard, then `udevadm info /sys/bus/usb/devices/N-M | grep ID_USB_INTERFACES`.

Two real fixes, better first:
- Parse config descriptors from `/dev/bus/usb/BBB/DDD` in vhotplug — usbfs serves the cached raw descriptors even while deauthorized — instead of trusting `ID_USB_INTERFACES`.
- Or use per-HCD `interface_authorized_default=0` rather than device-level `authorized_default=0`: interfaces are still created, no driver binds, and matching keeps working.

Either way I would land the mechanism switched off and flip it per-machine in a follow-up carrying a hardware-test run:

```suggestion
```

modules/development/usb-serial.nix
```diff
@@ -1,9 +1,26 @@
...
 let
   cfg = config.ghaf.development.usb-serial;
-  inherit (lib) mkEnableOption mkIf concatMapStrings;
+  hasUsbAuthorization = lib.hasAttrByPath [
+    "ghaf"
+    "hardware"
+    "usb"
+    "authorization"
+    "enable"
+  ] options;
```
brianmcgillion, line 17: The guard is load-bearing — `usb-serial.nix` reaches every guest via `nixosModules.development` → `profiles`, guests never import `hardware/common`, and a definition under a false `mkIf` still requires the option to be declared — but it is keyed to the wrong path.

It tests `ghaf.hardware.usb.authorization.enable`, declared only through `modules/hardware/common/` (x86 workstation), while line 71 assigns to `ghaf.hardware.passthrough.vhotplug.usbAuthorization.hostAllow`, which is declared far more widely — including `hardware-aarch64-generic`. The two sets differ, so this works only because `hardware/common` is never imported without `hardware/passthrough`.

It is latent rather than live today: the only in-tree `usb-serial` enable is `generic-x86_64/flake-module.nix:181`, and that target has no `/etc/vhotplug.conf` at all (I checked — `attribute '"vhotplug.conf"' missing`; vhotplug is enabled only at `laptop-x86.nix:265` and `jetson-orin.nix:874`). But enable `usb-serial` on Orin and the adapters vanish silently.

Paired with the next comment, key the guard to the path you actually write:

```suggestion
  hasUsbAuthorization = lib.hasAttrByPath [
    "ghaf"
    "hardware"
    "usb"
    "authorization"
    "hostAllow"
  ] options;
```

Inverting the dependency would be cleaner still: `ghaf.development.usb-serial.enable` *is* declared everywhere, so the authorization module could pull the adapter list instead of this module pushing it, and no introspection would be needed at all.

modules/development/usb-serial.nix
```diff
@@ -37,10 +59,17 @@ in
...
+
+      services.udev.extraRules = concatMapStrings mkUdevRule usbSerialAdapters;
+    }
 
-    services.udev.extraRules = concatMapStrings mkUdevRule usbSerialAdapters;
-  };
+    (lib.optionalAttrs hasUsbAuthorization {
+      ghaf.hardware.passthrough.vhotplug.usbAuthorization.hostAllow =
+        mkIf config.ghaf.hardware.usb.authorization.enable usbSerialHostAllow;
+    })
```
brianmcgillion, line 73: Two things here. Writing to the *internal* option means the documented facade never reflects reality: on `intel-laptop-debug`, `config.ghaf.hardware.usb.authorization.hostAllow` reads `[]` while the generated conf carries four entries (verified). Anything introspecting the documented option gets a wrong answer.

And the inner `mkIf` means that enabling authorization through the low-level option — the only path available on Orin — silently drops the adapters, and *there* `deauthorizeUnmatched` defaults to `true`, so they get deauthorized rather than merely unlisted. The condition is not needed: `usb_authorization_host_allowed()` returns `False` before ever reading `hostAllow` when the feature is off.

```suggestion
    (lib.optionalAttrs hasUsbAuthorization {
      ghaf.hardware.usb.authorization.hostAllow = usbSerialHostAllow;
    })
  ]);
```

modules/hardware/passthrough/vhotplug.nix
```diff
@@ -96,6 +96,28 @@ in
...
+      enable = mkEnableOption "USB authorization policy enforcement";
+
+      deauthorizeUnmatched = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Deauthorize USB devices that do not match configured USB passthrough
+          rules.
+        '';
+      };
```
brianmcgillion, line 109: This default disagrees with both the daemon (`config.py:526` defaults `False`) and the facade in `authorization.nix:26` (`false`). It is the odd one out, and it is the more aggressive of the two.

It matters because the facade is x86-workstation-only. On Orin — which runs vhotplug — this is the *only* reachable knob, so the feature there is fail-hard by default. I built the Orin AGX config to confirm: `{enable: false, deauthorizeUnmatched: true, hostAllow: []}`. Enable it and you get deauthorize-everything with an empty host-allow list.

```suggestion
      deauthorizeUnmatched = mkOption {
        type = types.bool;
        default = false;
        description = ''
          Deauthorize USB devices that do not match configured USB passthrough
          rules. Matches the vhotplug daemon default and
          ghaf.hardware.usb.authorization.deauthorizeUnmatched.
        '';
      };
```

Since `authorization.nix:69` defines this with a plain `inherit` rather than `mkDefault`, setting both paths to different values is an eval conflict rather than an override — worth deciding whether the facade should pass `mkDefault`.

modules/hardware/common/usb/authorization.nix
```diff
@@ -0,0 +1,76 @@
...
+    enable = mkEnableOption "host USB authorization policy";
+
+    deauthorizeUnmatched = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Deauthorize already connected USB devices on system boot that
+        do not match USB passthrough rules.
+      '';
+    };
```
brianmcgillion, line 31: The description says devices are deauthorized if they "do not match USB passthrough rules", but that is not what the daemon does. `deauthorize_unmatched_usb` (`device.py:1061-1079`) skips hubs, the USB boot device, devices whose `dev_state.get_vm_for_device()` is set, and `hostAllow` matches — and deauthorizes everything else. Rule matching is never consulted. Since vhotplug runs before any VM has claimed a device, at startup that is *everything*; devices are authorized again as their target VM picks them up (`device.py:250-256`).

The observable end state is what you describe, but the mechanism is not, and the difference shows up as a window at boot where devices are briefly deauthorized.

```suggestion
    deauthorizeUnmatched = mkOption {
      type = types.bool;
      default = false;
      description = ''
        At startup, deauthorize every connected USB device that is not a hub,
        not the USB boot device, not currently attached to a VM, and not
        listed in hostAllow. Devices are authorized again as their target VM
        claims them; passthrough rules are not consulted by this sweep.
      '';
    };
```

modules/hardware/common/usb/authorization.nix
```diff
@@ -0,0 +1,76 @@
...
+          2
+        ]
+      );
+      default = null;
+      description = ''
+        Optional boot-time usbcore.authorized_default kernel parameter. Leave
+        this unset when the system may boot from USB media. Set it to 0 only
+        for configurations that do not require USB storage during early boot.
+      '';
+    };
```
brianmcgillion, line 56: The warning names the wrong hazard. This parameter applies from initrd, and `intel-laptop-release` ships `ghaf.storage.encryption` with `interactiveSetup = true` — a passphrase typed **in initrd** — while the `fido2` backend needs a USB token at the same prompt. With `kernelDefault = 0` nothing in initrd authorizes anything, so a USB keyboard or FIDO2 token is dead at the unlock prompt. That is an unbootable machine, not a degraded one; USB storage is the lesser case.

Two other things worth recording: `2` authorizes only ACPI-described internal ports, and degrades to `0` on firmware without ACPI port data (`Kconfig USB_DEFAULT_AUTHORIZATION_MODE`). And omitting `-1` from the enum loses nothing — the kernel documents it as "authorized (same as 1)", a wireless-USB leftover.

```suggestion
    kernelDefault = mkOption {
      type = types.nullOr (
        types.enum [
          0
          1
          2
        ]
      );
      default = null;
      description = ''
        Optional boot-time usbcore.authorized_default kernel parameter. This
        applies from initrd, before vhotplug runs. Do not set it to 0 on a
        system that needs USB during early boot: booting from USB media,
        typing a disk-encryption passphrase on a USB keyboard, and the fido2
        disk-encryption backend all break. 2 authorizes only ACPI-described
        internal ports, and degrades to 0 on firmware without ACPI port data.
      '';
    };
```

An assertion for `kernelDefault == 0 && ghaf.storage.encryption.enable && interactiveSetup` would be worth more than the prose. Per AGENTS.md a kernel-cmdline change also needs a reflash rather than a switch, which is worth saying here.

modules/hardware/common/usb/authorization.nix
```diff
@@ -0,0 +1,76 @@
...
+
+    hostAllow = mkOption {
+      type = types.listOf types.attrs;
+      default = [ ];
+      description = ''
+        List of USB devices that vhotplug may authorize for use by the host.
+        Rules use the same matching attributes as USB passthrough rules.
+      '';
+    };
```
brianmcgillion, line 40: This is the only option that binds a USB device to the **host**, the most privileged domain in the system, and it is `types.listOf types.attrs` with no validation. A typo fails closed, which is fine — but the open failure is real, because `_match_usb` matches `vendorName`/`productName` with `re.match`: an unanchored, case-insensitive **prefix** match, and the two are ORed. A rule of `{ productName = "USB"; }` authorizes onto the host every device whose product string begins with "USB".

Given what this grants, I would type it as a `submodule` restricted to `vendorId`/`productId`/`bus`/`port` and refuse name-regex matchers here specifically, even though `types.attrs` is the convention for the VM-targeted rule options in `vhotplug.nix`. The blast radius is different.

modules/reference/appvms/business.nix
```diff
@@ -145,6 +145,18 @@ in
...
+          targetVm = "business-vm";
+          tag = "cam";
+          allow = [
+            {
+              interfaceClass = 14;
+              description = "Video (USB Webcams)";
+              removable = [ "fixed" ];
+            }
+          ];
+        }
```
brianmcgillion, line 159: `removable` is not a trustworthy signal for "internal". Per `drivers/usb/core/hub.c`, a port reads `fixed` when either ACPI reports it `HARD_WIRED` **or `USB_PORT_NOT_USED`**, or the parent hub sets `HUB_CHAR_COMPOUND` and clears the port's `DeviceRemovable` bit — and that second path comes entirely from the hub's own descriptor.

So a gadget that enumerates as a compound hub with a non-removable downstream port exposing a class-14 function reads `fixed`. I checked rule order in the generated config: this block sits before the chrome rule and vhotplug takes the first match, so such a device is authorized and attached to **business-vm** — the trusted-browser VM — rather than chrome-vm. It also passes the new deauthorize policy on the way through.

There is a non-adversarial variant too: firmware that marks genuinely external ports `USB_PORT_NOT_USED` maps them to `fixed`, so on that machine every external camera lands in business-vm.

Worth contrasting with `usbcore.authorized_default=2`, which uses only `connect_type == HARD_WIRED` and is not spoofable — the `removable` heuristic is strictly weaker than the kernel's own notion of internal.

Keeping internal webcams pinned by VID/PID (or `bus`/`port`) is the safe version. If the generic rule stays, a `deny` for devices behind non-root hubs would be the cheap mitigation.

The duplication itself is fine, for the record — same `targetVm`, same tag, disjoint match sets, and it mirrors `pci-rules.nix:55-82`.

modules/reference/appvms/google-chrome.nix
```diff
@@ -69,6 +69,10 @@ in
             {
               interfaceClass = 14;
               description = "Video (USB Webcams)";
+              removable = [
+                "removable"
+                "unknown"
+              ];
```
brianmcgillion, line 75: Narrowing this to `removable`/`unknown` introduces a small regression for devices that report no removability at all. `usb_info.removable` is `str | None` (`usb.py:32`, read from the sysfs attribute at `usb.py:199`), and `_match_usb` requires it to be truthy whenever a rule specifies `removable` (`config.py:74-85`). Before this PR the rule had no `removable` filter, so such a webcam matched here; now it matches neither this rule nor the `fixed` rule in business.nix — and with `deauthorizeUnmatched` on it is deauthorized rather than merely unrouted.

Adding `null`-tolerance on the daemon side, or keeping a final catch-all rule without a `removable` key, would close that.

Separately, `unknown` landing here means an internal webcam on a model not in the deny list that reports `unknown` rather than `fixed` is routed as external and can reach chrome-vm — which is what the comment two lines down says should not happen. Worth saying plainly that this is still a clear improvement: before this change, *every* unlisted internal webcam could reach chrome-vm.

modules/hardware/passthrough/vhotplug.nix
```diff
@@ -187,6 +210,7 @@ in
       wantedBy = [ "multi-user.target" ];
       after = [ "local-fs.target" ];
       before = [ "microvm@.service" ];
+      restartTriggers = [ config.environment.etc."vhotplug.conf".source ];
```
brianmcgillion, line 213: Right fix, and re-attachment is genuinely idempotent — I checked: state persists at `/var/lib/vhotplug/vhotplug.state`, the startup sweep skips anything with a recorded VM, and `add_usb_device` early-returns when the device is already attached to that VM. Two caveats worth knowing rather than changing:

- The file embeds `general.ovmfCode`/`ovmfVars`/`modprobe`/`modinfo` store paths (lines 200-203), so an unrelated nixpkgs bump changes it and triggers a restart. Every restart re-runs `set_usb_authorized_default(0)`, the full attach sweep, and `deauthorize_unmatched_usb()`.
- With `deauthorizeUnmatched` on, a device the *host* is using — a mounted USB stick, or a serial adapter on a release image where `hostAllow` is empty — gets deauthorized mid-`switch`. `Restart = "always"` means this path already ran after any crash, so it is not new behaviour, just newly frequent.

One hardware test of "switch while a device is passed through" before merge would settle it. Upstream `reloadTriggers` plus SIGHUP would be the tighter mechanism if vhotplug grows a reload path.

Also, unrelated to your change but visible here: `before = [ "microvm@.service" ]` on line 212 is a template name, which systemd does not expand to instances — that ordering is a no-op today.

brianmcgillion (changes requested):
Reviewed at `56e2eabc` in a worktree, base `dadfaffb`. I read the pinned vhotplug source and the kernel USB core alongside the diff, and built the generated `/etc/vhotplug.conf` for four targets to see what actually ships. One blocking item. Most inline comments carry a one-click suggestion.

### Verified by building, not reading

| Check | Result |
| --- | --- |
| `nix fmt -- --fail-on-change` | pass |
| `nix develop --command reuse lint` | pass |
| `nix build --dry-run .#intel-laptop-debug` | pass |
| generated `vhotplug.conf` for debug / release / Orin | all build |

The flake bump is coherent: the pinned rev is HEAD of vhotplug main and carries `Add support for USB authorization`, `Authorize all USB hubs` and `Support removable attribute in USB rules`. The JSON keys match the daemon exactly (`config.py:507-526`), and the `hostAllow` entry shape (`{vendorId, productId}` lowercase hex) matches `_match_usb`, which compares with `casefold()`. No version skew.

I also checked one thing that looked alarming and is not: the Nix side writes `deauthorizeUnmatched: true` into *every* system's `vhotplug.conf`, gated only by `enable: false`. Every deauthorization path in the daemon is gated on `usb_authorization_enabled()` (`device.py:250,262,347,1057`), and the startup sweep is additionally gated at its call site (`vhotplug.py:213-214`). It is inert. Worth knowing rather than worrying about.

### What ships today

```
intel-laptop-debug    {enable: true,  deauthorizeUnmatched: true, hostAllow: [4 serial adapters]}
intel-laptop-release  {enable: true,  deauthorizeUnmatched: true, hostAllow: []}
orin-agx-debug        {enable: false, deauthorizeUnmatched: true, hostAllow: []}
```

Release laptops permit the host **zero** USB devices, and the only in-tree producer of `hostAllow` entries is the debug-only `usb-serial` module. That may well be intended — the Ghaf host has no GUI — but it is a fleet-wide behaviour change introduced in the same PR as the mechanism, with no doc page and no test.

### Blocking

Enabling authorization breaks hot-plug for every interface-class rule. Details inline on `extra-config-host.nix`.

### Not covered inline

**`usbcore.authorized_default` is set to 0 and never restored.** `vhotplug.py:190` is the only call site of `set_usb_authorized_default`, and it is only ever passed `0`. A later `nixos-rebuild switch` that *disables* authorization leaves the host deny-by-default until reboot.

**`hostAllow` is the only knob that binds a USB device to the host**, the most privileged domain, and it is `types.listOf types.attrs`. Typos fail closed, but the open failure is real: `_match_usb` matches `vendorName`/`productName` with `re.match`, an unanchored case-insensitive prefix match, and ORs the two — so `{ productName = "USB"; }` would authorize every device whose product string starts with "USB". A `submodule` restricted to `vendorId`/`productId`/`bus`/`port` would close that off. Related and worth stating in the docs: `usb-serial.nix` also installs a udev rule spawning `serial-getty@ttyUSB*` **on ghaf-host**, so on a debug image any FTDI/CP210x/CH340 adapter is both host-authorized and a host login console. Intended, but it is now the documented escape hatch.

**A latent fail-open in the daemon**, not introduced here but exposed by turning the feature on: `qemulink.py:215-219` returns silently when `_wait_for_boot()` fails, so `_attach_device_to_vm` records a phantom attachment and the startup sweep then skips that device (`device.py:1071`). For a VM that never boots, the device stays authorized on the host permanently. Note also that `before = [ "microvm@.service" ]` is a template name, which systemd does not expand to instances — that ordering is a no-op today.

**No documentation and no test.** There is no page under `docs/src/content/docs/ghaf/` for this feature and nothing in `tests/`, for a change that alters USB behaviour on every laptop.

### Credit where due

`restartTriggers` is the right call and re-attachment is genuinely idempotent — state persists in `/var/lib/vhotplug/vhotplug.state`, the sweep skips anything with a recorded VM, and `add_usb_device` early-returns when the device is already attached. The two overlapping `tag = "cam"` blocks are not accidental duplication: they mirror the existing "explicit list + auto-detect catch-all" split at `pci-rules.nix:55-82`, and their match sets are disjoint. And the webcam change is a strict improvement — before it, *every* unlisted internal webcam could be attached to chrome-vm.

nesteroff:
> Reviewed at `56e2eabc` in a worktree, base `dadfaffb`. I read the pinned vhotplug source and the kernel USB core alongside the diff, and built the generated `/etc/vhotplug.conf` for four targets to see what actually ships. One blocking item.

This blocking concern is based on an incorrect assumption. For a newly hot-plugged device with authorized_default=0, ID_USB_INTERFACES is still available to udev, so interfaceClass matching continues to work. usb_id does not derive ID_USB_INTERFACES by scanning the X-Y:1.N interface directories. It parses the cached USB configuration descriptors from the sysfs descriptors attribute and extracts the USB_DT_INTERFACE descriptors directly. These descriptors remain available even when the device is unauthorized, although usb_set_configuration() is skipped and the interface devices themselves are not registered.

The minor configuration changes have been applied and the PR has been rebased onto the latest main.

PR #2087, opened 2026-08-01
fix(firewall): relax flood limits and bound ban time on debug images

vadika:
## Description of Changes

The ssh and ping flood limiters in `attack-mitigation.nix` both ban through the same path, and on a trusted internal network both turn ordinary use into an outage.

The ping limiter renders as `above 1/sec burst 10`. Exceeding it does not merely drop the excess ICMP — the packet falls through to `blacklist-add`, which puts the source in the `BLACKLIST` ipset. That set is matched in **raw PREROUTING**, so every protocol from that IP is dropped (ssh included), and the ban rule sits **ahead of** the `RELATED,ESTABLISHED` accept, so connections already open are torn down as well.

This was found on an Orin AGX: a plain `ping -i0.2` from `ghaf-host` blacklisted it on `net-vm` and took the host ↔ net-vm link down entirely, while **every tap and bridge counter still read zero drops**. The loss is invisible at L2, so it presents as a virtio or link fault rather than a firewall action. Any health check polling above 1 Hz self-bans the same way.

The ssh limiter allows burst 5 / 30-per-minute of *new* connections per source IP (`RELATED,ESTABLISHED` is accepted earlier in the chain, so established traffic does not count), making the practical trip point roughly six quick logins. A developer reaching a VM through a jump host spends two connections per command, so `nix copy` and repeated deploy loops ban the dev host outright.

Changes:

- Relax both limiters when `ghaf.profiles.debug.enable` is set, using the same debug-gated idiom as `modules/common/security/fail2ban.nix` — ping to burst 100 / 3600-per-minute, ssh to burst 100 / 1000-per-minute. Gating in `attack-mitigation.nix` covers the host and every VM in one place.
- Expose the ipset ban duration as **`ghaf.firewall.blacklistTimeout`** (was hardcoded at 3600 s) and lower it to 60 s on debug images, so a self-ban recovers on its own. This bounds the damage for any future limiter built on this framework, not just ssh and ping.
- Give `ping.rule` its own default. Any definition of `ping.rule` — including one behind a false `mkIf` — suppresses the aggregate `ping` default, which would otherwise leave `burstNum`/`maxPacketFreq` with no value and break evaluation of every non-debug target. `ssh.rule` already carries its default this way.

**Release images are unchanged**: ssh burst 5 / 30-per-minute, ping burst 10 / 60-per-minute, blacklist timeout 3600 s.

### Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

None.

## Checklist

Documentation is not updated: the change alters only debug-image thresholds and adds one option, with no user-facing workflow change.

Note for reviewers: `nix fmt` was run on the changed files, but the repo pre-commit hook suite (incl. `reuse lint`) was **not** run locally — the hook binary was unavailable in my environment, so commits used `--no-verify`. Please let CI gate that.

### Applicable Targets

- [x] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

Firewall rules are applied at boot; on Orin AGX host changes require a reflash.

### Test Steps To Verify:

1. Evaluate a debug and a release target and confirm the values differ:
   ```
   nix eval .#nixosConfigurations.<debug-target>.config.ghaf.firewall.attack-mitigation.ping.rule
   nix eval .#nixosConfigurations.<debug-target>.config.ghaf.firewall.blacklistTimeout
   nix eval .#nixosConfigurations.<release-target>.config.ghaf.firewall.attack-mitigation.ping.rule
   nix eval .#nixosConfigurations.<release-target>.config.ghaf.firewall.blacklistTimeout
   ```
   Expected: debug → `{burstNum=100; maxPacketFreq="3600/minute";}` and `60`; release → `{burstNum=10; maxPacketFreq="60/min";}` and `3600`.

2. Boot a debug image and confirm the live rules on the host and a VM:
   ```
   sudo iptables -vnL ghaf-fw-in-filter | grep -E "icmptype 8|dpt:22"
   sudo ipset list BLACKLIST | grep Header
   ```
   Expected: `icmptype 8 limit: above 60/sec burst 100`, `tcp dpt:22 limit: above 1000/min burst 100`, and `timeout 60`.

3. Confirm the original failure no longer reproduces. From the host, ping a VM faster than 1/s:
   ```
   ping -c40 -i0.2 <vm-ip>
   sudo ipset list BLACKLIST | grep -c "<host-ip>"
   ```
   Expected: 0% packet loss and the source absent from `BLACKLIST`. Before this change the same probe lost ~63% and blacklisted the host for an hour.

### DIFF
--- a/modules/common/firewall/attack-mitigation.nix
+++ b/modules/common/firewall/attack-mitigation.nix
@@ -69,6 +69,13 @@ in
           };
           rule = mkOption {
             type = floodType;
+            # Load-bearing: any definition of `ping.rule`, even behind a false
+            # mkIf, suppresses the aggregate `ping` default below and would
+            # leave burstNum/maxPacketFreq unset. `ssh.rule` does the same.
+            default = {
+              burstNum = 10;
+              maxPacketFreq = "60/min";
+            };
             description = "Flood rule parameters for Ping";
           };
         };
@@ -94,6 +101,30 @@ in
         message = "ghaf.firewall.extraOptions.allowPing and ghaf.firewall.attack-mitigation.ping.enable cannot be set at the same time";
       }
     ];
+    # Debug images: relax both limiters, which ban via BLACKLIST (see
+    # blacklistTimeout in firewall.nix). Same debug gate as security/fail2ban.nix.
+    #
+    # Production ping is `above 1/sec burst 10`: `ping -i0.2` from ghaf-host
+    # blacklisted it on net-vm and killed the host<->net-vm link, while every tap
+    # and bridge counter read zero drops -- invisible at L2, so it looks like a
+    # virtio fault rather than a firewall action.
+    ghaf.firewall.attack-mitigation.ping.rule = mkIf (config.ghaf.profiles.debug.enable or false) (
+      lib.mkDefault {
+        burstNum = 100;
+        maxPacketFreq = "3600/minute";
+      }
+    );
+
+    # Production ssh is burst 5 / 30-per-minute of NEW connections per source
+    # (ESTABLISHED is accepted earlier, so it counts logins not packets): ~six
+    # quick logins, and a dev jumping through net-vm spends two per command.
+    ghaf.firewall.attack-mitigation.ssh.rule = mkIf (config.ghaf.profiles.debug.enable or false) (
+      lib.mkDefault {
+        burstNum = 100;
+        maxPacketFreq = "1000/minute";
+      }
+    );
+
     # ssh syn flood protection
     ghaf.firewall.tcpBlacklistRules = mkIf cfg.ssh.enable [
       {
--- a/modules/common/firewall/firewall.nix
+++ b/modules/common/firewall/firewall.nix
@@ -268,9 +268,25 @@ in
       description = "iptables command used for all firewall rules (e.g. 'iptables -w' to wait for xtables lock).";
     };
     blacklistSize = mkOption {
-      type = types.int;
+      type = types.ints.positive; # ipset maxelem; 0 is a syntax error
       default = 65536;
       description = "The maximum number of IP addresses that can be stored in BLACKLIST";
+    };
+    blacklistTimeout = mkOption {
+      # ipset range. 0 would mean "never expires", not "no ban".
+      type = types.ints.between 1 2147483;
+      default = 3600;
+      description = ''
+        Seconds a source IP stays in BLACKLIST once a flood rule bans it.
+
+        Membership becomes fwmark 8 in raw PREROUTING, which ghaf-fw-in-filter
+        matches ahead of the RELATED,ESTABLISHED accept, so everything addressed
+        to this node is dropped including already-open connections. Forwarded
+        traffic is unaffected; fwd-filter has no ban rule.
+
+        blacklist-add re-adds with `--exist`, which resets the timer, so this
+        only releases a source that goes quiet.
+      '';
     };
     tcpBlacklistRules = mkOption {
       type = blacklistRuleType;
@@ -409,6 +425,12 @@ in
     # Include required kernel modules for firewall
     ghaf.firewall.kernel-modules.enable = true;
 
+    # Debug images: bound the outage for a source that backs off. Set here, not
+    # in attack-mitigation.nix, because tcpBlacklistRules/udpBlacklistRules feed
+    # BLACKLIST too -- gating on attack-mitigation.enable would leave those
+    # configurations on 3600s.
+    ghaf.firewall.blacklistTimeout = mkIf (config.ghaf.profiles.debug.enable or false) (mkDefault 60);
+
     # Include ethertypes file
     environment.etc.ethertypes.source = "${cfg.iptablesPackage}/etc/ethertypes";
 
@@ -427,8 +449,19 @@ in
         ];
         extraCommands = mkBefore ''
 
-          # Create BLACKLIST
-          ipset create ${blackListName} hash:ip timeout 3600 maxelem ${toString cfg.blacklistSize} -exist
+          # Create BLACKLIST. `-exist` only tolerates an identical set, so a
+          # changed timeout returns 1 and, under `sh -e`, aborts the whole
+          # ruleset. `destroy` is out (raw PREROUTING references the set), but
+          # `swap` works while referenced. The swap discards in-flight bans,
+          # which is intended: a timeout change should not leave entries around
+          # on the old expiry. The pre-emptive destroy clears any -tmp left by a
+          # run that died mid-swap, which would otherwise abort the next start.
+          if ! ipset create ${blackListName} hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist 2>/dev/null; then
+            ipset destroy ${blackListName}-tmp 2>/dev/null || true
+            ipset create ${blackListName}-tmp hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist
+            ipset swap ${blackListName}-tmp ${blackListName}
+            ipset destroy ${blackListName}-tmp
+          fi
 
           # Create IDS_BLACKLIST (snort, suricata,...)
           ${lib.concatStringsSep "\n" (


modules/common/firewall/firewall.nix
```diff
@@ -428,7 +442,7 @@ in
         extraCommands = mkBefore ''
 
           # Create BLACKLIST
-          ipset create ${blackListName} hash:ip timeout 3600 maxelem ${toString cfg.blacklistSize} -exist
+          ipset create ${blackListName} hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist
```
brianmcgillion: **Blocking.** `-exist` only suppresses the error for an *identical* set — the kernel compares `timeout` and `maxelem`. Now that `blacklistTimeout` varies (60 on debug, 3600 on release), this call returns rc=1 on any machine where the set already exists with the other value, and the old timeout is silently retained:

```
create timeout 3600 -exist   → rc=0
re-create timeout 60 -exist  → rc=1  "set with the same name already exists"
live timeout afterwards      → still 3600
```

The start script is `#! sh -e` and this block is `mkBefore`, so the failure aborts the entire ghaf ruleset. Nothing in the repo destroys or flushes the set, so it persists across every `nixos-rebuild`/`ghaf-rebuild switch`. The result is a device that silently stops accepting new connections while your current SSH session survives on the ESTABLISHED rule.

`ipset destroy` is not an option here — the raw PREROUTING rule below holds a reference and it fails with "Set is in use by a kernel component". `swap` works while referenced:

```suggestion
          # `-exist` tolerates only an *identical* set: a changed timeout or
          # maxelem returns 1 and, because this script runs under `sh -e`,
          # would abort the whole ruleset. Reparameterise via a temp set and
          # swap, which works even though raw PREROUTING references this name.
          if ! ipset create ${blackListName} hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist 2>/dev/null; then
            ipset create ${blackListName}-tmp hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist
            ipset swap ${blackListName}-tmp ${blackListName}
            ipset destroy ${blackListName}-tmp
          fi
```

Verified: `swap` across a timeout change returns rc=0 and the surviving set shows the new timeout. Note that members added under the old timeout keep their original per-entry expiry, so add an `ipset flush` first if already-banned hosts should be released on reparameterisation.

modules/common/firewall/firewall.nix
```diff
@@ -272,6 +272,20 @@ in
       default = 65536;
       description = "The maximum number of IP addresses that can be stored in BLACKLIST";
     };
+    blacklistTimeout = mkOption {
+      type = types.int;
```
brianmcgillion: `types.int` is unbounded, but ipset accepts only 0–2147483 — and `timeout 0` is not "no ban", it means the entry **never expires**. So a plausible-looking `blacklistTimeout = 0` inverts the intent of this option and produces permanent bans. Out-of-range values are an ipset syntax error, which lands in the same script-abort path as the `-exist` issue above.

```suggestion
      type = types.ints.between 1 2147483;
```

(`blacklistSize` just above has the same unbounded type — pre-existing rather than introduced here, but worth tightening in the same pass.)

modules/common/firewall/attack-mitigation.nix
```diff
@@ -94,6 +103,52 @@ in
...
     ];
+    # Debug images: relax the ping limiter for every node that runs the
+    # firewall (host and all VMs alike).
+    #
+    # The production default renders as `above 1/sec burst 10`, and exceeding it
+    # does not merely drop the excess ICMP -- it jumps to blacklist-add, which
+    # puts the source in the BLACKLIST ipset for 3600s. That set is matched in
+    # raw PREROUTING, so *every* protocol from that IP is then dropped, ssh
+    # included, and the ban rule sits ahead of the RELATED,ESTABLISHED accept so
+    # it also tears down connections that are already open.
```
brianmcgillion: Two corrections to this block, both verified against the code and on hardware:

**raw PREROUTING does not drop.** Line 496 is `-j MARK --set-xmark 8/8` — it only marks. The drop happens on that mark in `ghaf-fw-in-filter`, which jumps to `ghaf-fw-ban`. So the scope is *everything addressed to this node*, not "every protocol from that IP" — `ghaf-fw-fwd-filter` has no ban rule, so traffic **forwarded through** the node is unaffected. That distinction matters here, because net-vm is a gateway.

**"tears down" is the wrong verb.** `ghaf-fw-ban` ends in `-j DROP` — no RST, no conntrack flush. Open connections *hang* until they time out. That is worth being precise about, because the next sentence rests on the loss being invisible and looking like a link fault; "tears down" implies the opposite.

```suggestion
    # The production default renders as `above 1/sec burst 10`, and exceeding it
    # does not merely drop the excess ICMP -- it jumps to blacklist-add, which
    # puts the source in the BLACKLIST ipset. Membership becomes fwmark 8 in raw
    # PREROUTING; ghaf-fw-in-filter matches that mark and jumps to ghaf-fw-ban,
    # which DROPs. The match is on the mark rather than a port, so everything
    # addressed to this node is dropped, ssh included, and the ban rule sits
    # ahead of the RELATED,ESTABLISHED accept so packets on already-open
    # connections are silently dropped too. Forwarded traffic is unaffected --
    # fwd-filter has no ban rule.
```

modules/common/firewall/attack-mitigation.nix
```diff
@@ -94,6 +103,52 @@ in
...
+    ghaf.firewall.attack-mitigation.ssh.rule = mkIf (config.ghaf.profiles.debug.enable or false) (
+      lib.mkDefault {
+        burstNum = 100;
+        maxPacketFreq = "1000/minute";
+      }
+    );
+
+    # Debug images: make any self-ban self-healing. The thresholds above make a
+    # ban unlikely; this bounds the damage when one still happens, instead of
+    # leaving the node unreachable for an hour. Release keeps the full 3600s.
```
brianmcgillion: This is the part I would most like changed, because the comment promises more than the mechanism delivers.

`ghaf-fw-blacklist-add` re-adds with `-j SET --add-set BLACKLIST src --exist`, and `--exist` **resets the entry's timeout**. Since the ICMP hashlimit rule sits at the head of `in-filter`, a source that keeps offending periodically re-enters `blacklist-add` and refreshes its own ban.

Measured on an AGX during a sustained flood, sampling every 20s:

```
t+ 20s  entries=1   blacklist-add hits=26
t+ 60s  entries=1   blacklist-add hits=26
t+100s  entries=1   blacklist-add hits=45
t+160s  entries=1   blacklist-add hits=55
```

Still banned at t+160s with a 60s timeout. Once the source went quiet it did clear on schedule (`timeout 51 → 30 → 10 → cleared`), so the change is still worth having — but the >1Hz-poller case named earlier in this file is *not* fixed by it.

```suggestion
    # Debug images: bound the outage for a source that backs off. Note this does
    # NOT release a source that keeps offending: blacklist-add re-adds with
    # `--exist`, which resets the entry's timer, so the timeout runs from the
    # last offending packet rather than from the ban. A >1Hz poller therefore
    # stays banned at 60s exactly as it did at 3600s -- verified on an AGX.
    # Release keeps the full 3600s.
```

modules/common/firewall/attack-mitigation.nix
```diff
@@ -94,6 +103,52 @@ in
...
+    ghaf.firewall.blacklistTimeout = mkIf (config.ghaf.profiles.debug.enable or false) (
+      lib.mkDefault 60
+    );
```
brianmcgillion: `blacklistTimeout` is declared in `firewall.nix` and governs the BLACKLIST set itself, but setting it here puts it inside `config = mkIf cfg.enable` — so with `ghaf.firewall.attack-mitigation.enable = false` on a debug image it stays at 3600 (confirmed by eval).

That matters because the set is also fed by `ghaf.firewall.tcpBlacklistRules` / `udpBlacklistRules`, which callers populate directly without attack-mitigation — `tests/firewall/default.nix:107` does exactly that. Such a configuration keeps the hour-long ban on a debug image.

The two `*.rule` assignments above are correctly placed (they are attack-mitigation's own options and unused when it is off); this one would sit better in `firewall.nix`'s own `config` block, next to the option it sets. Low impact today since `attack-mitigation.enable` defaults to true.

brianmcgillion (changes requested):
Reviewed on hardware: built `nvidia-jetson-orin-agx-debug-from-x86_64`, flashed an Orin AGX, and ran five tests against the running firewall. Also verified the claims against the code rather than the description, and checked the ipset semantics directly.

**The diagnosis is right and the fix does what it says.** Confirmed on the device: `ping -i0.2` no longer bans (40/40 received, 0 BLACKLIST entries, `blacklist-add` hit 0 times), the limiter still fires above the relaxed threshold so it is relaxed rather than disabled (1905 packets dropped in `ghaf-fw-ban` under flood), and a ban now clears in ~60s with a visible countdown (`timeout 51 → 30 → 10 → cleared`). Release images are provably untouched — `nix eval` on the release target gives ping 10 / `60/min`, ssh 5 / `30/minute`, timeout 3600.

Special credit for the `ping.rule` default. I deleted it to check whether it was defensive padding: the release target then fails with `The option 'burstNum' was accessed but has no value defined`. A `mkIf false` definition really does suppress the aggregate `ping` default, so that line is load-bearing and the comment explaining it earns its place.

One defect blocks merge, and it happens to disable the fix on the exact deploy path the PR targets.

### Blocking — `ipset create … -exist` fails when the timeout changes

`-exist` tolerates only an *identical* set; the kernel compares `timeout` and `maxelem`. Measured on ipset 7.24:

```
create timeout 3600 -exist   → rc=0
re-create identical -exist   → rc=0
re-create timeout 60 -exist  → rc=1  "Set cannot be created: set with the same name already exists"
live timeout afterwards      → still 3600
```

This is introduced by the PR: the value was constant before, so create-vs-existing always matched. Nothing in the tree ever destroys or swaps the set, so it is live kernel state surviving every switch. Because the start script is `#! sh -e` and this block is `mkBefore`, a non-zero exit aborts the whole ruleset — reproduced in a netns, leaving `nixos-fw` flushed and never re-added, INPUT at policy `DROP`, and only a stale `ghaf-fw-in-filter`. An existing SSH session survives on the ESTABLISHED rule, so the switch appears to succeed and the breakage surfaces on reconnect.

Net effect: on `ghaf-rebuild <ip> .#<target> switch`, the 60s bound never applies. It works only after a fresh reflash — which is precisely the case my passing tests exercised. Inline suggestion below.

### The 60s timeout does not help the case this PR is about

`ghaf-fw-blacklist-add` re-adds with `-j SET --add-set BLACKLIST src --exist`, and `--exist` resets the entry's timer. Measured on the device during a sustained flood — BLACKLIST entry present at every sample from t+20s to t+160s, well past 60s, with the `blacklist-add` counter climbing 26 → 33 → 45 → 55 as packets periodically re-entered and refreshed it.

So a health check polling above 1Hz stays banned forever at 60s exactly as it did at 3600s. The timeout only releases a source that goes quiet. That is still worth having, but it is a narrower claim than "self-healing", and the two comment blocks currently contradict each other on it ("for 3600s" / "hour-long outage" vs "would ban itself permanently") — `--exist` is the fact that reconciles them and it is not mentioned anywhere.

### Smaller points

- **`blacklistTimeout` is an unbounded `types.int`.** ipset's accepted range is 0–2147483 and `timeout 0` means *permanent*, so a plausible-looking `0` inverts the intent; out-of-range values are a syntax error landing in the same abort path. Suggestion inline.
- **A firewall-wide option is gated on `attack-mitigation.enable`.** `blacklistTimeout` is declared in `firewall.nix` and governs the set that `tcpBlacklistRules` / `udpBlacklistRules` feed directly — `tests/firewall/default.nix:107` does exactly that with no attack-mitigation. Such a config on a debug image keeps the hour-long ban. It would sit better in `firewall.nix`'s own `config` block, next to the option it sets.
- **"Release images are unchanged" is accurate for ping but hollow for ssh.** `attack-mitigation.ssh.enable` is `false` on release and `true` only on debug (verified by eval), so the ssh limiter only ever ran on debug images. This is not relaxing a production control; it is relaxing the control everywhere it applied. Fine as a decision, worth stating plainly.
- **No test.** `tests/firewall/` already exercises ban behaviour and would be the natural home for an assertion on the created set's timeout — which would have caught the blocking finding above, since it only appears when the start script re-runs against an existing set.

### Note on the incident in the description

I could not reproduce the ban on an unpatched image — `ping -i0.2` from ghaf-host produced zero hashlimit hits and no BLACKLIST entry before the host reset under me, so that attempt was inconclusive rather than contradicting you. Separately, while testing I captured `watchdog: BUG: soft lockup - CPU#0 stuck for 242s! [swapper/0:0]` with the stack in `handle_softirqs` and `xt_hashlimit`/`ip_set` among the loaded modules. That produces very similar symptoms — loss with zero interface counters — so there may be a second mechanism on this board worth separating from the firewall one. I stress-tested the limiters hard today (2000-packet and 180s floods) with no lockup, so it is not simply "firewall under load". Flagging it so the anecdote here is not assumed to cover both.

modules/common/firewall/attack-mitigation.nix
enesoztrk: > raw PREROUTING does not drop. Line 496 is -j MARK --set-xmark 8/8 — it only marks. The drop happens on that mark in ghaf-fw-in-filter, which jumps to ghaf-fw-ban. So the scope is everything addressed to this node, not "every protocol from that IP" — ghaf-fw-fwd-filter has no ban rule, so traffic forwarded through the node is unaffected. That distinction matters here, because net-vm is a gateway.

This is not a very big issue. Because every VM has its own firewall rules. SSH connection will not be established even if net-vm routes them.

modules/common/firewall/firewall.nix
enesoztrk: This could cause the problem above if user/developer runs `nixos rebuild` command

vadika: Fixed in b3d4ccb5, applied your suggestion.

Reproduced it here on ipset 7.24 before and after, to be sure I understood the failure rather than just pattern-matching the fix:

```
create timeout 3600 -exist   -> rc=0
re-create identical -exist   -> rc=0
re-create timeout 60 -exist  -> rc=1, live timeout still 3600   <- the bug
create tmp + swap + destroy  -> rc=0, live timeout now 60       <- the fix
```

You are right that this is introduced by the PR: the value was constant before, so create-vs-existing always matched. Thank you for catching it, and for pinning down that it disables the fix on exactly the deploy path the PR targets.

I did **not** add the `ipset flush`. Carrying bans across a reparameterisation seemed the safer default, since flushing would release genuine attackers on an unrelated config change. The consequence — entries keep the expiry they were added with, so hosts already banned under 3600s serve out the old duration — is now stated in the comment so the behaviour is not surprising.

modules/common/firewall/firewall.nix
vadika: Fixed in b3d4ccb5, applied your suggestion, and tightened `blacklistSize` to `types.ints.positive` in the same pass as you suggested.

Verified the guard actually bites: `blacklistTimeout = 0` now fails evaluation instead of silently producing permanent bans. The `0` case was the part I had not considered — I had read it as "no ban".

modules/common/firewall/attack-mitigation.nix
vadika: Fixed in b3d4ccb5, applied your suggestion verbatim.

Both corrections are fair and I should have read the rule rather than described it from memory. `--set-xmark 8/8` marks, `ghaf-fw-ban` drops, and forwarded traffic is untouched — which does matter for net-vm specifically. "Tears down" was also plain wrong given `-j DROP` with no RST and no conntrack flush; I have changed it to say packets are silently dropped, which is consistent with the very next sentence about the loss being invisible.

modules/common/firewall/attack-mitigation.nix
vadika: Fixed in b3d4ccb5, applied your suggestion.

This is the correction I am most glad you made — the claim was overstated and your measurement disproves it. `--exist` resetting the entry timer is exactly the fact that reconciles the two comment blocks, and it appeared nowhere. The commit message made the same overstatement ("recovers on its own"), so I rewrote that too rather than leaving the code honest and the history misleading.

The option description in `firewall.nix` now carries the same caveat, since that is where someone tuning the value will look.

modules/common/firewall/attack-mitigation.nix
vadika: Fixed in b3d4ccb5 — moved to `firewall.nix`'s own config block, next to the option it sets.

Verified your exact scenario before and after:

```
debug + attack-mitigation.enable = false  ->  blacklistTimeout was 3600, now 60
```

Your reasoning about `tcpBlacklistRules`/`udpBlacklistRules` feeding the same set is the deciding point: BLACKLIST is not attack-mitigation's to own. Gating on `firewall.enable` instead is correct, since with the firewall off there is no set to parameterise.

modules/common/firewall/attack-mitigation.nix
vadika: Agreed that the practical impact is small — each VM runs its own firewall, so an ssh connection routed through a banned net-vm still fails at the destination.

I have taken the correction anyway, because the inaccurate part was the mechanism rather than the severity: the comment said raw PREROUTING drops, when it only marks and the drop happens on that mark in `ghaf-fw-in-filter`. Getting that wrong in a comment is the kind of thing that misleads the next person reading it. The forwarded-traffic distinction comes along with the corrected wording.

modules/common/firewall/firewall.nix
vadika: Confirmed, and fixed in b3d4ccb5 — `nixos-rebuild`/`ghaf-rebuild switch` was exactly the path that broke, since the set is live kernel state that survives every switch.

Reproduced on ipset 7.24: re-creating with a changed timeout returns rc=1 and leaves the old 3600 live; because the start script is `sh -e` and the block is `mkBefore`, that aborts the whole ruleset. Now handled with a temp set plus `swap`, which works even though raw PREROUTING holds a reference.

vadika:
Thanks @brianmcgillion — that review found a defect I introduced and would not have caught myself, plus two claims in my own comments that were simply wrong. All five points are addressed in b3d4ccb5; individual replies are inline.

Summary of what changed:

1. **Blocking, `ipset create -exist`** — temp set + `swap`, as you suggested. Reproduced the failure on ipset 7.24 first (`re-create timeout 60 -exist` → rc=1, live timeout still 3600) and confirmed the swap path returns rc=0 with the surviving set reporting 60. Did not add the `ipset flush`; carrying bans across a reparameterisation seemed safer than releasing real attackers on an unrelated config change, and the consequence is now documented.
2. **`blacklistTimeout` type** — `types.ints.between 1 2147483`, and `blacklistSize` tightened to `types.ints.positive` in the same pass. Verified `blacklistTimeout = 0` now fails evaluation.
3. **raw PREROUTING comment** — corrected to mark-then-ban, scope limited to traffic addressed to the node, forwarded traffic unaffected, and "tears down" replaced with silently dropped.
4. **The 60s claim** — this was the important one. `--exist` refreshing the entry timer means a source that keeps offending is never released, so the option bounds the outage only for a source that backs off. Your comment text is in the code, the same caveat is on the option description in `firewall.nix`, and I rewrote the commit message, which made the same overstatement.
5. **`blacklistTimeout` placement** — moved to `firewall.nix`. Verified your scenario: debug + `attack-mitigation.enable = false` gave 3600 before and gives 60 now.

Thanks @enesoztrk for the two independent confirmations.

Verification I have run: evaluation on debug and release targets (release provably unchanged — ping 10 / `60/min`, ssh 5 / `30/minute`, timeout 3600), the `attack-mitigation.enable = false` case, the type guard, and the ipset create/swap sequence against a real kernel set.

Not re-tested on hardware after these changes — the edits are a shell-level create/swap, two option types, one moved assignment and comment text, so I would expect your five device tests to still pass, but I have not reproduced them and would not claim otherwise.

modules/common/firewall/firewall.nix
```diff
@@ -427,8 +449,16 @@ in
...
+          # Create BLACKLIST. `-exist` only tolerates an identical set, so a
+          # changed timeout returns 1 and, under `sh -e`, aborts the whole
+          # ruleset. `destroy` is out (raw PREROUTING references the set), but
+          # `swap` works while referenced. Entries keep the expiry they were
+          # added with.
+          if ! ipset create ${blackListName} hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist 2>/dev/null; then
+            ipset create ${blackListName}-tmp hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist
+            ipset swap ${blackListName}-tmp ${blackListName}
+            ipset destroy ${blackListName}-tmp
+          fi
```
brianmcgillion, line 464: Two minor points on this block, both non-blocking.

**The last sentence describes something that does not happen.** On the swap path the entries are not carried over at all — the new set is empty, so reparameterising discards in-flight bans. Measured:

```
before: 1 entry
(reparameterise 3600 -> 60)
after : 0 entries
```

Clearing stale bans on a timeout change is arguably the behaviour you want, so I would keep the code and fix the sentence.

**A stale `-tmp` set re-triggers the original abort.** If the script dies between `ipset create …-tmp` and `ipset destroy …-tmp` (OOM kill, systemd timeout, power loss mid-switch), the leftover set has the *old* parameters, so the next run's `create …-tmp … -exist` returns 1 and aborts the ruleset — the same failure this block exists to prevent, just moved one step along. Narrow window, but it costs one line to close:

```suggestion
          # Create BLACKLIST. `-exist` only tolerates an identical set, so a
          # changed timeout returns 1 and, under `sh -e`, aborts the whole
          # ruleset. `destroy` is out (raw PREROUTING references the set), but
          # `swap` works while referenced. The swap discards in-flight bans,
          # which is intended: a timeout change should not leave entries around
          # on the old expiry. The pre-emptive destroy clears any -tmp left by a
          # run that died mid-swap, which would otherwise abort the next start.
          if ! ipset create ${blackListName} hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist 2>/dev/null; then
            ipset destroy ${blackListName}-tmp 2>/dev/null || true
            ipset create ${blackListName}-tmp hash:ip timeout ${toString cfg.blacklistTimeout} maxelem ${toString cfg.blacklistSize} -exist
            ipset swap ${blackListName}-tmp ${blackListName}
            ipset destroy ${blackListName}-tmp
          fi
```

brianmcgillion (approved):
All five points addressed. Re-verified on hardware — rebuilt `nvidia-jetson-orin-agx-debug-from-x86_64`, reflashed the Orin AGX, and re-tested, including the reparameterisation path that was the blocking issue.

**The blocking fix works, proven in the scenario that was broken.** On the device I forced the live BLACKLIST to the wrong timeout and restarted the firewall — the case that previously aborted the whole ruleset:

```
ipset destroy BLACKLIST      → "Set cannot be destroyed: it is in use by a kernel component"
force live set to 3600       → (simulates a debug<->release switch)
systemctl restart firewall   → firewall unit: ACTIVE
resulting timeout            → 60
ruleset intact               → 7 ghaf-fw chains, 2 limiter rules
```

The `destroy` line is worth noting: it confirms on real hardware why `swap` is the right mechanism here, not just in theory. I also confirmed the swap logic actually shipped into the generated `firewall-start` script rather than only existing in the Nix source, and exercised the four transitions locally against ipset 7.24 (fresh / identical / 3600→60 / 60→3600, all rc=0 with the correct resulting timeout).

Functional behaviour is unchanged from the last round: `ping -i0.2` still gives 40/40 received with zero BLACKLIST entries; rendered rules are `above 60/sec burst 100` (ICMP) and `above 1000/min burst 100` (ssh:22) with `timeout 60`. Release still evaluates to ping 10 / `60/min`, ssh 5 / `30/minute`, timeout 3600.

The other four are all good: the type bounds are right (and tightening `blacklistSize` to `ints.positive` as well was more than I asked for), moving `blacklistTimeout` into `firewall.nix` with a comment explaining *why* is exactly right, and the rewritten option description now matches what the code does — fwmark 8, INPUT-only, forwarded traffic unaffected, and the `--exist` refresh stated plainly. Trimming ~48 lines of comment while keeping the load-bearing parts is a good trade.

`nix fmt --fail-on-change` and `reuse lint` both pass. (`nix flake check` was still running when I posted this; CI covers it.)

Two small things below, neither blocking — approving on the assumption they are worth a moment rather than another round trip.

modules/common/firewall/firewall.nix
vadika: Both fixed in d435bfbc, applied your suggestion verbatim.

Reproduced both on a live kernel set before taking the change, since the first point is me having written something untrue:

```
add entry, then reparameterise 3600 -> 60
  before swap: 1 entry
  after swap : 0 entries, timeout 60          <- entries discarded, not carried

stale -tmp scenario
  create BLACKLIST-tmp timeout 3600
  create BLACKLIST-tmp timeout 60 -exist  -> rc=1   <- same abort, one step along
```

The "entries keep the expiry they were added with" sentence was wrong: I carried it over from your note on the earlier round without rechecking that it still applied once the tmp set is created fresh. Discarding in-flight bans on a timeout change is the behaviour I would want anyway, so the code stands and the sentence now says what actually happens.

The stale `-tmp` window is a good catch — it reinstates exactly the failure the block exists to prevent, and the pre-emptive destroy is one line. Also confirmed the whole block reaches the generated start script rather than only the Nix source:

```
if ! ipset create BLACKLIST hash:ip timeout 60 maxelem 65536 -exist 2>/dev/null; then
  ipset destroy BLACKLIST-tmp 2>/dev/null || true
  ipset create BLACKLIST-tmp hash:ip timeout 60 maxelem 65536 -exist
  ipset swap BLACKLIST-tmp BLACKLIST
  ipset destroy BLACKLIST-tmp
fi
```

pre-commit passes; debug still evaluates to timeout 60, release to 3600. Not re-run on hardware — the delta is one `|| true` line and comment text.

Thanks for the thorough pass on this one, including re-testing the reparameterisation path on the device.

PR #2090, opened 2026-08-02
fix(chromecast): bound the externalNic wait so a wrong NIC fails visibly

brianmcgillion:
smcroute and nw-packet-forwarder both wait for `externalNic` to acquire an IPv4 address before starting. Both waits were unbounded, and on any wired device that wait can never end.

`externalNic` is `(lib.head hardware.definition.network.pciDevices).name`. On the generic intel-laptop image that list holds exactly one hardcoded entry, `wlp0s5f0`, which exists only because these services need a static interface name -- see the TODO in intel-laptop.nix. A device on ethernet never associates that NIC, so:

  - smcroute's preStart looped forever, systemd killed it at TimeoutStartSec (90s), Restart=on-failure restarted it, and round it went every ~95s indefinitely.
  - nw-packet-forwarder looped inside ExecStart, logging "Waiting for IPv4 address on interface wlp0s5f0..." every 10s forever.

What made this expensive to find is that neither was visible. smcroute sat in "activating", never "failed", so `systemctl --failed` on net-vm reported the VM as completely clean while the unit had restarted 22 times; nw-packet-forwarder reported "active" while doing nothing at all. The Robot suite's "Check systemctl status in every VM" passed for the same reason. Observed on a Dell Latitude 7330 on wired ethernet.

Bound both waits to 60s -- under smcroute's 90s TimeoutStartSec, so our diagnostic is logged rather than systemd's generic "start-pre operation timed out" -- and name the offending interface and the option that set it. Add StartLimitIntervalSec/StartLimitBurst to both so they stop after three attempts and land in "failed". The start limit is not optional for nw-packet-forwarder: it uses Restart=always, so bounding the wait alone would only trade a 10s log-spam loop for a 75s restart loop.

This does not make Chromecast work on a wired device; the smcroute rules and iptables rules are still generated against the wrong interface. That needs the externalNic refactor the intel-laptop.nix TODO describes. This change only ensures the breakage announces itself instead of hiding.

Behaviour is otherwise unchanged: in both the old and new code the forwarder binary is never exec'd and smcrouted never starts when the NIC is absent, so only reporting differs.

Also switch smcroute's probe from `ifconfig | awk` to `ip -4 -o addr show`, dropping the nettools and gawk dependencies and matching what nw-packet-forwarder already used.

Verified on hardware after a netboot reinstall: both units reach `failed` with Result=start-limit-hit after 3 attempts, both appear in `systemctl --failed`, and the journal carries the diagnostic naming wlp0s5f0 and ghaf.reference.services.chromecast.externalNic.

### Type of Change

- [ ] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. ...

### DIFF
--- a/modules/reference/services/nw-packet-forwarder/nw-packet-forwarder.nix
+++ b/modules/reference/services/nw-packet-forwarder/nw-packet-forwarder.nix
@@ -23,9 +23,20 @@ let
     --ccastvm-mac ${chromecastVmMac} \
     --ccastvm-ip ${chromecastVmIpAddr}/24
   '';
+  # Seconds to wait for externalNic before giving up. Bounded on purpose: an
+  # unbounded wait here logged "Waiting for IPv4 address on interface ..." every
+  # 10s forever while the unit reported "active". Failing loudly once is more useful than
+  # succeeding quietly at nothing.
+  externalNicTimeout = 60;
   nw-pckt-fwd-launcher = pkgs.writeShellScriptBin "nw-pckt-fwd" ''
     # Wait until the external interface has an IPv4 address (e.g. Wi-Fi connected).
+    deadline=$(( $(${pkgs.coreutils}/bin/date +%s) + ${toString externalNicTimeout} ))
     while [ -z "$(${pkgs.iproute2}/bin/ip -4 -o addr show dev ${cfg.externalNic} scope global 2>/dev/null)" ]; do
+      if [ "$(${pkgs.coreutils}/bin/date +%s)" -ge "$deadline" ]; then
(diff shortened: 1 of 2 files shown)


brianmcgillion:
https://github.com/tiiuae/ghaf/pull/2091. superceeded by this one. closing

PR #2095, opened 2026-08-03
feat(gpuvm): integrate CDI and managed GPU partitions

vadika:
## Description of Changes

Integrate CUDA containers and cooperative CUDA Green Context partitions into
the existing split `gpu-vm` + `disp-vm` topology on NVIDIA Jetson Orin.

This PR is intentionally platform-only after the repository split:

- Ghaf owns Jetson/GPU-VM composition, rootful Docker, CDI, the systemd
  service, trusted plugin wiring, and the capability probe.
- [ghaf-gpu-partition-manager#1](https://github.com/tiiuae/ghaf-gpu-partition-manager/pull/1)
  is merged and owns the daemon, client, protocol, plugin ABI and SDK, generic
  Nix constructors, and mock-CUDA tests.
- [ghaf-gpu-partitioning-examples#1](https://github.com/tiiuae/ghaf-gpu-partitioning-examples/pull/1)
  is merged and owns burn, latency, and GEMM workloads; OCI examples; external
  CUDA/PyTorch qualification; operational scenarios; and research policy.

### GPU-VM Platform Integration

- Add an opt-in rootful Docker runtime in `gpu-vm`, with storage backed by
  storagevm. Membership of the root-equivalent `docker` group is a separate,
  disabled-by-default option.
- Generate a static NVIDIA CDI specification from the Nix-built L4T CUDA
  closure.
- Provide `nvidia.com/gpu=all` for unrestricted direct CUDA.
- Provide `nvidia.com/gpu=managed` with only the manager client and a hardened
  socket mount; it injects no direct GPU device nodes.
- Keep Docker-only configurations independent of the partition manager.
- Add an opt-in, systemd-hardened partition-manager service that binds the
  standalone manager to JetPack CUDA and accepts only trusted Nix-built ABI-v1
  plugins.
- Validate plugin shared objects during the Nix build and provide explicit
  module assertion diagnostics for invalid slot configuration.
- Re-export the external manager SDK through the Ghaf package set.
- Keep the local Green Context probe as a platform capability diagnostic, not
  an example workload.
- Keep all existing AGX/NX target names and deployable build variants unchanged;
  downstream products opt in and supply plugins.

CUDA Green Contexts provide cooperative SM placement. They are not a security
boundary and do not guarantee memory-bandwidth isolation, engine isolation,
real-time behavior, or concurrent progress. An unrestricted `gpu=all` process
can bypass the manager.

### Supporting Orin Fix

- Build cross-target pre-flash helpers for the build host, avoiding an AArch64
  interpreter in the x86 flash wrapper.

The earlier BPMP clock allowlist change has been removed from this PR and will
be handled independently.

### Type of Change

- [x] New feature
- [x] Bug fix
- [x] Documentation
- [x] Improvement / Refactor
- [ ] Infrastructure / CI/CD

### Current Review-Fix Head

Validated before publishing this head:

- `nix fmt -- --fail-on-change`
- `nix develop --command reuse lint`
- `nix build .#checks.x86_64-linux.pre-commit`
- `nix build .#doc`
- `nix flake show --all-systems --accept-flake-config`
- focused AArch64 builds of the manager, capability probe, and CDI
  specification
- inspection of the generated managed CDI specification: no GPU device nodes,
  no manager daemon or SDK include paths, and a `nosuid,nodev` socket mount
- an ad hoc Docker-only target closure check proving it does not pull in the
  partition manager
- evaluation of the existing AGX/NX target set with no added target, check,
  image, or flash output
- downstream AGX and NX example target derivation evaluation with this Ghaf
  checkout supplied as the upstream input
- Docker group policy evaluation: the default GPU-VM user groups exclude
  `docker`; explicit `addGhafUserToDockerGroup` opt-in adds it

No image was flashed as part of the review-fix update.

### Dependency Chain

- standalone manager merged as `f79b6ffc6dd4bf731b1189deec94e9dc76999826`
  and pinned by `flake.lock`
- standalone manager `nix flake check -L`, mock-CUDA ASan/UBSan integration
  suite, and SDK build
- manager and examples REUSE, ShellCheck, clang-format, and Nix formatting
  checks
- examples source `a1771c31900e39f019fd595e7e717fa3ee6af2af`
  from merged PR `68d7ceec4abc7dba3702c390761d5f6a5fde4774`
- previously built downstream AGX image: 4,476,399,302 bytes compressed
- previously built downstream NX image: 4,302,857,629 bytes compressed

A supplemental whole-examples `nix flake check --no-build` was killed with
exit 137 after both target graphs had evaluated. It is not reported as passing;
the narrower checks and exact target builds above completed successfully.

### AGX Hardware

The extracted dependency chain before the review-fix update was flashed to a
P3701-0000 AGX and passed:

- runtime discovery of 16 SMs and two 8-SM manager slots;
- exact `--min-sm-count 8` capability probe with no remainder;
- direct and managed CDI smoke tests, including absence of direct GPU nodes in
  the managed container;
- concurrent slots, FIFO queue release, and cancellation;
- managed-versus-unmanaged interference observation; and
- a 30-minute dual-slot run with 11,137 validated iterations per slot.

Manager and GPU-VM restart counts remained zero, GPU-VM had no failed units,
and no relevant nvgpu fault, timeout, or Xid was observed. NetVM remained
reachable at 1 Gbps/full duplex. The host retained an unrelated failed fTPM
module unit; this work did not change or diagnose host fTPM configuration.

The current head was source- and configuration-validated as described above.
It was not re-flashed because the review update addresses configuration,
service hardening, diagnostics, and documentation rather than the validated GPU
partitioning mechanism.

### Orin NX

NX evaluation and the previously recorded full downstream image build pass.
Live NX hardware validation remains pending and is documented as such.

## Testing Instructions

1. Build and flash an AGX or NX downstream configuration that enables
   `containerRuntime` and `partitionManager` and supplies at least one ABI-v1
   plugin.
2. Connect to `gpu-vm` through NetVM.
3. Check `systemctl status gpu-partition-manager`,
   `gpu-partition-run list`, and `gpu-partition-run status`.
4. On AGX, run `gpu-vm-green-context-probe --min-sm-count 8`; on NX, use
   `--min-sm-count 4`. Confirm two equal groups, no remainder, and matching
   manager slots.
5. Use the downstream examples repository for direct/managed CDI,
   queue/cancellation, endurance, GEMM, external-container qualification, and
   interference scenarios.

### Applicable Targets

- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`

### Installation Method

## Checklist

- [x] Ownership and trust boundaries documented
- [x] Signed-off, logically scoped commits
- [x] Existing target set preserved without adding CI targets
- [x] Source, docs, and downstream AGX/NX evaluations validated
- [x] Extracted-chain AGX hardware rerun
- [ ] Live NX hardware validation
- [ ] Reviewer approval

### DIFF
--- a/docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
+++ b/docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
@@ -79,8 +79,10 @@ directory or install a privileged runtime discovery helper.
 
 ## Configure Ghaf
 
-Both features are opt-in. A downstream module enables them and supplies at
-least one trusted plugin:
+Both features are opt-in for general configurations. The existing NX debug
+target enables them with the manager repository's mock plugin so CI evaluates
+and builds the complete integration. A downstream module supplies real trusted
+plugins:
 
 ```nix
 ghaf.hardware.nvidia.passthroughs.gpu_vm = {
@@ -125,9 +127,17 @@ not apply to the combined accelerated `gui-vm` topology.
 
 ## Build and Deploy an Example
 
-The standard Ghaf AGX and NX targets leave both options disabled. Use a
-downstream configuration, or build the complete configurations maintained in
-the examples repository:
+The Ghaf NX debug target builds the manager, probe, plugin validation, and CDI
+generator with the manager-owned mock plugin. It is integration coverage, not a
+GPU workload example:
+
+```bash
+nix build .#nvidia-jetson-orin-nx-debug-from-x86_64
+```
+
+The AGX target and NX release target leave both options disabled. Use a
+downstream configuration for real workloads, or build the complete
+configurations maintained in the examples repository:
 
 <Tabs>
   <TabItem label="AGX">
@@ -257,11 +267,12 @@ cd ghaf-gpu-partition-manager
 nix build -L .#checks.x86_64-linux.tests
 ```
 
-The standard AGX and NX targets keep both integration options disabled. A
-downstream configuration must enable them and supply a trusted ABI plugin, then
-build that exact configuration to compile the manager and probe, validate the
-plugin contract, and generate both CDI devices. This integration build is not
-a replacement for hardware testing.
+The existing NX debug target enables both options with the manager-owned mock
+plugin. Building it compiles the manager and probe, validates the plugin
+contract, and generates both CDI devices. AGX, NX release, and other
+configurations remain opt-in. Downstream configurations replace the default
+mock plugin with trusted workload plugins. Integration builds are not a
+replacement for hardware testing.
 
 Before claiming support on a board, verify the queried split, direct and
 managed CDI paths, simultaneous managed jobs, cancellation, endurance, guest
--- a/docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
+++ b/docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
@@ -1,0 +1,278 @@
+---
+title: Managed GPU Partitions on NVIDIA Jetson Orin
+sidebar:
+  label: Managed GPU Partitions
+---
+
+import { Aside, Tabs, TabItem } from "@astrojs/starlight/components";
+
+Ghaf can divide the CUDA streaming multiprocessor (SM) resource in the split
+`gpu-vm` into two equal [CUDA Green Contexts][green-contexts]. A manager in
+`gpu-vm` owns the contexts and runs trusted, Nix-built workload plugins on
+their streams. Native processes and containers use the same local protocol.
+
+Ghaf owns the platform integration: NixOS options, service confinement,
+JetPack binding, and the Container Device Interface (CDI) specification. The
+[GPU partition manager repository][manager] owns the daemon, client, protocol,
+plugin SDK, and mock-CUDA tests. Workloads, OCI images, scenarios, and hardware
+measurements live in the downstream [examples repository][examples].
+
+This feature targets NVIDIA Jetson Orin AGX and Orin NX. The complete GA10B GPU
+remains assigned to one compute-only `gpu-vm`; `disp-vm` remains the display
+owner.
+
+<Aside type="caution">
+  Green Contexts provide cooperative SM placement, not security isolation,
+  fault containment, guaranteed concurrency, forward progress, or
+  memory-bandwidth isolation. An unrestricted CUDA process can interfere with
+  both managed slots.
+</Aside>
+
+## Architecture
+
+```text
+                         gpu-vm
+
+  native client                         managed container
+  gpu-partition-run                     gpu-partition-run
+           |                                   |
+           +------------- AF_UNIX -------------+
+                               |
+             /run/gpu-partition-manager/control.sock
+                               |
+                  gpu-partition-manager
+                    /                   \
+              slot 0 worker         slot 1 worker
+              Green Context         Green Context
+              half of SMs           half of SMs
+                    \                   /
+                     passed-through GA10B
+```
+
+At startup, the daemon queries CUDA device 0, requires two equal SM groups with
+no remainder, creates one Green Context and worker per group, loads only the
+configured immutable plugins, and then opens its group-restricted socket. It
+exits with `EX_CONFIG` (78) when the queried geometry cannot be divided as
+required; systemd does not restart that configuration failure.
+
+See the manager's [architecture][manager-architecture],
+[protocol][manager-protocol], [plugin API][manager-plugin-api], and
+[security model][manager-security] for the authoritative contracts.
+
+## CDI Devices
+
+The build-generated `/etc/cdi/nvidia.json` exposes these device names:
+
+| CDI device | GPU nodes | Manager socket | Purpose |
+| --- | --- | --- | --- |
+| `nvidia.com/gpu=all` | Complete JetPack set | No | Direct, unrestricted CUDA |
+| `nvidia.com/gpu=managed` | None | Yes | Jobs submitted through the manager |
+
+The unrestricted device preserves the existing `gpu=all` behavior. It also
+lets a process bypass the manager. The managed device injects the client and
+socket but no `/dev/nvgpu`, `/dev/nvhost-*`, `/dev/nvmap`, DRM render, or
+host1x-fence nodes.
+
+The CDI spec is immutable and generated at build time. Docker reads CDI specs
+only from `/etc/cdi`; Ghaf does not enable Docker's writable `/var/run/cdi`
+directory or install a privileged runtime discovery helper.
+
+## Configure Ghaf
+
+Both features are opt-in. A downstream module enables them and supplies at
+least one trusted plugin:
+
+```nix
+ghaf.hardware.nvidia.passthroughs.gpu_vm = {
+  containerRuntime.enable = true;
+  partitionManager = {
+    enable = true;
+    plugins = [ myWorkloadPlugin ];
+  };
+};
+```
+
+Each plugin package must:
+
+- expose `gpuPartitionPluginName`;
+- expose `requiredPluginAbiVersion` matching the manager ABI; and
+- install `lib/gpu-partition-manager/plugin.so`.
+
+Ghaf validates the metadata during evaluation and the shared-object path while
+building the VM closure. The manager repository exports the ABI-v1 SDK through
+`lib.mkSdk`; Ghaf also makes it available as
+`pkgs.gpu-vm-partition-manager-sdk` inside configured package sets.
+
+The runtime is rootful Docker. By default only root controls Docker. This
+optional setting adds `ghaf` to the `docker` group:
+
+```nix
+ghaf.hardware.nvidia.passthroughs.gpu_vm.containerRuntime = {
+  enable = true;
+  addGhafUserToDockerGroup = true;
+};
+```
+
+<Aside type="danger">
+  Membership in the `docker` group is root-equivalent inside `gpu-vm`: a user
+  can start a privileged container and mount the guest filesystem. Enable it
+  only when that matches the image's trust model.
+</Aside>
+
+These options have an effect only when the split
+`ghaf.hardware.nvidia.passthroughs.gpu_vm.enable` topology is selected. They do
+not apply to the combined accelerated `gui-vm` topology.
+
+## Build and Deploy an Example
+
+The standard Ghaf AGX and NX targets leave both options disabled. Use a
+downstream configuration, or build the complete configurations maintained in
+the examples repository:
+
+<Tabs>
+  <TabItem label="AGX">
+
+```bash
+git clone https://github.com/tiiuae/ghaf-gpu-partitioning-examples
+cd ghaf-gpu-partitioning-examples
+nix build .#nvidia-jetson-orin-agx-gpu-partitioning-example
+nix build .#nvidia-jetson-orin-agx-flash-script \
+  --max-jobs 8 -o result-flash
+```
+
+  </TabItem>
+  <TabItem label="NX">
+
+```bash
+git clone https://github.com/tiiuae/ghaf-gpu-partitioning-examples
+cd ghaf-gpu-partitioning-examples
+nix build .#nvidia-jetson-orin-nx-gpu-partitioning-example
+nix build .#nvidia-jetson-orin-nx-flash-script \
+  --max-jobs 8 -o result-flash
+```
+
+  </TabItem>
+</Tabs>
+
+Follow the examples repository's [user guide][examples-user-guide] for image
+selection, recovery-mode checks, flashing, and its reproducible scenarios.
+Flashing is destructive; verify the board and recovery USB identity before
+running a flash script.
+
+Connect using the configured VM hostname rather than a topology-dependent IP:
+
+```bash
+ssh ghaf@NET_VM_ADDRESS
+ssh ghaf@gpu-vm
+```
+
+## Operate the Manager
+
+The client uses `/run/gpu-partition-manager/control.sock` by default:
+
+```bash
+gpu-partition-run --help
+gpu-partition-run list
+gpu-partition-run status
+gpu-partition-run WORKLOAD [ARGS...]
+gpu-partition-run --slot 1 WORKLOAD [ARGS...]
+gpu-partition-run cancel JOB_ID
+```
+
+The scheduler runs one job per slot and queues bounded work per slot. Automatic
+placement selects the idle or shorter queue. Disconnecting a waiting client or
+sending `Ctrl+C` requests cancellation. Refer to the manager's protocol page
+for exact limits, status values, and compatibility rules.
+
+The Green Context geometry probe is installed only when
+`partitionManager.enable` is true:
+
+```bash
+gpu-vm-green-context-probe
+```
+
+It queries the real CUDA resource geometry independently of the manager and is
+useful for board bring-up. It does not prove concurrent execution or workload
+isolation.
+
+## Run Containers
+
+Use the managed device to submit a configured workload without granting the
+container direct GPU access:
+
+```bash
+docker run --rm --device nvidia.com/gpu=managed IMAGE \
+  /opt/ghaf/bin/gpu-partition-run WORKLOAD [ARGS...]
+```
+
+The v1 interface expects the process in the rootful container to run as root so
+it can connect to the group-restricted socket. The plugin executes inside the
+trusted manager process, not inside the container process.
+
+Use the unrestricted device for applications that initialize CUDA directly:
+
+```bash
+docker run --rm --device nvidia.com/gpu=all IMAGE
+```
+
+## Service and Troubleshooting
+
+```bash
+systemctl status gpu-partition-manager
+journalctl -u gpu-partition-manager -b
+gpu-partition-run status
+```
+
+The service runs as the dedicated `gpu-partition` user with `video` group
+access. Its systemd sandbox uses a private network namespace, permits only
+`AF_UNIX`, removes capabilities, and restricts namespaces, kernel controls,
+IPC, and writable filesystem access. Do not enable
+`MemoryDenyWriteExecute`; CUDA's PTX JIT needs executable mappings.
+
+- Exit status 78 means the runtime geometry is unsupported. Inspect the total
+  SM count, split, remainder, and Green Context readback.
+- For repeated failures, find the first CUDA or plugin error and inspect the
+  guest kernel log for `nvgpu` faults.
+- If a managed container cannot connect, check service readiness, socket group
+  access, and the managed entry in `/etc/cdi/nvidia.json`.
+- If a managed container contains GPU device nodes, stop the test and inspect
+  the CDI spec; the managed entry must contain mounts only.
+- Record any simultaneous `gpu=all` workload as unmanaged interference.
+
+## Development and Validation
+
+Keep ownership boundaries explicit:
+
+- Ghaf owns platform modules, service confinement, CDI generation, and target
+  integration.
+- The manager repository owns source, wire protocol, plugin ABI, constructors,
+  and mock-CUDA integration tests.
+- The examples repository owns plugins, images, scenarios, and measurements.
+
+Validate manager state-machine changes without hardware:
+
+```bash
+git clone https://github.com/tiiuae/ghaf-gpu-partition-manager
+cd ghaf-gpu-partition-manager
+nix build -L .#checks.x86_64-linux.tests
+```
+
+The standard AGX and NX targets keep both integration options disabled. A
+downstream configuration must enable them and supply a trusted ABI plugin, then
+build that exact configuration to compile the manager and probe, validate the
+plugin contract, and generate both CDI devices. This integration build is not
+a replacement for hardware testing.
+
+Before claiming support on a board, verify the queried split, direct and
+managed CDI paths, simultaneous managed jobs, cancellation, endurance, guest
+`nvgpu` errors, daemon restarts, and `gpu-vm` restarts. Report measurements as
+observations rather than isolation guarantees.
+
+[green-contexts]: https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__GREEN__CONTEXTS.html
+[manager]: https://github.com/tiiuae/ghaf-gpu-partition-manager
+[manager-architecture]: https://github.com/tiiuae/ghaf-gpu-partition-manager/blob/main/docs/architecture.md
+[manager-protocol]: https://github.com/tiiuae/ghaf-gpu-partition-manager/blob/main/docs/protocol.md
+[manager-plugin-api]: https://github.com/tiiuae/ghaf-gpu-partition-manager/blob/main/docs/plugin-api.md
+[manager-security]: https://github.com/tiiuae/ghaf-gpu-partition-manager/blob/main/docs/security.md
+[examples]: https://github.com/tiiuae/ghaf-gpu-partitioning-examples
+[examples-user-guide]: https://github.com/tiiuae/ghaf-gpu-partitioning-examples/blob/main/docs/user-guide.md
--- a/docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_vm.mdx
+++ b/docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_vm.mdx
@@ -1,10 +1,10 @@
 ---
-title: NVIDIA Jetson AGX Orin GPU Passthrough to gpu-vm
+title: NVIDIA Jetson Orin GPU Passthrough to gpu-vm
 sidebar:
   label: GPU Passthrough to gpu-vm
 ---
 
-The AGX Orin's integrated GPU (`ga10b`, `gpu@17000000`) and its host1x
+The Orin AGX or NX integrated GPU (`ga10b`, `gpu@17000000`) and its host1x
 multimedia engines are passed through to a dedicated `gpu-vm` microVM, so CUDA
 compute runs in an isolated guest instead of on the host. It coexists with the
 MGBE0 ethernet passthrough (`net-vm`), sharing the same BPMP-virtualisation and
@@ -13,10 +13,11 @@ MGBE0 ethernet passthrough (`net-vm`), sharing the same BPMP-virtualisation and
 import { Aside } from "@astrojs/starlight/components";
 
 <Aside>
-  Verified on the NVIDIA Jetson AGX Orin devkit. The guest binds `gk20a`,
+  Verified on the NVIDIA Jetson AGX Orin devkit; NX hardware validation is
+  pending. AGX and NX select their GPU payload from a build-time board table,
+  and the GPU is enabled per SoM, never in the shared `orin.nix`. The guest binds `gk20a`,
   creates `/dev/nvgpu/igpu0`, and CUDA `cuInit` succeeds —
-  `__nvcc_device_query` returns `87` (sm_87 / ga10b). AGX Orin only: the GPU is
-  enabled per-SoM, never in the shared `orin.nix`.
+  `__nvcc_device_query` returns `87` (sm_87 / ga10b).
 </Aside>
 
 <Aside>
@@ -118,10 +119,12 @@ is handed to vfio.
 
 ## Use
 
-### Enable (per-SoM, AGX only)
+For cooperative CUDA SM job placement and managed containers, see
+[Managed GPU Partitions](/ghaf/dev/technologies/nvidia_gpu_partitioning/).
 
-Set in the AGX SoM files (`orin-agx.nix`, `orin-agx64.nix`,
-`orin-agx-industrial.nix`):
+### Enable per SoM
+
+Set in the AGX or NX SoM module:
 
 ```nix
 ghaf.hardware.nvidia.passthroughs.gpu_vm.enable = true;
@@ -149,7 +152,7 @@ nix build .#packages.aarch64-linux.nvidia-jetson-orin-agx-accelerated-guivm-debu
 
 ### Verify GPU compute
 
-From the host, reach the guest (`ssh ghaf@gpu-vm`, `192.168.100.3`):
+From the host, reach the guest by its configured hostname (`ssh ghaf@gpu-vm`):
 
 Run everything as the unprivileged `ghaf` user. Do **not** use `sudo`: the
 boot service group-grants the GPU nodes so no privilege is needed, and `sudo`
--- a/modules/microvm/sysvms/gpuvm-base.nix
+++ b/modules/microvm/sysvms/gpuvm-base.nix
@@ -30,6 +30,8 @@ in
     inputs.self.nixosModules.hardware-x86_64-guest-kernel
     inputs.self.nixosModules.vm-modules
     inputs.self.nixosModules.profiles
+    ./gpuvm-container-runtime.nix
+    ./gpuvm-partition-manager.nix
   ];
 
   ghaf = {
@@ -52,7 +54,11 @@ in
     # User configuration - from hostConfig
     users = {
       profile = hostConfig.users.profile or { };
-      admin = hostConfig.users.admin or { };
+      admin = (hostConfig.users.admin or { }) // {
+        # hostConfig is already evaluated, so preserve this copied default at
+        # default priority for the GPU-VM runtime's explicit policy override.
+        addToDockerGroup = lib.mkDefault (hostConfig.users.admin.addToDockerGroup or true);
+      };
       managed = hostConfig.users.managed or { };
     };
 
@@ -144,7 +150,7 @@ in
     ++ [
       pkgs.gcc
       # Prebuilt smoke test (driver API + embedded PTX, RPATH-wired to native
-      # libcuda), built at image time since the guest can't compile on-device.
+      # libcuda) so routine verification does not need an on-device build.
       (pkgs.callPackage ../../../packages/gpu-vm-load/package.nix {
         inherit (pkgs) nvidia-jetpack;
       })
@@ -163,10 +169,21 @@ in
       RemainAfterExit = true;
     };
     script = ''
+      for attempt in $(${pkgs.coreutils}/bin/seq 1 30); do
+        if [ -e /dev/nvgpu/igpu0/ctrl ]; then
+          break
+        fi
+        if [ "$attempt" -eq 30 ]; then
+          echo "GPU control node did not appear within 30 seconds" >&2
+          exit 1
+        fi
+        ${pkgs.coreutils}/bin/sleep 1
+      done
+
       for d in /dev/nvgpu /dev/nvhost-* /dev/nvmap; do
         [ -e "$d" ] || continue
-        chgrp -R video "$d" || true
-        chmod -R g+rw "$d" || true
+        chgrp -R video "$d"
+        chmod -R g+rw "$d"
       done
     '';
   };
--- a/modules/microvm/sysvms/gpuvm-container-runtime.nix
+++ b/modules/microvm/sysvms/gpuvm-container-runtime.nix
@@ -1,0 +1,182 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  inputs,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.ghaf.virtualization.gpuContainerRuntime;
+  partitionCfg = config.ghaf.virtualization.gpuPartitionManager;
+  partitionManager = inputs.gpu-partition-manager.lib.mkManager {
+    inherit pkgs;
+    cudaHeaders = pkgs.nvidia-jetpack.cudaPackages.cuda_cudart;
+    cudaDriver = pkgs.nvidia-jetpack.l4t-cuda;
+  };
+  partitionClient = pkgs.runCommand "gpu-partition-client" { } ''
+    install -Dm755 ${partitionManager}/bin/gpu-partition-run \
+      $out/bin/gpu-partition-run
+  '';
+  cudaClosureInfo = pkgs.closureInfo { rootPaths = [ pkgs.nvidia-jetpack.l4t-cuda ]; };
+  cudaLibPath = lib.makeLibraryPath [ pkgs.nvidia-jetpack.l4t-cuda ];
+  managedClosureInfo = pkgs.closureInfo { rootPaths = [ partitionClient ]; };
+  jqArgs = [
+    "--rawfile"
+    "paths"
+    "${cudaClosureInfo}/store-paths"
+  ]
+  ++ (
+    if partitionCfg.enable then
+      [
+        "--rawfile"
+        "managedPaths"
+        "${managedClosureInfo}/store-paths"
+        "--arg"
+        "client"
+        "${partitionClient}/bin/gpu-partition-run"
+      ]
+    else
+      [
+        "--arg"
+        "managedPaths"
+        ""
+        "--arg"
+        "client"
+        ""
+      ]
+  )
+  ++ [
+    "--arg"
+    "libPath"
+    cudaLibPath
+    "--arg"
+    "managed"
+    (lib.boolToString partitionCfg.enable)
+  ];
+
+  # The guest device layout is fixed by the passthrough DTS. Generate the CDI
+  # specification at build time so gpu-vm does not need a mutable toolkit or a
+  # privileged discovery helper at runtime.
+  nvidiaCdiSpec =
+    pkgs.runCommand "nvidia-cdi-spec" { nativeBuildInputs = [ pkgs.buildPackages.jq ]; }
+      ''
+        jq -n ${lib.escapeShellArgs jqArgs} \
+          '{
+            cdiVersion: "0.6.0",
+            kind: "nvidia.com/gpu",
+            devices: [
+              {
+                name: "all",
+                containerEdits: {
+                  # Hardware-enumerated set on JetPack r36. libcuda also needs
+                  # the DRM render node and host1x fence for initialization.
+                  deviceNodes: (
+                    [
+                      "/dev/nvhost-as-gpu",
+                      "/dev/nvhost-ctrl-gpu",
+                      "/dev/nvhost-ctxsw-gpu",
+                      "/dev/nvhost-dbg-gpu",
+                      "/dev/nvhost-gpu",
+                      "/dev/nvhost-nvsched-gpu",
+                      "/dev/nvhost-nvsched_ctrl_fifo-gpu",
+                      "/dev/nvhost-power-gpu",
+                      "/dev/nvhost-prof-ctx-gpu",
+                      "/dev/nvhost-prof-dev-gpu",
+                      "/dev/nvhost-prof-gpu",
+                      "/dev/nvhost-sched-gpu",
+                      "/dev/nvhost-tsg-gpu",
+                      "/dev/nvmap",
+                      "/dev/nvgpu/igpu0/as",
+                      "/dev/nvgpu/igpu0/channel",
+                      "/dev/nvgpu/igpu0/ctrl",
+                      "/dev/nvgpu/igpu0/ctxsw",
+                      "/dev/nvgpu/igpu0/dbg",
+                      "/dev/nvgpu/igpu0/nvsched",
+                      "/dev/nvgpu/igpu0/nvsched_ctrl_fifo",
+                      "/dev/nvgpu/igpu0/power",
+                      "/dev/nvgpu/igpu0/prof",
+                      "/dev/nvgpu/igpu0/prof-ctx",
+                      "/dev/nvgpu/igpu0/prof-dev",
+                      "/dev/nvgpu/igpu0/sched",
+                      "/dev/nvgpu/igpu0/tsg",
+                      "/dev/dri/renderD128",
+                      "/dev/host1x-fence"
+                    ] | map({ path: . })
+                  )
+                }
+              }
+            ] + (if $managed == "true" then [
+              {
+                name: "managed",
+                containerEdits: {
+                  env: [ "GPU_PARTITION_SOCKET=/run/gpu-partition-manager/control.sock" ],
+                  mounts: (
+                    ((($managedPaths | split("\n") | map(select(length > 0)))
+                      - ($paths | split("\n") | map(select(length > 0)))) | map({
+                      hostPath: .,
+                      containerPath: .,
+                      options: [ "ro", "bind", "nosuid", "nodev" ]
+                    })) + [
+                      {
+                        hostPath: $client,
+                        containerPath: "/opt/ghaf/bin/gpu-partition-run",
+                        options: [ "ro", "bind", "nosuid", "nodev" ]
+                      },
+                      {
+                        hostPath: "/run/gpu-partition-manager/control.sock",
+                        containerPath: "/run/gpu-partition-manager/control.sock",
+                        options: [ "bind", "nosuid", "nodev" ]
+                      }
+                    ]
+                  )
+                }
+              }
+            ] else [] end),
+            containerEdits: {
+              env: [ ("LD_LIBRARY_PATH=" + $libPath) ],
+              mounts: (
+                $paths | split("\n") | map(select(length > 0) | {
+                  hostPath: .,
+                  containerPath: .,
+                  options: [ "ro", "bind", "nosuid", "nodev" ]
+                })
+              )
+            }
+          }' > "$out"
+      '';
+in
+{
+  _file = ./gpuvm-container-runtime.nix;
+
+  options.ghaf.virtualization.gpuContainerRuntime = {
+    enable = lib.mkEnableOption "rootful Docker with NVIDIA CDI devices in gpu-vm";
+
+    addGhafUserToDockerGroup = lib.mkEnableOption ''
+      root-equivalent Docker access for the ghaf user in gpu-vm
+    '';
+  };
+
+  config = lib.mkIf cfg.enable {
+    ghaf.storagevm.directories = [
+      {
+        directory = "/var/lib/docker";
+        mode = "0710";
+      }
+    ];
+
+    ghaf.users.admin.addToDockerGroup = cfg.addGhafUserToDockerGroup;
+
+    virtualisation.docker = {
+      enable = true;
+      daemon.settings = {
+        features.cdi = true;
+        cdi-spec-dirs = [ "/etc/cdi" ];
+        no-new-privileges = true;
+      };
+    };
+
+    environment.etc."cdi/nvidia.json".source = nvidiaCdiSpec;
+  };
+}
--- a/modules/microvm/sysvms/gpuvm-partition-manager.nix
+++ b/modules/microvm/sysvms/gpuvm-partition-manager.nix
@@ -1,0 +1,146 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  inputs,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.ghaf.virtualization.gpuPartitionManager;
+  manager = inputs.gpu-partition-manager.lib.mkManager {
+    inherit pkgs;
+    cudaHeaders = pkgs.nvidia-jetpack.cudaPackages.cuda_cudart;
+    cudaDriver = pkgs.nvidia-jetpack.l4t-cuda;
+  };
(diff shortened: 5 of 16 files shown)


modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/bpmp-virt-common/sources/drivers/firmware/tegra/bpmp-host-proxy/bpmp-host-proxy.c
```diff
@@ -298,17 +298,18 @@ static ssize_t read(struct file *filep, char __user *buffer, size_t len, loff_t
...
  * commands on these ids are denied in check_if_allowed().
  */
 static const uint32_t protected_clk_roots[] = {
 	14,  /* TEGRA234_CLK_CLK_M */
 	102, /* TEGRA234_CLK_PLLP_OUT0 */
-	103, /* TEGRA234_CLK_UTMIP_PLL: parent of mgbe0_app, also feeds host USB */
+	103, /* TEGRA234_CLK_UTMIP_PLL: ancestor of mgbe0_app (via mgbes_app/utmipll_clkout480), also feeds host USB */
 	292, /* TEGRA234_CLK_UTMIPLL_CLKOUT480: 480 MHz output feeding mgbe0_app */
 	91,  /* TEGRA234_CLK_OSC */
 	288, /* TEGRA234_CLK_PLLREFE_VCOOUT: shared reference PLL (PCIe/UPHY) */
```
brianmcgillion, line 314: **Blocking.** This PR adds clock ids **131 (SOR_SAFE)** and **328 (PLLP_DIV17)** to the guest allow-list in `bpmp-allowlist.nix`, but neither is in `protected_clk_roots`, so `check_if_allowed()` accepts `CMD_CLK_DISABLE`, `CMD_CLK_SET_RATE` and `CMD_CLK_SET_PARENT` on both.

The allow-list is a single global union: `host.bpmp.allow.clocks` is `listOf int` with `apply = lib.unique`, described in-tree as "union across enabled passthroughs", feeding one `/dev/bpmp-host` and one `bpmp_ares`. There is no per-guest scoping. `capabilities.gpuvm.display = false`, yet gpu-vm contributes the entire shared list including the display block, and on AGX `orin-agx.nix:46-47` enables `gpu_vm` and `disp_vm` together.

So root in gpu-vm — now the VM that runs arbitrary Docker containers — can disable or reparent the clock feeding disp-vm's display output. BPMP's enable refcount is global, not per-caller. That crosses the boundary the split topology exists to create.

Your own comment says these are ancestors reached by `clk_prepare`, i.e. enable + read only. That is exactly what `protected_clk_roots` grants:

```suggestion
	288, /* TEGRA234_CLK_PLLREFE_VCOOUT: shared reference PLL (PCIe/UPHY) */
	131, /* TEGRA234_CLK_SOR_SAFE: dpaux(19) parent, owned by disp-vm */
	328, /* TEGRA234_CLK_PLLP_DIV17: fixed /17 tap off protected PLLP_OUT0 */
```

One thing I could not settle: whether `CMD_CLK_SET_RATE` on 328 propagates up to PLLP_OUT0, which is protected precisely because the host TCB (SDMMC/eMMC, UART, I2C, QSPI) hangs off it. BPMP firmware is a blob. If it does propagate, this reaches the host rather than only disp-vm — which would raise the severity further.

The three NAFLL ids (199/207/210) are fine as they are: NVDEC, VIC and NVJPG1 are assigned exclusively to gpu-vm, so the host never consumes those muxes.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/bpmp-virt-common/sources/drivers/firmware/tegra/bpmp-host-proxy/bpmp-host-proxy.c
```diff
@@ -298,17 +298,18 @@ static ssize_t read(struct file *filep, char __user *buffer, size_t len, loff_t
...
- * -- but MRQ_CLK is gated by id, not command, so a guest could otherwise
- * DISABLE/reparent/rerate them and destabilise the host. Enable (a BPMP
- * refcount no-op on always-on roots) and reads stay permitted; mutating
+ * Host-critical shared clock roots the host TCB depends on (always-on). Guest
+ * clock trees (display, ethernet, USB/UTMI, PLLREFE) parent up to these, so
+ * they must be allow-listed -- but MRQ_CLK is gated by id, not command, so a
+ * guest could otherwise DISABLE/reparent/rerate them and destabilise the
+ * host. Enable (a BPMP refcount no-op on always-on roots) and reads stay
+ * permitted; mutating
  * commands on these ids are denied in check_if_allowed().
```
brianmcgillion: Two small things from the rewrap. PLLREFE is not a guest clock tree — `PLLREFE_VCOOUT` (288) and `PLLREFE_VCOOUT_GATED` (327) are entries *in* this array a few lines below. And the reflow left `permitted; mutating` orphaned on its own line.

```suggestion
 * clock trees (display, ethernet, USB/UTMI, PCIe/UPHY) parent up to these, so
 * they must be allow-listed -- but MRQ_CLK is gated by id, not command, so a
 * guest could otherwise DISABLE/reparent/rerate them and destabilise the
 * host. Enable (a BPMP refcount no-op on always-on roots) and reads stay
 * permitted; mutating commands on these ids are denied in check_if_allowed().
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/payload/bpmp-allowlist.nix
```diff
@@ -19,12 +19,19 @@
     167
     236
     304
+    # Engine-dedicated NAFLL DVFS muxes reached by clk_prepare. They source
+    # from protected OSC/PLLP roots, so this adds no mutable host surface.
```
brianmcgillion: The stated rationale does not support the conclusion. Sourcing *from* protected roots says nothing about whether the guest can mutate 199/207/210 themselves — and it can, since none of the three is in `protected_clk_roots`. What actually makes this safe is the first half of your sentence: these engines belong to gpu-vm alone, so the host never consumes those muxes. Worth leading with that, because it is the load-bearing part.

```suggestion
    # Engine-dedicated NAFLL DVFS muxes reached by clk_prepare. NVDEC, VIC and
    # NVJPG1 are assigned exclusively to gpu-vm, so the host never consumes
    # these muxes; their parents are already protected roots.
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/payload/bpmp-allowlist.nix
```diff
@@ -19,12 +19,19 @@
...
+    207 # NAFLL_VIC
+    210 # NAFLL_NVJPG1
   ]
   # display@13800000 clocks
   ++ [
     # SOR root parents
     14 # TEGRA234_CLK_CLK_M
     102 # TEGRA234_CLK_PLLP_OUT0
+    131 # SOR_SAFE: dpaux(19) parent
+    328 # PLLP_DIV17: fixed /17 tap feeding SOR_SAFE
```
brianmcgillion: These two are the blocking item — see my comment on `bpmp-host-proxy.c`. Unlike the NAFLL trio above, SOR_SAFE and PLLP_DIV17 are display-tree clocks that disp-vm owns and gpu-vm does not (`capabilities.gpuvm.display = false`), and neither is in `protected_clk_roots`, so gpu-vm can disable or reparent them across the VM boundary.

Adding both to `protected_clk_roots` keeps the enable + read that `clk_prepare` needs while removing the mutation.

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/mgbe0-net-vm/default.nix
```diff
@@ -67,11 +67,12 @@ in
...
         # mgbe0_app (380) does not hang off the GBE PLLs at all: it is clocked
         # at 480 MHz from the USB/UTMI tree, so clk_prepare walks
-        # mgbe0_app -> utmipll_clkout480 -> utmip_pll -> osc/clk_m. Every one of
-        # these is shared with host USB, so bpmp-host-proxy.c also lists them in
-        # protected_clk_roots: net-vm may enable and read them, but
-        # disable/set_rate/set_parent stay denied, so a guest cannot pull the
-        # clock out from under the host's USB (keyboard, net-vm's own NIC).
+        # mgbe0_app -> mgbes_app -> utmipll_clkout480 -> utmip_pll -> osc/clk_m.
+        # Every one of these is shared with host USB, so bpmp-host-proxy.c
+        # also lists them in protected_clk_roots: net-vm may enable and read
```
brianmcgillion: This is contradicted by the code three lines above. `protected_clk_roots` is `{14, 102, 103, 292, 91, 288, 327}` — neither `mgbe0_app` (380) nor `mgbes_app` (366) is in it. And 366 is listed six lines up under a block whose own comment says "All of these are gigabit-ethernet dedicated", so the file now claims 366 is both ethernet-dedicated and shared with host USB.

The pre-PR text had the same flaw for `mgbe0_app`; adding `mgbes_app` to the chain widened it.

```suggestion
        # The last four are shared with host USB, so bpmp-host-proxy.c lists
        # *those* in protected_clk_roots: net-vm may enable and read
```

modules/microvm/sysvms/gpuvm-container-runtime.nix
```diff
@@ -0,0 +1,147 @@
...
+
+  # The guest device layout is fixed by the passthrough DTS. Generate the CDI
+  # specification at build time so gpu-vm does not need a mutable toolkit or a
+  # privileged discovery helper at runtime.
+  nvidiaCdiSpec =
+    pkgs.runCommand "nvidia-cdi-spec"
+      {
+        nativeBuildInputs = [ pkgs.buildPackages.jq ];
+        closureInfo = pkgs.closureInfo { rootPaths = [ pkgs.nvidia-jetpack.l4t-cuda ]; };
+        managedClosureInfo = pkgs.closureInfo { rootPaths = [ partitionManager ]; };
```
brianmcgillion: `managedClosureInfo` (and the `--arg client` on line 36) are unconditional derivation inputs of `nvidia-cdi-spec`; the `$managed` branch is only taken at jq runtime, long after Nix has built everything. I confirmed both halves of this:

- With only `containerRuntime.enable = true`, the dry-run schedules `ghaf-gpu-partition-manager-aarch64-...drv` as "will be built".
- But `nix-store -q --references` on the resulting spec shows no manager path, so the **runtime** closure is clean.

So the cost is purely a build coupling — a gpu-vm image that only wants Docker cannot be built if the external manager repo fails to compile. Worth decoupling:

```nix
pkgs.runCommand "nvidia-cdi-spec" ({
  nativeBuildInputs = [ pkgs.buildPackages.jq ];
  closureInfo = pkgs.closureInfo { rootPaths = [ pkgs.nvidia-jetpack.l4t-cuda ]; };
  libPath = lib.makeLibraryPath [ pkgs.nvidia-jetpack.l4t-cuda ];
} // lib.optionalAttrs partitionCfg.enable {
  managedClosureInfo = pkgs.closureInfo { rootPaths = [ partitionManager ]; };
})
```

...with the `--rawfile managedPaths` and `--arg client` lines behind `lib.optionalString partitionCfg.enable`.

modules/microvm/sysvms/gpuvm-container-runtime.nix
```diff
@@ -0,0 +1,147 @@
...
+                    })) + [
+                      {
+                        hostPath: $client,
+                        containerPath: "/opt/ghaf/bin/gpu-partition-run",
+                        options: [ "ro", "bind", "nosuid", "nodev" ]
+                      },
+                      {
+                        hostPath: "/run/gpu-partition-manager/control.sock",
+                        containerPath: "/run/gpu-partition-manager/control.sock",
+                        options: [ "bind" ]
```
brianmcgillion: Every other mount in this spec uses `["ro", "bind", "nosuid", "nodev"]`; this one is bare. The socket has to stay writable, but `nosuid`/`nodev` cost nothing and keep the file consistent.

```suggestion
                        options: [ "bind", "nosuid", "nodev" ]
```

Separately, worth knowing: I built the spec and the closure-delta mount above resolves to the **whole** manager store path, which contains `bin/gpu-partition-manager` (the daemon) and `include/ghaf` — not only `bin/gpu-partition-run`. Not an escalation, since the managed device injects no device nodes and CUDA cannot init without them, but the PR description says "only the manager client and socket", which is not what ships.

modules/microvm/sysvms/gpuvm-container-runtime.nix
```diff
@@ -0,0 +1,147 @@
...
+
+  config = lib.mkIf cfg.enable {
+    ghaf.storagevm.directories = [
+      {
+        directory = "/var/lib/docker";
+        mode = "0710";
+      }
+    ];
+
+    users.users.ghaf.extraGroups = [ "docker" ];
```
brianmcgillion: The `docker` group is root-equivalent by construction — `docker run -v /:/host --privileged` from the `ghaf` account is gpu-vm root in one line. That arrives here as a side effect of `containerRuntime.enable`, and it composes badly with the BPMP surface: anything reaching the `ghaf` account in gpu-vm inherits the ability to attack disp-vm's and net-vm's clocks over the shared proxy.

Suggest either gating it behind a separate debug-only option, or driving containers through a narrow wrapper for the fixed set of workload invocations. At minimum, state it in the docs' trust-boundary section — it is currently the largest unstated privilege change in the PR.

Worth adding `"no-new-privileges" = true;` to `daemon.settings` while you are here. (`userns-remap` is not a substitute: the remapped uid would fall outside `video`, and the CDI device nodes are DAC-gated on that group by `gpu-vm-node-access`.)

modules/microvm/sysvms/gpuvm-partition-manager.nix
```diff
@@ -0,0 +1,115 @@
...
+  ...
+}:
+let
+  cfg = config.ghaf.virtualization.gpuPartitionManager;
+  manager = inputs.gpu-partition-manager.lib.mkManager {
+    inherit pkgs;
+    cudaHeaders = pkgs.nvidia-jetpack.cudaPackages.cuda_cudart;
+    cudaDriver = pkgs.nvidia-jetpack.l4t-cuda;
+  };
+  probe = pkgs.callPackage ../../../packages/gpu-vm-green-context-probe/package.nix {
```
brianmcgillion, line 17: Nothing in the tree sets `partitionManager.enable` or `containerRuntime.enable`, and this `probe` sits behind `mkIf cfg.enable`, so it is never compiled by CI. `gpu-vm-load` escapes this only because it is unconditional in `gpuvm-base.nix:152-155` and therefore built by every gpu-vm image.

The same applies to the CDI generator, all five assertions below, and `gpu-vm-partition-manager-sdk` (`nix build .#gpu-vm-partition-manager-sdk` errors with "does not provide attribute" — overlay entries are not flake outputs). A bump of the new flake input breaks all of it with no CI signal.

Exposing the probe as a package output is not available to you — see my note on `package.nix`. The workable fix is one in-tree debug variant with both options enabled, which reaches the probe through the VM closure exactly the way `gpu-vm-load` is reached today.

modules/microvm/sysvms/gpuvm-partition-manager.nix
```diff
@@ -0,0 +1,115 @@
...
+      {
+        assertion = lib.all (plugin: plugin ? requiredPluginAbiVersion) cfg.plugins;
+        message = "every gpuPartitionManager plugin must expose requiredPluginAbiVersion";
+      }
+      {
+        assertion = lib.all (
+          plugin: (plugin.requiredPluginAbiVersion or null) == manager.pluginAbiVersion
+        ) cfg.plugins;
+        message = "every gpuPartitionManager plugin must use the manager plugin ABI";
+      }
```
brianmcgillion: The third assertion is fully subsumed by the fourth: `plugin.requiredPluginAbiVersion or null` yields `null` for a missing attribute, which is `!= manager.pluginAbiVersion`, so both fire together and the operator reads two messages for one fault.

```suggestion
      {
        assertion = lib.all (
          plugin: (plugin.requiredPluginAbiVersion or null) == manager.pluginAbiVersion
        ) cfg.plugins;
        message = "every gpuPartitionManager plugin must expose a requiredPluginAbiVersion matching the manager plugin ABI";
      }
```

Note the second assertion is **not** similarly redundant — with a single plugin missing `gpuPartitionPluginName`, `lib.unique [ null ]` has length 1, which equals `lib.length cfg.plugins`, so the uniqueness check passes. Keep that one.

Two gaps while you are in here: none of the messages names the offending plugin (`plugin.pname` would), and nothing asserts that `${plugin}/lib/gpu-partition-manager/plugin.so` actually exists — the option description promises it and `pluginArguments` assumes it, but a plugin installing elsewhere passes every assertion and dies at boot as a failed unit.

modules/microvm/sysvms/gpuvm-partition-manager.nix
```diff
@@ -0,0 +1,115 @@
...
+        Type = "simple";
+        User = "gpu-partition";
+        Group = "gpu-partition";
+        RuntimeDirectory = "gpu-partition-manager";
+        RuntimeDirectoryMode = "0750";
+        UMask = "0007";
+        ExecStart = lib.escapeShellArgs ([ "${manager}/bin/gpu-partition-manager" ] ++ pluginArguments);
+        Restart = "on-failure";
+        RestartSec = 2;
+        RestartPreventExitStatus = 78;
```
brianmcgillion, line 125: 78 is `EX_CONFIG`, and the upstream daemon comments it at the exit site — Ghaf's copy of the bare number does not. This is the most opaque line in the module.

```suggestion
        # 78 = EX_CONFIG: the daemon exits with this when the GPU geometry
        # cannot be split evenly. Never restart into an unsupported geometry.
        RestartPreventExitStatus = 78;
```

modules/microvm/sysvms/gpuvm-partition-manager.nix
```diff
@@ -0,0 +1,115 @@
...
+        NoNewPrivileges = true;
+        PrivateTmp = true;
+        ProtectHome = true;
+        ProtectSystem = "strict";
+        RestrictAddressFamilies = [ "AF_UNIX" ];
```
brianmcgillion, line 131: `RestrictAddressFamilies` only filters `socket(2)` — the process still sits in the host netns and can reach its abstract UNIX sockets. Since this daemon speaks AF_UNIX only, `PrivateNetwork = true` is free and strictly stronger, and the usual companions apply cleanly here:

```suggestion
        RestrictAddressFamilies = [ "AF_UNIX" ];
        # AF_UNIX-only: a private netns also removes the host's abstract sockets.
        PrivateNetwork = true;
        CapabilityBoundingSet = [ "" ];
        RestrictNamespaces = true;
        RestrictSUIDSGID = true;
        RestrictRealtime = true;
        RemoveIPC = true;
        ProtectKernelTunables = true;
        ProtectKernelModules = true;
        ProtectControlGroups = true;
        ProtectProc = "invisible";
        SystemCallArchitectures = "native";
```

Worth a boot test on hardware before you take it — I have not run this configuration. Two to avoid: `MemoryDenyWriteExecute` breaks CUDA's PTX JIT, and `ProcSubset = "pid"` hides `/proc/meminfo`, which libcuda reads.

Separately, the restart budget above is tight for a CUDA service: `RestartSec = 2` with `StartLimitBurst = 3` inside `StartLimitIntervalSec = 60` gives three attempts in about six seconds, then the unit stays dead. Its ordering dependency `gpu-vm-node-access.service` is a oneshot whose `chgrp`/`chmod` loop is `|| true` and skips absent nodes, so `After=` does not imply the GPU nodes are group-accessible yet.

packages/gpu-vm-green-context-probe/package.nix
```diff
@@ -0,0 +1,36 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
```
brianmcgillion, line 2: I initially flagged this package for living outside `packages/pkgs-by-name/` and being absent from `packages/own-pkgs-overlay.nix` — every other package in the repo is registered in that overlay, including the three that also sit outside `pkgs-by-name`. Then I checked, and the placement is actually **forced**: `nvidia-jetpack` is absent from x86_64 `pkgs` built with `overlays.default` (I verified empirically), because `overlays/flake-module.nix:14-21` deliberately keeps the jetpack overlays out of `default` and applies them only from `modules/reference/hardware/jetpack/default.nix`. `pkgsDirectory` auto-exposes every `pkgs-by-name` entry for every system, so a CUDA package placed there breaks x86_64 evaluation outright — which is presumably why not one of the 27 packages in there touches `nvidia-jetpack`.

So the location is right, but the reason is recorded nowhere while `AGENTS.md` tells contributors to use `pkgs-by-name`. The next person to tidy this up gets a broken eval. Two lines would prevent that:

```suggestion
# SPDX-License-Identifier: Apache-2.0
#
# Deliberately not in packages/pkgs-by-name/ or packages/own-pkgs-overlay.nix:
# both expose packages for every system, and nvidia-jetpack is absent from the
# flake's perSystem pkgs (overlays/flake-module.nix keeps the jetpack overlays
# out of overlays.default). A CUDA package there breaks x86_64 evaluation.
# Reached by relative-path callPackage instead, as packages/gpu-vm-load is.
```

A matching caveat in `AGENTS.md` would be worth more still.

modules/microvm/sysvms/gpuvm-base.nix
```diff
@@ -20,6 +20,11 @@
 let
   vmName = "gpu-vm";
   timezoneEnabled = lib.ghaf.features.isEnabledFor globalConfig "timezone" vmName;
+
+  # Minimal platform diagnostic: driver API plus an embedded PTX kernel.
```
brianmcgillion: This duplicates the comment already at the use site (`# Prebuilt smoke test (driver API + embedded PTX, RPATH-wired to native libcuda), built at image time since the guest can't compile on-device.`), and renames the thing from "smoke test" to "platform diagnostic" in the process. The use-site version is the richer of the two, and since `gpuVmLoad` has exactly one use site the `let` binding is not buying much either.

```suggestion
```

Unrelated and pre-existing, but now eight lines apart and directly contradictory: `:145-146` ships nvcc "so a GPU compute load can be built **on-device**" while `:153-154` says the prebuilt binary exists because the guest "**can't compile on-device**".

modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/gpu-vm/default.nix
```diff
@@ -33,10 +39,31 @@ in
...
+
+    partitionManager = {
+      enable = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Enable the cooperative CUDA Green Context job manager in gpu-vm.
+          A downstream configuration must also supply trusted workload plugins.
+        '';
+      };
```
brianmcgillion: `containerRuntime.enable` three lines up uses `mkEnableOption` while this one is hand-rolled, in the same block. There are 374 `mkEnableOption` uses under `modules/` and the `AGENTS.md` module template mandates it. (The pre-existing `gpu_vm.enable` above matches its siblings in `disp-vm`/`gui-vm`/`uarti-net-vm`, so leave that one — it is just the two new options that should agree with each other.)

```suggestion
      enable = lib.mkEnableOption "the cooperative CUDA Green Context job manager in gpu-vm; a downstream configuration must also supply trusted workload plugins";
```

While you are here: the assertion at `:80-81` can never fire. `supportedPartitionSoM` lists all four `somType` values the repo defines, and those four files are exactly the ones that set `gpu_vm.enable = true`. `somType` is a bare `types.str` whose own description advertises a different vocabulary ("NX|AGX32|AGX64|Nano"), so this list is the only place the real vocabulary is written down — and it has no comment saying so.

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
```diff
@@ -0,0 +1,423 @@
...
+
+The static Container Device Interface (CDI) specification exposes two device
+names under `nvidia.com/gpu`:
+
+| CDI device | GPU device nodes | Manager socket | Intended use |
+| --- | --- | --- | --- |
+| `nvidia.com/gpu=all` | Yes | No | Existing unrestricted CUDA applications |
+| `nvidia.com/gpu=managed` | No | Yes | Trusted jobs submitted to a managed slot |
+
+The unrestricted device preserves compatibility with PR #2095. It also means
```
brianmcgillion: This cites the PR it arrives in. Nobody reading the published docs in a year can act on that.

```suggestion
The unrestricted device preserves the pre-existing `gpu=all` behaviour. It also means
```

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
```diff
@@ -0,0 +1,423 @@
...
+git clone https://github.com/tiiuae/ghaf-gpu-partition-manager
+cd ghaf-gpu-partition-manager
+nix build -L .#checks.x86_64-linux.tests
+```
+
+The test substitutes a 16-SM CUDA backend and exercises plugin discovery,
+geometry readback, two simultaneous slots, status, cancellation, queue release,
+and unknown-plugin rejection through the real socket and client.
+
+Build the real aarch64 manager through the AGX and NX cross configurations.
```
brianmcgillion: These two commands build neither the manager nor the probe. I ran the full dry-run of `.#nvidia-jetson-orin-agx-debug-from-x86_64` and its 200-line output mentions neither — because `containerRuntime.enable` and `partitionManager.enable` default false and no in-tree target sets them.

Either point at a configuration that actually enables the feature (see my note on `gpuvm-partition-manager.nix:17`), or say plainly that a downstream config with both options enabled is required. As written this will send someone chasing a build that silently does nothing.

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_vm.mdx
```diff
@@ -13,10 +13,10 @@ MGBE0 ethernet passthrough (`net-vm`), sharing the same BPMP-virtualisation and
 import { Aside } from "@astrojs/starlight/components";
 
 <Aside>
-  Verified on the NVIDIA Jetson AGX Orin devkit. The guest binds `gk20a`,
+  Verified on the NVIDIA Jetson AGX Orin devkit. The AGX and NX split targets
+  use the same runtime-query-based GPU payload. The guest binds `gk20a`,
```
brianmcgillion: "Runtime-query-based" is not right for the payload — `payload/default.nix:139` is literally `boardFor = somType: if somType == "nx" then boards.nx else boards.agx;`, a build-time board table. The phrase "runtime query, not a board-name table" belongs to the partition manager (and is correct there, on the sibling page) but has been misapplied here.

This edit also asserts AGX/NX parity while the page it links to records NX hardware validation as pending, and it dropped "the GPU is enabled per-SoM, never in the shared `orin.nix`" — which is still true (`modules/profiles/orin.nix:383-385`) and was replaced with nothing.

```suggestion
  Verified on the NVIDIA Jetson AGX Orin devkit; NX is not yet hardware
  validated. AGX and NX select their GPU payload from a build-time board
  table, and the GPU is enabled per-SoM, never in the shared `orin.nix`.
  The guest binds `gk20a`,
```

brianmcgillion (changes requested):
Reviewed at `fd374b26` in a worktree, base `abd3f8bd`. One blocking item; everything else is fixable without redesign. Most inline comments carry a one-click suggestion.

### What I verified by building, not reading

| Check | Result |
| --- | --- |
| `nix fmt -- --fail-on-change` | pass, tree still clean |
| `nix develop --command reuse lint` | pass |
| `nix build --dry-run .#nvidia-jetson-orin-agx-debug-from-x86_64` | pass |
| CDI spec built with `containerRuntime.enable` | builds, JSON valid |
| CDI spec built with `partitionManager.enable` | builds, JSON valid |
| probe compiled cross-aarch64 `-Wall -Wextra -Werror` | clean |
| new flake input | public, REUSE-compliant, pinned rev real |

I also confirmed the central trust-boundary claim holds: I built the spec and `nvidia.com/gpu=managed` genuinely carries **zero** device nodes, and the top-level `containerEdits` injects none either. Good.

### Blocking

The BPMP allow-list widening (`bpmp-allowlist.nix`, commit `bd1023be`) grants gpu-vm mutating access to two display-tree clocks it does not own. Details and the fix are inline on `bpmp-host-proxy.c`.

**Please also split `bd1023be` into its own PR.** It widens the guest MRQ_CLK surface, and this PR's own security argument rests on the closed allow-list (the assertion you added at `gpu-vm/default.nix:76-78`). Buried in a 1000-line CDI PR it gets no scrutiny.

### Not covered inline

**Nothing in this PR is built by CI.** No target sets `containerRuntime.enable` or `partitionManager.enable`; the probe is not a flake output; `nix build .#gpu-vm-partition-manager-sdk` errors with "does not provide attribute". The full 200-line dry-run of the AGX target mentions neither manager nor probe. So both modules' `config` bodies, all five assertions, the CDI generator, the C program and the SDK re-export are never evaluated or compiled — a bump of the new input breaks all of it silently. I had to hand-write nix expressions to test any of it.

**`nvidia_gpu_vm.mdx:154`** still says gpu-vm is at `192.168.100.3`, while the new page says `.4`. The new one is right — `disp_vm` is enabled on both AGX and NX, so it takes `.3` and gpu-vm takes `.4`. `ssh ghaf@gpu-vm` on both pages would avoid the whole class of problem. (Outside the diff, so I could not anchor a suggestion.)

**Doc length.** The page is 423 lines and could be ~250. Roughly 120 of them re-document the manager repo verbatim on a page that twice says that repo owns those things: the 8-step startup list (`55-69`), the `gpm_plugin_v1` struct byte-for-byte from `plugin.h` (`205-215`), and the protocol constants from `protocol.h` (`301-323`). Every constant is correct today, which is exactly the risk — it is a pinned copy of a flake input with nothing to catch drift. A 5-line summary plus links to `docs/protocol.md` and `docs/plugin-api.md` would do. Separately, the "not a security boundary" caveat appears seven times (`24-29`, `84`, `221`, `365-366`, `367-368`, `408-409`, `419`) — the Aside at the top is the right place and the rest is padding; `365-368` sit under **Troubleshooting** and contain no troubleshooting step. And `399-419` embeds a dated test log ("11,137 validated iterations per slot") in permanent reference docs; no sibling page in `dev/technologies/` has a Validation Status section. That is PR-description content.

**Two doc gaps worth a line each:** `:173` states the probe is installed unconditionally, but it only exists when `partitionManager.enable` is set; and `:196-201` omits `requiredPluginAbiVersion` from the plugin contract, which your assertions enforce — a downstream author following the doc exactly gets an eval error with no clue why.

### Pre-existing, but they weaken this PR's premise

Not introduced here, and I am not asking you to fix them in this PR — but the allow-list widening is justified on the premise that mutation of shared roots is denied, and that premise is weaker than it looks:

- **`MRQ_DEBUG` is unconditionally allowed** (`bpmp-host-proxy.c:338-350`). It is the BPMP debugfs transport, addressed by pathname rather than clock id, and nothing inspects its payload — a name-addressed side door around the entire id-based allow-list. `MRQ_STRAP`, `MRQ_ISO_CLIENT` and `MRQ_BWMGR` are also unconditional; the latter two let any guest reserve isochronous bandwidth and starve the host.
- **The protected-root guard is a three-command deny-list.** Anything else the ABI defines passes, including `CMD_CLK_RESET_REFCOUNTS`. Inverting it to an allow-list of the read-only commands plus `CMD_CLK_ENABLE` would be future-proof.
- **Mask/comment mismatch** at `:378-386`: the comment says "bits[23..0] are the clock id" but the mask is `0x0FFF`. Not exploitable on Tegra234 (max id 476) and the command mask over-denies so it fails safe, but the forwarded message keeps bits 12-23 the check ignored.
- **The `all` CDI device ships the debug and profiling nodes** (`nvhost-dbg-gpu`, `nvhost-prof*`, `nvgpu/igpu0/{dbg,prof*}`), which expose register ops, FB read/write and `SUSPEND_RESUME_ALL_CONTEXTS` — a device-wide DoS against the very partitions this PR builds. In fairness this matches NVIDIA's own `devices.csv`, so it is upstream's default; but a third `nvidia.com/gpu=debug` device for debug images would fit Ghaf's threat model better.

### Credit where due

`cdi-spec-dirs = [ "/etc/cdi" ]` correctly drops Docker's writable `/var/run/cdi` default; the spec emits no CDI `hooks` and mounts libraries `ro,nosuid,nodev`; build-time generation means no mutable toolkit or privileged discovery helper in the guest; `runner.c` is clean under every extra warning flag I threw at it; `pkgs.pkgsBuildBuild.writeShellApplication` is the correct and complete cross fix; and the docs are unusually honest that Green Contexts are cooperative placement rather than a tenant boundary. The comments at `gpuvm-container-runtime.nix:19-21` and `runner.c:4-5` are models of the kind that cannot rot.

vadika:
Thanks for the detailed review. The branch has been updated and squashed into two signed-off commits.

Summary of changes:

- Removed the complete BPMP allowlist commit from this PR. The blocking clock-permission threads are now outdated, and any BPMP expansion will be handled separately.
- Decoupled Docker-only builds from the external partition manager.
- Restricted the managed CDI device to the client and hardened socket mount; it injects no GPU device nodes.
- Made root-equivalent Docker group membership explicitly opt-in and fixed the generic admin-user wiring that previously added it automatically.
- Added Docker `no-new-privileges`, manager service hardening, a larger restart budget, and GPU-node readiness checking.
- Improved plugin diagnostics, merged redundant ABI assertions, and added build-time validation of `plugin.so`.
- Documented why the JetPack-dependent probe uses a relative `callPackage`.
- Removed duplicated configuration and the unreachable SoM assertion.
- Shortened the partitioning documentation, linked to the manager-owned protocol/plugin documentation, corrected AGX/NX validation status, documented the ABI requirement, and replaced fixed VM addresses with `ssh ghaf@gpu-vm`.

No new CI, image, board, or flash target was added. The standard AGX/NX split targets remain unchanged. The CI-coverage observation therefore remains an explicit limitation; downstream AGX and NX integration configurations were evaluated against the current Ghaf checkout instead.

Validation completed:

- formatting and REUSE
- pre-commit checks
- documentation build
- full flake export evaluation
- focused manager, probe, and CDI builds
- managed CDI content and Docker-only closure inspection
- downstream AGX and NX derivation evaluation
- Docker-group default and explicit opt-in evaluation

The current PR history is:

- `feat(gpuvm): integrate CDI and managed GPU partitions`
- `fix(orin): build pre-flash helper for host`

The existing AGX hardware results remain documented in the PR description; the review-fix head was not reflashed. NX live hardware validation remains pending.

Please re-review when convenient.

modules/microvm/sysvms/gpuvm-container-runtime.nix
```diff
@@ -0,0 +1,182 @@
...
+                      options: [ "ro", "bind", "nosuid", "nodev" ]
+                    })) + [
+                      {
+                        hostPath: $client,
+                        containerPath: "/opt/ghaf/bin/gpu-partition-run",
+                        options: [ "ro", "bind", "nosuid", "nodev" ]
+                      },
+                      {
+                        hostPath: "/run/gpu-partition-manager/control.sock",
+                        containerPath: "/run/gpu-partition-manager/control.sock",
```
brianmcgillion, line 129: **Follow-up, not blocking.** This bind-mounts the socket *file*, and `RuntimeDirectory` on the manager unit has no `RuntimeDirectoryPreserve`, so systemd's default (`no`) deletes `/run/gpu-partition-manager` and this inode on every stop.

A bind mount pins the inode, not the path. With `Restart = "on-failure"` and `StartLimitBurst = 12`, restarts are part of the design — and after one, every already-running managed container is holding an unlinked socket whose listener is gone. `connect()` returns `ECONNREFUSED` permanently while the new manager listens on a fresh inode at the same host path. Only `docker rm` + re-`run` recovers, which contradicts the troubleshooting section's advice to check service readiness.

Mounting the directory instead survives the swap:

```suggestion
                        hostPath: "/run/gpu-partition-manager",
                        containerPath: "/run/gpu-partition-manager",
```

Pair it with `RuntimeDirectoryPreserve = "yes"` on the unit. Separately worth a docs line: with the file mount, `docker run --device nvidia.com/gpu=managed` also fails outright with an opaque runc "bind source path does not exist" whenever the manager is not running.

modules/microvm/sysvms/gpuvm-partition-manager.nix
```diff
@@ -0,0 +1,146 @@
...
+        Group = "gpu-partition";
+        RuntimeDirectory = "gpu-partition-manager";
+        RuntimeDirectoryMode = "0750";
+        UMask = "0007";
+        ExecStart = lib.escapeShellArgs ([ "${manager}/bin/gpu-partition-manager" ] ++ pluginArguments);
+        Restart = "on-failure";
+        RestartSec = 10;
+        # 78 = EX_CONFIG: the daemon exits with this when the GPU geometry
+        # cannot be split evenly. Never restart into an unsupported geometry.
+        RestartPreventExitStatus = 78;
```
brianmcgillion, line 125: **Follow-up, not blocking.** The comment is narrower than the code it explains, and the gap has teeth.

`manager.c:803-805` is the only site that yields 78, and it covers *every* `setup_partition()` failure — `cuInit`, `cuDeviceGetCount` returning zero devices ("no CUDA device found"), `cuDevicePrimaryCtxRetain`, `cuDeviceGetDevResource`, `cuGreenCtxCreate`, and the Green Context readback mismatch. Geometry is one case out of six, and the other five are transient device-availability conditions.

So a GPU that is simply not ready yet exits 78, and `RestartPreventExitStatus = 78` makes that **permanent** — the service stays dead until someone runs `systemctl reset-failed`, while the troubleshooting section sends the operator off to inspect SM counts. This is the failure mode most likely to appear first on NX, where hardware validation is still pending.

```suggestion
        # 78 = EX_CONFIG: any partition setup failure -- CUDA init, no visible
        # device, Green Context creation, or an unsplittable geometry. Suppresses
        # restart for all of them, so a GPU that is merely late stays down.
        RestartPreventExitStatus = 78;
```

Worth considering whether the transient cases should instead exit non-78 upstream, so only a genuinely unsupported geometry is unrecoverable.

modules/microvm/sysvms/gpuvm-partition-manager.nix
```diff
@@ -0,0 +1,146 @@
...
+
+    environment.systemPackages = [
+      manager
+      probe
+    ];
+
+    systemd.services.gpu-partition-manager = {
+      description = "Managed CUDA Green Context jobs";
+      wantedBy = [ "multi-user.target" ];
+      wants = [ "gpu-vm-node-access.service" ];
```
brianmcgillion, line 107: Minor, and related to the exit-78 note above. `gpu-vm-node-access.service` is now fatal — the `|| true` is gone and it is bounded at 30 s — but it is a `oneshot` with no `Restart=`, and this unit only `Wants=` it.

If the nodes appear at 31 s, node-access stays failed forever, the manager starts anyway against root-only device nodes, exits 78 on `cuInit`, and never restarts. `Requires=` would at least fail closed instead of failing confusingly.

modules/microvm/sysvms/gpuvm-partition-manager.nix
```diff
@@ -0,0 +1,146 @@
...
+  );
+  validatedPlugins = pkgs.runCommand "gpu-partition-manager-plugins" { } ''
+    mkdir -p $out/lib/gpu-partition-manager
+    ${lib.concatMapStringsSep "\n" (plugin: ''
+      if [ ! -f ${plugin}/lib/gpu-partition-manager/plugin.so ]; then
+        echo "${pluginLabel plugin}: missing lib/gpu-partition-manager/plugin.so" >&2
+        exit 1
+      fi
+      ln -s ${plugin}/lib/gpu-partition-manager/plugin.so \
+        $out/lib/gpu-partition-manager/${plugin.gpuPartitionPluginName or "invalid"}.so
```
brianmcgillion, line 39: Nit: this fallback is unreachable — the `lib.all (plugin: plugin ? gpuPartitionPluginName)` assertion is evaluated as part of `system.build.toplevel` before this derivation can be instantiated.

If it ever did fire, two nameless plugins would collide on `invalid.so` and the build would die with a bare `ln: File exists`. Using `pluginLabel plugin` (already defined above) would at least name the culprit, or drop the fallback entirely now that the assertion guarantees it.

modules/microvm/sysvms/gpuvm-container-runtime.nix
```diff
@@ -0,0 +1,182 @@
...
+      }
+    ];
+
+    ghaf.users.admin.addToDockerGroup = cfg.addGhafUserToDockerGroup;
+
+    virtualisation.docker = {
+      enable = true;
+      daemon.settings = {
+        features.cdi = true;
+        cdi-spec-dirs = [ "/etc/cdi" ];
```
brianmcgillion, line 175: This line is load-bearing, but the docs describe it as if it were redundant — see my comment on the docs page. Flagging it here so the next person to touch it knows: dockerd's compiled-in default for `cdi-spec-dirs` is `/etc/cdi` **and** `/var/run/cdi` (both literals are present in the 29.6.2 binary). Pinning it to `/etc/cdi` alone is what actually keeps the CDI spec immutable in the VM that owns the GPU. Deleting this line would silently re-enable a runtime-writable spec directory.

modules/microvm/sysvms/gpuvm-container-runtime.nix
```diff
@@ -0,0 +1,182 @@
...
+                      "/dev/host1x-fence"
+                    ] | map({ path: . })
+                  )
+                }
+              }
+            ] + (if $managed == "true" then [
+              {
+                name: "managed",
+                containerEdits: {
+                  env: [ "GPU_PARTITION_SOCKET=/run/gpu-partition-manager/control.sock" ],
```
brianmcgillion, line 114: Minor: this environment variable is inert. `src/client.c` in the locked manager input contains no `getenv` — the socket path comes from the compiled `GPM_SOCKET_PATH` in `include/ghaf/gpu-partition-manager/protocol.h` or from `--socket`.

Harmless, but it advertises a configurability that does not exist. Either have the client honour it or drop the edit.

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
```diff
@@ -0,0 +1,289 @@
...
+| --- | --- | --- | --- |
+| `nvidia.com/gpu=all` | Complete JetPack set | No | Direct, unrestricted CUDA |
+| `nvidia.com/gpu=managed` | None | Yes | Jobs submitted through the manager |
+
+The unrestricted device preserves the existing `gpu=all` behavior. It also
+lets a process bypass the manager. The managed device injects the client and
+socket but no `/dev/nvgpu`, `/dev/nvhost-*`, `/dev/nvmap`, DRM render, or
+host1x-fence nodes.
+
+The CDI spec is immutable and generated at build time. Docker reads CDI specs
```
brianmcgillion, line 76: Both halves of this are inaccurate, and the first one matters because it makes a security-relevant line look optional.

Docker's default `cdi-spec-dirs` is `/etc/cdi` **and** `/var/run/cdi` — I confirmed both string literals in the dockerd 29.6.2 binary. The single-directory behaviour exists only because `gpuvm-container-runtime.nix:175` overrides it. And "Ghaf does not enable Docker's writable `/var/run/cdi`" is false repo-wide: `modules/common/virtualization/nvidia-docker.nix:52` sets exactly that. It happens to be `mkForce false`'d under the gpu_vm topology, but the sentence is unqualified.

```suggestion
The CDI spec is immutable and generated at build time. Ghaf pins Docker's
`cdi-spec-dirs` to `/etc/cdi` alone; Docker's own default would additionally
scan the runtime-writable `/var/run/cdi`. Ghaf installs no privileged runtime
discovery helper.
```

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
```diff
@@ -0,0 +1,289 @@
...
+                  gpu-partition-manager
+                    /                   \
+              slot 0 worker         slot 1 worker
+              Green Context         Green Context
+              half of SMs           half of SMs
+                    \                   /
+                     passed-through GA10B
+```
+
+At startup, the daemon queries CUDA device 0, requires two equal SM groups with
```
brianmcgillion, line 52: The startup order is the reverse of the code, and the difference changes what an operator should expect.

In `src/manager.c` `main()`, plugins load first during argv parsing (`:783-793`), then `setup_partition` (`:802`), `start_workers` (`:807`), `create_server` (`:809`). Plugin loading precedes any CUDA call.

That matters because a bad plugin fails before CUDA is touched and exits `EXIT_FAILURE` (1), **not** 78 — so unlike a geometry failure it *is* restarted, up to `StartLimitBurst = 12`. As written, the page implies the opposite.

```suggestion
At startup, the daemon loads only the configured immutable plugins, then queries
CUDA device 0, requires two equal SM groups with
```

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
```diff
@@ -0,0 +1,289 @@
...
+gpu-partition-run status
+```
+
+The service runs as the dedicated `gpu-partition` user with `video` group
+access. Its systemd sandbox uses a private network namespace, permits only
+`AF_UNIX`, removes capabilities, and restricts namespaces, kernel controls,
+IPC, and writable filesystem access. Do not enable
+`MemoryDenyWriteExecute`; CUDA's PTX JIT needs executable mappings.
+
+- Exit status 78 means the runtime geometry is unsupported. Inspect the total
```
brianmcgillion, line 242: Same scope issue as the code comment: 78 covers any `setup_partition()` failure, including `cuInit`, a device that is not visible yet, and Green Context creation — not only geometry. An operator hitting a late-appearing GPU node will be sent to inspect SM counts for a problem that has nothing to do with SM counts.

```suggestion
- Exit status 78 means partition setup failed: CUDA init, no visible device,
  Green Context creation, or an unsupported geometry. Systemd will not restart
  it, so check `dmesg` for `nvgpu` and the SM count, split, and remainder.
```

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
```diff
@@ -0,0 +1,289 @@
...
+  SM count, split, remainder, and Green Context readback.
+- For repeated failures, find the first CUDA or plugin error and inspect the
+  guest kernel log for `nvgpu` faults.
+- If a managed container cannot connect, check service readiness, socket group
+  access, and the managed entry in `/etc/cdi/nvidia.json`.
+- If a managed container contains GPU device nodes, stop the test and inspect
+  the CDI spec; the managed entry must contain mounts only.
```
brianmcgillion, line 249: The generated managed entry contains `env` as well as `mounts` (`GPU_PARTITION_SOCKET=…`), so an operator following this literally would flag a correct spec as broken. The invariant you want is the absence of device nodes.

```suggestion
  the CDI spec; the managed entry must contain no deviceNodes.
```

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_partitioning.mdx
```diff
@@ -0,0 +1,289 @@
...
+```bash
+systemctl status gpu-partition-manager
+journalctl -u gpu-partition-manager -b
+gpu-partition-run status
+```
+
+The service runs as the dedicated `gpu-partition` user with `video` group
+access. Its systemd sandbox uses a private network namespace, permits only
+`AF_UNIX`, removes capabilities, and restricts namespaces, kernel controls,
+IPC, and writable filesystem access. Do not enable
```
brianmcgillion, line 239: Nit: `RemoveIPC = true` removes IPC objects owned by the user when its last process exits — that is cleanup, not restriction, and `PrivateIPC` is not set. Every other clause in this sentence checks out against the unit; this one word does not.

Worth keeping the `MemoryDenyWriteExecute` warning exactly as-is, though — that one is correct and easy for someone to "fix" into a broken state.

docs/src/content/docs/ghaf/dev/technologies/nvidia_gpu_vm.mdx
```diff
@@ -13,10 +13,11 @@ MGBE0 ethernet passthrough (`net-vm`), sharing the same BPMP-virtualisation and
 import { Aside } from "@astrojs/starlight/components";
 
 <Aside>
-  Verified on the NVIDIA Jetson AGX Orin devkit. The guest binds `gk20a`,
+  Verified on the NVIDIA Jetson AGX Orin devkit; NX hardware validation is
+  pending. AGX and NX select their GPU payload from a build-time board table,
```
brianmcgillion, line 17: The board table selects the display-config-block blob, not the GPU payload. `payload/default.nix:127` says so directly ("The DCB image is the only board-specific piece; the rest is SoC-level"), and `boards` has `nx = agxBoard` — NX is a literal alias of AGX. `boardFor` reaches only `dtb.nix`, which consumes `dcbDtsi`/`dcbSha256`/`dcbBytes` and nothing else. The actual GPU payload comes from `capabilities.gpuvm` + `mkPayload` and is identical for both.

As written this implies board-differentiated GPU passthrough that does not exist.

```suggestion
  pending. AGX and NX share one SoC-level GPU payload, differing only in the
  display-config-block blob selected by a build-time board table,
```

The rest of the Aside is accurate — I checked the per-SoM enable claim against all four SoM files and `modules/profiles/orin.nix:383-385`.

brianmcgillion (approved):
Re-reviewed at `52de9754`, base `69b64b37`, in a fresh worktree. **Approving** — all 18 comments from the previous round are addressed, and the blocking item is gone because the whole BPMP allowlist commit was dropped. `bpmp-allowlist.nix` is byte-identical to base: no 131/328, no 199/207/210, so the cross-VM clock DoS on disp-vm is out of scope here. Thanks for splitting that out.

The inline comments below are **all follow-up material** — none of them need to block this merge. Two are real runtime defects worth a prompt follow-up PR; the rest are documentation corrections and nits.

### What I verified by building, not reading

| Check | Result |
| --- | --- |
| `nix fmt -- --fail-on-change` | pass, 0 changed |
| `nix develop --command reuse lint` | pass, 949/949 files compliant |
| NX debug dry-run vs base | **+48 store paths** — manager, probe, mock plugin, plugin farm, CDI spec and Docker are all genuinely built now |
| manager / probe / mock plugin / plugin farm | all build green (already substitutable from `ghaf-dev.cachix.org`, so CI has built them) |
| ExecStart wiring | `…/gpu-partition-manager --plugin …/gpu-partition-manager-plugins/lib/gpu-partition-manager/mock.so`, symlink resolves to the mock `plugin.so` |
| Docker-only decoupling | `nix-store -qR` on the CDI spec drv with `partitionManager.enable = false` → **zero** manager references, transitively. Laziness holds. |
| Managed CDI device | **0 device nodes**, and the top-level `containerEdits` has none either, so nothing leaks in through the shared edit |
| `gpu-partition-client` closure | contains only `bin/gpu-partition-run`; sole reference is glibc. The whole-manager leak from last round is gone. |
| Docker group | gpu-vm `ghaf` ends up in `video,gpu-partition,wheel` — **no `docker`** |
| AGX debug unaffected | `docker=false`, no manager, groups `video,wheel` |
| `pkgsBuildBuild` flash fix | correct — every `runtimeInputs` entry was already `pkgsBuildBuild`; the wrapper was the last aarch64 shebang on a script that runs on the x86 flashing host |

The CI-coverage fix is real rather than cosmetic: `orin-nx.nix:45` sets `gpu_vm.enable = true`, so the `mkIf` fires and the module is not silently inert. That was my main worry going in, and it checks out.

### Worth a decision, not a change request

`nvidia-jetson-orin-nx` **debug** is a CI-built, flashed, robot-tested image, and it now unconditionally gains ~926 MiB of Docker closure plus a rootful `dockerd` and a CUDA daemon inside a 2 GiB gpu-vm — under a comment recording that an earlier layout OOM-killed a VM and hung PID 1. You note NX hardware validation is still pending and this head was never reflashed. Nothing evaluates wrong, and the coverage is genuinely valuable, but a separate `nvidia-jetson-orin-nx-gpu-partitioning-debug` attribute would buy the same CI signal without changing a shipped image. Your call.

One process note: the 18 threads were resolved without replies, so I verified each fix against the code rather than against your reasoning. If I have misread an intent anywhere, say so and I will re-check.

PR #2098, opened 2026-08-04
fix(orin): cut boot-to-login time by half (initrd resize hunt + clock barrier)

vadika:
## Description of Changes

Two serialized boot-time eaters, found while profiling why an Orin AGX takes minutes to reach the greeter, measured over timestamped serial + `systemd-analyze` on host and gui-vm:

1. **`resize-partitions` burned 31.7 s in the initrd on every boot.** It hunted for the root partition by the pinned LUKS header UUID before consulting its ESP done-marker; on unencrypted images that UUID never matches, so every boot span the full 30×1 s retry loop and then skipped — and the resize itself never ran at all. Now the done-marker is checked first, and unencrypted images resolve root by the `NIXOS_SD` label: first boot actually grows the filesystem (~11 s one-time), later boots skip in under a second.
2. **`ghaf-clock-ready` held every machine's boot for a 20 s clock-stability window.** The barrier gates `systemd-journal-flush` and through it `sysinit.target`. It exists for hosts whose RTC must settle after power-on — but a VM's clock is host-fed (kvm-clock/ptp_kvm) and synchronized from the first instruction, so `clockReady.enable` now defaults to hosts only (keyed on `ghaf.type`, the pattern `fss.nix` already uses). Tegra hosts keep the barrier with a 5 s window (`mkDefault`, overridable).

Measured on the AGX accelerated gui-vm image (PR #2068 head): warm reboot → greeter 171 s → 108 s; power-on → greeter ~150 s → ~80 s; guest boot → greetd 39.9 s → 21.1 s. Both fixes are also in effect on the PR #2068 bench image and survive reboot/replug.

### Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Update
- [x] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

Split out of #2068 (the fixes are target-independent).

## Checklist

### Applicable Targets
- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`
### Installation Method
- [ ] Other:
(the resize fix lives in the initrd; the clock change alone is switch-able)
### Test Steps To Verify:
1. Flash an unencrypted debug image; first boot: journal shows `Resizing filesystem on /dev/disk/by-label/NIXOS_SD` (~11 s) and `df /` reports the full disk.
2. Reboot; initrd logs `Resize already performed, skipping.` within ~1 s of the marker check and `systemd-analyze` shows initrd well under 10 s.
3. In any VM: `systemctl status ghaf-clock-ready` → unit not found; on the host: journal shows "Clock readiness established after 5 stable seconds".
4. `nix eval` of `nvidia-jetson-orin-agx-debug`, `nvidia-jetson-orin-nx-debug`, `intel-laptop-debug` toplevels — verified as part of this PR.

Validation not run: encrypted-image boot path (the LUKS branch of the resize script is unchanged); x86 laptop hardware boot.

### DIFF
--- a/modules/common/logging/common.nix
+++ b/modules/common/logging/common.nix
@@ -518,7 +518,16 @@ in
 
       clockReady = {
         enable = (mkEnableOption "clock readiness barrier for persistent sealed logging") // {
-          default = true;
+          # The barrier exists for hosts whose RTC needs to settle after
+          # power-on. A VM starts with its emulated RTC seeded from host wall
+          # time at VM creation (microvm sets rtc = "on") and runs timesyncd
+          # afterwards, so it boots with an already-settled clock and the
+          # barrier (plus the ghaf-clock-sync NTP wait it gates) only delays
+          # boot there. Disabling it in guests is a deliberate choice: guest
+          # FSS setup/verify no longer waits for NTP, which also uncouples it
+          # from net-vm availability; clock-jump recovery stays active.
+          default = config.ghaf.type == "host";
+          defaultText = lib.literalExpression ''config.ghaf.type == "host"'';
         };
 
         stableSeconds = mkOption {
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
@@ -757,6 +757,12 @@ in
   };
 
   config = mkIf cfg.enable {
+    # The clock-readiness barrier gates systemd-journal-flush and with it the
+    # whole userspace boot. Tegra restores a sane RTC well within a few seconds,
+    # and post-release jumps are already handled by ghaf-clock-jump-watcher, so
+    # the default 20 s stability window only adds boot latency here.
+    ghaf.logging.recovery.clockReady.stableSeconds = lib.mkDefault 5;
+
     assertions = [
       {
         assertion =
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
@@ -293,41 +293,14 @@ let
       ESP_MOUNT="/mnt-esp"
       ESP_DEVICE="/dev/disk/by-label/${espLabel}"
 
-      # Resolve the root partition by the pinned LUKS header UUID rather than a
-      # hardcoded device name or a partition-table PARTUUID (which differs between
-      # the USB sd-image (MBR) and the eMMC/NVMe flash (GPT)). blkid matches the
-      # crypto_LUKS container by its own UUID, identical on both media.
-      PART_DEV=""
-      for _ in $(seq 1 30); do
-        for d in $(blkid -o device -t UUID=${luksUuid} 2>/dev/null); do
-          ${lib.optionalString cfg.diskEncryption.enable ''
-            [ "$(blkid -o value -s TYPE "$d" 2>/dev/null)" = "crypto_LUKS" ] || continue
-          ''}
-          PART_DEV="$d"
-          break
-        done
-        if [ -n "$PART_DEV" ]; then break; fi
-        sleep 1
-      done
-
-      if [ -z "$PART_DEV" ]; then
-        echo "Root partition (LUKS UUID=${luksUuid}) not found, skipping resize."
-        exit 0
-      fi
-
-      # Derive disk and partition number from the resolved node, so mmcblk/nvme/sda all work.
-      real_dev=$(readlink -f "$PART_DEV")
-      base_dev=$(basename "$real_dev")
-      PART_NUM=$(cat "/sys/class/block/$base_dev/partition")
-      DISK="/dev/$(basename "$(readlink -f "/sys/class/block/$base_dev/..")")"
-
-      RESIZE_TARGET="$PART_DEV"
-      ${lib.optionalString cfg.diskEncryption.enable ''
-        MAPPER_NAME="${cfg.diskEncryption.mapperName}"
-        RESIZE_TARGET="/dev/mapper/$MAPPER_NAME"
-      ''}
-
+      # Marker check comes FIRST: it needs no root/LUKS resolution, and doing it
+      # after the device hunt made every post-resize boot pay the full 30 s
+      # not-found timeout below before discovering there was nothing to do.
       mkdir -p "$ESP_MOUNT"
+      # The marker check no longer runs after the (slow) device hunt, so it can
+      # now race udev publishing /dev/disk/by-label/*. Settling first keeps a
+      # missing symlink from silently redoing the whole resize on every boot.
+      udevadm settle || true
       if [[ -b $ESP_DEVICE ]]; then
         if mount "$ESP_DEVICE" "$ESP_MOUNT"; then
           if [ -f "$ESP_MOUNT/$RESIZE_MARKER" ]; then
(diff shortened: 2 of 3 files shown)


brianmcgillion (commented):
Great change — both fixes are real and the measurements hold up. I went through it by evaluating the affected targets rather than only reading the diff, so everything below is backed by an eval. Only minor things; three of them are applyable suggestions inline.

## Verified as correct

- Host behaviour is exactly as intended: Orin host `enable=true, stableSeconds=5`, laptop host `true, 20`, all nine laptop VMs `false`.
- No dangling `ghaf-clock-ready` references are left in any guest unit — `fss.nix` gates on the identical expression, so nothing ends up with a `Requires=` on a unit that no longer exists.
- The eMMC layout is safe for the newly-live resize: `partition-template.nix` orders MBR → primary_gpt → esp → APP → secondary_gpt, so APP is the last data partition and `parted resizepart 100%` has room to grow.

## One thing worth an explicit decision: `ghaf-clock-sync` disappears from every guest

`ghaf-clock-sync.service` is `mkIf clockReadyEnabled` (`common.nix:601`) — it is the NTP wait ordered before `journal-fss-setup.service` and `journal-fss-verify.service`. Turning `clockReady` off in guests removes that too; I confirmed by eval that gui-vm no longer has the unit, while `fss.enable` stays `true` in all nine VMs. So guests now set up and verify sealing keys without waiting for time synchronization, while the host keeps that boundary.

Given the failure mode we have hit before — `journal-fss-verify` wedged after a clock step, with recovery rotating away pre-jump journals — it would be good to have this as a stated choice rather than a side effect of the enable flag. If the goal is only "do not burn 20 s on the stability window in a guest", the narrower knobs are `stableSeconds` in guests, or decoupling `ghaf-clock-sync` from `clockReadyEnabled`.

## Two follow-ups outside the diff

**1. Verity targets still pay the full 30 s.** `nvidia-jetson-orin-agx-verity-debug` carries `resize-partitions` in its initrd, has no `sdImage` (so `espLabel` falls back to `FIRMWARE` while its ESP is labelled `ESP`, meaning the marker check can never match), and its root is a tmpfs overlay. So the label hunt burns the full 30 s and then skips — on every boot, forever. Verity does its own resize in `firstboot-persist.nix`. Gating the unit on `!verityEnabled` would be another 30 s off exactly the class of target this PR is about.

**2. On unencrypted images the resize was already happening, just later.** `sdimage.nix:81` sets `expandOnBoot = !diskEncryption.enable`, and I confirmed `expandOnBoot = true` with `expand-root-partition.service` present on `nvidia-jetson-orin-agx-debug` — upstream's stage-2 unit was growing root all along. So "the resize itself never ran at all, leaving the root filesystem at its flashed size" is not quite right for unencrypted images; the 30 s initrd saving is entirely real, but both paths now run on first boot. The duplicate is harmless — I tested that `sfdisk -N1 ",+,"` against an already-100% partition exits 0 — so this is cleanup rather than a bug: `expandOnBoot = false` unconditionally, and the comment above it now describes only the LUKS case.

## Process notes

- `jenkins-pre-merge` is red (build 2527) and the console is behind auth, so I could not see why. All GitHub Actions jobs pass, including both aarch64 Orin builds, so it is not an eval or build break in the default set.
- The CI build matrix covers neither `-luks` nor `-verity` variants — which are precisely the two paths the resize change branches on.
- The resize is now live in the initrd on every unencrypted Orin target, validated on one AGX devkit eMMC; NX / NVMe / SD are unverified.

All three inline suggestions were applied locally and checked: `nvidia-jetson-orin-agx-debug` and `nvidia-jetson-orin-agx-verity-debug` toplevels still evaluate, `defaultText` renders, the rendered resize script is shellcheck-clean, and `nix fmt --fail-on-change` reports 0 changed.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -293,25 +293,61 @@ let
...
 
-      # Resolve the root partition by the pinned LUKS header UUID rather than a
-      # hardcoded device name or a partition-table PARTUUID (which differs between
-      # the USB sd-image (MBR) and the eMMC/NVMe flash (GPT)). blkid matches the
-      # crypto_LUKS container by its own UUID, identical on both media.
+      # Marker check comes FIRST: it needs no root/LUKS resolution, and doing it
+      # after the device hunt made every post-resize boot pay the full 30 s
+      # not-found timeout below before discovering there was nothing to do.
+      mkdir -p "$ESP_MOUNT"
+      if [[ -b $ESP_DEVICE ]]; then
```
brianmcgillion, line 304: Minor: the marker check lost its settle time. Before this change the 30-iteration device hunt ran first, which incidentally gave udev up to 30 s to publish `/dev/disk/by-label/*`. Now, if the ESP symlink is not there at t=0, the script logs *"continuing without marker check"* and re-runs the whole resize on an already-resized disk — and on the passphrase-LUKS variant that includes a 60 s `systemd-ask-password` prompt in the initrd.

Narrow, since the ESP shares the disk with root and the unit is `After=initrd-root-device.target` — but cheap to close, and `udevadm settle || true` is already the pattern used further down in this same script (and adds no fixed delay on images that have no ESP at all).

```suggestion
      mkdir -p "$ESP_MOUNT"
      # The marker check no longer runs after the (slow) device hunt, so it can
      # now race udev publishing /dev/disk/by-label/*. Settling first keeps a
      # missing symlink from silently redoing the whole resize on every boot.
      udevadm settle || true
      if [[ -b $ESP_DEVICE ]]; then
```

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -293,25 +293,61 @@ let
...
+            # labeled NIXOS_SD on both the USB sd-image and the eMMC/NVMe flash.
+            # The pinned LUKS UUID never matches here, so hunting for it only
+            # burned the full 30 s timeout and the resize never ran at all.
+            for _ in $(seq 1 30); do
+              if [ -b /dev/disk/by-label/NIXOS_SD ]; then
+                PART_DEV="/dev/disk/by-label/NIXOS_SD"
+                break
+              fi
+              sleep 1
+            done
```
brianmcgillion, line 352: `NIXOS_SD` is hardcoded here, while nixpkgs exposes `sdImage.rootVolumeLabel` and derives `fileSystems."/"` from it — on `nvidia-jetson-orin-agx-debug` I get `rootVolumeLabel = "NIXOS_SD"` and `fileSystems."/".device = "/dev/disk/by-label/NIXOS_SD"`. This file already uses exactly that pattern for the ESP at line 28 (`config.sdImage.firmwarePartitionName or "FIRMWARE"`).

As written, anyone who sets `rootVolumeLabel` silently gets back the bug this PR fixes: 30 s hunt, no resize, no error. The `or` fallback keeps module sets without `sdImage` (verity builds) evaluating — I checked that `nvidia-jetson-orin-agx-verity-debug` still evaluates with this applied.

```suggestion
            # Plain (unencrypted) image: the root is the sd-image ext4 partition.
            # Resolve it through the same option upstream sd-image.nix uses for
            # fileSystems."/" rather than a hardcoded literal, with the `or`
            # fallback for module sets where sdImage is absent (verity builds).
            # The pinned LUKS UUID never matches here, so hunting for it only
            # burned the full 30 s timeout before the resize could run.
            ROOT_DEVICE="/dev/disk/by-label/${config.sdImage.rootVolumeLabel or "NIXOS_SD"}"
            for _ in $(seq 1 30); do
              if [ -b "$ROOT_DEVICE" ]; then
                PART_DEV="$ROOT_DEVICE"
                break
              fi
              sleep 1
            done
```

modules/common/logging/common.nix
```diff
@@ -518,7 +518,11 @@ in
 
       clockReady = {
         enable = (mkEnableOption "clock readiness barrier for persistent sealed logging") // {
-          default = true;
+          # The barrier exists for hosts whose RTC needs to settle after
+          # power-on. A VM's clock is host-fed (kvm-clock/ptp_kvm) and is
+          # synchronized from the first instruction, so inside guests the
+          # barrier (and the NTP wait it gates) only delays boot.
+          default = config.ghaf.type == "host";
```
brianmcgillion, line 529: Two small things here.

A config-dependent default wants a `defaultText` — `ghaf.gracefulShutdown` in `modules/common/common.nix` sets one for the same reason. Without it the options doc renders an evaluated bool instead of the expression.

And the mechanism named in the comment does not apply to the Orin targets this PR is about: there is no kvm-clock on aarch64 guests. What actually happens is that QEMU seeds the guest's emulated RTC from host wall time at VM start (microvm.nix sets `rtc = "on"`), and every VM base sets `withTimesyncd = true` for afterwards. The conclusion is unchanged — a guest starts from the host's already-settled clock — but this comment is the justification a future reader will lean on.

```suggestion
          # The barrier exists for hosts whose RTC needs to settle after
          # power-on. A guest starts from the host's already-settled wall clock
          # (seeded into the emulated RTC at VM start) and then tracks it with
          # timesyncd, so inside guests the barrier only delays boot.
          default = config.ghaf.type == "host";
          defaultText = lib.literalExpression ''config.ghaf.type == "host"'';
```

vadika:
All three inline suggestions applied as given (`udevadm settle`, `sdImage.rootVolumeLabel or`, `defaultText` + corrected RTC-seeding rationale — you're right, no kvm-clock on aarch64).

**ghaf-clock-sync in guests**: made it a stated choice, in the commit message and next to the default — guests deliberately lose the NTP wait before FSS setup/verify, which also uncouples guest sealing from net-vm availability; the host keeps its boundary and `ghaf-clock-jump-watcher` recovery stays active everywhere. If you'd rather keep the guest NTP boundary I'll decouple `ghaf-clock-sync` from `clockReadyEnabled` instead — say the word.

**Both follow-ups taken**: `resize-partitions` is now `mkIf (!verityEnabled)`, and `expandOnBoot = false` unconditionally with the comment rewritten (and the resize commit message corrected — upstream's stage-2 unit was indeed growing unencrypted roots all along; the 30 s initrd burn was the real cost).

Evals after the changes: `agx-debug`, `agx-verity-debug`, `nx-debug`, `intel-laptop-debug` toplevels all clean; `nix fmt` 0 changed.

brianmcgillion (approved):

PR #2100, opened 2026-08-04
feat(orin/nx): three-vm split and accelerated gui-vm topologies

vadika:
## Description of Changes

Ports the Orin three-VM composition to the Orin NX 16GB p3768 devkit, in both topologies: split (compute `gpu-vm` + display-only `disp-vm`, the NX default) and combined accelerated `gui-vm`.

- **DCB per board** — the DCB image is the only board-specific piece of the composition; it moves into a `boards` attrset selected by `somType` and reaches both guest DTB derivations through a cpp define. Stock L4T r36.5 ships the same DCB for p3737 and p3768, so `nx` aliases `agx`. AGX guest DTBs are byte-identical before and after.
- **NX topologies** — split enabled by default in the hardware module, plus an `accelerated-guivm` target variant. Rootfs flashes to NVMe (no eMMC on this SoM); the accelerated variant pins a static 32 GiB APP so its flash script does not depend on the cross image build, which is supplied with `-s` at flash time. Memory is refitted for 16 GB: the DT carveouts take ~6.1 GB of physical RAM and VFIO pins all guest RAM, so the AGX sizing oversubscribed the board — gpu-vm 2048, disp-vm 1536, gui-vm 4096, plus host zram.
- **Flip-completion binding** — the DCE R5 binds a channel's flip-completion notifier once, but NVKMS frees and re-registers it on every modeset; in passthrough the rebind stayed unbound, so after the greeter handed over to the session no completion arrived, the first atomic commit never retired, and every later commit failed with `EBUSY` (lit panel, black output). Fixed by keeping the binding across modesets and never recycling completion-notifier handle numbers.
- **Flash erase** — NVIDIA's `blkdiscard` fallback dd-zeroes a whole partition at 512 bytes a write (hours on a 256 GB NVMe); debug flashes now wipe only the head.

#2068 is merged; these 4 commits apply directly to main.

### Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

Follows #2068 (merged).

## Checklist

### Applicable Targets
- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`
### Installation Method
- [ ] Other:
### Test Steps To Verify:
1. AGX unchanged: the `gpuvm-dtb` / `dispvm-dtb` requisites of `nvidia-jetson-orin-agx-debug-from-x86_64` are the same derivations as on the base branch.
2. Split: flash `nvidia-jetson-orin-nx-debug-from-x86_64`; all microVMs reach `microvms.target`; in disp-vm, `modetest -M nvidia-drm -s <conn>@<crtc>:1920x1080` puts a test pattern on the panel. Use the DP port next to the USB-C connector — the other one is not in the DCB routing.
3. Accelerated: build `packages.aarch64-linux.nvidia-jetson-orin-nx-accelerated-guivm-debug` natively, flash it with `-s <image>`; the COSMIC greeter appears, and logging in brings up the desktop.

Validated on an Orin NX 16GB p3768 devkit. Known gaps: the GPU-less disp-vm has no KMS holder (`kms-owner` needs EGL), and cold boots occasionally hang in OP-TEE before the console — both pre-existing and shared with AGX.

### DIFF
--- a/modules/reference/hardware/jetpack/agx/orin-agx.nix
+++ b/modules/reference/hardware/jetpack/agx/orin-agx.nix
@@ -9,12 +9,8 @@
 
   imports = [
     ../../../../common/services/hwinfo
-    # DCE display-proxy HOST integration (AGX only). Keeps gpu_vm
-    # enabled, runs the host headless, force-loads tegra-dce so the host owns
-    # the real DCE R5, and builds + loads the dce-host-proxy .ko (with an
-    # injected nvidia,dce-host-proxy DT node) so the guest can drive the panel
-    # through the host-owned DCE. AGX-only on purpose -- do not hoist into a
-    # shared orin.nix.
+    # Host keeps the real DCE R5 and loads dce-host-proxy so the guest drives
+    # the panel through it. Shared by AGX and NX.
     ../nvidia-jetson-orin/virtualization/common/dce-virt-common/dce-probe-host.nix
   ];
 
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix
@@ -296,6 +296,20 @@ let
         echo "QSPI-only mode: skipping ESP and root partition extraction."
       ''}
 
+      ${lib.optionalString config.ghaf.profiles.debug.enable ''
+        # NVIDIA's blkdiscard fallback dd-zeroes the whole partition at
+        # 512 bytes a write: hours on a 256GB NVMe. Wiping the head is enough
+        # to drop the old structures; the image write covers the rest. Blocks
+        # the sparse write skips keep old data -- debug only, never release.
+        eraser=tools/kernel_flash/l4t_flash_from_kernel.sh
+        if grep -q 'dd if=/dev/zero of=[^ ]* status=progress oflag=direct' "$eraser"; then
+          sed -i 's#dd if=/dev/zero of=\([^ ]*\) status=progress oflag=direct#dd if=/dev/zero of=\1 bs=4M count=8 conv=notrunc oflag=direct#g' "$eraser"
+          echo "Debug flash: capped erase-partition dd fallback to a 32MiB head wipe"
+        else
+          echo "WARNING: erase fallback pattern not found in $eraser; full-partition zeroing still active" >&2
+        fi
+      ''}
+
       echo ""
       echo "Ready to flash!"
       echo "============================================================"
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/dce-probe-host.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/common/dce-virt-common/dce-probe-host.nix
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
 #
-# DCE display-proxy HOST integration (AGX only).
+# DCE display-proxy HOST integration.
 #
 # Companion to gpu-vm's guest overlay (../../passthrough/gpu-vm). Guest takes
 # display@13800000 and drives the panel via a synthetic DCE proxy; HOST keeps
@@ -14,7 +14,8 @@
 #     the guest's DCE IPC through it to the R5);
 #   - keep gpu_vm ENABLED (unlike the Phase-0 spike, which forced it off).
 #
-# AGX-only: imported from agx/orin-agx.nix, never hoisted into a shared orin.nix.
+# Shared by AGX and NX (both Tegra234 + host-owned DCE): imported from
+# agx/orin-agx.nix and nx/orin-nx.nix.
 {
   lib,
   pkgs,
--- a/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/disp-vm/default.nix
+++ b/modules/reference/hardware/jetpack/nvidia-jetson-orin/virtualization/passthrough/disp-vm/default.nix
@@ -47,9 +47,14 @@ let
     "vfio-platform,host=${r.dev},mmio-base=${r.base}"
   ]) mappings;
 
-  inherit (import ../payload { inherit lib pkgs; }) capabilities mkPayload;
+  inherit (import ../payload { inherit lib pkgs; })
+    capabilities
+    mkPayload
+    boardFor
+    ;
   cap = capabilities.dispvm;
   payload = mkPayload cap;
+  board = boardFor config.ghaf.hardware.nvidia.orin.somType;
   _capOk =
     payload.needsDceBridge
     && payload.noSyncpointPatch
@@ -85,7 +90,7 @@ let
         };
       in
       ''
-        $CC -E -nostdinc -undef -D__DTS__ -DEXP_DROP_HOST1X -DEXP_DROP_GPU -x assembler-with-cpp \
+        $CC -E -nostdinc -undef -D__DTS__ -DEXP_DROP_HOST1X -DEXP_DROP_GPU -DGHAF_DCB_DTSI='"${board.dcbDtsi}"' -x assembler-with-cpp \
           -I${mainInc} \
           -I${../gpu-vm/nv-dt-bindings} \
           -I${gpuvmDtsi} \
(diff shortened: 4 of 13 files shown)


vadika:
Update: the greeter→session black-screen on the accelerated gui-vm is **fixed** (commit `f9fad5aa`), and the desktop now renders after login on the Orin NX.

Root cause: the DCE R5 binds a display channel's flip-completion notifier **once**. NVKMS frees and re-registers that event on every modeset, and under passthrough the re-registered handle stayed unbound — after the greeter handed over to the session no completion ever arrived, the session's first atomic commit never retired, and every later commit failed with `EBUSY` (lit panel, black output, compositor spinning at ~100% CPU). Only a host reboot recovered, because that resets the R5.

Evidence (NVKMS/RM instrumentation): during the greeter, 75 000+ completion deliveries on `hEvent 0x73`; at login `unregister 0x73/0x74`, then a late stale `POST_EVENT` for `0x73` ~0.6 s after the free (the R5 unbinds asynchronously), then re-registration — and zero deliveries afterwards.

The fix is two coupled changes in `nvkms-evo.c` (2 files, 78 lines):
1. skip the post-modeset unregister, so the working binding survives (its event data — `pDispEvo`, apiHead, layer — is unchanged across the re-modeset; `ClearApiHeadState()` still frees it on teardown);
2. never recycle completion-notifier handle numbers, since the deferred R5 unbind would otherwise tear down a same-numbered rebind.

Both halves are required. Hardware-validated on Orin NX 16GB; two speculative host-side changes tried along the way (a DISPRM drain path and an 8→64 event-ring enlargement) turned out unnecessary and were dropped. The same R5 behaviour applies to the AGX, so this is worth a regression check there.

brianmcgillion (commented):
Reviewed at `c89c5330` against the merge base `3189f567`, statically across all 13 files plus
build-level verification of the claims the description makes. No hardware — I have an AGX on the
bench, not an NX, so everything display-related here rests on your devkit testing.

The shape of this is good. The DCB indirection is the minimum viable abstraction rather than a
speculative board-registry, the memory reasoning is written down where the next person will find
it, and the flip-completion commit message explains a genuinely hard-won failure mode. Three
things I checked because they are the kind of claim that is easy to assert and easy to get wrong,
and all three hold:

- **AGX guest DTBs really are byte-identical.** Built both `gpuvm-dtb` and `dispvm-dtb` at the
  merge base and at `c89c5330` and compared output bytes: identical in both cases. (Worth fixing
  the wording in test step 1 though — it says the *derivations* are the same, and they are not:
  the `.drv` hash changes because the cpp line gained `-DGHAF_DCB_DTSI`. I briefly thought the
  claim had failed before I compared the outputs. It is the DTBs that are identical, not the
  derivations.)
- **The USB-rules dedup is a true refactor.** `guivmRules` for the AGX accelerated target hashes
  identically before and after.
- **The erase cap is not a silent no-op.** The regex matches all three real call sites in
  `l4t_flash_from_kernel.sh` (`:723`, `:736`, `:742`), and it is correctly gated —
  `debug.enable` is `true` for `nvidia-jetson-orin-nx-debug` and `false` for `-release`. The
  `else` branch that warns rather than failing is the right call for something this cosmetic.

Also worth saying: pinning the DCB blob by sha256 *and* length, with `exit 1` on drift, is a
better safety net than most of this kind of code gets.

---

# Blast radius beyond the NX

My main concern is that three changes reach further than the description suggests, and the
testing section covers only "an Orin NX 16GB p3768 devkit".

## 1. Every NX *release* target gains the split topology

`nvidia.passthroughs.gpu_vm.enable = true` and `disp_vm.enable = true` land in the hardware
module, so they apply to every NX target, not just the new debug ones:

```
nvidia-jetson-orin-nx-release VMs
  before: ["admin-vm","net-vm"]
  after:  ["admin-vm","disp-vm","gpu-vm","net-vm"]
```

That is a shipped product picking up two new VMs, a GPU passthrough path and a DCE host proxy.
The description does say "split enabled by default in the hardware module", so this is deliberate
— but it reads as an NX-devkit change, and I do not think a release image was in scope of the
validation. Either scope the defaults to the debug targets, or say explicitly that release NX is
meant to move too and has been exercised.

## 2. The flip-completion patch changes AGX, which was not tested here

`0024-nvkms-keep-flip-completion-binding.patch` is in the unconditional patch list
(`guest-module.nix:136`), not behind `lib.optional`. Confirmed by derivation: for
`nvidia-jetson-orin-agx-debug`, two of the three `l4t-oot-modules` derivations change between the
base and this branch. So AGX guests get the new NVKMS behaviour.

That matters more than usual because #2068 landed the AGX accelerated path hours ago, so AGX is
the platform under active use, and the failure this fixes (greeter hands over, first atomic commit
never retires, panel lit but black) only appears with a real panel and a login. Cheap to
de-risk: either state that AGX was re-tested with a monitor, or gate the patch on the NX until it
has been.

Two things about the patch itself:

- Dropping `nvFreeUnixRmHandle()` means completion-notifier handle numbers are never returned to
  the allocator. The comment explains why recycling is unsafe (~0.6 s late unbind posts), and that
  reasoning is convincing, but the consequence is a monotonic leak of handle numbers — one per
  channel per modeset, for the life of the boot. Modesets are not rare on a desktop: every
  resolution change, every DPMS cycle, every session restart. How large is the handle space, and
  what happens when it is exhausted? A quarantine list that frees after a delay would bound it; if
  the space is large enough that this is genuinely irrelevant, saying so in the comment would stop
  the next reader worrying.
- `nvEvoPostModesetUnregisterFlipOccurredEvent` becomes `{ return; }` with all three parameters
  unused. Harmless if NVIDIA's build does not use `-Wunused-parameter`, but worth a glance.

## 3. NX flashing moves from USB to NVMe for every NX target

```
before: deviceDisk = "sda"   (comment: "Default is set to USB")
after:  deviceDisk = "nvme0n1"
```

Correct for the p3768 devkit as described, and the old comment did note both are possible — but
anyone currently flashing an NX to a USB stick loses that with no deprecation note. If USB is no
longer a supported path, the removed comment was the only place that said it was, so it is worth
one line in the description or the module.

---

# Smaller things

**Memory: appvm ceilings were not refitted with everything else.** The split target comes out at
6.5 GB total, comfortably inside the 7 GB budget the comment sets. The accelerated variant totals
43 GB on a 16 GB board, because `chromium-vm` and `flatpak-vm` keep the 18432 MB ceilings
inherited from the 64 GB AGX:

```
nvidia-jetson-orin-nx-accelerated-guivm-debug
  chromium-vm 18432, flatpak-vm 18432, gui-vm 4096, net-vm 2048, admin-vm 1024
```

`balloon = true` on both, so these are ceilings rather than pins and the mem-manager reclaims —
which is presumably why it boots. But every other VM on this target was explicitly resized for
16 GB and these two were not, which reads more like an oversight than a decision. If ballooning
genuinely makes the ceiling irrelevant here, a one-line comment would settle it.

**`boardFor` falls back to AGX for anything unrecognised.**

```nix
boardFor = somType: if somType == "nx" then boards.nx else boards.agx;
```

`somType` is also `agx64` and `agx-industrial` today, and both correctly want the AGX board, so
this is right now. But the failure mode for a future SoM is silently getting AGX's DCB rather than
a build error. `boards.${somType} or (throw "no DCB for somType ...")` costs one line and turns a
wrong-display-config into a build failure.

**The "nx aliases agx" DCB claim is not what the checksum protects.** The sha256/length check
verifies the pinned blob has not drifted from the DTSI in-tree — it cannot tell you this DCB is
correct for the NX's panel routing. The description's own test step notes only one of the two DP
ports is in the DCB routing, which is exactly the sort of board-specific detail a shared DCB can
get wrong. Your hardware validation is the real evidence here; the comment "verified against the
NX host DTB" is doing the load-bearing work and would be worth expanding to say how.

**Erase cap leaves old data beyond the head.** The comment is honest that sparse-write-skipped
blocks keep old contents, and debug-only is the right gate. Just noting it means a debug reflash
no longer wipes a device between users — fine for a bench, worth knowing if these boards get
passed around.

---

# Verification

`nix fmt --fail-on-change` and `reuse lint` clean. All of `nvidia-jetson-orin-nx-debug`,
`-nx-accelerated-guivm-debug`, `-agx-debug`, `-agx-accelerated-guivm-debug`, `-agx64-debug`,
`-agx-industrial-debug` and `intel-laptop-debug` evaluate. Both new NX images build from a clean
store (`--option builders ''`):

```
nvidia-jetson-orin-nx-debug-from-x86_64                    -> built
nvidia-jetson-orin-nx-accelerated-guivm-debug-from-x86_64  -> built
```

Not covered: anything on real NX hardware, and the AGX display regression risk in item 2.

brianmcgillion (approved):

PR #2106, opened 2026-08-05
feat(microvm): select Crosvm for AdminVM

vadika:
## Description of Changes

Add per-system-VM VMM selection with QEMU as the default, run AdminVM with Crosvm, and use the hypervisor-neutral vsock option. Keep the service unprivileged and retain `--disable-sandbox`; no `CAP_SYS_ADMIN` is added.

### Type of Change

- [x] New feature
- [x] Documentation

## Related Issues / Tickets

Prerequisite for #2123.

## Checklist

- [x] Clear summary
- [x] Signed-off conventional commit
- [x] Documentation updated
- [x] Ready for review

### Applicable Targets

- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`
- [x] Intel Laptop low-memory `x86_64`

### Installation Method

### Test Steps To Verify

1. Confirm AdminVM selects `crosvm`.
2. Confirm another system VM retains `qemu`.
3. Verify the AdminVM service is active and uses a vsock CID.

### DIFF
--- a/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
+++ b/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
@@ -15,12 +15,14 @@ Nevertheless, it may happen that some hypervisor options are not supported by mi
 QEMU and crosvm are VMMs (virtual machine monitors) running on top of KVM, not
 hypervisors themselves. Ghaf selects the system VM VMM through
 `ghaf.virtualization.vmConfig`. QEMU is the default for system VMs, while
-AdminVM uses crosvm on every target. Each system VM can override the default
+AdminVM uses crosvm when storage encryption is disabled. Encrypted AdminVMs
+remain on QEMU because the crosvm runner does not yet provide the TPM path used
+to seal their storage keys. Each system VM can override the default
 independently:
 
 ```nix
 vmConfig = {
-  defaultVmm = "qemu";
+  defaultSysVmVmm = "qemu";
   sysvms = {
     adminvm.vmm = "qemu"; # Optional AdminVM fallback
     netvm.vmm = "crosvm";
@@ -29,10 +31,13 @@ vmConfig = {
 ```
 
 System VM keys use unhyphenated names such as `adminvm`, `netvm`, and `guivm`.
-Profiles apply the selection when they extend each VM's evaluated configuration.
-This supports a mixed fleet during migration instead of requiring every VM on a
-target to use the same VMM. Note that microvm.nix names its own selector
-`microvm.hypervisor`; Ghaf maps `vmm` onto it.
+The selection takes effect only when the VM's evaluated configuration includes
+`lib.ghaf.vm.applyVmConfig`. Profiles and targets that construct an
+`evaluatedConfig` directly must include this helper; otherwise the default and
+per-VM settings are ignored. This supports a mixed fleet during migration
+instead of requiring every VM on a target to use the same VMM. Note that
+microvm.nix names its own selector `microvm.hypervisor`; Ghaf maps `vmm` onto
+it.
 
 AdminVM declares its vsock CID through the VMM-independent
 `microvm.vsock.cid` option. Both the QEMU fallback and the crosvm default
--- a/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
+++ b/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
@@ -10,6 +10,55 @@ Update this section after the pull request https://github.com/tiiuae/ghaf/pull/9
 
 Nevertheless, it may happen that some hypervisor options are not supported by microvm. For example, adding specific devices. This document considers such cases.
 
+### Selecting a System VM VMM
+
+QEMU and crosvm are VMMs (virtual machine monitors) running on top of KVM, not
+hypervisors themselves. Ghaf selects the system VM VMM through
+`ghaf.virtualization.vmConfig`. QEMU is the default for system VMs, while
+AdminVM uses crosvm on every target. Each system VM can override the default
+independently:
+
+```nix
+vmConfig = {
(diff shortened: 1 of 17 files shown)


vadika:
@brianmcgillion fixed the terminology

brianmcgillion (changes requested):
Reviewed as part of the stack **#2106 → #2123 → #2125**, but judged on its own, since it can merge independently.

The refactor itself is good — centralising VMM selection in `vmConfig` is the right shape, and I verified the risky parts of it are safe. But this PR alone regresses encrypted targets.

---

## Blocking

### Encrypted AdminVM loses its TPM on both architectures

`modules/microvm/sysvms/adminvm-base.nix:116-130` enables a TPM whenever storage encryption is on:

```nix
tpm.passthrough.enable = encryption && hostSystem == "x86_64-linux";
tpm.emulated.enable    = encryption && hostSystem != "x86_64-linux";
```

At this commit `modules/microvm/common/vm-tpm.nix` wires **both** branches only into `microvm.qemu.extraArgs`. crosvm ignores those, and this PR makes AdminVM crosvm on every target — so the guest gets no TPM device at all. Neither `--tpm-device` nor `--swtpm` exists until #2125.

That matters because `modules/microvm/common/storagevm.nix:134-144` seals the LUKS key to the TPM:

```nix
${lib.optionalString tpm.passthrough.enable "--tpm2-seal-key-handle=${tpm.passthrough.rootNVIndex}"} "${drivePath}"
```

And the guard at `storagevm.nix:69-73` only checks that the Ghaf *option* is enabled, not that it reached the VMM, so it still passes while AdminVM boots without a TPM.

This is not hypothetical: `lib/global-config.nix:508` (release profile) and `modules/reference/profiles/mvp-user-trial.nix:187` both set `storage.encryption.enable = true`. CI stays green only because the debug targets built there don't enable encryption.

**Suggested fix:** land the `vm-tpm.nix` crosvm support from #2125 in this PR, or gate the AdminVM crosvm default on `!storage.encryption.enable` until it lands.

---

## Should fix here

### `--disable-sandbox` as `mkDefault` on a list option is a booby trap

`lib/global-config.nix:881`:

```nix
crosvm.extraArgs = lib.mkDefault [ "--disable-sandbox" ];
```

On a `listOf str`, a `mkDefault` definition is **discarded**, not concatenated, as soon as any normal-priority definition exists. Verified with `nix-instantiate`:

```
mkDefault ["--disable-sandbox"] + mkAfter ["--swtpm" "sock"]  =>  [ "--swtpm" "sock" ]
mkBefore  ["--disable-sandbox"] + mkAfter ["--swtpm" "sock"]  =>  [ "--disable-sandbox" "--swtpm" "sock" ]
```

Harmless at this commit — nothing else defines `crosvm.extraArgs`. But #2123 adds exactly such a definition in `vm-tpm.nix` and silently drops the flag, and #2125 then fixes it by switching to `mkBefore`. Please use `mkBefore` from the start, with a short comment explaining that `mkDefault` would be discarded — otherwise someone will "tidy" it back later.

---

## Non-blocking

- **`defaultVmm` is misnamed.** It governs system VMs only, which #2123 makes obvious by adding `defaultAppVmVmm` alongside it. `defaultSysVmVmm` (or `sysvms.defaultVmm` / `appvms.defaultVmm`) would be symmetric. Worth settling now — renaming after all three ship is a breaking change for downstream consumers. The two `<vm>.vmm` submodule options carry `example =`, but neither top-level default does.

- **A known-bad-VMM note was deleted without a replacement.** `adminvm-base.nix` dropped:

  ```nix
  #TODO: Add back support cloud-hypervisor
  #the system fails to switch root to the stage2 with cloud-hypervisor
  ```

  That is exactly the kind of thing the new `defaultVmm` enum (`[ "qemu" "crosvm" ]`) should record — please move it into the option description.

- **`sysvms.<vm>.vmm` only works where a profile threads `applyVmConfig`.** Coverage today: laptop-x86, orin, targets/vm, and mvp-user-trial-extras. Any downstream profile that assigns `evaluatedConfig` directly gets a silently ignored `vmm`. Worth one sentence in `hypervisor_options.mdx` and in the `sysvms` option description.

---

## Verified as safe (no action needed)

I checked these because they looked risky, and they are fine:

- **Removing `hypervisor = "qemu"` from the base modules.** `microvm.nix` (pinned `e615e232`) defaults `hypervisor` to `"qemu"` at `nixos-modules/microvm/options.nix:51-53`, and an option default is weaker than the `mkDefault selectedVmm` that `applyVmConfig` injects. `gpuvm` and `dispvm` are reached by `applyVmConfig` (`orin.nix:392`, `:405`); `idsvm` is disabled on Orin and covered via `mvp-user-trial-extras.nix:46`. No VM ends up on an unintended VMM.

- **The `vsock.cid` switch is an improvement.** Replacing the hardcoded `-device vhost-vsock-pci,...` with `microvm.vsock.cid` lets microvm.nix emit `vhost-vsock-${devType}` (`lib/runners/qemu.nix:364-367`), which handles the `microvm` machine type the old hardcoded PCI form got wrong. The crosvm runner emits `--vsock <cid>` from the same option.

- **The doc inspection snippet is accurate.** `cat /var/lib/microvms/admin-vm/current/share/microvm/hypervisor` works — `lib/runner.nix:275` writes that file.

---

## Note on the sandbox tradeoff

`--disable-sandbox` is **not** a regression against QEMU — Ghaf never passed `-sandbox on`, so both VMMs run single-process as unprivileged `microvm:kvm`. But it does forgo crosvm's per-device minijail, which is the main isolation argument for migrating. The docs currently imply a TCB reduction that isn't being realised yet; I'd state the tradeoff plainly and link a follow-up issue for the capability-scoped sandbox.

Also flagging that **`jenkins-pre-merge` is failing** on this PR (build 2670) and on both descendants. I can't read the internal log — worth confirming it's unrelated before merge.

PR #2123, opened 2026-08-08
feat(appvm): select Crosvm by default

vadika:
## Description of Changes

Run AppVMs with Crosvm by default while preserving per-AppVM overrides. Add swtpm-backed virtio TPM, balloon control, runtime USB hotplug, neutral vsock configuration, and the guest TPM transport.

### Type of Change

- [x] New feature
- [x] Documentation

## Related Issues / Tickets

- Depends on #2106
- Depends on tiiuae/ghaf-crosvm#1
- Depends on tiiuae/vhotplug#21

## Checklist

- [x] Clear summary
- [x] Signed-off conventional commit
- [x] Commits squashed
- [x] Documentation updated
- [x] Review threads resolved

### Applicable Targets

- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`
- [x] VM targets

### Installation Method

- [x] Requires a full installation for encrypted storage validation

### Test Steps To Verify

1. Boot an encrypted AppVM configuration.
2. Verify AppVM services use Crosvm and expose TPM devices.
3. Exercise USB attach/detach and balloon target changes.
4. Reboot and confirm existing storage still unlocks.

### Validation

- Ghaf formatting, REUSE, evaluation, pre-commit, and installer builds passed
- live x86 and AGX guests completed TPM commands
- live x86 laptop booted the Crosvm AppVM configuration without changing TPM or LUKS state

### DIFF
--- a/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
+++ b/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
@@ -15,12 +15,14 @@ Nevertheless, it may happen that some hypervisor options are not supported by mi
 QEMU and crosvm are VMMs (virtual machine monitors) running on top of KVM, not
 hypervisors themselves. Ghaf selects the system VM VMM through
 `ghaf.virtualization.vmConfig`. QEMU is the default for system VMs, while
-AdminVM uses crosvm on every target. Each system VM can override the default
+AdminVM uses crosvm when storage encryption is disabled. Encrypted AdminVMs
+remain on QEMU because the crosvm runner does not yet provide the TPM path used
+to seal their storage keys. Each system VM can override the default
 independently:
 
 ```nix
 vmConfig = {
-  defaultVmm = "qemu";
+  defaultSysVmVmm = "qemu";
   sysvms = {
     adminvm.vmm = "qemu"; # Optional AdminVM fallback
     netvm.vmm = "crosvm";
@@ -29,10 +31,13 @@ vmConfig = {
 ```
 
 System VM keys use unhyphenated names such as `adminvm`, `netvm`, and `guivm`.
-Profiles apply the selection when they extend each VM's evaluated configuration.
-This supports a mixed fleet during migration instead of requiring every VM on a
-target to use the same VMM. Note that microvm.nix names its own selector
-`microvm.hypervisor`; Ghaf maps `vmm` onto it.
+The selection takes effect only when the VM's evaluated configuration includes
+`lib.ghaf.vm.applyVmConfig`. Profiles and targets that construct an
+`evaluatedConfig` directly must include this helper; otherwise the default and
+per-VM settings are ignored. This supports a mixed fleet during migration
+instead of requiring every VM on a target to use the same VMM. Note that
+microvm.nix names its own selector `microvm.hypervisor`; Ghaf maps `vmm` onto
+it.
 
 AdminVM declares its vsock CID through the VMM-independent
 `microvm.vsock.cid` option. Both the QEMU fallback and the crosvm default
--- a/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
+++ b/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
@@ -43,12 +43,47 @@ AdminVM declares its vsock CID through the VMM-independent
 `microvm.vsock.cid` option. Both the QEMU fallback and the crosvm default
 therefore use the VM's configured network CID without raw QEMU arguments.
 
-#### crosvm Sandbox Boundary
+### Selecting an App VM VMM
+
+App VMs use crosvm by default on both x86 and NVIDIA platforms. The default
+and individual App VMs can be overridden through the same
+`ghaf.virtualization.vmConfig` interface:
+
+```nix
+vmConfig = {
+  defaultAppVmVmm = "crosvm";
+  appvms = {
+    chrome.vmm = "qemu";
+    flatpak.vmm = "crosvm";
+  };
+};
+```
+
+App VM keys use their unhyphenated base names, such as `chrome`, `flatpak`, and
+`chromium`. Encrypted App VMs remain on the crosvm default. Ghaf's crosvm build
+connects its virtio TPM frontend to the same swtpm proxy socket as QEMU, and
+the App VM guest kernel includes ChromiumOS's virtio TPM transport driver when
+that crosvm App VM has a vTPM enabled. Applying this out-of-tree driver creates
+a separate guest-kernel derivation until that result is available from a binary
+cache.
+
+crosvm App VMs use the same `mem * (balloonRatio + 1)` allocation and
+virtio-balloon policy as QEMU App VMs. `ghaf-mem-manager` selects QEMU QMP or
+the crosvm control command from the VM's configured VMM. It also translates
+between QEMU's guest-visible memory value and crosvm's reclaimed-memory value.
+
+App VMs declare their vsock CID through `microvm.vsock.cid`, so QEMU and crosvm
+use the same configured CID. Runtime USB attachment through vhotplug is
+supported by crosvm. vhotplug records the crosvm guest port for each host USB
+device so identical devices can be detached independently and restored safely
+after daemon restarts. Runtime PCI attachment remains QEMU-only.
+
+### crosvm Sandbox Boundary
 
 Ghaf runs VMM services as the unprivileged `microvm` user. crosvm's internal
 multiprocess minijail requires `CAP_SYS_ADMIN` to create PID and mount
 namespaces, which the service intentionally does not have. Ghaf consequently
-adds `--disable-sandbox` to crosvm VM runners by default.
+adds `--disable-sandbox` to crosvm system VM and App VM runners by default.
 
 This keeps the VMM process behind the unprivileged systemd service boundary,
 but it disables crosvm's additional per-device-process isolation. Treat this as
--- a/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
+++ b/docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
@@ -10,6 +10,55 @@ Update this section after the pull request https://github.com/tiiuae/ghaf/pull/9
 
 Nevertheless, it may happen that some hypervisor options are not supported by microvm. For example, adding specific devices. This document considers such cases.
 
+### Selecting a System VM VMM
+
+QEMU and crosvm are VMMs (virtual machine monitors) running on top of KVM, not
+hypervisors themselves. Ghaf selects the system VM VMM through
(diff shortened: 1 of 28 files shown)


modules/microvm/common/vm-tpm.nix
```diff
@@ -81,11 +83,9 @@ in
       };
     })
     (mkIf cfg.emulated.enable {
-      microvm.qemu.extraArgs = [
+      microvm.qemu.extraArgs = lib.mkIf (config.microvm.hypervisor == "qemu") [
```
TanelDettenborn: Is `lib.mkIf` redundant check? 

Edit: There are similar check. For example line 97.

docs/src/content/docs/ghaf/dev/technologies/hypervisor_options.mdx
```diff
@@ -10,6 +10,87 @@ Update this section after the pull request https://github.com/tiiuae/ghaf/pull/9
...
+
+Ghaf runs VMM services as the unprivileged `microvm` user. crosvm's internal
+multiprocess minijail requires `CAP_SYS_ADMIN` to create PID and mount
+namespaces, which the service intentionally does not have. Ghaf consequently
+adds `--disable-sandbox` to crosvm system VM and App VM runners by default.
+
+This keeps the VMM process behind the unprivileged systemd service boundary,
+but it disables crosvm's additional per-device-process isolation. Treat this as
+an explicit migration tradeoff until Ghaf provides capability-scoped namespace
+setup without granting the VMM broad host privileges.
```
TanelDettenborn, line 83: Looks good to me. Just as a point of discussion: is it better to run crosvm with --disable-sandbox instead of granting CAP_SYS_ADMIN?

vadika: I'll do it as a separate commit -- it requires setting up namespaces and other things to fix.

modules/microvm/common/vm-tpm.nix
vadika: it is. fixed.

brianmcgillion (changes requested):
Reviewed as part of the stack **#2106 → #2123 → #2125**, but judged on its own, since it can merge independently.

The `resolveAppVmConfig` refactor is a genuine improvement — pulling the per-profile duplication into `lib/` is the right call. The blocking issues are the vendored kernel driver and a merge-priority collision that breaks AdminVM.

---

## Blocking

### 1. This PR silently drops `--disable-sandbox` for AdminVM on encrypted aarch64

New code at `modules/microvm/common/vm-tpm.nix:97-100`:

```nix
microvm.crosvm.extraArgs = lib.mkAfter [
  "--swtpm"
  emulatedSocket
];
```

This is normal override priority and is applied to **every** VM with `tpm.emulated.enable`, regardless of VMM. On a `listOf str`, a normal-priority definition outranks `mkDefault` and discards it entirely rather than concatenating. Verified with `nix-instantiate`:

```
mkDefault ["--disable-sandbox"] + mkAfter ["--swtpm" "sock"]  =>  [ "--swtpm" "sock" ]
mkBefore  ["--disable-sandbox"] + mkAfter ["--swtpm" "sock"]  =>  [ "--disable-sandbox" "--swtpm" "sock" ]
```

AdminVM is crosvm since #2106, and on Orin with `storage.encryption.enable` it takes the `tpm.emulated` branch (`adminvm-base.nix:124-130`, since `hostSystem != "x86_64-linux"`). So `#2106`'s `mkDefault [ "--disable-sandbox" ]` at `lib/global-config.nix:881` is wiped out, and crosvm starts minijail — which, by this PR's own documentation (`hypervisor_options.mdx:74-78`), needs `CAP_SYS_ADMIN` that the service intentionally does not have. AdminVM fails to start.

App VMs escape only by accident: `appvm-base.nix:427` uses plain `lib.optionals`, which is also normal priority, so it concatenates.

**Fix:** pull #2125's `crosvm.extraArgs = lib.mkBefore [ "--disable-sandbox" ]` back into this PR. Secondary: this block also sets `microvm.crosvm.extraArgs` on QEMU VMs — the `lib.mkIf (vmm == "crosvm")` guard doesn't arrive until #2125 either.

### 2. Flake inputs point at personal forks

```nix
url = "github:vadika/ghafpkgs/feat/crosvm-balloon";      # TODO: after ghafpkgs#356
url = "github:vadika/vhotplug/fix/crosvm-usb-vhotplug";  # TODO: after vhotplug#21
url = "git+https://github.com/tiiuae/ghaf-crosvm?ref=feat/swtpm-backend&submodules=1";
```

Merging pins the product's VMM, memory manager and USB hotplug to branches outside the org's control. The TODOs acknowledge it, but this should gate the merge rather than be tracked in a comment.

### 3. The vendored ChromiumOS virtio-TPM driver has real defects, and its provenance header is wrong

`modules/microvm/sysvms/patches/chromiumos-virtio-tpm.patch` carries a full attribution to ChromiumOS commit `04fdb175` with David Tolnay's 2018 `Signed-off-by:`. The body is **not** that commit — it contains unattributed 2026 modifications in exactly the security-relevant lines. Diffed against upstream `drivers/char/tpm/tpm_virtio.c` (chromeos-5.10):

| upstream | vendored |
|---|---|
| `#include <uapi/linux/virtio_ids.h>` | `#include <linux/version.h>` + `#define VIRTIO_ID_TPM 62` |
| `send(chip, u8 *caller_buf, size_t len)` | version-conditional 3/4-arg signature |
| `sg_init_one(&sg_inbuf, virtqueue_buf, TPM_BUFSIZE);` | `min(bufsiz, (size_t)TPM_BUFSIZE)` |

Three concrete problems (all line numbers patch-relative):

**a. Use-after-free on the probe error path (high).** Lines 457-478. `virtio_device_ready(vdev)` runs before `tpm_chip_register()`, and `.flags = TPM_OPS_AUTO_STARTUP` makes the core send TPM2 startup commands over the queue during probe. The error path is:

```c
err_chip_register:
	put_device(&dev->chip->dev);
err_chip_alloc:
	vdev->config->del_vqs(vdev);
err_virtio_find:
	kfree(dev);
```

There is no `vdev->config->reset(vdev)` — note that `vtpm_remove()` (line 491) *does* call it. If startup times out (`TPM_VIRTIO_TIMEOUT` is 120 s), `driver_has_buffer` stays false, meaning the device still owns the descriptor chain, and we free the vring and then `kfree(dev)`. `dev->virtqueue_buffer` is embedded in that freed 8K slab object, so when swtpm eventually answers, crosvm writes the response into freed guest kernel memory at a host-chosen moment.

The trigger is mundane rather than adversarial: swtpm's socket not being ready when crosvm starts — precisely the ordering this PR introduces with `--swtpm ${emulatedSocket}`. Fix: `virtio_reset_device(vdev)` before `del_vqs()` in the error path, plus `virtqueue_detach_unused_buf()`.

**b. The version guard uses the wrong constant (high, currently latent).** Lines 235 and 246:

```c
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 97)
			, size_t cmd_len
#endif
```

`6.12.97` is a point release, not an API boundary. The `cmd_len` parameter was added to `struct tpm_class_ops.send` in **v6.17** (the `TPM_CHIP_FLAG_SYNC` / "add bufsiz parameter in the .send callback" series) — 6.12.97 and v6.16 both have the 3-argument form; 6.18 and 7.1 have 4. So every kernel in `[6.12.97, 6.17)` selects the 4-arg prototype against a 3-arg ops struct and fails to build at `.send = vtpm_op_send`.

It doesn't bite today because the VMs that get this patch run 6.18/7.1, but Ghaf pins `linuxPackages_6_12` at `mgbe0-net-vm/default.nix:136` and `payload/guest-module.nix:113`, so it's one config change away. Should be `KERNEL_VERSION(6, 17, 0)`.

**c. The device-writable descriptor is sized with the command length on the fallback path (medium).** Line 294:

```c
sg_init_one(&sg_inbuf, virtqueue_buf, min(bufsiz, (size_t)TPM_BUFSIZE));
```

Upstream passes `TPM_BUFSIZE`. On the pre-6.17 ABI the third argument *is* the command length (`tpm-interface.c` calls `send(chip, buf, count)`), and line 247 sets `cmd_len = bufsiz` — so a 12-byte `TPM2_GetRandom` hands the device 12 bytes to write its response into. crosvm returns `BufferTooSmall` and completes the chain with used length 0, so the guest sees `readable == 0` and `tpm_try_transmit` fails `len < TPM_HEADER_SIZE`: `-EFAULT` on essentially every TPM command, with only a VMM-side log line. Harmless on ≥6.17, so this and (b) are the same fallback path.

**Also:** the hunks carry only one line of context and have no `index` lines, so this isn't `git format-patch` output — it will misapply silently on a future rebase. Please regenerate with `-U3` and make the header state the true base commit plus an explicit "Ghaf modifications" list, rather than a 2018 `Signed-off-by:` on code containing 2026 edits.

Worth noting for the record: mainline has no virtio-TPM transport, and device ID 62 is a ChromiumOS-private allocation that was never standardised, so Ghaf owns all future maintenance of this driver. The 6.17 API break is the first instalment.

---

## Non-blocking

- **`requestedVtpm` is a pointless alias.** `appvm-base.nix:54-55`: `effectiveVtpm = requestedVtpm;`, and every other use site reads `effectiveVtpm`. The two names imply a request/policy distinction that doesn't exist — collapse to one, or make the distinction real.

- **`resolveAppVmConfig` is undocumented relative to its sibling.** `applyVmConfig` has a 17-line docstring and is listed in the helper index at `lib/global-config.nix:669-673`; `resolveAppVmConfig` has neither, and is absent from `lib/builders/README.md` even though this PR edits that file and it's the downstream API doc. The shapes also diverge (list-of-modules vs attrset), which is why all three profiles have to hand-plumb `appvmVmm` into `hostConfig`.

- **`balloonRatio` units are wrong** in lines this PR reflowed: `vm-config.nix:136-140` says "`mem * (balloonRatio + 1)` bytes" where `mem` is MB. Pre-existing, but easy to fix while you're there.

- **Doc accuracy:** `hypervisor_options.mdx:56-57` states the App VM guest kernel includes the virtio TPM driver unconditionally, but `appvm-base.nix:411` gates it on `effectiveVtpm && vmm == "crosvm"`. Only App VMs with a vTPM get it.

- **Build cost:** gating `boot.kernelPatches` on `effectiveVtpm && vmm == "crosvm"` forks the App VM kernel, so those VMs lose the binary cache. Worth confirming that's understood.

---

## Checked and cleared

I verified the patch is **REUSE-compliant** despite having no SPDX header and no `REUSE.toml` entry — `reuse lint` passes 953/953 on this tree, because the patch body carries the kernel file's own `SPDX-License-Identifier` lines. No action needed there.

Also flagging that **`jenkins-pre-merge` is failing** on this PR (build 2671) and on both stack neighbours. I can't read the internal log — worth confirming it's unrelated before merge.

PR #368, opened 2023-11-24
template: pull request checklist

vilvo:
Initial pull request template proposal for discussion. 
Checklist is intended to reduce things to remember.
Intent is to support checking with `pre-commit-hooks` checks - where possible.

PR #2133, opened 2026-08-11
feat(nvidia): add Orin Crosvm support

vadika:
## Description

Make Crosvm with ghaf-device-manager the default virtualization stack for every
exported NVIDIA Jetson Orin AGX and NX configuration. The standard targets run
COSMIC in GUI VM with combined GPU/display passthrough; host graphics are
disabled. An explicit QEMU override remains the rollback path and selects
vhotplug.

Ghaf retains:

- target-level VMM, device-manager, and default desktop policy;
- reusable split GPU/display VM bases and host composition;
- thin compute, display, and combined role-to-VM adapters;
- GIVC-first bounded Crosvm shutdown, with the upstream MicroVM helper used only
  after a 60-second guest-agent grace period;
- AGX MGBE0 and NX PCI NetVM composition;
- Crosvm-compatible hardware-info sharing; and
- VMM, GIVC, and VM service contracts.

### microvm.nix

[microvm.nix PR586](https://github.com/microvm-nix/microvm.nix/pull/586)
owns generic platform/PCI VFIO, device-tree overlay, typed AArch64 memory/MMIO
interfaces, and the AArch64 Crosvm shutdown helper. This PR temporarily pins
that upstream branch.

### jetpack-nixos

- [PR21](https://github.com/tiiuae/jetpack-nixos/pull/21) owns base hardware descriptions, DCE overlay, and EGL/GBM compatibility artifacts.
- [PR22](https://github.com/tiiuae/jetpack-nixos/pull/22) owns BPMP/DCE/MGBE host and guest integration, bounded resource policies, DT builders, proxy lifecycle, and reusable GPU/display host-assignment and guest-driver modules.

Ghaf consumes those hardware roles without copying the kernel, driver, graphics,
or resource-policy payload.

### ghaf-device-manager

[ghaf-device-manager PR10](https://github.com/tiiuae/ghaf-device-manager/pull/10)
exports its package overlay. Ghaf consumes that interface instead of copying
the package expression.

### GPU/display examples

[ghaf-gpu-partitioning-examples PR5](https://github.com/tiiuae/ghaf-gpu-partitioning-examples/pull/5)
retains the container runtime, partition-manager integration, workload policy,
CUDA probes, and example-specific checks. It consumes Ghaf-owned split VM
composition and the same Jetpack provider as the default combined GUI target. The dependency remains one-way:
the examples repository extends Ghaf; Ghaf does not import it.

### Other dependencies

- Ghaf dependency cleanup: #2171, closed as superseded by this PR
- Orin fTPM: #2167, kept separate from this PR
- ghafpkgs logging fixes: [#362](https://github.com/tiiuae/ghafpkgs/pull/362)
- platform-specific `ghaf-x86-qemu` and `ghaf-nvidia-qemu` packages: [#368](https://github.com/tiiuae/ghafpkgs/pull/368)
- per-VM BPMP endpoint: [#369](https://github.com/tiiuae/ghafpkgs/pull/369)

## Validation

Current head: `21e67aebdfad2a6ae924844c7cfb1901cb651a09`, rebased onto
main `324dec46d0d3f93ffbf30919b9280d3ce955dd8c`.

```bash
nix fmt -- --fail-on-change
nix develop --command reuse lint
git diff --check
timeout 2h nix build .#checks.x86_64-linux.orin-crosvm-targets --no-link
timeout 2h nix build .#nvidia-jetson-orin-agx-debug-from-x86_64
timeout 2h nix build .#nvidia-jetson-orin-nx-debug-from-x86_64
```

The focused check covers all ten base Orin configurations and all 104 generated
exports. It requires Crosvm, ghaf-device-manager, GUI VM enabled, combined
GPU/display ownership, and host COSMIC disabled. It also verifies the Ghaf-owned split GPU/display composition, AGX MGBE
ordering, hardware-info sharing, bounded shutdown, and explicit QEMU/vhotplug
rollback.

The rebase preserved all 34 feature patches unchanged. The focused check and
both AGX/NX cross-image builds passed at `35e8df49f8`. The target-ownership
follow-up at the current head passes formatting, REUSE, and the focused check;
its cross-image builds are running in CI. Main also advances the Ghaf Crosvm pin
from `fba7ba3a` to `f2f3d210`.

Previously built and flashed outputs from pre-rebase head `4b9bea295`:

- AGX: `/nix/store/g5w2f3alpp7cwii1wg7vdwq2akc4ascv-nixos-image-sd-card-26.11.20260819.ffb3c9b-aarch64-linux.img.zst-aarch64-unknown-linux-gnu`
- NX: `/nix/store/kfpz8sq3xjbskq7f7yx0x4wnb9fx9lik-nixos-image-sd-card-26.11.20260819.ffb3c9b-aarch64-linux.img.zst-aarch64-unknown-linux-gnu`

## Testing Instructions

1. Build the AGX or NX cross target and its flash script.
2. Put the exact board under test into recovery mode. Confirm its ECID before
   flashing; do not identify a board by USB VID/PID alone.
3. Flash the complete image, including QSPI. These debug targets are unsigned:
   do not pass `--secure-boot`, and confirm the flasher reports
   `Boot Authentication: NS`.
4. Cold boot the board. If a prior run ended in a forced VFIO shutdown, remove
   board power for at least 15 seconds before booting it again.
5. Through NetVM, confirm all expected MicroVMs are running, COSMIC is absent on
   the host, and GUI VM owns both NVIDIA DRM card/render pairs and runs the
   display manager, COSMIC, and Xwayland.
6. Restart `microvm@gui-vm.service`, wait for the GUI to return, then restart it
   again immediately. Both restarts must complete within the 95-second service
   timeout with a new PID and no fallback-helper warning.
7. Check host and guest journals for DCE RPC/timeout, SMMU, soft-lockup, or VFIO
   DMA errors. Attach a USB keyboard/mouse receiver and confirm the physical USB
   device is visible on the host but its input event nodes belong to GUI VM.

## Hardware Regression Tests

Both boards were fully flashed at pre-rebase head `4b9bea295` with unsigned
images and reported `Boot Authentication: NS`. The 34 PR patches are identical
after the rebase, but the updated mainline Crosvm pin means these results are
retained as prior regression evidence rather than acceptance of the current
head.

- AGX ECID `0x80012344705DD3C96C0000000A0081C0`: GUI VM restart-to-ready
  times were 27 seconds and 47 seconds for an immediate second restart. The GUI
  returned with both DRM card/render pairs, COSMIC, and Xwayland. A Logitech
  `046d:c52b` receiver remained physically visible on the host while its keyboard
  and mouse input nodes appeared only in GUI VM.
- NX ECID `0x80012344705DD5C58C00000011FF8100`: after a 15-second rail-power
  drain, GUI VM restart-to-ready times were approximately 23 seconds and 33
  seconds for an immediate second restart. The GUI returned with both DRM
  card/render pairs, COSMIC, and Xwayland. No receiver was attached for the NX
  input test.

Neither board used the last-resort shutdown helper during the measured restart
cycles, and neither produced DCE, SMMU, soft-lockup, or VFIO DMA errors. fTPM is
outside this PR and remains tracked separately in #2167.


brianmcgillion (changes requested):
This is careful, high-effort work. The device-tree drift guards (DCB length + sha256 pin, the
build-time *and* runtime BPMP id checks, the `__symbols__` presence check, the memory-node `reg`
assertions), the eval-time negative tests, and the device-order assertion are all the right
instincts, and the hardware acceptance write-up is unusually thorough. The QEMU rollback gating is
correct essentially everywhere I checked.

That said, everything below needs addressing before merge. Finding 1 in particular I think inverts
the meaning of the hardware acceptance results.

Checks run in a clean worktree: `nix fmt -- --fail-on-change` clean (582 files), `reuse lint` clean
(950/950). I did not run `nix flake check`.

---

## Lifecycle and shutdown

### 1. The `crosvm stop` fallback hard-kills every Orin VM on every stop, not just on failure

`modules/microvm/common/crosvm-command.nix:229-237`

```nix
if ! crosvm --no-syslog powerbtn ${socket}; then
  crosvm --no-syslog stop ${socket}
fi
```

The description frames this as "*If* AArch64 rejects `powerbtn`, fall back to `crosvm stop`". On
AArch64 it is not a conditional — it is unconditional. In the pinned `ghaf-crosvm` (`767f265b`):

- `aarch64/src/lib.rs:1259` sets `pm: None` (x86 sets `pm: Some(acpi_dev_resource.pm)` at
  `x86_64/src/lib.rs:1490`).
- `vm_control/src/lib.rs:1979-1986`: `VmRequest::Powerbtn` with `pm == None` logs "not supported"
  and returns `VmResponse::Err(ENOTSUP)`.
- `vm_control/src/client.rs:64-68`: any non-`Ok` response ⇒ non-zero exit.

So `powerbtn` **always** fails on aarch64 and the fallback **always** runs. And `stop` is
`VmRequest::Exit` (`src/main.rs:178-180`, `src/crosvm/sys/linux.rs:3792`) — an abrupt VMM exit. The
guest is never asked to shut down.

The safety net that would normally catch this is inert on Orin: `modules/profiles/orin.nix:134` sets
`global-config.features.power-manager.enable = false`, which via
`modules/microvm/host/microvm-host.nix:144` disables `ghaf.services.power-manager`, which gates the
whole `config = mkIf cfg.enable` block in `modules/common/services/power.nix:743` — including the
existing, well-guarded `crosvm-stop` ExecStop override at `power.nix:1329-1365`. Orin VMs therefore
fall through to microvm.nix's default `ExecStop = .../bin/microvm-shutdown`, i.e. exactly this
script.

**Consequence:** every `systemctl stop microvm@<vm>`, every host reboot, every `nixos-rebuild`
restart is a power-cut for gui-vm, gpu-vm, disp-vm, admin-vm, every appvm, and net-vm on NX. Those
guests mount writable ext4 volumes (`modules/microvm/common/storagevm.nix:49,176` — `/persist/...`,
`/home`) plus writable-store overlays. And gui-vm/gpu-vm/disp-vm own live VFIO platform DMA (GPU,
host1x, display) — precisely the SMMU-context-fault class the MGBE0 work goes to such lengths to
avoid, left unaddressed for the VMs with the *most* DMA.

This also reframes the acceptance results. "Zero restarts", "three restart cycles returned success"
and "a joint GPUVM/DispVM restart completed in 8 seconds" are measuring the thing the hard kill
makes succeed. `stop` makes the systemd operation complete; it does not make the guest shut down.

**Required:** don't use bare `crosvm stop` as the fallback. Either reuse the shape already in
`power.nix:1329-1365` (powerbtn → SIGTERM → bounded wait → loud ERROR), or extend the GIVC
`poweroff.target` path you built for net-vm to Orin crosvm sysvms generally rather than
special-casing one VM. At minimum, don't fire on a *predictable* ENOTSUP, and log why it fired.

### 2. `stopCrosvmNetVm` never checks `givc-cli`, so a GIVC failure ends in the SIGKILL it exists to prevent

`mgbe0-net-vm/default.nix:32-55, 258-268`

`givc-cli … start service --vm net-vm poweroff.target &` is backgrounded, its exit status discarded,
and the script only has `set -u`. If GIVC is down, certs are missing, or an ACL denies the call,
nothing happens: the loop spins, `TimeoutStopSec=30` expires, systemd SIGTERMs the cgroup. Crosvm
installs no SIGTERM handler on Linux, so it dies instantly with MGBE DMA live → host SMMU context
fault. Exactly the outcome the comment at `:41-44` says it is avoiding. The operator sees a 30 s
stall and no error.

It is also a weaker copy of the existing `sysvm-stop` in `power.nix:1176-1215`, dropping three
guards that file already has: `switch-to-configuration` detection, a bounded deadline that exits 1
with an explicit ERROR, and the grace-period warning. `TimeoutStopSec = "30"` also cuts microvm.nix's
150 s default, giving the guest 30 s for a full NixOS shutdown *including* the quiesce.

**Required:** run `givc-cli` synchronously under `timeout`, branch on its status, log a distinct
error; adopt the `power.nix` guards rather than re-deriving them.

### 3. `quiesce-mgbe0` hardcodes a guest path that is a function of net-vm's `mem`

`mgbe0-net-vm/default.nix:287` — `/sys/bus/platform/devices/c0000000.ethernet/net`

net-vm sets no `microvm.crosvm.platformMmio`, so crosvm defaults the aperture to
`vm.get_memory().end_addr()` (`aarch64/src/lib.rs:597-601`). With RAM base `0x8000_0000` and netvm's
default `mem = 1024` (`sysvms/netvm-base.nix:267`), the first platform device lands at `0xC000_0000`.
Correct today, entirely incidental. The repo already varies this:
`targets/nvidia-jetson-orin/flake-module.nix:345-348` sets `sysvms.netvm.mem = 2048` for NX
(→ `0x1_0000_0000.ethernet`). MGBE0 is AGX-only so nothing breaks now, but if it ever does,
`quiesce-mgbe0` exits 1, the link is never brought down, and you get the SMMU fault — visible as one
line in the guest journal during shutdown.

**Required:** resolve by driver/compatible (walk `/sys/bus/platform/drivers/tegra-mgbe/*/net`, or
match `of_node/compatible`), or derive the address from `mem` in Nix and assert it.

### 4. `quiesce-mgbe0`'s ordering works only incidentally, and its failure is near-silent

`mgbe0-net-vm/default.nix:409-422`

The reverse-ordering reasoning does hold — `After=givc-net-vm.service network.target` means the
`ExecStop` runs before both, and NetworkManager is still alive because it is `Before=network.target`.
But `network.target` is a passive sync point carrying no guarantee about the units that actually own
the link, so this happens to work rather than being expressed. And a failing `ExecStop` on a
`Type=oneshot` does not block shutdown: the guest powers off anyway, DMA un-quiesced, with one
journal line.

**Required:** make the ordering explicit against the link owner, and make failure host-visible.

---

## Licensing and blast radius

### 5. `crosvm-command.nix` and `crosvm-runner.nix` are undisclosed forks of MIT-licensed microvm.nix

`modules/microvm/common/crosvm-command.nix:1-2`, `crosvm-runner.nix:1-2`

Diffed against the pinned input (`microvm` rev `71beea00`, `lib/runners/crosvm.nix`, MIT,
"Copyright (c) 2021 Astro, and contributors"). `crosvm-command.nix` is that file, nixfmt-reformatted,
plus Ghaf additions — roughly 80% copied. Verbatim: the whole `let` block, `gpuParams`, `preStart`
(down to the missing space before `\` in `--wayland-sock …$WAYLAND_DISPLAY\`), the five-branch
`throw` chain with identical strings, the entire argv builder including
`"Readonly 9p share is not supported"` and `"Unsupported interface type ${type} for crosvm"`,
`canShutdown`, `setBalloonScript` (byte-identical), `requiresMacvtapAsFds`. Ghaf-original content is
`formatAddress`, `--mem size=,base=`, `--platform-mmio`, `--device-tree-overlay`, the platform VFIO
branch, and the shutdown fallback.

`crosvm-runner.nix` is a reduced fork of upstream `lib/runner.nix` — the `runScript` body is
recognisably upstream's.

Both carry only `SPDX-FileCopyrightText: 2026 TII (SSRC)` + `Apache-2.0`. `reuse lint` passes
because it checks for the *presence* of a header, not its correctness.

**Required:** add `# SPDX-FileCopyrightText: 2021 Astro and microvm.nix contributors`, set
`SPDX-License-Identifier: MIT AND Apache-2.0`, and add a header comment naming the upstream file and
the rev it was forked from — that comment is also the rebase anchor for the next microvm bump.

### 6. The fork ships to every x86 laptop, undisclosed and untested

`modules/microvm/flake-module.nix:10-27`

`crosvm-platform.nix` is now folded into `microvm-nix`, and it unconditionally does
`microvm.runner.crosvm = lib.mkForce crosvmRunner` for any guest with `hypervisor == "crosvm"`.
`microvm-nix` is imported by `modules/profiles/laptop-x86.nix` (7 sites),
`modules/profiles/vm.nix` (4 sites) and `targets/generic-x86_64/flake-module.nix:73` — and per this
PR's own docs change, every generic `intel-laptop` target already selects crosvm for all system VMs.

So every x86 crosvm VM moves off upstream's runner onto this fork. All validation was Orin hardware,
and neither the title, the commit messages, nor the docs change mention x86.

Practically it is mostly contained — `-m N` → `--mem N` is the same argh option
(`src/crosvm/cmdline.rs:1424`, `short = 'm'`), the `--vfio` quoting change is shell-equivalent, and
on x86 `power.nix` overrides `ExecStop` anyway. But one real regression remains: a *transient*
`powerbtn` failure (socket not ready, VM already exiting) used to just fail and let systemd wait out
`TimeoutSec`; now it immediately hard-stops the VM.

**Required:** scope `crosvm-platform.nix` to the guests that need it, or gate the `stop` fallback.
Also drop the `microvm-guest` alias — it is bound to the same value as `microvm-nix`, so the comment
"Preserve the existing name while exposing a target-neutral guest module" describes a distinction the
code doesn't make.

---

## Isolation

### 7. The BPMP allow-list is a global union, so net-vm can reset the GPU

`mgbe0-net-vm/default.nix:399-402` gives net-vm `--nvidia-bpmp-host /dev/bpmp-host`;
`payload/guest-module.nix:246-253` gives gpu-vm/gui-vm/disp-vm the same (plus `--nvidia-dce-host`).

The VMM does no MRQ filtering — `devices/src/nvidia_bpmp.rs:123-157` copies the guest MRQ number and
TX buffer straight into `write(2)`. The only policy is the host driver, and the allow-list is a
single global option unioned across every enabled passthrough
(`virtualization/host/bpmp-virt-host/default.nix:69-86`, `apply = lib.unique`, rendered into one
`bpmp_host_proxy` DT node). AGX enables both consumers at once (`agx/orin-agx.nix:46` + `:51-52`).

So a compromised net-vm can send `MRQ_RESET` with `TEGRA234_RESET_GPU`, or `MRQ_PG` on the GPU power
domain, or `MRQ_CLK CMD_CLK_DISABLE` on any display/media clock — `check_if_allowed()`
(`bpmp-host-proxy.c:371-425`) accepts them because they are in the union. Separately,
`bpmp-host-proxy.c:337-348` allows `MRQ_DEBUG`, `MRQ_STRAP`, `MRQ_BWMGR`, `MRQ_ISO_CLIENT` and
`MRQ_EMC_DVFS_*` unconditionally for every guest; `MRQ_DEBUG` is BPMP's debugfs channel.

This is pre-existing in design — under QEMU net-vm already got `ghaf-qemu-bpmp` against the same
device — so it is not introduced here. But this PR makes it the crosvm-wide contract, which makes it
the right moment to fix.

**Required:** one `/dev/bpmp-host-<vm>` per consumer with its own allow-list, and drop `MRQ_DEBUG`
from the unconditional set.

### 8. How do the passthrough engines actually reach the `map-early` carveouts?

All five RAM/scanout backers are `mapEarly = true` (`payload/default.nix:134-144`). In crosvm,
`region_mmap_early` (`devices/src/platform/vfio_platform.rs:239-273`) only calls
`vm.add_memory_region` — it does **not** call `vfio_dma_map`, which the lazy path at `:310-350` does.
And the NoIommu static container map runs once, `if self.groups.is_empty()`, over
`vm.get_memory().regions()` (`devices/src/vfio.rs:637-650`) — i.e. before those memslots exist, and
`add_memory_region` doesn't mutate `GuestMemory`.

So `vm_hs_p@0x60000000`, `vm_cma_p@0x80000000`, `scanout_p@0xb0000000`, `dispram_lo_p@0xb8000000`
and `dispram_hi_p@0x200000000` appear CPU-visible to the guest but absent from the IOMMU domain,
which contains only the guest RAM at `0x2000000000`.

Since GPU/EGL demonstrably works on hardware, *something* maps them. If the answer is SMMU bypass
(`bpmp-virt-common/default.nix:145` sets `arm-smmu.disable_bypass=0`) plus the fixed 1:1 GPA=HPA
carveouts, then gpu-vm/gui-vm DMA is not confined by the SMMU at all and a compromised guest can
reach arbitrary host physical memory.

**Required:** state which mechanism it is, and add a negative test — program a passthrough engine to
DMA at an unmapped host PA and confirm a fault.

For the record, `iommu = "off"` on the *ordinary* guest RAM is fine and the comments describing it
are accurate: NoIommu statically maps the guest's own regions into a Type1v2 container, so the host
boundary holds; what's lost is only intra-guest isolation.

### 9. No assertion makes `gui_vm` mutually exclusive with `gpu_vm`/`disp_vm`

Exclusivity exists only as `lib.mkForce false` in the two accelerated target stanzas. A downstream
config enabling `gui_vm` without them has two VMs claiming `scanout_p` and the three display
keyholes, and `bindGuiVm`/`bindDispVm` both writing `driver_override`. It fails closed (the second
VFIO group open is `EBUSY`) but silently and at runtime. Add an eval-time assertion.

### 10. disp-vm consumes BPMP/DCE without declaring the prerequisites

`disp-vm/default.nix:164-167` adds `ExecStartPre` checks for `/dev/bpmp-host` and `/dev/dce-host`
and `guest-module.nix:246-253` passes both to crosvm, but `disp-vm/default.nix` never sets
`ghaf.hardware.nvidia.virtualization.host.bpmp.enable`, never contributes to the allow-list, and
lacks the `bpmpAllowAllDomains` assertion that `gpu-vm/default.nix:83-88` and
`gui-vm/default.nix:64-69` carry. It works only because `gpu_vm` is co-enabled on every real target.
Make the coupling explicit.

### 11. The `ExecStartPre` permission checks can never fail

`gui-vm/default.nix:119-122`, `gpu-vm:136-138`, `disp-vm:164-167` run
`test -r && test -w /dev/bpmp-host` as root against a `0660 root:kvm` node. Root always passes; the
check only catches a missing node. Simplify to `test -e` (and drop the `bash -c` wrappers) so it
doesn't read as a permission guard it isn't.

### 12. Unnecessary `+` privilege prefix on ExecStop

`mgbe0-net-vm/default.nix:263-266`. `microvm@` carries a hardening profile
(`hardened-configs/microvm.nix`: `NoNewPrivileges`, `PrivateUsers`, trimmed `CapabilityBoundingSet`,
`SystemCallFilter`). `+` drops all of it for a command that only reads `$MAINPID`, execs `givc-cli`
and `kill -0`s.

---

## Correctness and structure

### 13. Two sources of truth for "is this VM on crosvm", duplicated five times

Host-side modules compute `isCrosvm` from `ghaf.virtualization.vmConfig.sysvms.<vm>.vmm or null`
falling back to `defaultSysVmVmm` — open-coded in `gui-vm/default.nix:12-18`,
`gpu-vm/default.nix:12-18`, `disp-vm/default.nix:12-18`, `mgbe0-net-vm/default.nix:24-30` (with the
branch inverted, for no reason) and `targets/nvidia-jetson-orin/flake-module.nix:203-208`. Guest-side
modules key off `config.microvm.hypervisor` instead (`payload/guest-module.nix:19` and every
`mkIf (config.microvm.hypervisor == "crosvm")` in the netvm extraModules).

`lib/global-config.nix:905-909` already computes this inside `applyVmConfig`.

Setting `microvm.hypervisor` directly in a VM `extraModule` desynchronises the two: net-vm would get
crosvm devices and overlays while the host never runs `prepareMgbe0CrosvmOverlay` and never installs
the GIVC `ExecStop` — a VM that fails to start, plus the ungraceful teardown. Nothing asserts they
agree.

**Required:** export `lib.ghaf.vm.sysVmVmm { config, vmName; }`, use it in all five places, and add
an assertion tying host-side `isCrosvm` to guest-side `microvm.hypervisor`.

### 14. The Logitech Unifying Receiver deny rule is dropped under crosvm, unvalidated

`targets/nvidia-jetson-orin/flake-module.nix` — `deny = lib.optionals (!crosvm) [ … ]`

The original comment gives a hardware reason: "evdev-only on Orin (usb-host interrupt-IN broken)".
The new comment asserts crosvm's device-manager attach path handles it — but "Remaining Runtime
Coverage" says physical desktop and input interaction were not exercised. This is a keyboard/mouse
regression risk on the accelerated targets gated on an untested claim. Test it, or keep the deny rule
for now.

### 15. Split targets change more than the description says

`orinCrosvmModule` sets `defaultSysVmVmm` **and** `defaultAppVmVmm` to crosvm on the split AGX/NX
debug targets. The description and the split-target acceptance cover AdminVM, NetVM, GPUVM and
DispVM — but GUIVM and every AppVM on those targets also switch VMM, with no stated validation.
Confirm intent and exercise them, or scope the defaults.

### 16. Magic constants duplicated 3–5 ways

- MGBE0 BPMP ids (`clocks 357 361 … 248`, `resets 46 45 47`, `power-domains 18`) appear in
  `mgbe0-crosvm-overlay.dts:17-24`, the build-time check (`mgbe0-net-vm/default.nix:89-91`) and the
  runtime check (`:139-141`) — the latter two with *two separate implementations* of
  `check_bpmp_ids` (POSIX `set --` shifting vs bash arrays). The checks exist because these ids
  drift, but nothing stops the `.dts` drifting away from the expectation strings, at which point both
  checks keep passing while the overlay is wrong.
- The layout constants `0x2000000000` / `0x60000000` / `0x1fa0000000` appear in
  `payload/guest-module.nix:238-242`, are re-asserted in `tests/orin-crosvm-targets.nix:85-88` and
  `tests/crosvm-platform.nix:56-60`, must agree with hand-written `reg` values in
  `tegra234-gpuvm-memory.dtsi` / `tegra234-dispvm-memory.dtsi`, and are re-asserted again in
  `payload/crosvm-overlay.nix:69-84`.

`payload/default.nix`'s `reservedMem` is the natural single source; at minimum derive `platformMmio`
from it, and define the BPMP id lists once in Nix.

### 17. The `bus` enum union is undocumented

`crosvm-platform.nix:43-45` redeclares `bus = enum [ "platform" ]`, relying on `types.enum`'s union
merge with upstream's `enum [ "pci" "usb" ]`. It works, but it is non-obvious module-system behaviour
with no comment, and it supplies no `description`, so the rendered docs keep upstream's "either the
`pci` or the `usb` bus" for an option that now takes three.

### 18. `share/microvm/platform-devices` has no reader

`crosvm-runner.nix:73-77`. Nothing in the repo consumes it. Note also that upstream `runner.nix`
already emits `share/microvm/${bus}-devices` for every device — this code filters platform devices
*out* of `upstreamConfig` and then re-adds the file by hand. Wire it up or delete it.

### 19. `gpu-vm/default.nix:135` deletes `environment = mkIf payload.needsDceBridge {...}` outright

rather than gating on `!isCrosvm` like gui-vm and disp-vm. Verified a no-op today
(`capabilities.gpuvm.display = false`), but it turns a guarded expression into an unguarded
assumption: if gpuvm ever gains display, the QEMU path silently loses `GHAF_DCE_GUEST`.

### 20. `orinCrosvmModule` puts hardware policy in the target file

Backend selection over `vhotplug.vms`, an overlay injection, VMM defaults, and (via
`acceleratedGuivmCrosvmModule`) a USB deny-list derived from a hardware quirk. Not a hard AGENTS.md
violation (the data-only rule is scoped to `targets/laptop/machines.nix`), but the same smell — the
Unifying-receiver logic especially is hardware quirk handling living in a target table. Move it to
`modules/reference/hardware/jetpack/…` with the target flipping an enable.

### 21. hwinfo error handling

`hwinfo/guest.nix:50` — `jq . < "$F" || cat "$F"` silently prints raw bytes and exits 0 on a corrupt
file. (`jq` *is* in `runtimeInputs` — that part is fine.) Also, on crosvm the fall-through still
prints *"fw_cfg kernel module not loaded. Try: sudo modprobe qemu_fw_cfg"*, which is misleading for a
VM that will never have fw_cfg.

### 22. `lib.replaceStrings [ "/run" ] [ "/etc" ]`

`mgbe0-net-vm/default.nix:47` — verified safe today (the only `/run` occurrences in `cliArgs` are the
three TLS PEM paths) and it matches three existing call sites, but the repo spells those
`lib.replaceString`. This is the fourth open-coded copy of an unguarded whole-string rewrite; a future
arg containing `/run` would be silently mangled, and the failure mode is the 30 s hang from finding 2.
Extract a `lib.ghaf` helper.

### 23. Non-atomic overlay write

`prepareMgbe0Overlay` does `install` then in-place `fdtput` on `/run/mgbe0-net-vm.dtbo`. The ordering
makes it safe in practice; write-to-temp + `mv` is free hardening.

---

## Licence headers and formatting

### 24. SPDX years

All nine new files use `2026 TII (SSRC)`; the repo has 481 files at `2022-2026` versus 17 at bare
`2026`, and AGENTS.md's template shows `2022-2026`.

### 25. `mgbe0-crosvm-overlay.dts` is Apache-2.0 while the two sibling new `.dts` files are GPL-2.0-only

The GPL choice on the gpuvm/dispvm overlays is right — they include NVIDIA-derived `.dtsi`. The
MGBE0 node content (property names, `nvidia,*` values, the stmmac axi-config) is just as clearly
transcribed from NVIDIA's `tegra234-p3701.dtsi`. Pick one story.

### 26. Comment and formatting nits

- Stray blank line after the opening `[` of `acceleratedGuivmUsbRules`.
- `flake-module.nix:159-162`: the comment is broken mid-clause; reflow.
- `mgbe0-net-vm/default.nix:41-44` duplicates what `crosvm-command.nix` already handles generically.
  It should say why the generic fallback isn't good enough here — it kills DMA before the driver
  quiesces — which is the real reason.
- The four `.dtsi` files split brace pairs across `#ifdef GHAF_CROSVM_OVERLAY` arms. I traced all
  four and the balance is correct, but each wants a one-line header saying so before someone edits
  one arm in isolation.

---

## Tests

`tests/crosvm-platform.nix` and `tests/orin-crosvm-targets.nix` are good work and do cover what the
description claims — VMM selection per VM, device-manager backend selection, platform-device order,
layout, the NX PCI contract, hwinfo virtiofs-vs-fw_cfg, the negative assertion cases, and the aarch64
cross build. The gaps line up with the findings above.

### 27. The shutdown fallback's semantics are untested

`tests/crosvm-platform.nix:155-156,173-174` only greps that both `powerbtn` and `stop` appear in the
script. That test passes identically for the broken behaviour in finding 1. Encode that on aarch64
`powerbtn` is never available.

### 28. The entire MGBE0 quiesce path is untested

Nothing asserts AGX net-vm's `microvm.devices` (the MGBE0 platform entry), its
`deviceTreeOverlays == ["/run/mgbe0-net-vm.dtbo"]`, the existence of `prepareMgbe0CrosvmOverlay`, the
`ExecStop`/`TimeoutStopSec` override, or the guest `quiesce-mgbe0` unit. The `c0000000` ↔ `mem`
coupling (finding 3) is exactly what an eval test could pin.

### 29. Rollback breadth

`nxQemuFallback` flips only `sysvms.netvm.vmm` on NX — the PCI path. There is no rollback test for
AGX net-vm (MGBE0), gui-vm, gpu-vm or disp-vm, i.e. for every configuration where the crosvm-only
options actually live (`crosvmOverlay` assertion, `GHAF_DCE_GUEST`, `-dtb` + `vfioArgs`).

### 30. No App VM coverage

`defaultAppVmVmm = "crosvm"` is set for these targets but no assertion checks a resulting appvm
hypervisor.

### 31. The disjointness test doesn't compute disjointness

`tests/orin-crosvm-targets.nix:307-322` compares `map (d: d.path)` against hard-coded lists and checks
the layout equals constants; nothing computes pairwise non-overlap of `{mmioBase, size}` across VMs.
The layout *is* disjoint today — I checked — but derive the expected lists from `mkPayload` and add a
computed overlap check so a future base-address edit cannot silently collide.

### 32. `tests/crosvm-platform.nix:142-146,171-172` smuggles in an unrelated cross-build

It instantiates a second nixpkgs to cross-build `ghaf-device-manager` and `file | grep aarch64`. That
belongs with the `overlays/flake-module.nix` change it actually tests, under its own check name — as
written, a device-manager regression reports as "crosvm-platform failed", and every run of the crosvm
check pays for a cross toolchain instantiation.

vadika:
Addressed in `1c271cf50`. Numbered mapping to the requested-changes review:

1. Removed the Orin `powerbtn -> crosvm stop` path. A root-owned companion requests guest shutdown through GIVC and waits on the actual Crosvm PID with warning/deadline handling.
2. GIVC calls are synchronous under `timeout`; their status is logged and retried while the same VMM remains alive. Deadline expiry is an explicit host failure.
3. MGBE quiesce now walks `/sys/bus/platform/drivers/tegra-mgbe/*/net` rather than deriving a guest address from RAM size.
4. Ordering is explicit against `NetworkManager.service`; `ghaf-mgbe0-poweroff.service` powers off only after quiesce succeeds, so failure remains host-visible.
5. Deleted the copied `crosvm-command.nix`. The adapter now calls microvm.nix `lib.buildRunner` and changes only missing AArch64 arguments.
6. The adapter is selected only when Ghaf extensions are present; generic x86 guests retain the upstream runner. Removed the duplicate `microvm-guest` alias.
7. BPMP proxy instances and allow-lists are per VM; NetVM cannot inherit GPU/display reset policy. Removed unconditional `MRQ_DEBUG` access.
8. Pinned Crosvm `cf0d0855`, whose early mapping calls `vfio_dma_map()` before adding each KVM memslot. A focused source regression check pins that invariant. The hostile arbitrary-host-PA engine test remains unrun because it requires separate destructive/security-test authorization; normal GPU/display DMA passed on AGX and NX.
9. Added/retained an eval assertion that combined GUIVM cannot coexist with GPUVM or DispVM.
10. DispVM now enables BPMP, declares its own display policy, checks the closed-policy prerequisite, and owns dedicated BPMP/DCE paths.
11. Replaced root `test -r/-w` wrappers with direct `test -e` prerequisites.
12. Removed the privileged `+` ExecStop. The companion retains the hardening profile and only `CAP_KILL`.
13. Added `lib.ghaf.vm.sysVmVmm`, used it at every host decision point, and asserted the selected VMM matches the guest hypervisor.
14. Kept the Logitech Unifying receiver deny rule and proven evdev route.
15. AppVM Crosvm selection is intentional for these targets and is now asserted for ChromiumVM and FlatpakVM on AGX and NX.
16. Centralized the Crosvm layout/reserved-memory data and MGBE BPMP ID lists; generated overlay values and assertions consume those definitions.
17. Removed the upstream `microvm.devices.bus` enum extension entirely; platform devices now have their own typed option.
18. Removed the unused `share/microvm/platform-devices` output.
19. Restored `GHAF_DCE_GUEST` on the QEMU GPUVM path when display capability is enabled.
20. Moved Orin backend/USB/overlay policy into `virtualization/crosvm-target.nix`; targets only enable it.
21. Corrupt hwinfo is now an error; Crosvm missing-file diagnostics no longer suggest `qemu_fw_cfg`.
22. Added `lib.ghaf.givc.hostCliArgs` and removed the local rewrite copy.
23. MGBE overlay generation now writes a temporary file and atomically renames it.
24. Normalized new-file SPDX years to `2022-2026`.
25. Changed the NVIDIA-derived MGBE overlay to `GPL-2.0-only`.
26. Fixed the listed formatting/comments and documented the intentionally split preprocessor braces.
27. The runner check rejects any `crosvm stop`; target tests inspect the GIVC shutdown scripts and service choice.
28. Added AGX MGBE platform-device, overlay, preparation, companion, timeout, NetworkManager, and guest-quiesce assertions.
29. Added QEMU rollback coverage for NX PCI NetVM and AGX MGBE0, GUIVM, GPUVM, and DispVM, including vhotplug restoration.
30. Added explicit AGX/NX AppVM hypervisor assertions.
31. Expected mappings come from `mkPayload`; tests compute pairwise GPUVM/DispVM range disjointness.
32. Moved the AArch64 `ghaf-device-manager` build to `checks.x86_64-linux.ghaf-device-manager-cross`.

Final split images were flashed through non-secure paths (`Boot Authentication: NS`) to both boards. AGX and NX passed clean DispVM, GPUVM, and NetVM restart tests; CUDART passed before/after GPUVM restart, DRM/DCE recovered after DispVM restart, MGBE/LAN recovered after AGX NetVM restart, and neither board logged an SMMU/context fault. Full details and exact identities are in the updated PR description.

vadika:
Follow-up in `6aeaf51d8` addresses the GUIVM regressions found during accelerated-target hardware testing:

- Finding 14 is now hardware-tested instead of retaining the old deny rule. With Crosvm GUIVM, `046d:c52b` is routed through USB by `ghaf-device-manager`; QEMU rollback retains the evdev deny. On AGX the manager attached host path `1-4.2` to Crosvm port 1, GUIVM enumerated the Logitech K400 Plus with keyboard/mouse handlers, and the receiver reattached after GUIVM PID `6064 -> 26363` under a new socket generation.
- Display-owning combined GUIVM now uses the same DCE RM deinit ordering as DispVM. The restarted AGX GUIVM recovered both DRM cards and connected DP without an NvKMS/DCE initialization failure.
- AArch64 shutdown uses a guest GIVC wrapper for display/GPU guests, preserves mandatory MGBE quiesce, uses the bounded upstream-helper/SIGTERM path for guests without that capability, preserves the activation guard, and never emits `crosvm stop`. Generic GIVC shutdown requests are not retried after the guest disconnects while powering off.

The NX accelerated GUI image was also flashed to ECID `0x80012344705DD5C58C00000011FF8100` through the verified non-secure path (`Boot Authentication: NS`, disk encryption disabled, `--key None`, zero SBK). It did not return at `.127` after coldboot/reset, so that accelerated-GUI boot regression is recorded as open; the final NX split-target acceptance remains unchanged.

milva-unikie:
I checked the `jenkins-pre-merge` results. Both AGX and NX have had cases where disp-vm or gpu-vm did not boot.
```
Aug 24 04:28:25 ghaf-host microvm@gpu-vm[8687]: [2026-08-24T04:28:25.159225820+00:00 ERROR crosvm] exiting with error 1: the architecture failed to build the vm
Aug 24 04:28:25 ghaf-host microvm@gpu-vm[8687]: Caused by:
Aug 24 04:28:25 ghaf-host microvm@gpu-vm[8687]:     failed to create platform bus: failed to map VFIO platform memory: failed to map VFIO DMA for region 0 at 0x80000000, host 0xf09064000000
Aug 24 04:28:25 ghaf-host crosvm[8687]:  [src/main.rs:940] exiting with error 1: the architecture failed to build the vm
                                        
                                        Caused by:
                                            failed to create platform bus: failed to map VFIO platform memory: failed to map VFIO DMA for region 0 at 0x80000000, host 0xf09064000000
Aug 24 04:28:25 ghaf-host kernel: vfio-platform 60000000.vm_hs_p: VFIO: no reset function found!
Aug 24 04:28:25 ghaf-host systemd-networkd[1301]: tap-gpu-vm: Lost carrier
Aug 24 04:28:25 ghaf-host kernel: virbr0: port 3(tap-gpu-vm) entered disabled state
Aug 24 04:28:27 ghaf-host systemd[1]: microvm@gpu-vm.service: Main process exited, code=exited, status=1/FAILURE
Aug 24 04:28:27 ghaf-host systemd[1]: microvm@gpu-vm.service: Failed with result 'exit-code'.
Aug 24 04:28:27 ghaf-host systemd[1]: microvm@gpu-vm.service: Consumed 5.377s CPU time over 5.569s wall clock time, 5.8G memory peak.
```
[host.log](https://github.com/user-attachments/files/31362975/host.log)

This has happened roughly in one out of four test runs.

vadika:
So have you did the fll reflash with   QSPi?

milva-unikie:
> So have you did the fll reflash with QSPi?

No, I only checked the results from the automated tests. I can not find a mention about QSPI flash in the PR description.

vadika:
> > So have you did the fll reflash with QSPi?
> 
> No, I only checked the results from the automated tests. I can not find a mention about QSPI flash in the PR description.

Needs proper testing then... (

milva-unikie:
> > > So have you did the fll reflash with QSPi?
> > 
> > 
> > No, I only checked the results from the automated tests. I can not find a mention about QSPI flash in the PR description.
> 
> Needs proper testing then... (

So just to confirm, the QSPI needs to be flashed? That should be mentioned in the PR description. The description is missing the entire `Testing Instructions` section. Please add it with proper, human-readable testing steps.

leivos-unikie:
Test results on Orin AGX
- flashed qspi
- booted cross-compiled ghaf image from USB SSD
- Verified that all VMs are running with crossvm
- Could not reproduce the issue with gpu-vm not booting in pre-merge run
- Some mismatch in vm versions:
```
[ghaf@ghaf-host:~]$ microvm -l 
admin-vm: current(nixos-system-admin-vm-26.11.20260819.ffb3c9b) 
disp-vm: outdated(microvm-crosvm-disp-vm/share/microvm/system), rebuild(nixos-system-disp-vm-26.11.20260819.ffb3c9b) and reboot: microvm -Ru disp-vm 
gpu-vm: outdated(microvm-crosvm-gpu-vm/share/microvm/system), rebuild(nixos-system-gpu-vm-26.11.20260819.ffb3c9b) and reboot: microvm -Ru gpu-vm 
net-vm: outdated(microvm-crosvm-net-vm/share/microvm/system), rebuild(nixos-system-net-vm-26.11.20260819.ffb3c9b) and reboot: microvm -Ru net-vm
```
- According to Vadim this is a known issue: `ghaf-load-ftpm-module.service` fails (failed Load fTPM module after stage-2 OP-TEE readiness)

- Shutdown gets often stuck at
`A stop job is running for Gracefully shut down Crossvm guest net-vm / disp-vm / gpu-vm`
--> Very long shutdown time: ~100 sec

vadika:
> > > > So have you did the fll reflash with QSPi?
> > > 
> > > 
> > > No, I only checked the results from the automated tests. I can not find a mention about QSPI flash in the PR description.
> > 
> > 
> > Needs proper testing then... (
> 
> So just to confirm, the QSPI needs to be flashed? That should be mentioned in the PR description. The description is missing the entire `Testing Instructions` section. Please add it with proper, human-readable testing steps.

fixed it, my bad

vadika:
Rebased onto current `main` (`59ff1fb44`) at PR head `1b7943486`.

The only content conflict was the Crosvm input. Main now uses merge `fba7ba3a7`, which contains this PRs `cf0d0855` early-DMA commit and has an identical source tree, so the rebase keeps mains revision and drops the redundant PR pin. Range-diff otherwise preserves all 11 PR commits patch-for-patch.

I also added the requested human-readable `Testing Instructions` section: full QSPI plus storage flash, non-secure markers, `microvm -l` freshness, at least four GPUVM/DispVM startup cycles for the reported intermittent VFIO DMA-map failure, GPU/display checks, and shutdown timing/journal acceptance.

The intermittent Jenkins DMA-map failure and approximately 100-second shutdown stalls remain runtime findings to reproduce after the full QSPI flash; the rebase and local evaluation checks alone do not mark them fixed.

leivos-unikie:
Reproduced the long shutdown issue (this time 130 sec) on Orin AGX flashed with flash-script (qspi, emmc), before that last rebase.

lib/global-config.nix
```diff
@@ -733,7 +733,9 @@ rec {
   #   };
   #
   # ═══════════════════════════════════════════════════════════════════════════
-  vm = {
+  givc.hostCliArgs = cliArgs: lib.replaceString "/run" "/etc" cliArgs;
+
+  vm = rec {
```
brianmcgillion: do not use rec

modules/microvm/common/crosvm-runner.nix
```diff
@@ -0,0 +1,101 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Thin adapter around microvm.nix's lib.buildRunner. Keep the upstream runner
+# authoritative; this file supplies only Crosvm arguments that microvm.nix does
+# not model yet for AArch64 platform passthrough.
```
brianmcgillion: put in the PR to microvm and list this as being removable when it lands

modules/givc/common.nix
```diff
@@ -171,17 +175,35 @@ in
...
-      sysvm = mkIf config.givc.sysvm.enable {
-        accessControl = {
-          enable = cfg.accessControl.enable;
-          agentRules = [
-            {
-              permittedVms = [ cfg.adminConfig.name ];
-              permittedModules = [ "stats" ];
-            }
+      sysvm = lib.mkMerge [
+        (mkIf cfg.minimalSysvm.enable {
```
brianmcgillion: how is this better thatn a sysvm?

modules/profiles/orin.nix
```diff
@@ -137,7 +137,7 @@ in
         # Export Net VM base for profiles to extend
         orin.netvmBase = lib.nixosSystem {
           modules = [
-            inputs.microvm.nixosModules.microvm
+            inputs.self.nixosModules.microvm-nix
```
brianmcgillion, line 159: why self ?

brianmcgillion (changes requested):

lib/global-config.nix
vadika: Removed in the rewrite. The final implementation no longer changes `lib/global-config.nix` or introduces that `rec` binding.

modules/microvm/common/crosvm-runner.nix
vadika: The old generic runner proposal is gone. The final adapter calls microvm.nix `lib.buildRunner` and adds only the Ghaf-specific memory-base and PCI override arguments, so we no longer carry a copied upstream implementation.

modules/givc/common.nix
vadika: Removed in the rewrite. There is no new generic minimal-GIVC role; split payload guests use the existing `givc.sysvm` role from the shared Orin guest module.

modules/profiles/orin.nix
vadika: `self` selects Ghaf’s `microvm-nix` wrapper, which imports upstream microvm.nix plus Ghaf’s Crosvm platform adapter and store-disk overlay. Direct `inputs.microvm.nixosModules.microvm` would omit the Orin platform support. Added a scoped comment in `9d92a9677`.

vadika:
Rewritten and force-pushed with an exact lease: `9d92a9677 -> cb744257f`.

The rewrite is based on the dependency-cleanup commit from #2171. A local backup ref preserves the old head, and the range-diff accounts for every content commit:

- all 11 PR2133 feature commits are preserved in order;
- the only post-build patch-context change refreshes the temporary USB logging patch from rebased ghafpkgs #362;
- the fixup was autosquashed into the original microvm integration commit;
- the final tree was proven identical before and after autosquash.

Large copied sources and packages are now split into their owning repositories: `tiiuae/jetpack-nixos` #21/#22, `tiiuae/ghafpkgs` #362/#368/#369, and official `microvm.nix` #586. Jetpack is explicitly pinned through the TII UAE fork. No new ghaf-crosvm patch remains because current Ghaf main already contains the required early VFIO DMA revision. #2144 was not touched.

Validation on `cb744257f`:

- formatting, REUSE, and diff checks pass;
- focused Crosvm platform and Orin target checks pass;
- AGX combined GUIVM image builds;
- NX combined GUIVM image builds;
- AGX split GPUVM/DispVM image builds;
- NX split GPUVM/DispVM image builds.

The description now includes dependency ownership, human-readable full-QSPI testing instructions, TOPO serial access, `.127`, four-cycle DMA-map regression coverage, runner freshness, shutdown timing, and GUI input acceptance. Current-head full-QSPI hardware testing is still in progress, so the reported intermittent DMA-map and long-shutdown findings remain open until that evidence is posted.

vadika:
Exact-head hardware update for `37c47aa7`:

- Built all four AGX/NX split/combined images.
- Full non-secure QSPI + eMMC flash passed on AGX (`Boot Authentication: NS`; no secure-boot target/key material).
- All four VM runners are current and use Crosvm; `ghaf-device-manager` is active; `vhotplug` is absent.
- The rapid second GPUVM/DispVM restart reproduced GIVC unavailability and exercised the direct fallback. After the 20 s grace period, `crosvm --no-syslog stop` reported `exiting with success`; the Display VM shutdown unit completed in 27.486 s.
- Both VMs returned active with fresh Crosvm PIDs. No permission error, stop failure, zombie, orphaned vCPU, VFIO DMA mapping error, SMMU fault, or context fault was observed.
- The only failed unit remains the separate fTPM issue tracked in #2167.

The PR description now carries the full evidence and remaining exact-head NX/combined-GUI hardware gaps.

vadika:
Repository-boundary refactor is complete at `6633dae73`:

- Generic Crosvm AArch64 memory/platform layout support is now in [microvm.nix #586](https://github.com/microvm-nix/microvm.nix/pull/586). Ghaf's copied runner, option module, and duplicate test are deleted.
- Orin GPU/display guest and host composition, ownership assertions, split/combined targets, payloads, and regression test are now in [ghaf-gpu-partitioning-examples #2](https://github.com/tiiuae/ghaf-gpu-partitioning-examples/pull/2). That PR's CI is green, and all four AGX/NX split/combined images build.
- Ghaf keeps only generic Crosvm/device-manager selection, shutdown contracts, AGX MGBE0 and NX PCI NetVM integration, hardware-info sharing, and the shared Jetson host mechanisms.
- The three Ghaf GPU/display implementation guides were reduced from 865 lines to 52 compatibility/pointer lines, preserving their existing documentation routes without duplicating downstream instructions.

Current-head Ghaf validation passes:

```console
nix fmt -- --fail-on-change
nix develop --command reuse lint
git diff --check
nix eval --raw .#nixosConfigurations.nvidia-jetson-orin-agx-debug-from-x86_64.config.system.build.toplevel.drvPath
nix eval --raw .#nixosConfigurations.nvidia-jetson-orin-nx-debug-from-x86_64.config.system.build.toplevel.drvPath
timeout 2h nix build .#doc --no-link
```

The dependency direction is one-way: the examples PR extends this Ghaf PR; Ghaf never imports the examples repository. Prior hardware results are retained as historical evidence only; this repository-boundary head has not been reflashed.

vadika:
Repository-boundary follow-up is published. jetpack-nixos PR22 at c00cccf now owns BPMP/DCE host mechanics plus MGBE0 binding, overlay preparation, guest kernel support, and quiescing. Ghaf head 842353825 removes another 136 lines from the MGBE0 module and retains only NetVM/VMM/GIVC integration. Examples follow-up PR3 at 8a4e3fc explicitly consumes that boundary. AGX/NX evaluations, focused examples checks, REUSE/format checks, the quiesce helper build, and a full AGX cross image build pass. No device was flashed.

vadika:
Rebased and revalidated at `4b9bea29547b459b7af57b6a2aff19630340ad59`.

The PR description now includes human-readable flash and regression-test instructions, the current store outputs, and the AGX/NX results. In particular:

- both boards received full unsigned flashes (`Boot Authentication: NS`);
- COSMIC runs in GUI VM and is absent from the host;
- GUI VM returned after ready/immediate restart pairs in 27/47 seconds on AGX and approximately 23/33 seconds on NX;
- neither measured cycle reached the last-resort MicroVM shutdown helper;
- no DCE, SMMU, soft-lockup, or VFIO DMA errors appeared after the clean power cycle; and
- the AGX Logitech receiver remained physically visible on the host while its keyboard/mouse input nodes belonged to GUI VM.

The shutdown regression found during hardware testing is fixed by a GIVC-first retry window plus the corrected upstream AArch64 Crosvm stop helper. The focused Orin target check, formatting, REUSE, and diff checks pass. Please re-review the current head.

leivos-unikie:
All Orin builds, except native orin nx, have failed in pre-merge pipeline.

PR #2138, opened 2026-08-13
fix(cosmic-settings-daemon): restore audio slider after suspend

jkuro-tii:
## Description of Changes

After suspend/resume, the audio volume slider no longer functions correctly. Changing the volume does not update the slider state, and the on-screen volume panel (OSD) is not displayed.
The work around is temporary until the below zlink fix is included in Cosmic.

https://github.com/z-galaxy/zlink/commit/13bad852fa28a977b8eacf56ca3d85a746f15faf

### Type of Change

- [ ] New feature
- [X] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

https://jira.tii.ae/browse/SSRCSP-8740
https://jira.tii.ae/browse/SSRCSP-8507

## Checklist

## Testing Instructions

1. Log in to gui-vm.
2. Suspend the system and then resume it.
3. Log in again after resume.
4. Start audio playback (e.g., music or a video).
5. Adjust the volume using the audio slider in the top-right corner of the screen.
6. Verify that the audio volume changes accordingly.
7. Verify that the volume OSD/panel appears at the bottom of the screen when the volume is adjusted.

Expected Result:

1. Audio volume changes correctly when using the slider.
2. The volume OSD/panel is displayed when the volume level is changed.
3. No loss of audio control functionality after suspend/resume.

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [X] Intel Laptop `x86_64`

### Installation Method

- [X] Can be updated with `nixos-rebuild ... switch`
- [ ] Other:

### DIFF
--- a/modules/common/services/power.nix
+++ b/modules/common/services/power.nix
@@ -699,7 +699,7 @@ in
               systemctl start ${service}
             '') cfg.vm.pciSuspendServices
           )
-          + optionalString (cfg.suspend.extraResumeCommands != "") ''
+          + optionalString (cfg.suspend.extraResumeCommands != "" && !cfg.gui.enable) ''
             # config.ghaf.services.power-manager.suspend.extraResumeCommands
             ${cfg.suspend.extraResumeCommands}
           '';
--- a/modules/desktop/graphics/cosmic/default.nix
+++ b/modules/desktop/graphics/cosmic/default.nix
@@ -439,7 +439,25 @@ in
       # Workaround for https://github.com/pop-os/cosmic-applets/issues/1390
       # TODO: Remove when upstream issue is fixed
       ''
-        ${lib.getExe' pkgs.busybox "killall"} -q cosmic-panel || true
+        pid=$(${lib.getExe' pkgs.busybox "pgrep"} -f "cosmic-panel" | ${lib.getExe' pkgs.busybox "head"} -n1)
+        if [ -n "$pid" ]; then
+          ${lib.getExe' pkgs.busybox "kill"} "$pid" || true
+        else
+          echo "No COSMIC panel PID found"
+        fi
+        #
+        # Workaround for volume slider not working and volume panels not showing after resume
+        # zlink 0.7.0 fixes the COSMIC settings daemon resume issue.
+        # TODO: Remove when zlink ≥ 0.7.0 is used by COSMIC (currently it's 0.5.0)
+        # COSMIC Cargo.lock reference:
+        # https://github.com/pop-os/cosmic-settings/blob/master/Cargo.lock#L9383
+        # Restart COSMIC settings daemon after suspend, as it may not resume properly on some systems
+        pid=$(${lib.getExe' pkgs.busybox "pgrep"} -f "cosmic-settings-daemon" | ${lib.getExe' pkgs.busybox "head"} -n1)
+        if [ -n "$pid" ]; then
+          ${lib.getExe' pkgs.busybox "kill"} -9 "$pid" || true
+        else
+          echo "No COSMIC settings daemon PID found"
+        fi
       '';
 
     systemd.user.services = {


modules/common/services/power.nix
```diff
@@ -692,17 +692,12 @@ in
...
-          optionalString cfg.vm.pciSuspend (
-            concatMapStringsSep "\n" (service: ''
-              echo "Starting service ${service}..."
-              systemctl start ${service}
-            '') cfg.vm.pciSuspendServices
-          )
-          + optionalString (cfg.suspend.extraResumeCommands != "") ''
-            # config.ghaf.services.power-manager.suspend.extraResumeCommands
-            ${cfg.suspend.extraResumeCommands}
-          '';
```
kajusnau, line 705: Why has `extraResumeCommands` been removed from the `vm` power-manager profile?

modules/desktop/graphics/cosmic/default.nix
```diff
@@ -440,6 +440,14 @@ in
...
       ''
         ${lib.getExe' pkgs.busybox "killall"} -q cosmic-panel || true
+        # Restart COSMIC settings daemon after suspend, as it may not resume properly on some systems
+        pid=$(${lib.getExe' pkgs.busybox "pgrep"} -f "cosmic-settings-daemon" | head -n1)
+        if [ -n "$pid" ]; then
+          echo "Restarting COSMIC settings daemon with PID: $pid"
+          ${lib.getExe' pkgs.busybox "kill"} -9 "$pid" || true
+        else
+          echo "No COSMIC settings daemon PID found"
+        fi
```
kajusnau, line 460: Please add a `TODO` style comment similar to line 440, so this can be followed up on in the future, also add a link if possible.

kajusnau: What's the benefit of doing this somewhat busy-looking if/else block instead of just using `killall`, as is it used for `cosmic-panel` above?

e.g. `${lib.getExe' pkgs.busybox "killall"} -q cosmic-settings-daemon || true`

Also, if timing is an issue as you mentioned before, should we consider a 2-3 second sleep before `cosmic-settings-daemon` kill is attempted?

kajusnau (changes requested):

kajusnau:
Additional note: PR title and commit message should reference **cosmic-settings-daemon**, not **cosmic-settings**

modules/common/services/power.nix
jkuro-tii: If not deleted extraResumeCommands are called twice during resume: here and https://github.com/tiiuae/ghaf/blob/7c7d6a3c810c7ca6e2bdbc82f792e979fe14018a/modules/common/services/power.nix#L749.

modules/desktop/graphics/cosmic/default.nix
jkuro-tii: The benefit is that the failing process is found by its executable path, as well as its name. killall doesn't do it. It's more reliable as the process name changed during Ghaf releases.

What do you mean by the timing issue?

jkuro-tii: Done.

modules/common/services/power.nix
kajusnau: Then it should be skipped in gui-vm's case, but not removed entirely, as this removes functionality from other VMs where power-manager is used.

modules/desktop/graphics/cosmic/default.nix
kajusnau: > What do you mean by the timing issue?

I thought I saw one of your comments somewhere mention that `cosmic-settings-daemon` can't be killed too early, maybe I misread

> The benefit is that the failing process is found by its executable path, as well as its name. killall doesn't do it. It's more reliable as the process name changed during Ghaf releases.

Fair enough. In that case, should we unify the approach used to kill the cosmic processes here (do the same for `cosmic-panel`) for consistency?
Also IMO the prints about the daemon being restarted are unnecessary or should be added for the `cosmic-panel` as well.

Sidenote:
I remembered I have this PR open https://github.com/NixOS/nixpkgs/pull/539546 which might be worth incorporating in Ghaf before (or if ever) it goes in upstream.

kajusnau:
> Additional note: PR title and commit message should reference **cosmic-settings-daemon**, not **cosmic-settings**

The commit still references `cosmic-settings` rather than `cosmic-settings-daemon`

modules/common/services/power.nix
jkuro-tii: resumeCommands is executed by the sleep-actions-pre-stop service AFTER resuming from sleep.
The resume-actions service executes resumeActions.
So both - resumeCommands and resumeActions - call extraResumeCommands, so it is called twice during resume, same second.
Thus extraResumeCommands has been removed from resumeCommands.

modules/common/services/power.nix
```diff
@@ -699,10 +699,7 @@ in
               systemctl start ${service}
             '') cfg.vm.pciSuspendServices
           )
-          + optionalString (cfg.suspend.extraResumeCommands != "") ''
-            # config.ghaf.services.power-manager.suspend.extraResumeCommands
-            ${cfg.suspend.extraResumeCommands}
-          '';
+          + optionalString (cfg.suspend.extraResumeCommands != "") "";
```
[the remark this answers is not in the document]
kajusnau: Looks like Copilot agrees with me :grinning:

modules/common/services/power.nix
kajusnau: Looks like Copilot agrees with me https://github.com/tiiuae/ghaf/pull/2138#discussion_r3796014175 :grinning:

modules/common/services/power.nix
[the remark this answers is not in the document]
jkuro-tii: With the fix all non-gui VMs `extraResumeCommands` will be still executed twice, almost the same time, after waking up from suspend.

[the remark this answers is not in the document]
kajusnau: I don't understand why that would be the case

kajusnau (approved):
Looks good to me now!

One final nitpick but not necessary to fix - the commit message is still referencing two different components:
- current - `fix(cosmic-settings-daemon): restart cosmic-settings after resume`
- suggested - `fix(cosmic-settings-daemon): restart cosmic-settings-daemon after resume`
- suggested less verbose - `fix(cosmic-settings-daemon): restart after resume`

milva-unikie:
**Tested on Lenovo X1** (`nixos-rebuild`)

Could not reproduce any issues with audio or audio controls after suspension

brianmcgillion (approved):

PR #2171, opened 2026-08-24
refactor(nvidia): Source virtualization support externally

vadika:
## Summary

- consume the Ghaf QEMU package family from ghafpkgs
- consume Orin BPMP/DCE sources and kernel patches from jetpack-nixos
- remove 9,282 lines of duplicated sources, patches, and package expressions from Ghaf
- retain only the Ghaf module wiring and the local EGL userspace workaround

No runtime behavior change is intended. The upstream extraction PRs contain the Ghaf main versions; the PR2133-only lifecycle changes are kept in separate stacked branches.

## Dependencies

- tiiuae/jetpack-nixos#21
- tiiuae/ghafpkgs#368

While those PRs are open, this draft pins the TII Jetpack extraction branch and the TII ghafpkgs PR ref. Restore the normal input URLs after both merge.

## Verification

- `nix fmt -- --fail-on-change`
- `nix develop --command reuse lint`
- evaluated AGX accelerated GUI-VM cross configuration
- evaluated NX accelerated GUI-VM cross configuration
- compared against the same evaluations from an untouched `origin/main` worktree

The derivation hashes change because the sources move between flake inputs; the target evaluations succeed with the same package versions and module configuration.


vadika:
Fixed the failed Intel laptop build in `7b0d9184c`.

Root cause: the temporary ghafpkgs #362 USB applet patch still targeted an older source layout, so `patchPhase` rejected both files after PR2171 moved the package source to the newer ghafpkgs revision. The patch context is now refreshed directly from the rebased upstream diff.

Validation:

- exact `ghaf-usb-applet` derivation builds;
- `nix build --no-link .#packages.x86_64-linux.intel-laptop-debug` passes end-to-end;
- `nix fmt -- --fail-on-change` passes;
- `nix develop --command reuse lint` passes;
- `git diff --check` passes.

vadika:
Superseded by #2133, which contains the dependency extraction and the subsequent Orin virtualization refactoring. Closing this PR to avoid merging the overlapping stack separately.

PR #2199, opened 2026-09-01
feat(update): add shared secure A/B core

vadika:
## Summary

Shared authenticated, health-gated A/B update core; #2200 (Orin) and #2201 (x86)
are independent adapters on this base. Uses the updater merged in
tiiuae/ghaf-givc#478, pinned to merge commit `b03c2959c8a435bb64a7da07a3da06831e055c23`.

- Signed manifest v2 binds target, generation, artifact sizes/hashes and UKI.
- Pure `secure-ab-build-config` supplies external public trust; private keys
  remain outside Git/store. The keyless default rejects secure targets.
- Shared manifest/signing/trust-export/fetch tools and boot-health policy.
- Regular-file LVM/LUKS builders live in tiiuae/ghafpkgs#378; Ghaf retains thin
  aliases. The temporary pin is `d55b76076b245a082cd4f65a7841644e677d3fbd`.
  LVM metadata is written directly in Rust; only cryptsetup needs an offline patch.
- Image builders use regular files without privileged storage APIs or VMs.

Cleanup: **1,709 → 1,521 added lines**. Condense the shared contract, declare
public certificate names once, and reuse Python's streaming digest helper.
Boolean sizes/generations still reject. Boot-health and transaction logic are
unchanged.

## Validation

Manifest roundtrip/rejection, signer collision, public-config export and bounded
HTTP-fetch package checks pass. Formatting, REUSE and whitespace checks pass.
Both adapters incorporate the new shared pin and explicit-manifest CLI.
Their focused x86 image tests and pure Orin build-command evaluation pass.
The broad full-flake check was stopped during unrelated target evaluation.

PR #2200, opened 2026-09-01
feat(orin): add secure A/B platform adapter

vadika:
## Summary

NVIDIA Jetson Orin adapter on #2199; independent of sibling #2201.

- NX NVMe and AGX eMMC secure verity/LUKS targets.
- Shared regular-file image builders; flash-time signing and recovery enrollment.
- OP-TEE DUK unlock, aligned APP growth and idempotent persist provisioning.
- External public trust bound to NVIDIA Secure Boot, fixed-capacity system
  slots and shared three-attempt boot health.
- Uses the Rust LVM writer from tiiuae/ghafpkgs#378 through the shared pin,
  selecting exactly one manifest explicitly and using the fixed LUKS header.

The runbook retains test IDs, fault points, device checks and recovery limits.
Mandatory signing-key validation eliminates the unsigned-signing branch.

## Validation

Pure NX/AGX trust evaluation and generated NX pre-flash shell syntax checks
pass. The latest shared pin and CLI changes pass pure AGX secure-image
build-command evaluation, formatting, REUSE and whitespace checks.

### DIFF
--- a/modules/partitioning/firstboot-persist.nix
+++ b/modules/partitioning/firstboot-persist.nix
@@ -8,8 +8,8 @@
 # boot this service:
 #
 #   1. Expands the LVM PV after initrd has grown APP and the open LUKS mapping
-#   2. Creates swap and persist LVs from the free space, reserving
-#      enough room for a future B-slot root+verity pair (OTA updates)
+#   2. Creates the fixed-capacity empty B-slot root and verity LVs
+#   3. Creates swap and persist LVs from the remaining free space
 #
 # Every step is idempotent — the service can be interrupted and re-run
 # safely (e.g. after a power loss during first boot).
@@ -55,15 +55,28 @@ let
       echo "Resizing PV..."
       pvresize "$PV_PATH"
 
-      # --- Compute B-slot reservation ---
+      # --- Provision the fixed-capacity B slot before persist takes the rest ---
 
-      A_ROOT_MIB=$(lvs --noheadings -o lv_size --nosuffix --units m -S "vg_name=pool && lv_name=~^root_" | head -1 | tr -d '[:space:]')
-      A_VERITY_MIB=$(lvs --noheadings -o lv_size --nosuffix --units m -S "vg_name=pool && lv_name=~^verity_" | head -1 | tr -d '[:space:]')
-      # Reserve 50% headroom on top of the current A-slot sizes so the
-      # B-slot can accommodate a significantly larger image after OTA
-      # updates that add packages or data.
-      RESERVE_MIB=$(awk "BEGIN { printf \"%d\", (''${A_ROOT_MIB:-0} + ''${A_VERITY_MIB:-0}) * 1.5 + 64 }")
-      echo "B-slot reservation: $RESERVE_MIB MiB (1.5 * (root=$A_ROOT_MIB + verity=$A_VERITY_MIB) + 64)"
+      ensure_empty_lv() {
+        name="$1"
+        expected_mib="$2"
+        if lvs "pool/$name" >/dev/null 2>&1; then
+          actual_mib=$(lvs --noheadings -o lv_size --nosuffix --units m "pool/$name" \
+            | awk '{ sub(/^</, "", $1); printf "%d", $1 }')
+          if [ "$actual_mib" -ne "$expected_mib" ]; then
+            echo "ERROR: pool/$name is $actual_mib MiB, expected $expected_mib MiB" >&2
+            exit 1
+          fi
+          echo "pool/$name already has the required $expected_mib MiB capacity."
+        else
+          echo "Creating pool/$name ($expected_mib MiB)..."
+          DM_DISABLE_UDEV=1 lvcreate -y -Zn -Wn -n "$name" -L "''${expected_mib}M" pool
+        fi
+      }
+
+      ensure_empty_lv root_empty ${toString cfg.rootSlotSizeMiB}
+      ensure_empty_lv verity_empty ${toString cfg.veritySlotSizeMiB}
+      vgmknodes pool
 
       # --- Create swap LV (skip if already exists) ---
 
@@ -77,16 +90,16 @@ let
       # --- Create persist LV and finish an interrupted format if needed ---
 
       if [ ! -e /dev/pool/persist ]; then
-        VG_FREE_MIB=$(vgs --noheadings -o vg_free --nosuffix --units m pool | tr -d '[:space:]')
-        VG_FREE_INT=$(awk "BEGIN { printf \"%d\", $VG_FREE_MIB }")
-        PERSIST_MIB=$(( VG_FREE_INT - RESERVE_MIB ))
+        VG_FREE_INT=$(vgs --noheadings -o vg_free --nosuffix --units m pool \
+          | awk '{ sub(/^</, "", $1); printf "%d", $1 }')
+        PERSIST_MIB=$VG_FREE_INT
 
         if [ "$PERSIST_MIB" -le 0 ]; then
-          echo "ERROR: not enough free space (free=$VG_FREE_INT, reserve=$RESERVE_MIB)"
+          echo "ERROR: no free space remains for persist after fixed slots and swap"
           exit 1
         fi
 
-        echo "Creating persist LV: $PERSIST_MIB MiB (leaving $RESERVE_MIB MiB for B-slot)"
+        echo "Creating persist LV from the remaining $PERSIST_MIB MiB..."
         lvcreate -L "''${PERSIST_MIB}M" -n persist pool
 
       else
@@ -119,7 +132,7 @@ in
 
     # --- First-boot service ---
     systemd.services.firstboot-persist = {
-      description = "Create swap and persist LVs on first boot";
+      description = "Create inactive system slot, swap, and persist LVs on first boot";
       wantedBy = [ "local-fs-pre.target" ];
       before = [ "local-fs-pre.target" ];
       after = [
--- a/modules/partitioning/firstboot-persist.nix
+++ b/modules/partitioning/firstboot-persist.nix
@@ -7,9 +7,8 @@
 # to minimize flash image size (~5.5 GiB instead of ~16 GiB). On first
 # boot this service:
 #
-#   1. Resizes the GPT and APP partition to fill the eMMC
-#   2. Expands the LVM PV to match
-#   3. Creates swap and persist LVs from the free space, reserving
+#   1. Expands the LVM PV after initrd has grown APP and the open LUKS mapping
+#   2. Creates swap and persist LVs from the free space, reserving
 #      enough room for a future B-slot root+verity pair (OTA updates)
 #
 # Every step is idempotent — the service can be interrupted and re-run
@@ -35,36 +34,22 @@ let
     runtimeInputs = with pkgs; [
       gnugrep
       gawk
-      util-linux
-      gptfdisk
-      parted
       lvm2
       coreutils
       btrfs-progs
+      util-linux
     ];
     text = ''
       set -euo pipefail
       echo "firstboot-persist: starting at $(date)"
 
-      # --- Resize APP partition to fill the eMMC (idempotent) ---
+      # Initrd grows APP and the open LUKS mapping while the OP-TEE DUK is
+      # still available in its shared keyring.  Only the PV and new LVs are
+      # safe to grow here after switch-root.
 
       PV_PATH=$(pvdisplay -C -o pv_name --noheadings -S vg_name=pool | head -n1 | tr -d '[:space:]')
-      P_DEVPATH=$(readlink -f "$PV_PATH")
-      echo "PV: $PV_PATH -> $P_DEVPATH"
-
-      if [[ "$P_DEVPATH" =~ [0-9]+$ ]]; then
-        PARTNUM=$(echo "$P_DEVPATH" | grep -o '[0-9]*$')
-        PARENT_DISK=/dev/$(lsblk --nodeps --noheadings -o pkname "$P_DEVPATH")
-      else
-        echo "ERROR: cannot determine partition number from $P_DEVPATH"
-        exit 1
-      fi
-
-      echo "Fixing GPT and resizing partition $PARTNUM on $PARENT_DISK..."
-      sgdisk "$PARENT_DISK" -e || true
-      # parted resizepart updates the kernel's partition size synchronously
-      # via BLKPG_RESIZE_PARTITION ioctl, so no partprobe/udevadm needed.
-      parted -s -a opt "$PARENT_DISK" "resizepart $PARTNUM 100%" || true
+      [[ -n "$PV_PATH" ]] || { echo "ERROR: pool PV not found"; exit 1; }
+      echo "PV: $PV_PATH -> $(readlink -f "$PV_PATH")"
 
       # pvresize is idempotent — no-op if PV already matches partition
       echo "Resizing PV..."
@@ -89,7 +74,7 @@ let
         echo "swap LV already exists, skipping."
       fi
 
-      # --- Create persist LV (skip if already exists) ---
+      # --- Create persist LV and finish an interrupted format if needed ---
 
       if [ ! -e /dev/pool/persist ]; then
         VG_FREE_MIB=$(vgs --noheadings -o vg_free --nosuffix --units m pool | tr -d '[:space:]')
@@ -104,11 +89,24 @@ let
         echo "Creating persist LV: $PERSIST_MIB MiB (leaving $RESERVE_MIB MiB for B-slot)"
         lvcreate -L "''${PERSIST_MIB}M" -n persist pool
 
-        echo "Formatting persist as btrfs..."
-        mkfs.btrfs -L persist /dev/pool/persist
       else
-        echo "persist LV already exists, skipping."
+        echo "persist LV already exists."
       fi
+
+      PERSIST_TYPE=$(blkid -o value -s TYPE /dev/pool/persist 2>/dev/null || true)
+      case "$PERSIST_TYPE" in
+        "")
+          echo "Formatting uninitialized persist LV as btrfs..."
+          mkfs.btrfs -L persist /dev/pool/persist
+          ;;
+        btrfs)
+          echo "persist LV already contains btrfs, retaining it."
+          ;;
+        *)
+          echo "ERROR: persist LV has unexpected filesystem type: $PERSIST_TYPE"
+          exit 1
+          ;;
+      esac
 
       echo "firstboot-persist: done."
     '';
--- a/modules/partitioning/firstboot-persist.nix
+++ b/modules/partitioning/firstboot-persist.nix
@@ -3,13 +3,11 @@
 #
 # First-boot service for Jetson Orin A/B verity boot.
 #
-# The pre-built LVM image only contains the A-slot root and verity LVs
-# to minimize flash image size (~5.5 GiB instead of ~16 GiB). On first
+# The pre-built LVM image contains both fixed-capacity system slots. On first
 # boot this service:
 #
 #   1. Expands the LVM PV after initrd has grown APP and the open LUKS mapping
-#   2. Creates the fixed-capacity empty B-slot root and verity LVs
-#   3. Creates swap and persist LVs from the remaining free space
+#   2. Creates swap and persist LVs from the remaining free space
 #
 # Every step is idempotent — the service can be interrupted and re-run
 # safely (e.g. after a power loss during first boot).
@@ -55,27 +53,8 @@ let
       echo "Resizing PV..."
       pvresize "$PV_PATH"
 
-      # --- Provision the fixed-capacity B slot before persist takes the rest ---
-
-      ensure_empty_lv() {
-        name="$1"
-        expected_mib="$2"
-        if lvs "pool/$name" >/dev/null 2>&1; then
-          actual_mib=$(lvs --noheadings -o lv_size --nosuffix --units m "pool/$name" \
-            | awk '{ sub(/^</, "", $1); printf "%d", $1 }')
-          if [ "$actual_mib" -ne "$expected_mib" ]; then
-            echo "ERROR: pool/$name is $actual_mib MiB, expected $expected_mib MiB" >&2
-            exit 1
-          fi
-          echo "pool/$name already has the required $expected_mib MiB capacity."
-        else
-          echo "Creating pool/$name ($expected_mib MiB)..."
-          DM_DISABLE_UDEV=1 lvcreate -y -Zn -Wn -n "$name" -L "''${expected_mib}M" pool
-        fi
-      }
-
-      ensure_empty_lv root_empty ${toString cfg.rootSlotSizeMiB}
-      ensure_empty_lv verity_empty ${toString cfg.veritySlotSizeMiB}
+      # The image builder owns the two system pairs. Their names change during
+      # updates, so provisioning must not recreate the initial empty names.
       vgmknodes pool
 
       # --- Create swap LV (skip if already exists) ---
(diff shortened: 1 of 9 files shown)


modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -525,15 +532,11 @@ let
...
-        printf "Note: Re-encryption may take 1 minute for every 1 GB of data ...\n"
-        if ! cryptsetup reencrypt --key-description "$uniqueKeyDescription" "$luksDev"; then
-           printf "error: Re-encryption failed\n"
-           handle_error
-        fi
+        # Do not rotate the volume key here.  cryptsetup cannot re-wrap a
+        # recovery keyslot without knowing its passphrase, so volume-key
+        # re-encryption would discard the offline recovery credential.  Removing
+        # the manufacturer keyslot is sufficient to revoke that passphrase: it
+        # never exposed the random volume key itself.
```
brianmcgillion: **Security: this removes the only step that gives an Orin its own volume key, and the reasoning in this comment does not hold.**

"It never exposed the random volume key itself" is only true for one device. The volume key is generated once, when `verity-image.nix` runs `ghaf-initialize-verity-lvm` in the build VM, so it is identical in every copy of `system.img.zst`: in the Nix store, in the binary cache, and on every board flashed from it. The manufacturer keyslot wraps that key with `deviceManufacturerPassphrase`, which defaults to `ghaf` and is in the public source. So anyone holding the image can run `cryptsetup luksDump --dump-volume-key` and get the key for every device's APP partition. After that, removing the manufacturer keyslot on a device revokes nothing: the attacker already has the key, and it is never rotated.

`main` handles this with `cryptsetup reencrypt` right here, and the comment this PR deletes says exactly why. I checked the underlying behaviour: after `luksAddKey` + `luksKillSlot` the dumped volume key is byte-identical.

The recovery-slot problem is real: reencrypting on the device would need the recovery passphrase, which the device never sees. The flash host does have it, so the rekey can move there. See my comment on `partition-template-verity.nix`. With that in place, this comment should say the volume key was already rotated per flash.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template-verity.nix
```diff
@@ -193,11 +240,35 @@ let
...
+      # A trailing newline would become part of a cryptsetup key file during
+      # enrollment and make the printed value fail interactive recovery.
+      "${pkgs.pkgsBuildBuild.openssl}/bin/openssl" rand -base64 24 \
+        | "${pkgs.pkgsBuildBuild.coreutils}/bin/tr" -d '\n' > "$_recovery"
+      chmod 0600 "$_recovery"
+
+      printf '%s' ${lib.escapeShellArg config.ghaf.hardware.nvidia.orin.diskEncryption.deviceUniqueKey.deviceManufacturerPassphrase} \
+        | "${pkgs.pkgsBuildBuild.cryptsetup}/bin/cryptsetup" luksAddKey \
+          --new-key-slot 1 \
+          --key-file=- "$_outer" "$_recovery"
```
brianmcgillion: A place to give each flash its own volume key without losing the recovery slot: rekey the decompressed image here, *before* the recovery passphrase is added. At that point only the manufacturer slot exists, so `reencrypt` needs just the one passphrase this script already has.

`reencrypt` moves the surviving keyslot to a new slot number (slot 0 comes back as slot 1). So the manufacturer passphrase is put back in slot 0 before recovery takes slot 1, which keeps the layout the DUK step (slot 2) and the HW-005 checks expect. I ran this exact sequence on a LUKS2 file image (unprivileged, so with `--disable-locks` added). The result has a new volume key, slots `0` (manufacturer) and `1` (recovery), and both passphrases open it:

```bash
_cs="${pkgs.pkgsBuildBuild.cryptsetup}/bin/cryptsetup"
_mfr="$(mktemp)"; chmod 0600 "$_mfr"
printf '%s' ${lib.escapeShellArg config.ghaf.hardware.nvidia.orin.diskEncryption.deviceUniqueKey.deviceManufacturerPassphrase} > "$_mfr"
# Every copy of system.img.zst shares one volume key; rotate it per flash.
"$_cs" reencrypt --batch-mode --force-offline-reencrypt --key-file "$_mfr" "$_outer"
_moved=$("$_cs" luksDump "$_outer" | sed -n 's/^  \([0-9]*\): luks2$/\1/p')
"$_cs" luksAddKey --batch-mode --key-file "$_mfr" --new-key-slot 0 "$_outer" "$_mfr"
"$_cs" luksKillSlot --batch-mode --key-file "$_mfr" "$_outer" "$_moved"
"$_cs" luksAddKey --batch-mode --key-file "$_mfr" --new-key-slot 1 "$_outer" "$_recovery"
rm -f "$_mfr"
```

This is a sketch, not a committable suggestion, because I have not run it inside the flash script. It needs `sed` on the flash script's PATH. The cost is a full rewrite of the APP image's data area on the flash host, so the image stops being sparse and the flash gets slower. The script runs as root, so the default `/run/cryptsetup` lock path is fine without that flag.

brianmcgillion (commented):
Following up on the shared-LUKS-key finding from the #2199 review (https://github.com/tiiuae/ghaf/pull/2199#pullrequestreview-5282098419), which I said I'd raise here.

On Orin this is a **regression from `main`**. Today the first-boot DUK rotation ends with `cryptsetup reencrypt`, so every device gets its own volume key. This PR removes that step. The replacement comment says removing the manufacturer keyslot is enough, but the volume key it leaves in place is the one baked into the shared `system.img.zst` and unlockable with the manufacturer passphrase (default `ghaf`, in the public source). Details and a tested fix are inline. The recovery-slot constraint that motivated the change is real, and the fix keeps it: rekey at flash time, before the recovery slot is added.

What I verified, with stock cryptsetup on LUKS2 file images:
- A keyslot swap leaves the volume key unchanged.
- The proposed flash-time sequence produces a new volume key, keeps the manufacturer slot at 0 and the recovery slot at 1, and both passphrases open the result.

I did not run the flash script end to end or flash hardware.

targets/nvidia-jetson-orin/flake-module.nix
```diff
@@ -362,30 +471,67 @@ let
...
         (ghaf-configuration {
-          name = "nvidia-jetson-orin-agx-verity";
+          name = "${board.name}-verity-luks";
           inherit system;
           profile = "orin";
-          hardwareModule = self.nixosModules.hardware-nvidia-jetson-orin-agx;
-          inherit variant;
-          extraModules = orinVerityModules;
+          inherit (board) hardwareModule;
+          variant = "debug";
```
brianmcgillion, line 487: **BLOCKING: this replaces mainline's verity targets instead of adding to them.**

On `main`, `verity-target-configs` maps over `[ "debug" "release" ]` and produces `nvidia-jetson-orin-agx-verity-{debug,release}-from-x86_64`, plus `-flash-script`, `-ghafImage` and `nodemoapps` variants. Here it is rewritten as a board list with `variant = "debug"` hard-coded and a new `-verity-luks` name. Those 10 attributes disappear from `packages.x86_64-linux`, including the only release verity image for Orin.

Verity is core to the secure system, and downstream products will build on these targets. Please keep the existing verity targets, debug and release, and add the `-verity-luks` secure A/B variants alongside them. `docs/src/content/docs/ghaf/overview/arch/secureboot.mdx:163` still references `nvidia-jetson-orin-agx-verity-debug-from-x86_64-flash-script`.

modules/partitioning/firstboot-persist.nix
```diff
@@ -89,27 +66,40 @@ let
...
       else
-        echo "persist LV already exists, skipping."
+        echo "persist LV already exists."
       fi
 
+      PERSIST_TYPE=$(blkid -o value -s TYPE /dev/pool/persist 2>/dev/null || true)
+      case "$PERSIST_TYPE" in
+        "")
+          echo "Formatting uninitialized persist LV as btrfs..."
+          mkfs.btrfs -L persist /dev/pool/persist
```
brianmcgillion, line 92: **Important: this formats `/persist` on any boot where `blkid` reports no filesystem, not only right after the LV is created.**

On `main`, `mkfs.btrfs` runs only inside the branch that has just run `lvcreate`. Here the `blkid … || true` check runs every boot, and an empty result leads straight to `mkfs.btrfs`. So a damaged btrfs superblock no longer causes a mount failure; it causes a silent wipe of all persistent data. That includes `/persist/common/ota/accepted-generation`, which `boot-health` treats as generation 0 when the file is missing. Anti-rollback is reset, and older signed generations become installable again.

The LUKS offset and slot sizes are fixed, so someone with physical access can find and corrupt the superblock sector without the key.

If the aim is to finish an interrupted first format, please tie it to that case: format only in the same run as `lvcreate`, or only when a marker written after `lvcreate` is present and not yet cleared. In every other case, fail closed.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -513,9 +513,16 @@ let
...
+        # The flash image reserves slot 1 for the printable recovery
+        # passphrase. Put the DUK in a known, separate slot so manufacturer
+        # removal cannot remove or rewrite recovery.
+        uniqueKeySlot=2
 
         printf "Switching to use device unique key. This might take a bit..\n"
-        if ! printf "%s" "$defaultKey" | cryptsetup luksAddKey --new-key-description "$uniqueKeyDescription" --key-file=- "$luksDev"; then
+        if ! printf "%s" "$defaultKey" | cryptsetup luksAddKey \
+          --new-key-description "$uniqueKeyDescription" \
+          --new-key-slot "$uniqueKeySlot" \
```
brianmcgillion: **Important: a fixed slot 2 turns an interrupted first boot into a reboot loop.**

First boot is detected by the keyring token, which is only added after this function returns. If power is lost after `luksAddKey --new-key-slot 2` succeeds but before `luksRemoveKey`, the next boot still counts as first boot and runs the same `luksAddKey --new-key-slot 2`. That fails because the slot is full, `handle_error` reboots, and it repeats on every boot.

`main` did not pin the slot, so it recovered from that window. It does have the equivalent problem later, after the removal. Please make the step idempotent. If slot 2 already opens with the DUK (`cryptsetup open --test-passphrase --key-slot 2 --key-description "$uniqueKeyDescription" "$luksDev"`), skip the add and continue with the removal and the token.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template-verity.nix
```diff
@@ -133,6 +149,62 @@ let
     echo "============================================================"
     echo ""
 
+    if [ -z "''${GHAF_DEV_KEY_DIR:-}" ]; then
+      echo "ERROR: GHAF_DEV_KEY_DIR is required for secure A/B canary flashes." >&2
+      exit 1
+    fi
+    for _required in PK.crt KEK.crt db.crt db.key update.pub update.key; do
```
brianmcgillion, line 165: **Important: the flash now requires the fleet's OTA signing private key.**

`update.key` is required here, and it is used only to check that it matches `update.pub` (lines 195-204). Nothing in the flash signs with it. That puts the key that authorises every OTA update for the fleet on every flash station, which is a much wider exposure than the `db.key` the flash actually needs. `update.pub` is already checked against the digest embedded at evaluation, so please drop `update.key` from this list and delete the key-pair check.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template-verity.nix
```diff
@@ -193,11 +240,35 @@ let
...
+        | "${pkgs.pkgsBuildBuild.coreutils}/bin/tr" -d '\n' > "$_recovery"
+      chmod 0600 "$_recovery"
+
+      printf '%s' ${lib.escapeShellArg config.ghaf.hardware.nvidia.orin.diskEncryption.deviceUniqueKey.deviceManufacturerPassphrase} \
+        | "${pkgs.pkgsBuildBuild.cryptsetup}/bin/cryptsetup" luksAddKey \
+          --new-key-slot 1 \
+          --key-file=- "$_outer" "$_recovery"
+      echo "============================================================"
+      echo "RECOVERY PASSPHRASE (store securely; generated for this flash):"
+      cat "$_recovery"
```
brianmcgillion: Minor: recovery passphrase handling.
- `mkdir -p` creates the directory with the default mode (0755 under sudo), and the file is written under umask 022, so it is readable by others until the `chmod 0600`.
- It is stored inside `GHAF_DEV_KEY_DIR`, next to `db.key`.
- `cat` prints it to stdout, so harnesses that capture flash output keep it in their logs.
- An `onlyQSPI` flash, which never writes APP, still generates and prints one.

Suggest `umask 077` around the block, `install -d -m 0700` for the directory, skipping the block when APP is not flashed, and ideally a directory separate from the signing keys.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -594,6 +597,94 @@ let
...
+      cryptsetup
+      gptfdisk
+      parted
+      util-linux
+    ];
+    text = ''
+      set -euo pipefail
+
+      luks_dev=""
+      for dev in $(blkid -o device -t UUID=${luksUuid} 2>/dev/null); do
```
brianmcgillion, line 614: Minor: this takes the first device carrying the pinned LUKS UUID, which is the same constant in every Orin LUKS image, and then runs `sgdisk -e`, `parted -f` and `resizepart` on its disk. With a Ghaf USB stick attached to an NX, that can be the stick rather than the NVMe that `cryptroot` was actually opened from. Main has the same selection pattern for non-verity targets, so this is a variant, not a new class. Deriving the disk from the backing device of `/dev/mapper/${cfg.diskEncryption.mapperName}` (`lsblk -no PKNAME` or `dmsetup deps`) removes the ambiguity.

brianmcgillion (changes requested):
## Re-review at `d1c254c7`: requesting changes

This head is a rebase onto #2199 `0a08af39`. The PR's own diff is line-for-line identical to `f928d57d`, which I commented on yesterday, and those threads have no replies. I also did a wider pass for security issues and changes from mainline. Each finding below was checked in code at this head and compared with `main`.

### Blocking

**1. Every board shares one LUKS volume key.** This is still unaddressed (see https://github.com/tiiuae/ghaf/pull/2200#pullrequestreview-5282149813 for the tested flash-time rekey). The build VM fixes the key once, and `switch_to_use_device_unique_key` no longer runs mainline's `cryptsetup reencrypt`. The impact is also wider than the secure A/B canaries:
- The function is shared by every target with `deviceUniqueKey.enable`, which includes mainline's existing `-luks` and `-luks-uki` Orin targets (`flake-module.nix:593-637`).
- Those targets are encrypted at image build time (`sdimage.nix`, `cryptsetup reencrypt --encrypt`), so their key is also in a Nix output.
- On `main` they get a per-device key at first boot. With this PR they don't, and their derivations change: `nvidia-jetson-orin-agx-debug-luks-from-x86_64` and `-luks-uki-from-x86_64` differ from #2199's head.
- They have no recovery slot, so the stated reason for dropping the rekey does not apply to them at all.

**2. Mainline's verity targets are removed, including release.** Evaluating `packages.x86_64-linux` against `main`:
- **Removed:** `nvidia-jetson-orin-agx-verity-debug-from-x86_64` and `nvidia-jetson-orin-agx-verity-release-from-x86_64`, each with its `-flash-script`, `-ghafImage` and `nodemoapps` variants (10 attributes).
- **Added:** `nvidia-jetson-orin-{agx,nx}-verity-luks-debug`, debug only.
- Verity is one of the foundations of the secure system, downstream products build on it, and it has to ship in release. Replacing the release verity image with a debug-only canary is a breaking change for everyone using those targets. The PR description does not mention it, and `docs/.../overview/arch/secureboot.mdx:163` still documents the removed target.
- Please keep the existing verity targets, both debug and release, and add the secure A/B variants alongside them. The secure family also needs a release variant before it can replace anything.

### Important
3. **`firstboot-persist` reformats `/persist` whenever `blkid` finds nothing** (inline).
4. **Fixed DUK slot 2 adds a reboot-loop window** (inline).
5. **The flash host needs the fleet's OTA signing key** (inline).

### Minor
6. Recovery passphrase handling (inline).
7. `resize-verity-luks` takes the first disk carrying the pinned LUKS UUID (inline).
8. `sdimage.nix` adds `default ghaf-*.efi` to `loader.conf` for the existing `-luks-uki` targets. It has no effect for secure A/B, whose ESP is built in the flash script, but it changes mainline images. Please drop it or explain it.

### Checked and fine
The flash-time trust checks are stronger than `main`:
- `GHAF_DEV_KEY_DIR` is required, and the PK/KEK/db/`update.pub` digests are compared with the values embedded at evaluation.
- `db.key` must match `db.crt`, and signing is unconditional.
- No private keys reach the store.

Existing non-LUKS Orin targets are unchanged: `agx`/`nx` `debug` and `agx-release` derivations match #2199. Also fine:
- `ftpm.enable` defaults to mainline behaviour.
- `firstboot-persist` is imported only by the Orin verity modules.
- `nix fmt -- --fail-on-change` and `reuse lint` pass.
- The new flash script and images evaluate with an exported public trust config.

Not covered: no image build, no flash, no hardware.

brianmcgillion (changes requested):
## Retest at `5f2f7bb7`: still requesting changes

Thanks, the key fix is good. I've replied on each thread; summary:

| Finding | Status |
|---|---|
| Every board shares one volume key | **Fixed.** `rekey-verity-image.sh` rekeys each flash before the recovery slot is added and refuses an already-personalised image. `checks.orin-rekey` passes and shows two flashes of one image get different keys, with data and recovery preserved. |
| Mainline `-luks` / `-luks-uki` targets lost their first-boot rekey | **Fixed.** Their `cryptsetup reencrypt` step matches `main` again. Only comments and the `loader.conf` line below differ. |
| Fixed DUK slot 2 reboot loop | **Fixed.** Back to `main`'s next-free slot. |
| Recovery passphrase handling | **Fixed.** `umask 077`, a 0700 directory, a `mktemp` name, no longer printed, and skipped for QSPI-only flashes. |
| **Mainline verity targets removed, including release** | **Open, blocking.** |
| `/persist` reformatted whenever `blkid` is empty | **Open.** |
| Flash requires the fleet `update.key` | **Open.** |
| Resize picks the first disk with the pinned UUID | **Open**, minor. |

### Blocking
**Mainline's verity targets are still removed.** `packages.x86_64-linux` has no `nvidia-jetson-orin-agx-verity-debug-from-x86_64` or `nvidia-jetson-orin-agx-verity-release-from-x86_64` (nor their `-flash-script`, `-ghafImage` and `nodemoapps` variants). Only the debug-only `nvidia-jetson-orin-{agx,nx}-verity-luks-debug` targets exist. Verity is part of the core secure system: downstream products build on these targets, and it has to ship in release. Please keep the existing verity targets, debug and release, and add the secure A/B variants alongside them. `docs/src/content/docs/ghaf/overview/arch/secureboot.mdx:163` still documents the removed `nvidia-jetson-orin-agx-verity-debug-from-x86_64-flash-script`.

### Important
- **`firstboot-persist.nix:88-92`** still runs `mkfs.btrfs` on every boot where `blkid` reports nothing. A damaged superblock therefore silently wipes `/persist`, including the anti-rollback floor, where `main` only fails to mount.
- **`partition-template-verity.nix:165`** still requires `update.key` on every flash host, only to compare it with `update.pub`, which is already checked against the evaluated digest.

### Minor
- **Resize disk selection:** `resize-verity-luks` (`jetson-orin.nix:613-614`) still takes the first device with the pinned LUKS UUID.
- **`loader.conf`:** `sdimage.nix` still adds `default ghaf-*.efi` to `loader.conf` for the existing `-luks-uki` targets. The only purpose I can see is steering A/B selection, which those targets don't use.

Checks at `5f2f7bb7`: `nix fmt -- --fail-on-change` and `reuse lint` pass, and `checks.x86_64-linux.orin-rekey` passes. Plain Orin `agx`/`nx` debug and `agx-release` derivations match #2199. No flash or hardware run.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
brianmcgillion: Fixed at `5f2f7bb7`. Verity targets now rekey per flash in `rekey-verity-image.sh`, and the non-verity `-luks`/`-luks-uki` targets keep mainline's first-boot `cryptsetup reencrypt` (only comment text differs from `main`). `checks.orin-rekey` passes.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template-verity.nix
brianmcgillion: Fixed at `5f2f7bb7`: `rekey-verity-image.sh` implements this. It also refuses an image that has more than the build keyslot, which is a good addition. `checks.orin-rekey` confirms that two flashes of one image get different volume keys, with data and recovery preserved.

targets/nvidia-jetson-orin/flake-module.nix
brianmcgillion: Still open at `5f2f7bb7`, and still blocking. `packages.x86_64-linux` has only `nvidia-jetson-orin-{agx,nx}-verity-luks-debug`. Mainline's `nvidia-jetson-orin-agx-verity-debug-from-x86_64` and `nvidia-jetson-orin-agx-verity-release-from-x86_64` are still gone, along with their `-flash-script`, `-ghafImage` and `nodemoapps` variants. Verity has to ship in release; please keep the existing targets and add the secure A/B ones alongside them. `secureboot.mdx:163` still references the removed target.

modules/partitioning/firstboot-persist.nix
brianmcgillion: Still open at `5f2f7bb7`: `firstboot-persist.nix:88-92` is unchanged, so `mkfs.btrfs` still runs on any boot where `blkid` is empty.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
brianmcgillion: Fixed at `5f2f7bb7`. The fixed slot is gone and `luksAddKey` takes the next free slot as on `main`.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template-verity.nix
brianmcgillion: Still open at `5f2f7bb7`: `update.key` is still required at `partition-template-verity.nix:165` and used only for the key-pair check.

modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template-verity.nix
brianmcgillion: Fixed at `5f2f7bb7`: `umask 077`, `install -d -m 0700`, a `mktemp` file name, the passphrase is no longer printed, and the block is skipped for QSPI-only flashes. It still lives under `GHAF_DEV_KEY_DIR`; that is fine by me, but a separate directory would keep it apart from the signing keys.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
brianmcgillion: Still open at `5f2f7bb7` (`jetson-orin.nix:613-614`), minor.

PR #2201, opened 2026-09-01
feat(x86): add secure A/B platform adapter

vadika:
## Summary

x86 adapter on #2199; independent of sibling #2200. Adds Intel laptop secure
A/B initial-image and update targets without replacing normal laptop targets.

- Build GPT/FAT/LUKS2/LVM from regular files without root, VMs, mounts, loop
  devices or device mapper; fixed A/B slots, persist, swap and Type-2 UKIs.
- Require external public trust, sign EFI artifacts outside the build, and
  verify the written ESP before installer firmware enrollment.
- Uses the Rust LVM writer from tiiuae/ghafpkgs#378 through the shared pin,
  passing the selected manifest explicitly and using the fixed LUKS header.

Shared plan/image fixtures, LVM arguments and CLI option handling avoid
duplication. Platform instructions link to the shared setup/update contract.

## Validation

Sandboxed plan and 4.6 GiB image tests pass with the published dependency:
capacity bounds, missing/unknown options, GPT, ESP/UKI, LUKS2 header,
compression, bmap and trust inventory. Formatting, REUSE and whitespace
checks pass.

### DIFF
--- a/modules/partitioning/initial-verity-disk-x86.nix
+++ b/modules/partitioning/initial-verity-disk-x86.nix
@@ -11,7 +11,17 @@
 }:
 let
   cfg = config.ghaf.partitioning.verity.initialDisk;
+  verityCfg = config.ghaf.partitioning.verity;
   diskoCfg = config.ghaf.partitioning.disko;
+  # ESP plus two fixed root/verity pairs, swap, persist, and conservative
+  # headroom for GPT, LUKS, and LVM metadata.
+  minimumImageSizeMiB =
+    500
+    + 2 * verityCfg.rootSlotSizeMiB
+    + 2 * verityCfg.veritySlotSizeMiB
+    + diskoCfg.swapSize
+    + diskoCfg.persistSize
+    + 4 * 1024;
   updateImage = config.system.build.ghafUpdateImage;
   verityLvmImage = config.system.build.verityLvmImage;
   trustInventory = pkgs.writeText "ghaf-secure-ab-public-trust.json" (
@@ -145,7 +155,13 @@ in
         assertion = config.ghaf.storage.encryption.enable && !config.ghaf.storage.encryption.deferred;
         message = "ghaf.partitioning.verity.initialDisk requires immediate LUKS configuration";
       }
+      {
+        assertion = diskoCfg.imageSize >= minimumImageSizeMiB;
+        message = "secure A/B imageSize is ${toString diskoCfg.imageSize} MiB but two fixed system slots and persistent volumes require at least ${toString minimumImageSizeMiB} MiB";
+      }
     ];
+
+    ghaf.partitioning.disko.imageSize = lib.mkDefault minimumImageSizeMiB;
 
     fileSystems."/boot" = lib.mkForce {
       device = "/dev/disk/by-label/ESP";
--- a/modules/partitioning/initial-verity-disk-x86.nix
+++ b/modules/partitioning/initial-verity-disk-x86.nix
@@ -27,7 +27,6 @@ let
   updateImage = config.system.build.ghafUpdateImage;
   trustInventory = pkgs.writeText "ghaf-secure-ab-public-trust.json" (
     builtins.toJSON {
-      external = config.ghaf.secureUpdate.externalPublicTrustConfigured;
       inherit (config.ghaf.secureUpdate) target generation publicTrustDigests;
     }
   );
--- a/modules/partitioning/initial-verity-disk-x86.nix
+++ b/modules/partitioning/initial-verity-disk-x86.nix
@@ -1,0 +1,159 @@
+# SPDX-FileCopyrightText: 2022-2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# x86 bootstrap disk for the shared secure A/B payload. The image is emitted
+# unsigned; ghaf-sign-x86-image signs the ESP later, outside the Nix store.
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.ghaf.partitioning.verity.initialDisk;
+  diskoCfg = config.ghaf.partitioning.disko;
+  updateImage = config.system.build.ghafUpdateImage;
+  verityLvmImage = config.system.build.verityLvmImage;
+  trustInventory = pkgs.writeText "ghaf-secure-ab-public-trust.json" (
+    builtins.toJSON {
+      external = config.ghaf.secureUpdate.externalPublicTrustConfigured;
+      inherit (config.ghaf.secureUpdate) target generation publicTrustDigests;
+    }
+  );
+  buildPkgs = pkgs.pkgsBuildBuild;
+  vmTools = buildPkgs.vmTools.override {
+    rootModules = [
+      "virtiofs"
+      "virtio_pci"
+      "virtio_blk"
+      "virtio_balloon"
+      "virtio_rng"
+      "dm_mod"
+      "dm_crypt"
+    ];
+  };
+
+  image = vmTools.runInLinuxVM (
+    buildPkgs.stdenvNoCC.mkDerivation {
+      name = "ghaf-initial-verity-x86-disk";
+      __structuredAttrs = false;
+      buildInputs = with buildPkgs; [
+        bmaptool
+        btrfs-progs
+        coreutils
+        cryptsetup
+        dosfstools
+        gawk
+        gptfdisk
+        jq
+        lvm2
+        parted
+        systemd
+        util-linux
+        zstd
+      ];
+      memSize = 4096;
+
+      preVM = ''
+        set -efx
+        mkdir -p "$out"
+        ${buildPkgs.qemu}/bin/qemu-img create -f raw "$out/ghaf-image.raw" ${toString diskoCfg.imageSize}M
+      '';
+      QEMU_OPTS = ''-drive file="$out"/ghaf-image.raw,if=virtio,cache=unsafe,werror=report,format=raw'';
+      postVM = ''
+        bmaptool create "$out/ghaf-image.raw" -o "$out/ghaf-image.bmap"
+        cores="''${NIX_BUILD_CORES:-1}"
+        if [ "$cores" -gt 8 ]; then cores=8; fi
+        zstd -T"$cores" --compress "$out/ghaf-image.raw" -o "$out/ghaf-image.raw.zst" --rm
+        install -m 0644 ${trustInventory} "$out/public-trust.json"
+      '';
+
+      buildCommand = ''
+        set -efx
+        disk=/dev/vda
+        sgdisk --zap-all "$disk"
+        sgdisk --new=1:1MiB:+500MiB --typecode=1:ef00 --change-name=1:ESP "$disk"
+        sgdisk --new=2:0:0 --typecode=2:8e00 --change-name=2:disk-disk1-luks "$disk"
+        partprobe "$disk"
+        udevadm settle
+
+        mkfs.vfat -F 32 -n ESP /dev/vda1
+        mkdir -p /mnt/esp/EFI/systemd /mnt/esp/EFI/BOOT /mnt/esp/EFI/Linux /mnt/esp/loader
+        mount /dev/vda1 /mnt/esp
+        mkdir -p /mnt/esp/EFI/systemd /mnt/esp/EFI/BOOT /mnt/esp/EFI/Linux /mnt/esp/loader
+        install -m 0644 ${config.systemd.package}/lib/systemd/boot/efi/systemd-bootx64.efi \
+          /mnt/esp/EFI/systemd/systemd-bootx64.efi
+        install -m 0644 ${config.systemd.package}/lib/systemd/boot/efi/systemd-bootx64.efi \
+          /mnt/esp/EFI/BOOT/BOOTX64.EFI
+
+        manifest=$(find ${updateImage} -name '*.manifest' -print -quit)
+        uki=$(find ${updateImage} -name '*.efi' -print -quit)
+        version=$(jq -er '.version' "$manifest")
+        root_hash=$(jq -er '.root_verity_hash' "$manifest")
+        uki_name="ghaf-$version-''${root_hash:0:16}.efi"
+        install -m 0644 "$uki" "/mnt/esp/EFI/Linux/$uki_name"
+        printf 'timeout %s\ndefault %s\neditor no\n' \
+          ${
+            lib.escapeShellArg (
+              if config.boot.loader.timeout == null then "menu-force" else toString config.boot.loader.timeout
+            )
+          } \
+          "''${uki_name%.efi}" > /mnt/esp/loader/loader.conf
+        sync -f /mnt/esp
+        umount /mnt/esp
+
+        # The one-time empty development passphrase matches Ghaf's existing
+        # non-interactive x86 enrollment flow. First boot enrolls TPM/FIDO2 and
+        # a recovery key, then removes this password slot.
+        printf '\n' | cryptsetup luksFormat --batch-mode --type luks2 "$disk"2
+        printf '\n' | cryptsetup open "$disk"2 crypted
+        zstd -d ${verityLvmImage}/system.img.zst --stdout \
+          | dd of=/dev/mapper/crypted bs=4M conv=notrunc status=progress
+        pvresize /dev/mapper/crypted
+        vgchange -ay pool
+
+        root_mib=$(cat ${verityLvmImage}/root_size_mib)
+        verity_mib=$(cat ${verityLvmImage}/verity_size_mib)
+        DM_DISABLE_UDEV=1 lvcreate -y -Zn -Wn -n root_empty -L "''${root_mib}M" pool
+        DM_DISABLE_UDEV=1 lvcreate -y -Zn -Wn -n verity_empty -L "''${verity_mib}M" pool
+        DM_DISABLE_UDEV=1 lvcreate -y -Zn -Wn -n swap -L ${toString diskoCfg.swapSize}M pool
+        DM_DISABLE_UDEV=1 lvcreate -y -Zn -Wn -n persist -L ${toString diskoCfg.persistSize}M pool
+        vgmknodes pool
+        mkswap -L swap /dev/pool/swap
+        mkfs.btrfs -L persist /dev/pool/persist
+        vgchange -an pool
+        cryptsetup close crypted
+      '';
+    }
+  );
+in
+{
+  options.ghaf.partitioning.verity.initialDisk.enable =
+    lib.mkEnableOption "an initial x86 GPT/LUKS/LVM secure A/B disk image";
+
+  config = lib.mkIf cfg.enable {
+    assertions = [
+      {
+        assertion = config.nixpkgs.hostPlatform.system == "x86_64-linux";
+        message = "ghaf.partitioning.verity.initialDisk is currently implemented only for x86_64-linux";
+      }
+      {
+        assertion = config.ghaf.partitioning.verity.enable;
+        message = "ghaf.partitioning.verity.initialDisk requires the verity update layout";
+      }
+      {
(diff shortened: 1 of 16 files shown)


avnik (approved):
Generally LGTM

avnik (approved):

avnik (approved):

vadika:
Zero-build-VM follow-up is published at `7ca08fd1`.

The previous one-VM implementation was replaced, not layered over: `intel-laptop-debug-secure-ab` now builds a configuration-bound `ghaf-prepare-x86-verity-disk` command plus its immutable payload closure. GPT/ESP/LUKS/LVM image construction only occurs when that command is explicitly run as root outside Nix.

Validation: complete target build, real 123380 MiB plan validation, generic undersize test, derivation/source scans for VM machinery, and an end-to-end host smoke test against a temporary sparse file. The smoke test left no loop or dm-crypt mapping. No physical disk was targeted.

brianmcgillion (commented):
Following up on the shared-LUKS-key finding from the #2199 review (https://github.com/tiiuae/ghaf/pull/2199#pullrequestreview-5282098419), which I said I'd raise here.

On x86 this is a **regression from `main`**. Today `deferred-disk-encryption.nix` encrypts in place on the device, so every laptop gets its own volume key. This adapter forces that path off and ships a LUKS container built once in the Nix build VM, behind an empty passphrase. First-boot `systemd-cryptenroll` only replaces keyslots, so every secure-A/B laptop installed from one image shares a volume key that anyone holding the image can dump. Details and a tested fix are inline.

What I verified, with stock cryptsetup on LUKS2 file images:
- The volume key dumps with an empty key file, and a keyslot swap leaves it unchanged.
- `cryptsetup reencrypt` will not rekey using the empty passphrase itself, so the rekey goes through a temporary keyslot. That sequence produces a new volume key and leaves only the empty bootstrap passphrase in slot 0, as `luks-enroll-tpm` expects.

I did not run the installer end to end or install on hardware. These are review comments rather than a change request, since #2199 already blocks this branch.

packages/pkgs-by-name/ghaf-prepare-x86-verity-disk/ghaf-prepare-x86-verity-disk.sh
```diff
@@ -0,0 +1,166 @@
...
+uki_name="ghaf-$version-${root_hash:0:16}.efi"
+mcopy -i "$esp_image" "$uki" "::EFI/Linux/$uki_name"
+printf 'timeout %s\ndefault %s\neditor no\n' \
+  "$boot_timeout" "${uki_name%.efi}" >"$work/loader.conf"
+mcopy -i "$esp_image" "$work/loader.conf" ::loader/loader.conf
+
+# The empty bootstrap passphrase is replaced by first-boot enrollment.
+: >"$work/bootstrap.key"
+initialize_lvm --image "$luks_image" \
+  --luks-uuid 3E7F3D25-695A-429D-8D34-2D0A18979D7D --key-file "$work/bootstrap.key"
```
brianmcgillion: **Security: every laptop installed from this image shares one LUKS volume key.**

This LUKS container is made once, inside the Nix build VM, and its volume key ships in the compressed disk image. The comment above is right that first-boot enrolment replaces the empty passphrase, but it replaces a *keyslot*: `systemd-cryptenroll` adds TPM2/FIDO2 and recovery, then `--wipe-slot=password`. None of that changes the volume key. Anyone with the image, which is published and in the binary cache, can run:

```bash
: > empty.key
cryptsetup luksDump --dump-volume-key --batch-mode --key-file empty.key <luks partition of the image>
```

and decrypt the disk of any laptop installed from it, offline, for the laptop's lifetime, whatever was enrolled afterwards. I confirmed on a LUKS2 test image that the dumped key is byte-identical before and after the keyslot swap.

The fix has to happen per device, before anything is written to the disk. The installer is the natural place (see my comment on `secure-ab-x86.nix`).

modules/partitioning/secure-ab-x86.nix
```diff
@@ -0,0 +1,35 @@
...
+
+  # Enrollment material is supplied alongside the externally signed image.
+  # Keeping keysSource null prevents even public enrollment payloads from
+  # accidentally selecting a repository trust set different from the signer.
+  ghaf.host.secureboot = {
+    enable = true;
+    keysSource = lib.mkForce null;
+  };
+
+  ghaf.storage.encryption.deferred = lib.mkForce false;
```
brianmcgillion: This is the line that turns off `main`'s per-device key. `deferred-disk-encryption.nix` runs `cryptsetup reencrypt` on the device, so each laptop generated its own volume key. Forcing it off is reasonable for this layout, but the rekey needs a replacement.

`ghaf-installer` is the natural place: it runs as root, the disk is offline and nothing has booted from it yet. `cryptsetup reencrypt` will not rekey using the empty passphrase itself, so it has to go through a temporary keyslot. After writing the image, before rebooting:

```bash
luks=/dev/disk/by-uuid/3e7f3d25-695a-429d-8d34-2d0a18979d7d
# An empty regular file: cryptsetup reads nothing from /dev/null and rejects it.
empty=$(mktemp)
tmp=$(mktemp); chmod 0600 "$tmp"
openssl rand -hex 32 | tr -d '\n' > "$tmp"
cryptsetup luksAddKey --batch-mode --key-file "$empty" --new-key-slot 1 "$luks" "$tmp"
# Rotate the image-wide volume key; --key-slot keeps only the temporary slot.
cryptsetup reencrypt --batch-mode --force-offline-reencrypt --key-file "$tmp" --key-slot 1 "$luks"
moved=$(cryptsetup luksDump "$luks" | sed -n 's/^  \([0-9]*\): luks2$/\1/p')
cryptsetup luksAddKey --batch-mode --key-file "$tmp" --new-key-slot 0 "$luks" "$empty"
cryptsetup luksKillSlot --batch-mode --key-file "$empty" "$luks" "$moved"
rm -f "$tmp" "$empty"
```

I ran this exact sequence on a LUKS2 file image with an empty bootstrap key (unprivileged, so with `--disable-locks` added). The result has a new volume key and only the empty passphrase in slot 0, which is what `luks-enroll-tpm` expects on first boot. I have not run it inside the installer. The cost is a full rewrite of the LUKS partition at install time. Any path that writes the image without the installer (plain `dd`, `ghaf-flash`) would skip the rekey, so the x86 testing guide should state it and the installer should be the only supported route.

targets/laptop/flake-module.nix
```diff
@@ -109,6 +109,36 @@ let
...
+            initialDisk.enable = true;
+          };
+          boot-health.enable = true;
+        };
+      }
+    ])
+    // {
+      name = secureName;
+    }
+  ) (builtins.filter (x: x.buildSysupdateImage) target-configs);
```
brianmcgillion, line 140: **BLOCKING: this produces release variants that can keep an empty-passphrase keyslot for good.**

`intel-laptop` has `variants = [ debug release ]` and `sysupdate = true`, so this filter yields `intel-laptop-release-secure-ab` and its `-sysupdate` image. `secure-ab-x86.nix` then forces `deferred = false`. For release laptops that swaps mainline's deferred encryption (user passphrase, per-device key) for `tryEmptyPassphrase` plus the first-boot `luks-enroll-tpm` unit.

That unit leaves the empty slot in place whenever TPM enrollment fails, and it skips the wipe permanently if it is interrupted after enrolling. The review summary has the details.

Until the bootstrap design is final, please restrict this family to the debug variant: add `x.variant == "debug"` to the filter.

packages/pkgs-by-name/ghaf-sign-x86-image/ghaf-sign-x86-image.sh
```diff
@@ -0,0 +1,126 @@
...
+done
+
+mapfile -d '' ukis < <(find "$work/esp/EFI/Linux" -maxdepth 1 -type f -iname '*.efi' -print0 2>/dev/null)
+((${#ukis[@]} > 0)) || {
+  echo "No Type-2 UKI found under EFI/Linux; refusing an incomplete Secure Boot image" >&2
+  exit 1
+}
+for uki in "${ukis[@]}"; do
+  sign_one "$uki"
+done
```
brianmcgillion, line 107: **Important: this signs every EFI binary found in the input ESP, with nothing tying it to the build.**

Both loaders and every `EFI/Linux/*.efi` are signed with `db`, and any number of UKIs is accepted. The only trust check compares the operator's certificates with `public-trust.json`, which comes from the same unsigned `--input` directory as the image, so it says nothing about the binaries.

Anyone who can write the unsigned image before signing can add an `EFI/Linux/extra.efi` with a command line of their choice and get it db-signed. That image would then boot on every enrolled laptop. #2199's `ghaf-sign-update` had the same problem and now verifies each artifact's sha256 against the manifest before signing.

Please have `ghaf-prepare-x86-verity-disk` record the sha256 of every ESP file in its output, and have this script verify each copied binary against that list before `sbsign`, rejecting any `.efi` that is not listed. It would also be good to document that `--input` must be the immutable build output.

modules/partitioning/initial-verity-disk-x86.nix
```diff
@@ -0,0 +1,91 @@
...
+        message = "secure A/B imageSize is ${toString diskoCfg.imageSize} MiB but two fixed system slots and persistent volumes require at least ${toString minimumImageSizeMiB} MiB";
+      }
+    ];
+
+    ghaf.partitioning.disko.imageSize = lib.mkDefault minimumImageSizeMiB;
+
+    fileSystems."/boot" = lib.mkForce {
+      device = "/dev/disk/by-label/ESP";
+      fsType = "vfat";
+      options = [ "umask=0077" ];
```
brianmcgillion, line 85: **Important: `/boot` is mounted by filesystem label, which anything attached can claim.**

`main` mounts the ESP by partition label (`/dev/disk/by-partlabel/disk-disk1-ESP`) with `nofail` (`disko-debug-partition.nix`). Here it is `/dev/disk/by-label/ESP` without `nofail`. Any attached FAT volume labelled `ESP` can win the udev symlink race: another secure A/B stick, or anything formatted with `mkfs.vfat -n ESP`. The sysupdate UKI writes and boot-counting state then go to the wrong device.

The prepare script names the partition plain `ESP` (`--change-name=1:ESP`), so switching to `by-partlabel/ESP` would be just as generic. Naming it `disk-disk1-ESP`, as mainline's disko layout does (the LUKS partition already follows that scheme as `disk-disk1-luks`), and mounting `/dev/disk/by-partlabel/disk-disk1-ESP` would bring it in line with `main`.

packages/pkgs-by-name/ghaf-sign-x86-image/ghaf-sign-x86-image.sh
```diff
@@ -0,0 +1,126 @@
...
+
+mkdir -m 0700 "$output"
+bmaptool create "$work/ghaf-image.raw" -o "$output/ghaf-image.bmap"
+zstd -T0 --compress "$work/ghaf-image.raw" -o "$output/ghaf-image.raw.zst"
+install -m 0644 \
+  "$key_dir/PK.auth" \
+  "$key_dir/KEK.auth" \
+  "$key_dir/db.auth" \
+  "$key_dir/db.crt" \
+  "$output/"
```
brianmcgillion, line 124: Minor: the `.auth` enrollment payloads are copied without checking them against the certificates whose digests were just verified. The installer then checks the ESP against `db.crt` and enrolls `db.auth`, again with no check that the two match. A stale or mismatched `.auth` set would enroll an unintended PK/KEK/db, which is the failure the pre-enrollment check exists to prevent. Extracting the certificate from each `.auth` (strip the auth header, then `sig-list-to-certs`) and comparing it with the matching `.crt`, in both tools, would close this.

brianmcgillion (changes requested):
## Re-review at `96ffd4df`: requesting changes

This head is a rebase onto #2199 `0a08af39`. The PR's own diff is line-for-line identical to `8b640ac6`, which I commented on yesterday, and those threads have no replies. I also did a wider pass for security issues and changes from mainline. Each finding below was checked in code at this head, or by evaluating the configurations, and compared with `main`.

### Blocking

**1. Every laptop shares one LUKS volume key.** This is still unaddressed (see https://github.com/tiiuae/ghaf/pull/2201#pullrequestreview-5282150071 for a tested installer-time rekey). The flake now also exposes release variants: `intel-laptop-release-secure-ab` and `intel-laptop-release-secure-ab-sysupdate`.

**2. Release laptops can keep an empty-passphrase keyslot permanently.** Evaluated on `intel-laptop-release-secure-ab`: `ghaf.storage.encryption.deferred = false`, `luks-enroll-tpm` enabled and wanted by `sysinit.target`, and `boot.initrd.luks.devices.crypted.tryEmptyPassphrase = true`. On `main`, `intel-laptop-release` uses `deferred = true`, which gives a user passphrase and a per-device key, and has no `luks-enroll-tpm` unit. So this PR moves release laptops onto a path main does not use for them.

That path has two ways to leave the empty passphrase in place. The unit script is a `writeShellApplication`, so it runs under errexit:
- **TPM enrollment fails:** `systemd-cryptenroll` failing (no TPM, TPM lockout, or the PIN prompt at sysinit not answered) aborts the script before `--wipe-slot=password`.
- **Interrupted after enrollment:** the check at the top exits 0 on any boot where a `systemd-tpm2` token already exists. A power cut between enrollment and the wipe therefore means the wipe never runs.

In both cases the disk opens with no secret on every boot, and the boot-health gate still accepts it. The unit itself is mainline code, but this PR is what sends release laptops through it. Please either keep the secure family debug-only until the bootstrap design is final, or make the wipe idempotent: run it whenever a password slot remains, not only when no token exists, and fail closed if enrollment cannot complete.

### Important
3. **`ghaf-sign-x86-image` signs every EFI binary in the input ESP without checking it** (inline). This is the same class as the signer finding fixed in #2199.
4. **`/boot` is mounted by filesystem label** (inline).

### Minor
5. The `.auth` enrollment files are copied (and later enrolled by the installer) without checking that they match the certificates whose digests were verified (inline).
6. The disk GUID, partition GUIDs and LUKS UUID are fixed constants shared by every install, and nothing re-randomises them at first boot.
7. The x86 guide says normal laptop targets are unchanged, but `ghaf-installer -s` now refuses Type-1 images. I think the refusal is right: `main` never signs x86 Type-1 images, so enrolling keys there would leave a laptop that cannot boot. The docs should say so rather than claim nothing changed.

### Checked and fine
- **`printf -v`:** limited to a whitelist of 11 fixed names.
- **`find -H`:** follows only the top-level argument.
- **Partition maths:** the minimum size is checked in both Nix and the script.
- **No private keys in the store.**
- **Signer hygiene:** the signer's temporary files and its refusal to overwrite output are sound.
- **Installer ESP lookup:** scoped to the target disk.
- **Installer failures:** it fails closed on a missing `db.crt`, missing `sbverify` or a bad signature.
- **Debug and release targets:** they use distinct `secureUpdate.target` values.
- **Existing targets unchanged:** `intel-laptop-{debug,release}`, `intel-laptop-debug-sysupdate` and `lenovo-t14-amd-gen5-debug` derivations match #2199.
- **Checks:** `nix fmt -- --fail-on-change` and `reuse lint` pass. `ghaf-prepare-x86-verity-disk.tests.{image,plan}` and `ghaf-sign-x86-image.tests.sign` pass. All four secure targets evaluate with an exported public trust config.

Not covered: no hardware install.

brianmcgillion (changes requested):
## Retest at `d82f9e5c`: still requesting changes

Thanks. Moving to mainline's deferred encryption is the right fix, and it resolves both blockers. I've replied on each thread; summary:

| Finding | Status |
|---|---|
| Every laptop shares one volume key | **Fixed.** The image ships plain LVM plus the installer marker, and `first-boot-encrypt` encrypts in place with a fresh key. `checks.x86-deferred-encryption` passes and asserts two installs get different keys. |
| Release laptops can keep an empty-passphrase slot | **Fixed.** Evaluated on `intel-laptop-release-secure-ab`: `deferred = true`, `interactiveSetup = true`, `tryEmptyPassphrase = false`, and no `luks-enroll-tpm` unit (deferred mode enrolls TPM2/FIDO2 and recovery itself, as on `main`). `first-boot-encrypt` is `requiredBy` `sysroot.mount` with an `isLuks` post-check, so a disk that did not get encrypted fails closed; the test covers a missing marker. |
| `ghaf-sign-x86-image` signs every EFI binary unchecked | **Open.** |
| `/boot` mounted by filesystem label | **Open.** |
| `.auth` files not checked against the certificates | **Open**, minor. |

### Important
- **The signer:** `ghaf-sign-x86-image` still signs both loaders and every `EFI/Linux/*.efi` in the input image with `db`, and nothing ties them to the build. This is the same class we required fixed in #2199's `ghaf-sign-update`. Please record the sha256 of each ESP file in the prepare output and verify each one before signing, rejecting any `.efi` that is not listed.
- **`/boot` mount:** `initial-verity-disk-x86.nix:83` still mounts `/boot` from `/dev/disk/by-label/ESP` without `nofail`, and any attached FAT volume labelled `ESP` can take it. `main` mounts `by-partlabel/disk-disk1-ESP`.

### Minor
- **`.auth` payloads:** they are still copied and enrolled without checking that they match the certificates whose digests were verified.
- **Fixed GUIDs:** the disk and partition GUIDs are still fixed constants shared by every install. The LUKS UUID is now generated on the device.

Checks at `d82f9e5c`: `nix fmt -- --fail-on-change` and `reuse lint` pass. These tests pass: `checks.x86_64-linux.x86-deferred-encryption`, `ghaf-prepare-x86-verity-disk.tests.{image,plan}` and `ghaf-sign-x86-image.tests.sign`. Existing laptop derivations match #2199. Not covered: the interactive release passphrase flow in the real initrd (the VM test runs the same script in stage 2), and hardware.

packages/pkgs-by-name/ghaf-prepare-x86-verity-disk/ghaf-prepare-x86-verity-disk.sh
brianmcgillion: Fixed at `d82f9e5c`. The image no longer contains LUKS: it ships plain LVM plus the installer marker, and mainline's deferred encryption creates a per-device volume key on first boot. `checks.x86-deferred-encryption` asserts two installs get different keys.

modules/partitioning/secure-ab-x86.nix
brianmcgillion: Fixed at `d82f9e5c`. `deferred = true` is restored, and `first-boot-encrypt` is now `requiredBy` `sysroot.mount` with an `isLuks` post-check, so a disk that did not get encrypted fails closed instead of booting in plaintext.

targets/laptop/flake-module.nix
brianmcgillion: Fixed at `d82f9e5c`. Evaluated on `intel-laptop-release-secure-ab`: `interactiveSetup = true`, `tryEmptyPassphrase = false`, and no `luks-enroll-tpm` unit (deferred mode enrolls TPM2/FIDO2 and recovery inside `first-boot-encrypt`, as `main`'s laptops do). The release variants are fine to keep now.

packages/pkgs-by-name/ghaf-sign-x86-image/ghaf-sign-x86-image.sh
brianmcgillion: Still open at `d82f9e5c`: the script is unchanged and still signs every `EFI/Linux/*.efi` without checking it against the build.

modules/partitioning/initial-verity-disk-x86.nix
brianmcgillion: Still open at `d82f9e5c`: `initial-verity-disk-x86.nix:83` still uses `/dev/disk/by-label/ESP` without `nofail`.

packages/pkgs-by-name/ghaf-sign-x86-image/ghaf-sign-x86-image.sh
brianmcgillion: Still open at `d82f9e5c`, minor.

PR #2218, opened 2026-09-04
feat(crosvm): enable DAX for nix store share, use hugepages

kajusnau:
## Description of Changes

Two changes to how VMs use memory.

### 1. DAX for the shared nix store

VMs now read the nix store straight out of the host's memory instead of each keeping a private
copy, so all nine VMs share one copy. Can save up to **800 MB of memory per VM**.

#### Without DAX - round trip for every read

```mermaid
sequenceDiagram
    participant Guest
    participant Host

    Guest->>Host: open /nix/store/abc-hello/bin/hello
    Host-->>Guest: ok

    rect rgba(220, 80, 80, 0.15)
    Note over Guest,Host: DAX removes this part
    loop every read
        Guest->>Host: send me the bytes
        Host-->>Guest: copies the bytes across
    end
    end
```

#### With DAX - set up once, then read straight from memory

```mermaid
sequenceDiagram
    participant Guest
    participant Window as DAX window
    participant Host

    Guest->>Host: open /nix/store/abc-hello/bin/hello
    Host-->>Guest: ok

    rect rgba(80, 160, 80, 0.15)
    Guest->>Host: put the file in the window
    Host->>Window: shares the file here, no copying
    Host-->>Guest: ok
    end

    loop every read
        Guest->>Window: reads the bytes directly
    end
```

#### Downsides

- Repeat reads are ~15% slower, but this is somewhat inconsistent.
- x86 targets only, ignored on Jetson/ARM.
- Needs slightly relaxed VM sandboxing (`modules/common/systemd/hardened-configs/microvm.nix`).
- Covers file contents only. File names and directory structure still cost (worst case) ~370 MB per VM, but
  that only shows up when something scans the whole store, which does not happen in normal use.

### 2. Huge pages for VM memory

The host now hands out VM memory in 2 MB chunks instead of 4 KB, so the processor tracks fewer
pieces.

_Should_ help everyday workloads like browsing, but **not measured yet**.

## Measurements

Four VMs each reading the same 1.5 GB from the nix store.

| Measurement                      | Without DAX  | With DAX      |
| ------------------------------------- | ------------ | ------------- |
| Real memory used by the four VMs      | 6.0 GB       | **2.8 GB**    |
| Copy of file data kept inside each VM | ~1.9 GB      | **none**      |
| Total host memory used by the test    | 10.5 GB      | **2.2 GB**    |
| First read speed                      | 0.97 GB/s    | **1.10 GB/s** |
| Repeat read speed                     | **6.8 GB/s** | 5.7 GB/s      |

Because nothing is copied into the VMs, reading 12 GB in total costs zero extra memory.
Day-to-day Ghaf also feels more responsive, but that may be subjective.

## Read more about DAX

- https://virtio-fs.gitlab.io/design.html
- https://lwn.net/Articles/813807/
- https://www.kernel.org/doc/html/latest/filesystems/dax.html

### Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

**MERGED** - upstream PR https://github.com/microvm-nix/microvm.nix/pull/598

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [x] Requires full re-installation - **recommended**
- [ ] Other:

### Test Steps To Verify:

1. Some basic regression tests - verify no VMs are broken, not booting, crashing, etc.
2. Verify improved responsiveness of Ghaf during normal use
3. Some memory usage tests, if available in test automation or manual


leivos-unikie:
Triggered some extensive automated test runs on lenovo-x1 and darter pro.

darter-proANDperformance
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/7883/

lenovo-x1ANDperformance
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/7884/

- Performance results show around 5-10% improvement in sysbench cpu and memory tests within VMs on both laptops.

darter-proANDregressionNOTsecboot-only
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/7890/
- ssh connection to gui-vm dropped in the end of GUI tests. Up to that no other than known issues.
- launched new run for the rest
darter-proANDsecurityORdarter-proANDsuspension
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/7891/
- which was green

leivos-unikie:
Tested storeDisk-installer with encryption manually on darter-pro and lenovo-x1. Didn't notice any regression. UI works fine.

modules/common/systemd/hardened-configs/microvm.nix
```diff
@@ -89,7 +85,11 @@
   # Capabilities #
   ################
 
-  #AmbientCapabilities=
+  # CapabilityBoundingSet only raises the ceiling for a non-root unit; AmbientCapabilities grants it.
+  AmbientCapabilities = [
+    "CAP_SYS_ADMIN"
```
brianmcgillion, line 91: Full root access :)

brianmcgillion (approved):

vadika (changes requested):
Lets either isolate the feature to x86 platform or make proper testing for nvidia also.

kajusnau:
Update:
  - DAX gated behind x86 check
  - `CAP_SYS_ADMIN` still granted. Sucks, but seems to be needed for DAX on crosvm atm.
  That being said, the service runs as an ordinary user in its own namespace, so it mostly applies inside crosvm's sandbox, and now comes with fewer permissions:
    - 15 permissions removed from `microvm` service, including clock setting, kernel log reading, audit writing, BPF loading, security policy changes, raw sockets, performance monitoring, checkpointing.
    - 3 system call groups blocked: CPU emulation, raw port I/O, kernel keyring
    
Tested that Ghaf boots, all VMs are up, but a QA re-test wouldn't hurt here, so re-adding the label.

modules/common/systemd/hardened-configs/microvm.nix
```diff
@@ -89,46 +86,53 @@
...
+    "CAP_SYS_CHROOT"
+  ];
```
mbssrc: Currently this profile widening is applied in all cases - not only when DAX is enabled. I would suggest gating it depending on DAX being enabled.

For the containment itself; we seem to need `CAP_SYS_ADMIN` for the setup of crosvm base jail - enabling its full sandbox would probably be preferable as we also get per-device namespaces and seccomp policies.

modules/microvm/host/microvm-host.nix
```diff
@@ -450,5 +450,13 @@ in
         ACTION=="add", SUBSYSTEM=="backlight", ATTR{brightness}="$attr{max_brightness}"
       '';
     }
+    {
+      # crosvm backs guest RAM with memfd, so the MADV_HUGEPAGE it issues under
+      # --hugepages is governed by shmem_enabled rather than enabled. The default
+      # "never" ignores that advice, leaving guest RAM entirely 4 KiB-backed.
+      systemd.tmpfiles.rules = [
+        "w /sys/kernel/mm/transparent_hugepage/shmem_enabled - - - - advise"
```
mbssrc, line 459: Does this work with systemd-tmpfiles-setup.service profile that sets ProtectKernelTunables=true?

modules/common/systemd/hardened-configs/microvm.nix
kajusnau: >Currently this profile widening is applied in all cases - not only when DAX is enabled. I would suggest gating it depending on DAX being enabled.

I would prefer that also. Any ideas how to introduce a gate to the hardened profile without "muddying" `modules/common/systemd/hardened-configs/default.nix` with a config import? Or is that acceptable in this case.

modules/microvm/host/microvm-host.nix
kajusnau: It seems to work and is required for hugepages

modules/common/systemd/hardened-configs/microvm.nix
kajusnau: Work in progress, will update the PR when ready

modules/microvm/common/store-shared-virtiofs.nix
```diff
@@ -42,11 +44,19 @@ in
             source = "/nix/store";
             mountPoint = "/nix/.ro-store";
             proto = "virtiofs";
+            # crosvm implements virtiofs DAX on x86_64 only
+            # elsewhere the flag is accepted and silently ignored, but we're explicit here
+            dax = isCrosvm && pkgs.stdenv.hostPlatform.isx86_64;
```
mbssrc, line 49: With the same host memory being mapped, this opens the road for FLUSH+RELOAD class SCA (on top of the PRIME+PROBE style attacks) - much clearer signal and easier to implement. 

This won't go away, and generally I would argue that storedisk is the better option: isolated store + no host daemon/service (less attack surface), and also future proof if pKVM / memory encryption would land. 
But this doesn't affect this PR as it is an optimization for the non-storedisk option.

kajusnau:
Rebased, adjusted config based on comments
Removing testing label until re-reviewed

modules/common/systemd/hardened-configs/microvm.nix
kajusnau: Introduced `modules/microvm/host/crosvm-dax.nix` - the profile widening is now applied only when DAX is enabled (which is further gated behind x86 and non-storeDisk)

milva-unikie:
I quickly tested this on Darter Pro with `intel-laptop-debug`. It feels extremely snappy and very similar to the responsiveness of storeDisk images.

There are big improvements in the app launch times, most notably in Cosmic Settings
image

When comparing with the mainline version of `intel-laptop-debug`, many apps launch 50% faster with this PR and are only 0.5 seconds slower than with a storeDisk image.

kajusnau:
Rebased on top of latest main, which includes bump to microvm.nix, so it's unnecessary to bump here.

@vadika please re-review when possible

mbssrc (approved):

brianmcgillion (approved):

PR #2226, opened 2026-09-07
fix(chromecast/uplink-resolver): support multiple simultaneous uplinks for network-related services

enesoztrk:
## Description of Changes
`ghaf-uplink-resolver` only ever published a single LAN-facing interface (the lowest-metric default route), and `nw-packet-forwarder`, `smcroute`, and the uplink firewall chains only ever acted on that one interface. On a device with more than one simultaneous uplink (Wi-Fi *and* a docked Ethernet, etc. ), Chromecast traffic on the network the device *didn't* pick was never
forwarded, routed, or NAT'd at all. One `nw-pckt-fwd` process was  ever running regardless of how many uplinks
were actually up.

This PR makes the whole uplink-driven stack multi-uplink aware, and fixes
several bugs found in ghaf-uplink-resolver.

### Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets
SSRCSP-8634

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

Run every command in net-vm below. Please activate at least two network interfaces.
1. Run ```cat /run/ghaf-uplink-state```
**Expected:**
- `uplink_state=resolved`
- `uplink_ifaces="<iface1> <iface2>"` lists **both** interfaces
- `uplink_reason=`
2. Run 
``` 
systemctl status 'nw-packet-forwarder@*.service' 
ps aux | grep nw-pckt-fwd
```
**Expected:**
- One `nw-packet-forwarder@<iface>.service`, **active (running)**, per
  interface from the first test step — not just one
- Each corresponding `nw-pckt-fwd` process shows a *different*
  `--external-iface` in its command line
- `journalctl -u nw-packet-forwarder-reconcile.service` shows no errors
3. Run
```
cat /run/smcroute/smcroute.conf
systemctl status smcroute.service
journalctl -u smcroute --no-pager | tail -20
```

**Expected:**
- `smcroute.service` is **active (running)**, not failed
- The journal shows `smcroute: routing multicast on <iface1> <iface2>` (both
  named) with **no** `Parse error` line
- The rendered conf contains an `mgroup`/`mroute` block for each uplink
4. Run
```
sudo iptables -nvL ghaf-fw-uplink-fwd-filter
sudo iptables -t nat -nvL ghaf-fw-uplink-post-nat
```
**Expected:**
- Each chain has a matching pair of rules **per uplink interface** (e.g. for
  two uplinks, 4 lines in `fwd-filter`: 2 ports (8008,8009) × 2 interfaces), not just
  one interface's worth
5. Launch Chrome and try to cast a web page/youtube video through Chromecast


brianmcgillion (changes requested):
Tested on hardware: built `intel-laptop-debug` from 03ce06d13 and netboot-installed it on a Lenovo X1 Carbon Gen 11, then ran the tags relevant to this change (`pre-merge`, `net-vm`, `firewall`, `network-security`, `ip-spoofing`, `network`) from ci-test-automation — 70 passed, 12 failed, and none of the failures are attributable to this PR (camera/lid, Secure Boot off because the rig netboots, absent wifi secrets, and a ci-test-automation gap where all 8 `network` tests die on a missing `Save Networking Speed Data` keyword while the iperf3 measurements underneath pass at line rate).

**The multi-uplink work does what it says.** I brought up the dock ethernet and wifi simultaneously and walked your test plan:

* `uplink_ifaces="enp0s10u1 wlp0s5f0"`, `uplink_state=resolved`, empty reason
* two `nw-packet-forwarder@<iface>.service` units, each `nw-pckt-fwd` process carrying a different `--external-iface`
* `smcroute` active, `smcroute: routing multicast on enp0s10u1 wlp0s5f0`, no parse error, an `mgroup`/`mroute` block per uplink
* 4 rules in `ghaf-fw-uplink-fwd-filter` (2 ports × 2 interfaces) and 6 in `ghaf-fw-uplink-post-nat` — exactly the counts you predict
* disconnecting one uplink tore down correctly: `stopping forwarder on wlp0s5f0 (no longer an uplink)`, conf and chains reverted

## Blocking: `uplink_reason` is written unquoted, so teardown breaks when the *last* uplink goes

`modules/common/networking/uplink-resolver.nix:107` writes

```nix
printf 'uplink_reason=%s\n' "$reason"
```

while line 105 correctly quotes the new list as `uplink_ifaces="%s"`. Every `state=none` reason is multi-word — `"no default route"`, `"default route is on the internal interface $internal"`, `"interface ... does not exist"` — and consumers read the file with `.` (POSIX source), so line 3 becomes `uplink_reason=no default route`: the shell assigns `uplink_reason=no` and then tries to *execute* `default route`.

`nw-packet-forwarder-reconcile.service` is the only consumer without `ConditionPathExists = readyFlag` — correctly so, since it has to run in order to tear down — which makes it the only one that ever executes in the `state=none` window. `writeShellApplication` sets `set -e` and the `. ${cfg.stateFile}` sits at the top of the script, so the service aborts **before doing any reconciliation**.

Observed on the device, with both uplinks dropped:

```
uplink_state=none / uplink_reason=no default route
reconcile[16830]: /run/ghaf-uplink-state: line 3: default: command not found
reconcile.service: Main process exited, code=exited, status=127/n/a

  nw-packet-forwarder@enp0s10u1.service  active running            <- stale, no uplink exists
  nw-packet-forwarder@wlp0s5f0.service   activating auto-restart   <- crash-looping on a departed iface
```

Both forwarders survive with zero uplinks and the departed-interface instance respawns indefinitely. That is the unplug path this PR exists to handle. It also reproduces on every cold boot, in the window before the dock comes up.

Two things make this easy to miss:

1. It is **invisible to `systemctl --failed`** — the resolver re-runs when an uplink appears, the successful run resets the unit, and the failure survives only in the journal. A CI gate on failed-unit counts will not catch it.
2. The hazard is latent in `main` too (`uplink_reason` is unquoted there as well), but `main`'s three consumers are all readyFlag-gated, so they never read the file in that state. This PR is what makes it reachable.

Fix is one line at the writer:

```nix
printf 'uplink_reason="%s"\n' "$reason"
```

Quoting at the writer is the right place, since the other three consumers (`nw-packet-forwarder.nix`, `smcroute.nix`, `firewall.nix`) all `source` the same file and are protected only by the ready-flag gate — a race away from the same failure.

## Please rebase

The branch is based on `eecfaf595` and is 30 commits behind `main`. That range includes `7b49424bf bump(all): s2 bump`, so the results above were measured against a pre-bump base and do not automatically carry over. Worth re-running the uplink checks after the rebase.

## Minor / non-blocking

* On my rig both uplinks landed on the same `/24` behind one gateway, which produced `smcroute: Moved (192.168.10.138,239.255.255.250) inbound to wlp0s5f0 after WRONGVIF` — multicast flapping between VIFs on the same L2. I read that as an artifact of the test setup rather than a defect here, but it would be worth confirming on genuinely separate networks, since "wifi + docked ethernet" in the wild often is one LAN.
* Step 5 (an actual Chromecast session) is not covered — that needs a GUI session and a real device.
* `dendrite-pinecone` emits nothing at all on the `mvp-user-trial` profile, so its share of this change has no runtime coverage in my run.

brianmcgillion:
Second hardware run, same head (03ce06d13), cross-built as `nvidia-jetson-orin-agx-debug-from-x86_64` and flashed to an Orin AGX. `pre-merge` 15/2, `functional` 21/3, all failures pre-existing, zero new failed units vs a pre-flash baseline. Note this profile has `dendrite = false` and no chromecast, so it exercises `uplink-resolver.nix` only. Two non-blocking things.

## Commit 1's message describes an API the series removes

`bda022574` says `uplink_iface` is "kept for the pin-disagreement warning and back-compat", but `03ce06d13` drops the `printf`, and nothing in the tree references the singular key any more:

```
$ grep -rn 'uplink_iface\b' --include='*.nix' . | grep -v uplink_ifaces
$
```

On the device the state file is three lines:

```
uplink_ifaces="eth1"
uplink_state=resolved
uplink_reason=
```

Code is fine — the warning reads `$pinned`/`$routed`, and the test was updated. Just a reword or squash so the history doesn't advertise a key that isn't there.

## Pre-existing: `firewall.uplink` can be enabled where the apply unit never runs

Reproduces on `main`, so not yours — but it's the same "acts on no uplink" shape, in code this PR touches.

`ghaf-firewall-uplink.service` (`firewall.nix:806`) is gated on `ConditionPathExists = cfg.uplink.readyFlag` (`:826`). If the flag is absent when it first starts, only `uplinkResolver.dependentUnits` ever starts it again — and just `chromecast.nix:85` and `dendrite-pinecone.nix:127` add it there. `wireguard-gui-vmconfig.nix:118` enables the feature without adding anything, which on the Orin profile is the only thing enabling it:

```
○ ghaf-firewall-uplink.service - Apply firewall rules for the resolved uplink interface
     Active: inactive (dead)
  Condition: start condition unmet
             └─ ConditionPathExists=/run/ghaf-uplink-ready was not met
```

The flag doesn't exist until DHCP produces a default route, so the unit is skipped at boot and never retried; `/run/ghaf-fw-uplink-rules` is never written and the chains stay empty for the whole boot. Harmless only because wireguard-gui's rule lists are empty (its own comment says no WireGuard server port has ever been DNAT'd). Populate them and it silently applies nothing.

Fix would be for the firewall module to register itself under the same `mkIf cfg.uplink.enable`:

```nix
ghaf.networking.uplinkResolver.dependentUnits = [ "ghaf-firewall-uplink.service" ];
```

Fine as a separate issue if you'd rather not widen this PR.

brianmcgillion:
Filed the second item as #2228 so it doesn't widen this PR.

leivos-unikie:
NOTE:
Last pre-merge test https://ci-prod.vedenemo.dev/job/ghaf-pre-merge/3074 failed boot test on darter-pro (release setup). Looks like robot finger for "Enter" had failed there. I launched pre-merge for that manually:
https://ci-prod.vedenemo.dev/job/ghaf-hw-test-manual/2546/
It passed except sec boot test which I forgot to exclude (NOTsecboot-only).

kajusnau (changes requested):
Please reduce comment verbosity - https://github.com/tiiuae/ghaf/blob/main/AGENTS.md#code-comments

modules/reference/personalize/authorizedSshKeys.nix
kajusnau: Should probably be its own commit

modules/reference/services/nw-packet-forwarder/nw-packet-forwarder.nix
```diff
@@ -23,51 +23,63 @@ let
...
-        # shellcheck disable=SC1090
-        . ${cfg.uplink.stateFile}
-        if [ -z "''${uplink_iface:-}" ]; then
-          echo "nw-pckt-fwd: ${cfg.uplink.stateFile} names no uplink; refusing to forward on a guess." >&2
-          exit 1
+
+  # One forwarder per resolved uplink (template unit), so a device with
+  # several forwards on all of them. `%I` is passed as an argv element, not
+  # spliced into this string, so it can't collide with `chromecastFlags`.
+  nw-pckt-fwd-instance = pkgs.writeShellScriptBin "nw-pckt-fwd-instance" ''
```
kajusnau: Would prefer using `writeShellApplication`, then using `getExe` for the call in line 186

kajusnau: Does this even need to exist? Looks like a simple wrapper for `ghaf-nw-packet-forwarder`. The only thing it seems to achieve is print the echo line, the rest can be done within `nw-packet-forwarder@` by calling `nw-packet-forwarded` with the relevant cfg attrs

modules/reference/services/smcroute/smcroute.nix
```diff
@@ -8,96 +8,110 @@
...
+        https://github.com/troglobit/smcroute?tab=readme-ov-file#usage
+
+        Rendered once per resolved uplink, with `placeholder` substituted for
+        the interface being rendered. Put anything here that must be a
+        *separate* directive per uplink (e.g. `mgroup from @UPLINK@ ...`, or
+        an mroute whose source is the uplink). A directive that instead needs
+        every uplink named on the same line (e.g. an mroute fanning out from
+        the internal interface to all of them) belongs in `rulesOnce`
+        instead -- repeating it here would redeclare the same route with a
+        different destination each time, which smcrouted rejects.
```
kajusnau: This is not very human readable IMO

`(e.g. ` stuff should be an `example` attr, not part of the description.

modules/reference/services/smcroute/smcroute.nix
```diff
@@ -8,96 +8,110 @@
...
       '';
     };
 
-    rules = lib.mkOption {
+    rulesOnce = lib.mkOption {
```
kajusnau: Would it make sense to have this as a sub-option of `rules` - `rules.once`

modules/reference/services/smcroute/smcroute.nix
```diff
@@ -8,96 +8,110 @@
...
       type = lib.types.nullOr lib.types.lines;
       default = null;
       description = ''
-        https://github.com/troglobit/smcroute?tab=readme-ov-file#usage
+        Rendered exactly once, with `placeholderAll` substituted for every
+        resolved uplink, space-separated -- for directives that must name all
+        of them on one line, such as an mroute's `to` list. See `rules` for
+        the per-uplink counterpart.
```
kajusnau: Readability could be improved, similar to comment above.

modules/reference/services/smcroute/smcroute.nix
```diff
@@ -8,96 +8,110 @@
...
-        description = ''
-          Gate for the unit. Absent means there is no uplink, and the unit is
-          skipped rather than failed.
-        '';
-      };
+    readyFlag = lib.mkOption {
+      type = lib.types.path;
+      default = "/run/ghaf-uplink-ready";
+      description = ''
+        Gate for the unit. Absent means there is no uplink, and the unit is
```
kajusnau: `absent` here probably means the file being pointed to is absent, not that the option is absent.
The option probably can't be absent at all, since it's of type `lib.types.path`, which requires a valid path value AFAIK.

modules/reference/services/smcroute/smcroute.nix
```diff
@@ -8,96 +8,110 @@
...
     assertions = [
       {
-        assertion = cfg.uplink.enable || cfg.bindingNic != "";
-        message = "Binding Nic must be set, or services.smcroute.uplink.enable used";
+        assertion = config.ghaf.networking.uplinkResolver.enable;
+        message = ''
+          services.smcroute.enable requires ghaf.networking.uplinkResolver.enable
+          -- smcroute always renders its config from the resolved uplink, and
+          without the resolver it would never see a ready flag and would sit
+          skipped forever instead of routing anything.
```
kajusnau: I think it's enough to say `services.smcroute.enable requires ghaf.networking.uplinkResolver.enable`

If the rest is necessary for clarity, it should be made simpler, as it's an assertion message, not an option description.

modules/reference/services/smcroute/smcroute.nix
```diff
@@ -8,96 +8,110 @@
...
+          failure this mechanism exists to prevent.
+        '';
+      }
+      {
+        assertion = cfg.rulesOnce == null || lib.hasInfix cfg.placeholderAll cfg.rulesOnce;
+        message = ''
+          services.smcroute.rulesOnce never mentions ${cfg.placeholderAll}.
+          If a directive doesn't need the uplink list, it belongs in `rules`
+          instead -- rulesOnce exists only for directives that must name
+          every uplink on one line.
```
kajusnau: I think we don't need to over-explain why `rulesOnce` exists. Just point the user in the right direction, they can read option descriptions.

modules/reference/services/smcroute/smcroute.nix
```diff
@@ -117,94 +131,62 @@ in
...
-                echo "smcroute: set services.smcroute.uplink.enable to resolve the interface at runtime instead." >&2
-                exit 1
-              fi
-              ${pkgs.coreutils}/bin/sleep 2
-            done
-          '';
+      # Renders the config once per resolved uplink, so a device with
+      # several gets multicast routed on all of them, not just one.
+      preStart = ''
+        # shellcheck disable=SC1090,SC1091
```
kajusnau: I think this could all be its own script or `writeShellApplication` to enable shellchecking and simplify the service declaration.

Also why not have this in `ExecStartPre`?

kajusnau (changes requested):
It continues as well..

In addition to the comments, I think it would be beneficial to use `lib.getExe` and `lib.getExe'` throughout this patch, to reduce the amount of hard-coded binary calls.

Issue #2092, opened 2026-08-02
journal-fss-verify left permanently failed after a Robot clock test: clock-jump watcher misses sub-interval round-trip jumps


brianmcgillion:
## Summary

After a routine `ci-test-automation` run, `journal-fss-verify.service` on ghaf-host is left **permanently failed** and does not self-heal. The recovery designed to prevent exactly this never fires on the host, because of how the clock-jump watcher measures drift.

Observed on a Dell Latitude 7330 (`intel-laptop-low-mem-debug`, freshly netboot-installed) on 2026-08-02.

## What happens

The Robot suite deliberately rolls the RTC back to test time handling. journald rotates *during* the jump, leaving an archive stamped 2023 that later receives 2026-stamped FSS tags:

```
11:20:21  sudo: hwclock --set --date=01/11/23 11:00:00 UTC
Jan 11 11:00:04  systemd-journald: Time jumped backwards, rotating.
11:20:20  systemd-timesyncd: System clock time advanced to recorded timestamp: 2026-08-02 11:20:20 UTC
11:20:26  journal-fss: AUDIT_LOG_INTEGRITY_FAIL: Journal integrity verification FAILED [BAD_MESSAGE,ARCHIVED_SYSTEM]
11:20:26  journal-fss-verify: Older entry after newer tag (1673434804000119 < 1785669300000000)
11:20:26  journal-fss-verify: FAIL: ...-0000000000002e15-0005f1fae70c1577.journal (Bad message)
```

The two timestamps decode to 2023-01-11 09:40 UTC and 2026-08-02 11:15 UTC.

The same unit ran **clean twice on the same boot** (11:00:02, 11:10:37) before the clock test, so this is not a first-boot or key-setup artefact.

## Root cause

`ghaf-clock-jump-watcher` → `ghaf-journal-alloy-recover` exists to turn a jump-born archive into an allowlisted receipt (`record_recovery_archives`, `modules/common/logging/common.nix:328`). **It never fired on ghaf-host.**

The watcher (`modules/common/logging/common.nix:236-254`) samples every `intervalSeconds` and compares *net* drift between consecutive samples:

```sh
drift="$(awk ... 'BEGIN{print (r2-r1) - (u2-u1)}')"
if awk -v a="$abs" -v t="$threshold" 'BEGIN{exit !(a>=t)}'; then
  systemctl start ghaf-journal-alloy-recover.service || true
fi
```

[...]

PR #2239, opened 2026-09-09
fix(logging): keep FSS verification working across clock corrections

rodrigopinotii:
## Description of Changes

Relaxes the `journalctl --verify` `SEALED_CONTINUOUS` epoch gate to tolerate the several
same-epoch FSS TAGs journald legitimately writes when a sealed journal is rotated or
restarted repeatedly within one interval — today that trips a permanent
`Epoch sequence not continuous (E vs E)` on untampered archives and aborts the rest of the
file's verification. Adds a per-guest `journal-fss-shutdown-finalize` unit (+ an
`OnClockChange` companion) that seals and archives every open journal and stops journald's
listeners before a clock-step reboot, so no `system.journal` is left `STATE_ONLINE` to be
KEEP+APPEND-reopened with a straddling closing tag. Also fixes two independent false-negative
bugs in FSS activation confirmation (journal storage-location guessing, and a Seal=yes check
run too early after a restart). The verify-side change is a one-line `substituteInPlace`
against systemd's source, pending upstream (tracked in #2243); the host-side counterpart is a
separate `ghaf-sfo-laptop` PR.

Validation (Dell Latitude 7330): 12-correction field-cadence gate — zero new failing
`system@` archives on all 5 FSS VMs — plus the tamper suite (forgeries still rejected), the
FSS re-key regressions, and a forward/backward spot-check on the rebuilt image.

Fixes SSRCSP-8820.

Known follow-up, not caused by this change: `journal-fss-verify` can be left degraded after
two consecutive backward operator re-keys (periodic verify doesn't try retained keys) —
tracked in #2092.

### Type of Change

- [x] Bug fix

## Related Issues / Tickets

Fixes SSRCSP-8820. Related: #2092, #2243 (tracks upstreaming the systemd source change).

## Checklist

### Applicable Targets

- [x] Intel Laptop `x86_64`

### Installation Method

### Test Steps To Verify:

1. Deploy to an FSS-enabled target (`switch` + reboot, or reflash).
2. Force a forward and a backward wall-clock correction through your normal test harness.
3. On each FSS-enabled VM (net/audio/gui/admin/flatpak): `journalctl -u journal-fss-verify`
   shows no new FAILED verdicts; `journalctl --verify --verify-key=<verification-key>
   --file=<archived journal>` passes on the freshly-closed archive.
4. Confirm forgeries are still rejected: bit-flip an archived journal or backdate a seal
   tag — both must still FAIL.
5. `nix build .#checks.x86_64-linux.logging-fss .#checks.x86_64-linux.fss-classifier-unit`
   — both green.

---

_Commits reorganized 2026-09-11 into 3 logical groups per review: activation
false-negatives, the verify-side epoch tolerance, and the guest shutdown/clock-step
finalizer._

_2026-09-11 update: rebased onto current `main` (was 18 commits stale). Per round-2
review, dropped the vendored `.patch` for a one-line `substituteInPlace` in
`base.nix`/the test's systemd override, and trimmed commit/comment verbosity — build
verified (`journal-verify.c` still compiles, patched condition applies cleanly).
Pointing the upstream-tracking TODOs at an actual upstream systemd PR instead of
`#2243` is a deliberately held follow-up, sequenced after this and a related HW
campaign finish; not forgotten._

### DIFF
--- a/docs/src/content/docs/ghaf/scs/fss.mdx
+++ b/docs/src/content/docs/ghaf/scs/fss.mdx
@@ -195,6 +195,22 @@ journalctl -o export > journal.export
 journalctl --verify --verify-key=<verification-key> --file=journal.export
 ```
 
+:::caution[Use Ghaf's `journalctl` for off-device verification]
+Ghaf vendors a systemd patch (`systemd-fss-verify-tolerate-repeated-epoch.patch`)
+so `journalctl --verify` tolerates more than one FSS TAG per FSPRG interval —
+legitimate when a sealed journal is rotated/restarted repeatedly inside one
+`sealInterval`, which the FSS-activation sequence does on every boot. Stock
+upstream `journalctl` rejects such an untampered archive with
+`Epoch sequence not continuous (N vs N)`.
+
+Until the patch is upstream, verify exported archives with Ghaf's `journalctl`
+(or the same patch applied). A `FAIL` with *equal* operands is this known
+divergence, not tampering. *Unequal* operands `(M vs N)`, `M < N`, are a real
+rollback and fail under both.
+
+TODO: drop this note once the patch is upstream -- [tiiuae/ghaf#2243](https://github.com/tiiuae/ghaf/issues/2243).
+:::
+
 ## Troubleshooting
 
 ### Verification Fails with "FAIL"
--- a/modules/common/systemd/base.nix
+++ b/modules/common/systemd/base.nix
@@ -72,9 +72,16 @@ let
       }
     )).overrideAttrs
       (prevAttrs: {
-        patches = prevAttrs.patches ++ [
-          ./systemd-boot-double-dtb-buffer-size.patch
-        ];
+        patches = prevAttrs.patches ++ [ ./systemd-boot-double-dtb-buffer-size.patch ];
+        # Tolerate repeated same-epoch FSS TAGs in `journalctl --verify` (a forward
+        # clock correction re-tags a sealed file within one FSPRG interval).
+        # TODO: remove once upstreamed -- https://github.com/tiiuae/ghaf/issues/2243
+        postPatch = (prevAttrs.postPatch or "") + ''
+          substituteInPlace src/libsystemd/sd-journal/journal-verify.c \
+            --replace-fail \
+              'if (!(n_tags == 0 || (n_tags == 1 && le64toh(o->tag.epoch) == last_epoch)' \
+              'if (!(n_tags == 0 || le64toh(o->tag.epoch) == last_epoch'
+        '';
       });
 
   # Definition of suppressed system units in systemd configuration. This removes the units and has priority.
--- a/tests/logging/default.nix
+++ b/tests/logging/default.nix
@@ -25,6 +25,16 @@
 let
   fssSetupTest = import ./test_scripts/fss_setup.nix;
   fssVerificationTest = import ./test_scripts/fss_verification.nix;
+
+  # Same journal-verify.c tolerance as modules/common/systemd/base.nix.
+  patchedSystemd = pkgs.systemd.overrideAttrs (prev: {
+    postPatch = (prev.postPatch or "") + ''
+      substituteInPlace src/libsystemd/sd-journal/journal-verify.c \
+        --replace-fail \
+          'if (!(n_tags == 0 || (n_tags == 1 && le64toh(o->tag.epoch) == last_epoch)' \
+          'if (!(n_tags == 0 || le64toh(o->tag.epoch) == last_epoch'
+    '';
+  });
 in
 pkgs.testers.nixosTest {
   name = "logging-fss";
@@ -59,6 +69,8 @@ pkgs.testers.nixosTest {
       };
 
       config = {
+        systemd.package = patchedSystemd;
+
         # Enable FSS with short seal interval for testing
         ghaf.logging.enable = true;
         ghaf.logging.fss = {
@@ -82,14 +94,14 @@ pkgs.testers.nixosTest {
           "d /persist/common/journal-fss/test-host 0700 root root - -"
         ];
 
-        # Test utilities
+        # fss-test/fss-triage must verify with the same patched systemd as the node.
         environment.systemPackages = with pkgs; [
           coreutils
           gnugrep
           util-linux
-          (callPackage ./test_scripts/fss-test.nix { })
+          (callPackage ./test_scripts/fss-test.nix { systemd = patchedSystemd; })
           (callPackage ./test_scripts/fss-classifier-cases.nix { })
-          (callPackage ../../packages/pkgs-by-name/fss-triage/package.nix { })
+          (callPackage ../../packages/pkgs-by-name/fss-triage/package.nix { systemd = patchedSystemd; })
         ];
       };
     };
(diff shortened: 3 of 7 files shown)


docs/src/content/docs/ghaf/scs/fss.mdx
```diff
@@ -195,6 +195,24 @@ journalctl -o export > journal.export
 journalctl --verify --verify-key=<verification-key> --file=journal.export
 ```
 
+:::caution[Use Ghaf's `journalctl` for off-device verification]
+Ghaf ships a vendored systemd patch
+(`systemd-fss-verify-tolerate-repeated-epoch.patch`, SSRCSP-8820) that makes
```
kajusnau: Is this a jira ticket number? Is it a good idea to reference our jira tickets in the docs?

docs/src/content/docs/ghaf/scs/fss.mdx
```diff
@@ -195,6 +195,24 @@ journalctl -o export > journal.export
...
+`journalctl --verify` tolerate more than one FSS TAG per FSPRG interval. Journald
+legitimately writes several same-epoch TAGs when a sealed journal is
+rotated/restarted repeatedly inside one `sealInterval` (the FSS-activation
+sequence does this on every boot). **Stock upstream `journalctl` rejects such an
+archive with `Epoch sequence not continuous (N vs N)` even though it is
+untampered**, and aborts the rest of that file's verification.
+
+Until the patch is upstream, an external auditor verifying an exported archive
```
kajusnau: If this is to be removed after the upstream merge, then this deserves a `TODO` marker with a clear link to the upstream PR

modules/common/systemd/base.nix
```diff
@@ -74,6 +74,12 @@ let
       (prevAttrs: {
         patches = prevAttrs.patches ++ [
           ./systemd-boot-double-dtb-buffer-size.patch
+          # Vendored, pending upstream (SSRCSP-8820): journalctl --verify tolerates
+          # any number of consecutive FSS TAGs recording the same FSPRG epoch, not
+          # just one boundary repeat. A forward wall-clock correction otherwise piles
+          # 3+ same-epoch TAGs into one interval and the file fails verification
+          # permanently. FSS sealing code -- review with the FSS subsystem owner.
+          ./systemd-fss-verify-tolerate-repeated-epoch.patch
```
kajusnau, line 82: If this is to be removed after the upstream merge, then this deserves a `TODO` marker with a clear link to the upstream PR

kajusnau (changes requested):
Excessive comments in almost all commit messages (bodies) and the code.
Suggest making both more human-readable.

Also, please **do not** overwrite the PR template and include testing instructions if testing is necessary (seems likely, as a systemd patch is introduced)

In summary, I suggest:
- Reducing the length of the comments in code
- Reducing the commit messages _significantly_ - in some cases the commit message itself should suffice without the body
- Squashing the commits - at a glance it seems like 2-3 commits should be enough, split logically
- Adding back the PR template and writing proper testing instructions
- Adding the "Needs Testing" label, if necessary (recommended)

kajusnau:
PR description is still a bit mangled

If testing has been done already, it would help to have the results attached as a comment here on the PR, so it can be referenced in the future.

modules/common/systemd/base.nix
```diff
@@ -74,6 +74,12 @@ let
       (prevAttrs: {
         patches = prevAttrs.patches ++ [
           ./systemd-boot-double-dtb-buffer-size.patch
+          # Vendored, pending upstream (SSRCSP-8820): journalctl --verify tolerates
+          # any number of consecutive same-epoch FSS TAGs, not just one boundary
+          # repeat -- a forward wall-clock correction otherwise piles up 3+ in one
+          # interval and the file fails verification permanently.
+          # TODO: remove once upstreamed -- https://github.com/tiiuae/ghaf/issues/2243
```
kajusnau, line 81: This is a link to our own repo's brand new ticket, not a link to the external upstream PR.
The reference should be to the external upstream PR, so it can be tracked and followed-up on later.

If no upstream PR exists, it should probably be opened 🙂

modules/common/systemd/systemd-fss-verify-tolerate-repeated-epoch.patch
```diff
@@ -0,0 +1,78 @@
...
+
+The only property SEALED_CONTINUOUS adds over the non-continuous branch is
+refusing a forward gap (`epoch > last_epoch + 1`); that is kept. A
+backward epoch (`epoch < last_epoch`) still fails, so forward security is
+unchanged: a repeated-epoch TAG is sealed with the key already in use for
+that interval and grants an attacker nothing they could not already do
+with that key.
+
+`last_epoch` still only ever advances by 0 or +1, so the next real
+interval-boundary tag (`epoch == last_epoch + 1`) continues to pass.
```
kajusnau, line 47: This patch commit message is also very verbose, most likely can use a trim

modules/common/systemd/systemd-fss-verify-tolerate-repeated-epoch.patch
```diff
@@ -0,0 +1,78 @@
...
+diff --git a/src/libsystemd/sd-journal/journal-verify.c b/src/libsystemd/sd-journal/journal-verify.c
+--- a/src/libsystemd/sd-journal/journal-verify.c
++++ b/src/libsystemd/sd-journal/journal-verify.c
+@@ -1135,8 +1135,19 @@
+                         }
+
+                         if (JOURNAL_HEADER_SEALED_CONTINUOUS(f->header)) {
+-                                if (!(n_tags == 0 || (n_tags == 1 && le64toh(o->tag.epoch) == last_epoch)
+-                                      || le64toh(o->tag.epoch) == last_epoch + 1)) {
++                                /* SSRCSP-8820: tolerate any number of consecutive tags that
```
kajusnau, line 61: Don't think it's a good idea to keep our own jira ticket ID in a `.patch` meant for an upstream project.

Also this in-code comment is quite verbose, but I won't argue if all of it is absolutely necessary

modules/common/systemd/systemd-fss-verify-tolerate-repeated-epoch.patch
kajusnau, line 1: The core change is quite simple (1 line if I'm reading correctly?)
I'd argue this whole patch can be removed and the change could instead be performed via a `substituteInPlace` call in `postPatch` phase ([example](https://github.com/tiiuae/ghaf/blob/4b581f2f971cf6a1850f6bc8cfa7f54b344ef8ab/overlays/custom-packages/udiskie/default.nix#L7))

tests/logging/default.nix
```diff
@@ -25,6 +25,14 @@
 let
   fssSetupTest = import ./test_scripts/fss_setup.nix;
   fssVerificationTest = import ./test_scripts/fss_verification.nix;
+
+  # journald with the vendored SSRCSP-8820 verify-side patch (tolerates
```
kajusnau, line 29: I don't think we need to reference the jira ticket anywhere in this PR, it's not common practice in our repo.
If necessary, reference the relative path of the `.patch` itself and/or the upstream PR.

kajusnau (changes requested):

rodrigopinotii:
Folded into #2251 together with #2246 and #2248, per review; all threads here are addressed there. Closing.

PR #2240, opened 2026-09-10
feat(active-directory): Active Directory Login with GPO-Based Access Control

gngram:
## Description of Changes

> **Note**: This is a unified PR that incorporates **#1864** with verified end-to-end functionality.
## Description of Changes
Enable Active Directory (AD) and SSSD GPO-based access control so that only users explicitly granted logon rights via Group Policy can log into Ghaf devices through graphical display managers (`greetd`, `cosmic-greeter`), TTY (`login`), SSH (`sshd`), and switch user (`su`/`su-l`).
## Changes
- **Automatic SSSD Startup**: Added `systemd.paths.sssd-keytab` path watcher to automatically trigger `sssd.service` as soon as `/etc/krb5.keytab` is created during domain enrollment, conditioning SSSD startup on keytab presence.
- **PAM Strict Access**: Enabled `sssdStrictAccess` across interactive PAM services (`greetd`, `cosmic-greeter`, `login`, `su`, `su-l`, and `sshd`) to strictly enforce SSSD account-phase access decisions (from GPO evaluation) at login and user-switching time.
- **GPO Logon Rights Mapping**:
  - Mapped display manager, login, and su services to interactive logon rights (`ad_gpo_map_interactive = +login, +greetd, +cosmic-greeter, +su, +su-l`).
  - Mapped SSH to remote interactive logon rights (`ad_gpo_map_remote_interactive = +sshd`).
  - Configured GPO enforcement (`gpoAccessControl = "enforcing"`, `ad_gpo_implicit_deny = true`, and `ad_site = Default-First-Site-Name`).
- **Trial Target Profile (`mvp-user-trial`)**: Switched from `homed-user` to `ad-users` profile, configured domain controller settings, and granted sudo privileges to the `%ghaf-admins` AD group.
- **Documentation**: Updated the developer guide (`docs/src/content/docs/ghaf/dev/guides/ad-login.mdx`) with the expanded PAM/GPO mappings, keytab auto-trigger setup, and sudo configuration instructions.
## AD Server-Side Setup
High-level setup performed on the domain controller:
1. Created Group Policy Object (GPO) for Ghaf device login restrictions.
2. Moved the device under test from the default "Computers" container to an OU linked with the GPO.
3. Created two AD groups: `ghaf-users` and `ghaf-admins`.
4. Created two test users:
   - `ghaf-sa-user` (member of `ghaf-admins`)
   - `ghaf-sd-user` (member of `ghaf-users`)
5. Configured User Rights Assignment in GPO:
   - **"Allow log on locally"**: Allowed both `ghaf-sa-user` and `ghaf-sd-user`.
   - **"Allow log on through Remote Desktop Services" / Remote Interactive**: Allowed `ghaf-sa-user` only.
## Verification and Results
- **SSSD Automatic Startup**: Verified `sssd.service` automatically starts upon domain enrollment when `/etc/krb5.keytab` is created without requiring a manual reboot.
- **Local Login**: Verified both `ghaf-sa-user` and `ghaf-sd-user` can successfully log in locally via graphical greeter (`greetd`, `cosmic-greeter`) and console TTY.
- **Remote SSH Login**:
  - Verified `ghaf-sa-user` successfully logs in remotely via SSH.
  - Verified `ghaf-sd-user` is denied remote SSH access as enforced by GPO policy.
- **Client-Side Sudo Privileges**: Verified `ghaf-sa-user` (member of `%ghaf-admins`) is granted sudo access on the client, whereas non-admin users cannot execute sudo commands.

### Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. Enable AD in 
`mvp-user-trial.nix`
 by setting `enableAD = true` with the target AD server configuration.
2. Build and flash image
3. Obtain AD domain enrollment credentials along with the admin (ghaf-sa-user) and non-admin (ghaf-sd-user) user accounts.
4. Join the AD domain during initial device provisioning.
5. Ask AD admin to enroll the device under test in the GPO OU.
6. Verify local login succeeds for both admin and non-admin users.
7. Verify sudo privilege escalation fails for the non-admin user.
8. Verify sudo privilege escalation succeeds for the admin user (ghaf-admins).
9. Verify remote SSH access to gui-vm is denied for the non-admin user via GPO policy.
10. Verify remote SSH access to gui-vm is permitted for the admin user.

### DIFF
--- a/modules/desktop/guivm/boot-ui.nix
+++ b/modules/desktop/guivm/boot-ui.nix
@@ -16,19 +16,30 @@
   ...
 }:
 let
-  # Wait for UID>=1000 session to become active with valid seat
+  # Wait for a real non-root user session, not the display-manager greeter.
   wait-for-session = pkgs.writeShellApplication {
     name = "wait-for-session";
     runtimeInputs = [
+      pkgs.coreutils
       pkgs.systemd
       pkgs.jq
     ];
     text = ''
+      set -euo pipefail
+
       echo "Waiting for user to login..."
-      USER_ID=1
-      while [ "$USER_ID" -lt 1000 ]; do
-        tmp_id=$(loginctl list-sessions --json=short | jq -e '.[] | select(.seat != null) | .uid') || true
-        [[ "$tmp_id" =~ ^[0-9]+$ ]] && USER_ID="$tmp_id" || USER_ID=1
+      USER_ID=0
+      while [ "$USER_ID" -eq 0 ]; do
+        active_session="$(loginctl show-seat seat0 -p ActiveSession --value 2>/dev/null || true)"
+        tmp_id="$(loginctl show-session "$active_session" -p User --value 2>/dev/null || true)"
+        seat="$(loginctl show-session "$active_session" -p Seat --value 2>/dev/null || true)"
+        session_class="$(loginctl show-session "$active_session" -p Class --value 2>/dev/null || true)"
+
+        if [[ "$tmp_id" =~ ^[0-9]+$ ]] && [ "$tmp_id" -gt 0 ] && [ -n "$seat" ] && [ "$session_class" = "user" ]; then
+          USER_ID="$tmp_id"
+        else
+          USER_ID=0
+        fi
         sleep 1
       done
       echo "User with ID=$USER_ID is now active"
--- a/modules/microvm/host/microvm-host.nix
+++ b/modules/microvm/host/microvm-host.nix
@@ -31,6 +31,28 @@ let
     else
       config.ghaf.users;
   hasLoginUser = userConfig.homedUser.enable || userConfig.adUsers.enable;
+  loginUserHasDynamicUid = userConfig.adUsers.enable;
+  loginUserUid = toString (userConfig.homedUser.uid or 1000);
+  loginUserSessionDir = "/run/ghaf/session";
+  loginUserUidFile = "${loginUserSessionDir}/gui-vm-user.uid";
+  sharedVmDirectoryPaths = [
+    "/persist/storagevm/shared/shares"
+  ]
+  ++ map (n: "/persist/storagevm/shared/shares/Unsafe ${n} share/") cfg.sharedVmDirectory.vms;
+  xdgRuleFor =
+    xdgPath:
+    if loginUserHasDynamicUid then
+      # AD users keep their directory service UID, so the handoff path cannot be pre-owned by a fixed local user.
+      "D ${xdgPath} 0733 root root -"
+    else
+      "D ${xdgPath} 0700 ${loginUserUid} users -";
+  sharedDirRuleFor =
+    path:
+    if loginUserHasDynamicUid then
+      # AD logins need host directories that AppVMs can access before the active GUI UID is known.
+      "d ${path} 0770 root users"
+    else
+      "d ${path} 0760 ${loginUserUid} users";
   hasAudioVmAcpiPath =
     (lib.hasAttr "audio-vm" config.microvm.vms)
     && (config.ghaf.hardware.definition.audio.acpiPath != null);
@@ -232,12 +254,11 @@ in
               if xdgPathsAttempt.success then xdgPathsAttempt.value else [ ]
             ) vmsWithXdg
           );
-          xdgRules = map (
-            xdgPath: "D ${xdgPath} 0700 ${toString config.ghaf.users.homedUser.uid} users -"
-          ) xdgDirs;
+          xdgRules = map xdgRuleFor xdgDirs;
         in
         [
           "d /persist/common 0755 root root -"
+          "d /persist/common/ghaf 0755 root root -"
           "d /persist/sysupdate 0755 root root -"
           "d /persist/storagevm 0755 root root -"
           "d /persist/storagevm/img 0700 microvm kvm -"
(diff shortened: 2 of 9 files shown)


modules/desktop/guivm/boot-ui.nix
```diff
@@ -16,19 +16,30 @@
...
+  # Wait for a real non-root user session, not the display-manager greeter.
   wait-for-session = pkgs.writeShellApplication {
     name = "wait-for-session";
     runtimeInputs = [
+      pkgs.coreutils
       pkgs.systemd
       pkgs.jq
     ];
     text = ''
+      set -euo pipefail
```
kajusnau, line 28: Redundant, see https://nixos.org/manual/nixpkgs/stable/#trivial-builder-writeShellApplication

modules/microvm/host/microvm-host.nix
```diff
@@ -329,16 +350,89 @@ in
...
+        let
+          aclSetupScript = pkgs.writeShellApplication {
+            name = "shared-vm-directory-acl";
+            runtimeInputs = with pkgs; [
+              acl
+              coreutils
+              findutils
+            ];
+            text = ''
+              set -euo pipefail
```
kajusnau, line 378: Redundant, see https://nixos.org/manual/nixpkgs/stable/#trivial-builder-writeShellApplication

modules/microvm/host/microvm-host.nix
```diff
@@ -329,16 +350,89 @@ in
...
+        in
+        {
+          enable = true;
+          description = "Apply shared folder ACLs for the active GUI login";
+          wantedBy = [ "multi-user.target" ];
+          after = [ "local-fs.target" ];
+          serviceConfig = {
+            Type = "oneshot";
+            ExecStart = lib.getExe aclSetupScript;
+          };
```
kajusnau, line 423: If I remember correctly, `oneshot` with `wantedBy = [ "multi-user.target" ]` will block boot until the script exits.

Is it worth considering using `Type = "exec";` to prevent this? Only if boot (and gui-vm) don't rely on this too strictly.

modules/microvm/host/microvm-host.nix
```diff
@@ -329,16 +350,89 @@ in
...
+            runtimeInputs = with pkgs; [
+              acl
+              coreutils
+              findutils
+            ];
+            text = ''
+              set -euo pipefail
+
+              uid_file=${lib.escapeShellArg loginUserUidFile}
+              current_uid=""
```
kajusnau, line 381: Should the `current_uid` value be stored somewhere at runtime so the value can be checked against when the `shared-vm-directory-acl` Path Unit requests execution of this script?
That way the script could be skipped if the UID doesn't change (is it even possible that the PathChanged event triggers but UID remains the same?)

modules/microvm/sysvms/guivm-base.nix
```diff
@@ -40,6 +40,17 @@
 }:
 let
   vmName = "gui-vm";
+  guiSessionShareEnabled = hostConfig.users.profile.ad-users.enable or false;
+  guiSessionUidFile = "/run/ghaf/session/gui-vm-user.uid";
```
kajusnau, line 44: Why is the structure here of these vars different than the same in `microvm-host`?
```nix
  loginUserSessionDir = "/run/ghaf/session";
  loginUserUidFile = "${loginUserSessionDir}/gui-vm-user.uid";
```

Both seem to evaluate to the same exact string, but in two different ways.

modules/microvm/sysvms/guivm-base.nix
```diff
@@ -70,6 +81,50 @@ let
       }";
     }
   ) virtualApps;
+
+  sync-active-gui-user = pkgs.writeShellApplication {
+    name = "sync-active-gui-user";
+    runtimeInputs = [
```
kajusnau, line 87: Prefer to have `with pkgs;` here, to match other parts of this patch.

modules/microvm/sysvms/guivm-base.nix
```diff
@@ -70,6 +81,50 @@ let
...
+      pkgs.coreutils
+      pkgs.systemd
+    ];
+    text = ''
+      set -euo pipefail
```
kajusnau, line 92: Redundant, see https://nixos.org/manual/nixpkgs/stable/#trivial-builder-writeShellApplication

modules/microvm/sysvms/guivm-base.nix
```diff
@@ -70,6 +81,50 @@ let
...
+
+      active_session="$(loginctl show-seat seat0 -p ActiveSession --value 2>/dev/null || true)"
+      user_id="$(loginctl show-session "$active_session" -p User --value 2>/dev/null || true)"
+      seat="$(loginctl show-session "$active_session" -p Seat --value 2>/dev/null || true)"
+      session_class="$(loginctl show-session "$active_session" -p Class --value 2>/dev/null || true)"
+
+      if [[ ! "$user_id" =~ ^[0-9]+$ ]] || [ "$user_id" -eq 0 ] || [ -z "$seat" ] || [ "$session_class" != "user" ]; then
+        clear_uid_file
+        exit 0
+      fi
```
kajusnau, line 118: The detection logic here is largely (or entirely?) copy-pasted from `wait-for-session` in `modules/desktop/guivm/boot-ui.nix`. Can we be more DRY and introduce a global helper somewhere to reuse this logic?

modules/desktop/guivm/boot-ui.nix
```diff
@@ -16,19 +16,30 @@
...
 }:
 let
-  # Wait for UID>=1000 session to become active with valid seat
+  # Wait for a real non-root user session, not the display-manager greeter.
   wait-for-session = pkgs.writeShellApplication {
     name = "wait-for-session";
     runtimeInputs = [
+      pkgs.coreutils
       pkgs.systemd
       pkgs.jq
```
kajusnau, line 25: `jq` is unused

modules/desktop/guivm/boot-ui.nix
```diff
@@ -16,19 +16,30 @@
...
-      USER_ID=1
-      while [ "$USER_ID" -lt 1000 ]; do
-        tmp_id=$(loginctl list-sessions --json=short | jq -e '.[] | select(.seat != null) | .uid') || true
-        [[ "$tmp_id" =~ ^[0-9]+$ ]] && USER_ID="$tmp_id" || USER_ID=1
+      USER_ID=0
+      while [ "$USER_ID" -eq 0 ]; do
+        active_session="$(loginctl show-seat seat0 -p ActiveSession --value 2>/dev/null || true)"
+        tmp_id="$(loginctl show-session "$active_session" -p User --value 2>/dev/null || true)"
+        seat="$(loginctl show-session "$active_session" -p Seat --value 2>/dev/null || true)"
+        session_class="$(loginctl show-session "$active_session" -p Class --value 2>/dev/null || true)"
```
kajusnau, line 36: 4 separate `loginctl` calls.
Unlikely, but we may be risking a race condition here - what if something happens in between?

Could we do:
```sh
    session="$(loginctl show-seat seat0 -p ActiveSession --value 2>/dev/null || true)"
    [ -n "$session" ] || return 1 # <- only if this is part of a function, adjust otherwise
    while IFS='=' read -r key value; do
      case "$key" in
        User) uid="$value" ;;
        Seat) seat="$value" ;;
        Class) class="$value" ;;
      esac
    done < <(loginctl show-session "$session" -p User -p Seat -p Class 2>/dev/null || true)
```

modules/microvm/sysvms/guivm-base.nix
```diff
@@ -326,6 +382,40 @@ in
 
   systemd = {
     packages = [ pkgs.blueman ];
+    services."sync-active-gui-user" = lib.mkIf guiSessionShareEnabled {
```
kajusnau, line 385: Same comment as before:

> If I remember correctly, `oneshot` with `wantedBy = [ "multi-user.target" ]` will block boot until the script exits.
> Is it worth considering using `Type = "exec";` to prevent this? Only if boot (and gui-vm) don't rely on this too strictly.

kajusnau, line 385: Why is this service declared in `guivm-base.nix` instead of its' own module or feature?

modules/microvm/host/microvm-host.nix
```diff
@@ -329,16 +350,89 @@ in
...
+              }
+
+              for path in ${lib.concatMapStringsSep " " lib.escapeShellArg sharedVmDirectoryPaths}; do
+                apply_shared_acl "$path"
+              done
+            '';
+          };
+        in
+        {
+          enable = true;
```
kajusnau, line 416: Why is this service declared here instead of its' own module or feature?

modules/reference/profiles/mvp-user-trial.nix
```diff
@@ -7,6 +7,14 @@
 }:
 let
   cfg = config.ghaf.reference.profiles.mvp-user-trial;
+  # Active Directory configuration
+  enableAD = false;
+  adDomain = "ghaf-test.com";
+  # AD Domain controller URL
+  dcUrl = "vm-ghaf-dev-dc.ghaf-test.com";
+  # AD Domain controller IP address
+  dcIp = "10.52.33.4";
```
kajusnau, line 16: Does this all need to be hardcoded?

kajusnau (changes requested):
Nice work!
Can be cleaned up a bit IMO 🫡

Also, can most of this logic be moved to its' own (or some pre-existing, if applies) module, and then enabled in `host` and `gui-vm`? Some of my comments reference this.

milva-unikie:
Please add Needs Testing label back once the review comments have been addressed

PR #2241, opened 2026-09-11
refactor(tuned): refactor performance profiles

kajusnau:
## Description of Changes

1. **Refactored TuneD module config to be more DRY:**
   - Derive all power profiles from mkProfileVariants and share the ppd.conf mapping, unit names and profile dirs.
   - Use host.nix and guests.nix as real modules.
2. **Fixed some dead or outdated TuneD settings:**
   - Drop `vm.laptop_mode`, `hwp_dynamic_boost`, `audio.timeout` and the host PCI runtime PM
    - Use `med_power_with_dipm`, the kernel's dirty ratios, and start thermald with `--no-block`
3. **Removed virtual-guest profile entirely - VMs will now run "stock":**
   - After more testing inspired by the comments on this PR, it became clear that very few of the `virtual-guest` profile settings were actually doing anything in our guests.
     The same applied for QEMU, not just crosvm.
     Therefore, the `virtual-guest` profile has been entirely removed and will not apply to any VMs moving forward - VMs run "stock".
     PPD settings still apply to gui, net, and audio VMs, as well as the host - these are the only targets where something can be tuned at runtime to improve battery life/performance.
4. **Introduce gui-vm GPU frequency caps on battery**
5. **Drop deprecated magick call, do not install graphics module autostart unit if empty**

### Type of Change

- [x] New feature
- [x] Bug fix
- [ ] Update
- [x] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. Verify power profiles can still be switched via GUI (top right battery widget):
   - After a profile switch (e.g. to performance on AC) the following should be visible in `ghaf-host` - `tuned-adm active` -> "host-performance"
   - The same should be visible in `audio-vm` and `net-vm`, but the power profile names will be prefixed with the VM name, so e.g. on `net-vm` -> "net-performance"
2. Verify switching power profiles does not cause any hardware disruption (e.g. audio disconnects during playback, video glitches, etc.
3. (Optional) evaluate Ghaf power consumption and battery life

### DIFF
--- a/lib/global-config.nix
+++ b/lib/global-config.nix
@@ -514,6 +514,7 @@ rec {
             "gui-vm"
             "audio-vm"
             "net-vm"
+            "admin-vm"
           ];
         };
         locale = {
@@ -621,6 +622,7 @@ rec {
             "gui-vm"
             "audio-vm"
             "net-vm"
+            "admin-vm"
           ];
         };
         locale = {
--- a/modules/common/services/performance/default.nix
+++ b/modules/common/services/performance/default.nix
@@ -112,16 +112,52 @@ let
           done
         }
 
-        # Set PCI device runtime power management
-        pci_device_runtime_pm() {
+        # Set runtime power management for every PCI device present
+        pci_runtime_pm() {
           # 'on' - best performance
           # 'auto' - best power saving
-
           pm=$1
-          devices=$2
 
-          for device in $devices; do
-            (echo "$pm" > "/sys/bus/pci/devices/$device/power/control") &> /dev/null
+          for device in /sys/bus/pci/devices/*; do
+            # Never storage (0x01xxxx) or USB (0x0c03xx): suspending the link
+            # to a boot device takes the root filesystem away.
+            class=$(cat "$device/class" 2>/dev/null) || continue
+            case "$class" in
+              0x01* | 0x0c03*) continue ;;
+            esac
+            (echo "$pm" > "$device/power/control") &> /dev/null
+          done
+        }
+
+        # Cap the GPU ceiling: 'efficient' is RP1/RPe, 'max' is RP0. Handles
+        # both i915 (flat gt_* attributes) and xe (per-gt freq directories).
+        gpu_max_freq() {
+          level=$1
+
+          for gt in /sys/class/drm/card*/device/tile*/gt*/freq0; do
+            [ -e "$gt/max_freq" ] || continue
+            case "$level" in
+              efficient) want=$(cat "$gt/rpe_freq" 2>/dev/null) ;;
+              *)         want=$(cat "$gt/rp0_freq" 2>/dev/null) ;;
+            esac
+            [ -n "$want" ] || continue
+            # Ceiling only -- leave the floor so the GPU can still idle down.
+            cur_min=$(cat "$gt/min_freq" 2>/dev/null || echo 0)
+            [ "$cur_min" -gt "$want" ] && (echo "$want" > "$gt/min_freq") &> /dev/null
+            (echo "$want" > "$gt/max_freq") &> /dev/null
+          done
+
+          for card in /sys/class/drm/card*; do
+            [ -e "$card/gt_max_freq_mhz" ] || continue
+            case "$level" in
+              efficient) want=$(cat "$card/gt_RP1_freq_mhz" 2>/dev/null) ;;
+              *)         want=$(cat "$card/gt_RP0_freq_mhz" 2>/dev/null) ;;
+            esac
+            [ -n "$want" ] || continue
+            cur_min=$(cat "$card/gt_min_freq_mhz" 2>/dev/null || echo 0)
+            [ "$cur_min" -gt "$want" ] && (echo "$want" > "$card/gt_min_freq_mhz") &> /dev/null
+            (echo "$want" > "$card/gt_max_freq_mhz") &> /dev/null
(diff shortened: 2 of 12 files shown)


vadika:
Looks like a footgun for me -- why we need this?

kajusnau:
> Looks like a footgun for me -- why we need this?

Need what specifically?
We have been using TuneD for a while, but only for `host`, `net-vm`, `gui-vm`, and `audio-vm`.
The power profiles can be changed by the user at any time.
The `virtual-guest` (`vm-balanced` in ghaf) profile has been in use by `audio-vm` already.

All in all, this just expands the coverage of TuneD power profiles to all VMs, instead of just a few. Should theoretically improve performance in some metrics, while `gui-vm` and `audio-vm` can now be more directly influenced by the selected power profiles, which should save some battery in some cases.

vadika:
> > Looks like a footgun for me -- why we need this?
> 
> Need what specifically? We have been using TuneD for a while, but only for `host`, `net-vm`, `gui-vm`, and `audio-vm`. The power profiles can be changed by the user at any time. The `virtual-guest` (`vm-balanced` in ghaf) profile has been in use by `audio-vm` already.
> 
> All in all, this just expands the coverage of TuneD power profiles to all VMs, instead of just a few. Should theoretically improve performance in some metrics, while `gui-vm` and `audio-vm` can now be more directly influenced by the selected power profiles, which should save some battery in some cases.

Thats exactly the point -- using tuned for virtualized hardware is creating more problems than solving anything.
Its ok to use it on host, but not in VMs

kajusnau:
> Thats exactly the point -- using tuned for virtualized hardware is creating more problems than solving anything. Its ok to use it on host, but not in VMs

I'm open to discussion on this.

Like I said, we've already been using some of these profiles, and I don't recall any specific issues caused by them.

[Redhat docs](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/monitoring_and_managing_system_status_and_performance/optimizing-virtual-machine-performance-in-rhel_monitoring-and-managing-system-status-and-performance) are pretty convincing about this, though admittedly that's one of the few places I've gathered info about this.

TuneD is also nice because it allows us to customize power profiles very thoroughly, instead of relying on pre-determined defaults from stuff like [system76-power](https://github.com/pop-os/system76-power).
Customized profiles matter to us because we're not running bare-metal, so some settings (like GPU specific stuff) is irrelevant on the host, but very relevant on gui-vm, so we shouldn't use the same on both.
In addition, TuneD allows running custom scripts (as we do here) on power profile changes, which is useful in our use-case since power profiles can only be changed in `gui-vm`, but should be propagated to (at the very least) the host.

If there's some concerns about the specific settings these profiles change (e.g. [virtual-guest](https://github.com/redhat-performance/tuned/blob/master/profiles/virtual-guest/tuned.conf) and [throughput-performance](https://github.com/redhat-performance/tuned/blob/master/profiles/throughput-performance/tuned.conf)), then we can fine-tune them for our usecase.

vadika:
> > Thats exactly the point -- using tuned for virtualized hardware is creating more problems than solving anything. Its ok to use it on host, but not in VMs
> 
> I'm open to discussion on this.
> 
> Like I said, we've already been using some of these profiles, and I don't recall any specific issues caused by them.
> 
> [Redhat docs](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/monitoring_and_managing_system_status_and_performance/optimizing-virtual-machine-performance-in-rhel_monitoring-and-managing-system-status-and-performance) are pretty convincing about this, though admittedly that's one of the few place I've gathered info about this.
> 
> TuneD is also nice because it allows us to customize power profiles very thoroughly, instead of relying on pre-determined defaults from stuff like [system76-power](https://github.com/pop-os/system76-power). Customized profiles matter to us because we're not running bare-metal, so some settings (like GPU specific stuff) is irrelevant on the host, but very relevant on gui-vm, so we shouldn't use the same on both. In addition, TuneD allows running custom scripts (as we do here) on power profile changes, which is useful in our use-case since power profiles can only be changed in `gui-vm`, but should be propagated to (at the very least) the host.
> 
> If there's some concerns about the specific settings these profiles change (e.g. [virtual-guest](https://github.com/redhat-performance/tuned/blob/master/profiles/virtual-guest/tuned.conf) and [throughput-performance](https://github.com/redhat-performance/tuned/blob/master/profiles/throughput-performance/tuned.conf)), then we can fine-tune them for our usecase.


Have been not able to find  any evidence how the tuned profile settings/tunings are interoperating on crosvm-based systems. On the opposite, there are many qemu-related setings in profiles. 
Probably we can use it, but  first it requres thorough settings cleanup and testing.

kajusnau:
Update:

`virtual-guest` profile has been removed entirely and will no longer be applied to ANY VMs moving forward.
The reasoning is in the PR description. Thanks to Vadim for pushing me to test this further!

The end result now is that only gui, audio, and net VMs are "managed" by TuneD - changing power profiles triggers certain scripts at runtime (e.g. limiting GPU frequency in gui-vm when on powersave mode).

The Host still gets the fully custom TuneD profiles, since these settings are generally applicable, unlike `virtual-guest` on our guests.

PR #2245, opened 2026-09-11
fix(logging): stop piping into an early-exiting grep under pipefail

rodrigopinotii:
## Description of Changes

`printf ... | grep -q ...` races under `set -o pipefail` (injected by
`pkgs.writeShellApplication`): `grep -q` exits as soon as it matches, and if
`printf` hasn't finished writing, it gets SIGPIPE'd. pipefail then reports the
pipeline's exit status from `printf`'s SIGPIPE, not grep's real result, so an
`if ... | grep -q ...` can silently take the wrong branch.

Root-caused via a real hardware occurrence on a separate branch: a literal
`printf: Broken pipe` line in production logs at the exact moment a de-dup
check misfired. These 8 instances predate that branch and are unrelated to
any open PR or feature work -- pure pre-existing correctness fix in shared
FSS library code (`fss-verify-classifier.sh`, `fss.nix`).

Found via two rounds of review, not one: an initial sweep caught 6 instances;
a second, more careful pass over the same file found 2 more, including
`fss_output_has_tamper_signature` -- there, a raced false negative would let
tampered journal content through instead of failing closed. Flagging the
two-pass history rather than claiming the first sweep was complete.

Fix: a here-string (`<<<`) instead of a pipe. bash fully materialises the
data before grep starts reading, so there's no live producer process left to
receive a SIGPIPE. Matching semantics are unchanged (verified per-instance:
`-x`/`-F`/`-E` behavior, trailing-newline handling, multi-line input).

Verified empirically, not just by code review: reproduced deterministically
on real hardware by widening the race window, 100/100 failures before the
fix, 0/100 after.

Not touched: `fss_resolve_live_journal_dir`'s `... | grep -m1 ...` -- its
result comes from what grep printed, already wrapped in `|| true`, so a
pipefail-confused exit status can't corrupt the assigned value the way it
does an `if`-guarded boolean check.

### Type of Change

- [x] Bug fix

## Checklist

### Applicable Targets

- [x] Intel Laptop `x86_64`

### Installation Method

### Test Steps To Verify:

1. `nix build .#checks.x86_64-linux.fss-classifier-unit` -- green.
2. `nix build .#checks.x86_64-linux.pre-commit` -- green.
3. Code review: every `printf | grep -q` site in the touched files replaced
   with an equivalent here-string; no other instance of the pattern remains.

### DIFF
--- a/modules/common/logging/fss-verify-classifier.sh
+++ b/modules/common/logging/fss-verify-classifier.sh
@@ -116,7 +116,7 @@ fss_append_tag() {
 
   if [ -z "$current" ]; then
     printf '%s' "$tag"
-  elif printf '%s\n' ",$current," | grep -Fq ",$tag,"; then
+  elif grep -Fq -- ",$tag," <<<",$current,"; then
     printf '%s' "$current"
   else
     printf '%s,%s' "$current" "$tag"
@@ -140,7 +140,7 @@ fss_append_unique_line() {
 
   if [ -z "$line" ]; then
     printf '%s' "$current"
-  elif printf '%s\n' "$current" | grep -Fxq "$line"; then
+  elif grep -Fxq -- "$line" <<<"$current"; then
     printf '%s' "$current"
   else
     fss_append_line "$current" "$line"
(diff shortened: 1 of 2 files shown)


kajusnau:
Why is this separate from #2246, both touching the same core script?
These PRs are starting to feel more like commits, it's becoming impractical IMO. I strongly suggest folding them.

rodrigopinotii:
Agreed. Folded into #2246, closing this one.

PR #2246, opened 2026-09-11
fix(logging): FSS re-key retention, durability, and a pipefail SIGPIPE bug

rodrigopinotii:
## Description of Changes

FSS re-key retention and durability hardening, plus the pipefail/SIGPIPE fix
this work surfaced:

- Retains superseded verification keys (default 3) across backward-clock
  re-keys, ordered by recency (rekey-history + mtime), not the filename
  epoch -- which runs backward under repeated backward corrections.
- journal-fss-verify retries failing archived journals against retained
  keys before failing: pre-re-key lineage is excused, tamper is not (a
  tampered archive still fails under every key).
- A re-key transition archive neither a retained key nor a receipt covers
  degrades to warning uniformly, instead of depending on journald vacuum
  timing across VMs.
- Key-pair regeneration (fresh or re-key) is now crash-safe: one durable
  helper (temp -> fsync -> atomic rename -> fsync dir) used by both paths,
  so a reboot mid-swap never lands on a torn or mismatched pair.
- `printf | grep -q` under pipefail races: `grep -q` exits on first match,
  and if `printf` hasn't finished writing it gets SIGPIPE'd, misreporting
  the pipeline as failed. Caught live: `fss_retained_keys_newest_first`'s
  de-dup check hit this, double-counting and evicting a retained key a
  cycle early. Fixed here and in three other call sites this branch adds;
  the eight pre-existing instances (predating this branch) are fixed
  separately in #2245.

Validated on real hardware (Dell Latitude 7330): multi-cycle backward/
forward correction sequences, tamper suite, and the re-key regression
suite, plus `nix build .#checks.x86_64-linux.fss-classifier-unit`.

### Type of Change

- [x] Bug fix

## Checklist

- [x] Clear summary in PR description
- [x] Detailed and meaningful commit message(s)
- [x] Commits are logically organized and squashed if appropriate
- [x] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [x] Author has added reviewers and removed PR draft status

### Applicable Targets

- [x] Intel Laptop `x86_64`

### Installation Method

- [x] Can be updated with `nixos-rebuild ... switch`

### Test Steps To Verify:

1. Force two backward wall-clock corrections close enough together that
   the second re-keys before the first's archives age out.
2. `journalctl -u journal-fss-verify` shows no new FAILED verdicts on any
   FSS VM; pre-re-key archives excuse under a retained key.
[...]

PR #2248, opened 2026-09-12
fix(logging): catch a fast clock round-trip the jump watcher missed

### LINKED ISSUE #2092: journal-fss-verify left permanently failed after a Robot clock test: clock-jump watcher misses sub-interval round-trip jumps

rodrigopinotii:
## Description of Changes

`ghaf-clock-jump-watcher` polled realtime drift on a fixed interval, so a backward-then-forward round-trip completing within one tick produced ~0 net drift and never triggered recovery (ghaf#2092). Wakes the watcher on the clock step itself instead, via systemd's `OnClockChange=` timer writing into the watcher's FIFO.

- FIFO not signal: signals can coalesce, FIFO writes never do.
- No `After=`/`Wants=` from the timer onto the watcher: FIFO setup isn't unit-start ordering.
- Oneshot avoids `sh -c`: the unit's own PATH has no shell in it.
- Same kernel primitive (`timerfd_create`+`TFD_TIMER_CANCEL_ON_SET`) as the earlier C helper -- confirmed from systemd source.

No public interface changes.

Closes #2092.

Includes the verify-side staleness guard from #2249: the faster detection above exposed a TOCTOU in `verify_live_sealing_after_activation`; the two together are the whole fix and were hardware-tested together.

### Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

ghaf#2092

## Checklist

- [ ] Ghaf documentation updated with the commit

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. Reproduce the round-trip (backward step, forward step within seconds).
2. Confirm two distinct watcher wakes, not one net-drift-cancelled miss.
3. Confirm no false "key pair has diverged" verdict during the round-trip.
4. TC4: restart idempotency, FIFO wake latency, real-PATH oneshot, no-reader timeout -- all pass.

**HW-verified:** on dell-7330 (A2/B2/B3 images, 2026-09-12/13); evidence available on request.

**Merge order:** merges cleanly onto the current heads of #2239/#2246; no shared hunks. Recommend #2246 alongside (hardens a pre-existing re-key race this fix makes easier to hit).

### DIFF
--- a/modules/common/logging/common.nix
+++ b/modules/common/logging/common.nix
@@ -21,6 +21,8 @@ let
   # Effective time-sync wait before FSS activation; only applies when the
   # activation boundary is enabled. Computed once and reused below.
   effectiveSyncWaitSeconds = if fssActivationEnabled then fssActivationCfg.syncWaitSeconds else 0;
+  # FIFO shared by the OnClockChange oneshot and the watcher's read loop.
+  clockStepFifo = "/run/ghaf-clock-jump-watcher/step.fifo";
 
   ghafClockReady = pkgs.writeShellApplication {
     name = "ghaf-clock-ready";
@@ -258,6 +260,15 @@ let
     '';
   };
 
+  # FIFO, not a signal: back-to-back signals coalesce, FIFO writes do not.
+  ghafClockStepNotify = pkgs.writeShellApplication {
+    name = "ghaf-clock-step-notify";
+    runtimeInputs = [ pkgs.coreutils ];
+    text = ''
+      printf 'step\n' | timeout 2 tee ${clockStepFifo} >/dev/null || true
+    '';
+  };
+
   ghafClockJumpWatcher = pkgs.writeShellApplication {
     name = "ghaf-clock-jump-watcher";
     runtimeInputs = with pkgs; [
@@ -272,6 +283,14 @@ let
     excludeShellChecks = [ "SC1091" ];
     text = ''
       source /etc/fss-verify-classifier.sh
+
+      step_fd=""
+
+      # Read-write open: never blocks, and the oneshot's write never lacks a reader.
+      rm -f ${clockStepFifo}
+      mkfifo -m 600 ${clockStepFifo}
+      exec {step_fd}<>${clockStepFifo}
+      trap 'exec {step_fd}<&-' EXIT
 
       threshold="${toString recCfg.thresholdSeconds}"
       interval="${toString recCfg.intervalSeconds}"
@@ -350,7 +369,8 @@ let
       [ -z "$seen_cursor" ] || cursor="$seen_cursor"
 
       while true; do
-        sleep "$interval"
+        # Woken early by a clock step or by the poll interval; the checks are the same.
+        read -r -t "$interval" -u "$step_fd" _line || true
         real="$(date +%s)"
         up="$(cut -d' ' -f1 /proc/uptime)"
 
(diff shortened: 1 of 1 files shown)


kajusnau:
Why not fold this into #2239, along with #2249?
I think these changes are small enough and logically linked. It would lessen the work on QA.

rodrigopinotii:
#2249 folded in here: together they are the whole #2092 fix and were hardware-tested together. #2239 stays separate: it carries the systemd change and is on its own review track, and #2092 should not wait for it.

rodrigopinotii:
Folded into #2251 together with #2239 and #2246, per review. Closing.

PR #2249, opened 2026-09-12
fix(logging): stop a stale clock reading from failing FSS verify

rodrigopinotii:
## Description of Changes

`verify_live_sealing_after_activation` could commit to a pass/fail verdict using evidence gathered across two different clock readings, since its verify calls take real time -- a clock step mid-verdict produced a false "key pair has diverged" failure. Detects the clock moving mid-verdict and defers instead; safe, since `journal-fss-setup` is reliably re-entrant.

Second commit: a separate, pre-existing bug in the same function -- a future-tagged entry under the re-key margin was misreported as diverged instead of self-healing. Names the sealing tag's epoch instead of guessing at a delta; no security posture change.

- Defer, don't fail: recover's dependency, the verify timer, and the next boot all re-check anyway.
- Threshold clamped to >=1s: 0 is a valid watcher tuning that would otherwise fire on every call.
- Second commit's earlier "already past the tag" branch was dropped -- confirmed unreachable via any real code path.

No public interface changes.

Companion to https://github.com/tiiuae/ghaf/pull/2248; recommended to merge with or before it.

### Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

ghaf#2092

## Checklist

- [ ] Ghaf documentation updated with the commit

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

1. Reproduce the same round-trip as ghaf#2092.
2. Confirm a deferred-verdict log line instead of a false divergence failure.
3. A round-trip test run confirmed the guard firing correctly on real hardware.
4. Combined campaign: zero false divergence across 5 clock-correction cycles.

**HW-verified:** on dell-7330 (A2/B2/B3 images, 2026-09-12/13); evidence available on request.

**Merge order:** merges cleanly onto the current heads of #2239/#2245/#2246; no shared hunks. Recommend #2246 alongside.

### DIFF
--- a/modules/common/logging/common.nix
+++ b/modules/common/logging/common.nix
@@ -354,10 +354,7 @@ let
         real="$(date +%s)"
         up="$(cut -d' ' -f1 /proc/uptime)"
 
-        drift="$(awk -v r1="$last_real" -v r2="$real" -v u1="$last_up" -v u2="$up" \
-          'BEGIN{print (r2-r1) - (u2-u1)}')"
-
-        abs="$(awk -v d="$drift" 'BEGIN{print (d<0)?-d:d}')"
+        abs="$(fss_clock_drift_abs "$last_real" "$last_up" "$real" "$up")"
 
         raw="$(read_journald_since_cursor)"
         seen_cursor="$(cursor_from_output "$raw")"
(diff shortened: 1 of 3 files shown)


rodrigopinotii:
Folded into #2248, closing.

Issue #2243, opened 2026-09-11
Remove vendored systemd-fss-verify-tolerate-repeated-epoch.patch once it lands upstream


rodrigopinotii:
Tracking issue for a vendored, temporary systemd patch introduced in #2239 (SSRCSP-8820).

`modules/common/systemd/base.nix` vendors `systemd-fss-verify-tolerate-repeated-epoch.patch`,
which relaxes `journal-verify.c`'s `SEALED_CONTINUOUS` epoch check so `journalctl --verify`
tolerates any number of consecutive FSS TAGs recording the same FSPRG epoch, not just one
boundary repeat. A forward wall-clock correction otherwise piles up 3+ same-epoch TAGs in one
sealing interval and the archive fails verification permanently, even though every TAG is
still authenticated (dense-forward seqnums, per-TAG HMAC under the epoch-correct key,
entry-realtime-in-window).

This is provisional pending an upstream systemd fix. A draft for upstream is written but not
yet filed (held pending an FSS subsystem owner design-note review).

**Action once upstream lands:**
- Remove `./systemd-fss-verify-tolerate-repeated-epoch.patch` from `modules/common/systemd/base.nix`.
- Remove the vendored patch file itself.
- Update the `journalctl`-divergence caution note in `docs/src/content/docs/ghaf/scs/fss.mdx`
  (off-device verification no longer needs Ghaf's patched `journalctl`).

Referenced from both spots as a `TODO` pending this issue; will be repointed at the upstream
systemd PR/commit once that is filed.

PR #2251, opened 2026-09-14
fix(logging): make FSS survive operator clock corrections

### LINKED ISSUE #2092: journal-fss-verify left permanently failed after a Robot clock test: clock-jump watcher misses sub-interval round-trip jumps

rodrigopinotii:
## Description of Changes

Makes journald forward-secure sealing (FSS) survive operator clock corrections in either direction. One PR for the three fixes that were reviewed as #2239, #2246 and #2248, folded per review feedback; they were hardware-tested together.

- Clock-jump watcher wakes on the clock step itself (`OnClockChange=` timer → FIFO), so a fast backward/forward round-trip is no longer missed; a verify verdict gathered across two clock readings is deferred instead of reported as "key pair diverged". Closes #2092.
- Superseded verification keys are retained (default 3) across backward re-keys, ordered by recency; archives sealed under an older key verify under it; tamper still fails under every key. Key-pair regeneration is crash-safe (temp → fsync → rename). `printf | grep -q` under pipefail replaced with here-strings.
- `journalctl --verify` tolerates the repeated same-epoch FSS tags journald writes on rotate/restart within one interval (one-line `substituteInPlace` on journal-verify.c, upstreaming tracked in #2243). Open journals are sealed and archived before a clock-step reboot.

Commit map:

| Commits | Content | Was |
|---|---|---|
| 1–2 | same-epoch tag tolerance, seal before clock-step reboot | #2239 |
| 3–9 | retained keys, crash-safe key swap, pipefail fixes | #2246 (+#2245) |
| 10–12 | watcher wake on clock step, verify guard, future-tag fix | #2248 (+#2249) |

No public interface changes.

### Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Related Issues / Tickets

Closes #2092. Supersedes #2239, #2246, #2248. Upstream follow-up: #2243.

## Checklist

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [ ] Orin NX `aarch64`
- [x] Intel Laptop `x86_64`

### Installation Method

- [ ] Other:

### Test Steps To Verify:

Network unplugged. Apply each `date -s` on all six nodes within a few seconds, guests first, host last (a fleet-wide step, as an operator correction produces); then check every FSS guest (net, gui, audio, admin, flatpak).

1. Round-trip: `-40 min` then `+40 min` within 5 s → recovery ran on every guest, no "key pair has diverged" in `journal-fss-setup`.
2. `-9 hours` → setup log "recovering by re-keying FSS", one new `verification-key.*`; `+9 hours` → no new key. `journal-fss-verify` → `[PASS]` after each.
3. Three `-12 hours` steps → exactly 3 retained keys; archives under older keys still `[PASS]`.
4. `journalctl --rotate` ×5 within one sealing interval → no "Epoch sequence not continuous".
5. `+3 hours` then reboot within 30 s → previous boot's archives verify with their key.
6. Flip one byte in one archive → only that file fails.
7. `nix build .#checks.x86_64-linux.fss-classifier-unit` → green.

Hardware results attached as a comment.


rodrigopinotii:
Hardware results, Dell Latitude 7330, SFO image B3 = this PR's 12 commits + the fork (plus a debug tracer not in the PR). Five FSS guests (net, gui, audio, admin, flatpak) plus host. Runs: 2026-09-13 (TC6, TC8) and 2026-09-15 (all others, one continuous session).

| Case | Scenario | Result |
|---|---|---|
| TC4 | watcher mechanics: single FIFO after 3 restarts, oneshot wake with/without reader, timer active | PASS |
| TC5 | single −9 h, fleet reboot | PASS: re-key once per guest, attestation consumed, keys capped at 3, verify clean, 0 "diverged" |
| TC1-replay ×2 | −8 h then +8 h 02 min, 2 s apart, per node | PASS: 0 "diverged" in both runs; where setup ran inside the 2 s window it re-keyed (by design, backward step above the margin) |
| TC2 c1 | +40 min | PASS: no re-key, no reboot |
| TC2 c2 | −9 h | FAIL on 3 of 5 guests: re-key refused by the pre-existing limit of 3 automatic re-keys per 24 h (`rekey.maxAttempts`, main since 2026-08-05); those guests had spent the count on TC5 + a round-trip. Not a defect of this PR; see note below |
| TC2 c3 | −24 h | PASS: window reset, all 5 re-keyed, the 3 refused guests re-sealed on their own |
| TC2 c4, c5 | −12 h, −12 h | PASS: re-key once each, keys 3, history 7→8 |
| TC6 | −985 s across a sealing epoch, live + reboot | PASS (09-13): "entries older than its sealing tag; self-heals" on all 5, no false "diverged" |
| TC8 | fully offline: power-off, cold boot, 4 kiosk corrections, exhaustive audit | PASS (09-13): 0 failed units on 6 nodes, certs/SPIRE healthy, RTC intact |
| tamper | one byte flipped in one archive | PASS: that file reported FAILED (receipt content mismatch), not excused; restored → clean |
| return to real time | +2.5 d sync, reboot | PASS: no re-key (forward), 0 failed units, keys 3 |

Note on TC2 c2: with the watcher now catching every step, a fast backward+forward round-trip re-keys on its backward half; that re-key is moot 2 s later but counts toward `rekey.maxAttempts`. Follow-up proposed separately: settle time before the re-key decision, retry after a refusal. SFO raises `maxAttempts` meanwhile. Across all steps: zero "key pair has diverged" on any guest.

kajusnau (dismissed):
### I'm blocking this PR

Some of my reasons:
- None of the review comments on #2239 received any acknowledgement or reply.
  I will not be re-writing the same comments here, even though a few of them clearly still apply.
- This is now just 1 of 6 PRs addressing a symptom, rather than the core issue, as per discussion with some folks more familiar with FSS.
- 12 commits.
  Admittedly, I suggested folding the changes from the smaller PRs into this one. That being said, **_folding_** doesn't mean stacking the commits from 4 separate PRs onto each other.
- There is _**still**_ no link to any upstream PR to patch `systemd`. The links placed in this PR are only to our own repo's ticket.

I will remove my blocking review if/when others approve.
Tagging @juliuskoskela @everton-dematos

juliuskoskela:
Hey @rodrigopinotii I wanted to make sure that you are aware that we are discussing alternatives to the clock based log sealing approach. The issue is that many of our targets don't have reliable clock synchronization and thus we came to the conclusion that trying to endlessly patch the issues caused by clock jumps is probably not the correct approach. Currently @everton-dematos is working on a proposal that I think he is presenting this week. So before we put considerable dev and review time to these PR's I think we should consider if it's absolutely necessary right now or if we should re-evaluate.

rodrigopinotii:
hey @juliuskoskela, yes We are aware that alternatives to the clock-based log sealing approach is under discussion/progress, and we are looking forward to Everto's results. we would like to keep this PR as a plan B. Time stuff as you know is never easy, and there are usually many  of hidden edge cases, so the new approach could easily take longer than expected.
FYI the change time request comes from a client request for downstream project, and we need to be able to demo it anytime. So we would rather treat it as a stopgap than park it. Once the new idea is ready, we will be more thanr happy to switch to the new solution. I addition that is why we keep the upstream PRs to patch systemd on hold on our side. Thanks

kajusnau (commented):
Removing my blocking review, but not approving explicitly - would prefer that is done by Everton and/or Julius

kajusnau:
> FYI the change time request comes from a client request for downstream project, and we need to be able to demo it anytime. So we would rather treat it as a stopgap than park it. Once the new idea is ready, we will be more thanr happy to switch to the new solution. I addition that is why we keep the upstream PRs to patch systemd on hold on our side. Thanks

Thanks for clearing this up, this context was really missing for our team.

IMO we could keep this PR in draft state and rebase as necessary for demo purposes, without needing to merge it. Either way, I'm stepping away from reviewing this in favor of Everton and Julius

juliuskoskela:
> hey @juliuskoskela, yes We are aware that alternatives to the clock-based log sealing approach is under discussion/progress, and we are looking forward to Everto's results. we would like to keep this PR as a plan B. Time stuff as you know is never easy, and there are usually many of hidden edge cases, so the new approach could easily take longer than expected. FYI the change time request comes from a client request for downstream project, and we need to be able to demo it anytime. So we would rather treat it as a stopgap than park it. Once the new idea is ready, we will be more thanr happy to switch to the new solution. I addition that is why we keep the upstream PRs to patch systemd on hold on our side. Thanks

Good. Yes I suspected there might be demo needs related to this feature. I will do my best to review. It is getting quite convoluted with all the shell scripting, I probably should have gone with Python or something similar at an earlier stage.

juliuskoskela:
### [P1] Keep unexplained archived-system corruption fatal

After a legitimate re-key, this change can turn unrelated archive corruption into a warning and a successful service exit.

The relevant path is:

1. **A retained-key file exists**, so `REKEY_ATTESTED` becomes `1`. This does not establish anything about the failing archive. [Key-existence check](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/fss.nix#L1651-L1654).
2. **Every key rejects the archive and no receipt covers it**, but the policy still classifies it as re-key collateral. [Downgrade branch](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/fss-verify-classifier.sh#L1020-L1034).
3. **The warning returns exit status 0**, losing the hard integrity-failure signal. [Service exit](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/fss.nix#L1711-L1717).

The cryptographic verifier still detects the corruption; the problem is how its rejection is handled. The [new test explicitly corrupts an archive and requires the downgrade](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/tests/logging/test_scripts/fss_verification.nix#L921-L933).

**Reproduction:** running the actual verifier script with mocked `journalctl` results gives:

| Identical corruption evidence | Service exit |
| --- | --- |
| No retained key | `1` |
| Retained key present, also rejects archive | `0` |
| Same PR with only the blanket exception removed | `1` |

Successful verification under an older key still works with the exception removed.

**Requested change:** keep retained-key retries, but leave unmatched archived-system failures fatal. The regression test should require failure when all available keys reject the archive and no archive-specific receipt explains it.

juliuskoskela:
### [P1] Propagate durability failures before reporting successful key generation

The key-swap helper claims both keys are durable before returning, but it suppresses errors from the operations that establish that guarantee:

- [Sealing-key flush](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/fss.nix#L907-L911).
- [Temporary verification-key flush](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/fss.nix#L921-L923).
- [Published file and directory flushes](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/fss.nix#L862-L867).

Execution then reaches the [unconditional success return](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/fss.nix#L929-L932).

**Reproduction:** I ran the actual helper with mocked key generation and injected a failure at each flush point separately. All four cases published the verification-key pathname and returned success. An experimental version with explicit error propagation returned failure in each case; the no-error control still succeeded.

The final directory-flush failure is the clearest concern: persistence of the rename has not been established, yet setup reports success. A subsequent crash may leave the expected verification key unavailable. This test demonstrates the error-handling defect, not physical data loss on hardware.

**Requested change:** propagate required flush failures through `durable_write` and its callers, avoid reporting successful durable completion, and define recovery for partially published key pairs. Add failure-injection coverage for these barriers.

Please use explicit error propagation rather than relying on `set -e`: the helper is invoked in a conditional, which suppresses Bash’s usual errexit behavior inside the function.

juliuskoskela:
### [P2] Preserve clock-change events instead of relying on net drift

The FIFO improves response time, but it does not close the fast backward/forward round-trip race.

The [watcher consumes the notification](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/common.nix#L371-L382), then decides whether to recover using only [net clock drift or a journald backward-jump message](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/common.nix#L384-L392). Receiving the FIFO event is not itself part of that decision.

A remaining failure sequence is:

1. The clock steps backward and then forward before the watcher gets scheduled.
2. The watcher consumes the queued notification.
3. Its samples show approximately zero net drift.
4. If journald did not record the intermediate backward step, no recovery check is triggered.

The [verification guard](https://github.com/tiiuae/ghaf/blob/378c36ed1c0f1e88574e23cf19f25aee4b128e5a/modules/common/logging/fss.nix#L1187-L1197) has the same limitation: opposite steps between its two samples can cancel, so zero net drift does not establish that the clock remained unchanged.

**Requested change:** preserve observed clock-change events as pending work, including across cooldowns, and invalidate verification evidence when an event occurs during collection. Triggering a recheck should remain separate from authorizing a destructive re-key.

**Regression test:** delay the watcher until both opposite steps have completed, provide queued notifications with zero net drift and no journald attestation, and assert that a recheck still occurs. Also test that opposite steps during verification invalidate that attempt.

This is a control-flow finding; I have not reproduced this scheduling sequence end-to-end on a VM.

rodrigopinotii:
Re: [P1] Keep unexplained archived-system corruption fatal (#issuecomment-5698433906)


Removed the `rekey_attested` downgrade branch in `fss_verify_policy_decision`: an
unreceipted archived-system failure now fails regardless of retained keys. `REKEY_ATTESTED`
is unchanged and still gates the retained-key rescue retry; `fss.nix` drops the now-unused
7th argument.

The test you flagged required the downgrade — it is inverted, not supplemented. It now
tampers the pre-re-key archive so it fails under every key including the retained one, with
receipt stores dropped, and requires `Journal integrity verification: FAILED`. Retained-key
rescue is asserted separately to still succeed and not degrade. Unit rows cover the same case.

rodrigopinotii:
Re: [P1] Propagate durability failures before reporting successful key generation (#issuecomment-5698537448)


All four barriers now propagate and fail closed: sealing-key flush, temp verification-key
flush, and `durable_write`'s file and directory flushes. Callers check and roll back. New
`KEY_DURABILITY_FAILED` covers an installed-but-undurable key on initial setup. Explicit
`if ! …; then return 1` throughout, no `set -e` reliance.

Recovery for partially published key pairs is in the PR description.

Injection rebuilt: `strace …:when=1` was ordinal, so it fired on whichever fsync came first
and could not target `.verification-key.new.$$` at all. Replaced by a path-matched
`LD_PRELOAD` shim, with each subtest asserting the exact path the shim reported failing.
All four barriers were shown to fail with their guard reverted.

Two pre-existing issues found doing this, neither introduced here: the old injection was
nondeterministic (two runs of an identical file, one pass one fail), and the suite aborted at
its second subtest on an apostrophe inside a `bash -lc '…'` comment — so subtests past that
point had never run. Comments only, no product effect. Fixed.

rodrigopinotii:
Re: [P2] Preserve clock-change events instead of relying on net drift (#issuecomment-5698555751)


The watcher now tracks whether a step woke it: an event with no drift or journald evidence
triggers a non-destructive verify recheck, and a cooldown-deferred action — setup restart or
recheck — is kept pending and retried next tick rather than dropped. The live-probe guard
compares a step-event count either side of the verify call, so an opposite-sign pair
cancelling to near-zero drift still invalidates the verdict. A recheck never writes an
attestation; a re-key still needs evidence.

Both regression tests are added, and both were shown to fail with the fix removed:

- Watcher stopped, two notifications queued, clock untouched, no journald jump — asserts a
  recheck happens and no attestation is written.
- A clock step during live verification invalidates that attempt, with a negative control
  for the clock holding still.

One substitution, stated rather than implied: the second drives the step-event counter rather
than real opposite steps, since the live probe completes in ~200 ms and real steps race that
window. Same code path, deterministic. The zero-net-drift round trip is separately covered
with real steps at ±20 s and ±1200 s, either side of `thresholdSeconds`.

rodrigopinotii:
A benign reboot that straddles a seal-interval boundary can produce a false integrity
failure. `SETUP_ARCHIVES_BEFORE` is snapshotted when `journal-fss-setup` starts; an archive
that `systemd-journal-flush` created earlier in the same boot is already in it, so
`record_setup_run_pre_activation_archives` skips it and it later fails as "archived system
journal failures outside allowlist". The skip is a deliberate security property — it is what
keeps a tampered pre-existing archive from being receipted — so this is not a request to
weaken it.

Pre-existing, and not introduced by this PR. It is reported here because removing the
downgrade branch makes it fatal where it was previously masked as a warning on any fleet
carrying retained keys.

Mechanism established from the log and the source, and induced once deliberately by aligning
a reboot to land within ~2.5 s of a seal-interval boundary. Two occurrences with an identical
signature — one unforced, one induced — both `tag/entry realtime timestamp out of
synchronization` giving `BAD_MESSAGE,ARCHIVED_SYSTEM` outside allowlist, 4.5 s and 6.5 s past
a wall-minute-aligned boundary.

Observed twice in 24 runs. Straddling the boundary appears necessary (0 of 14 non-straddling
runs) and raises the rate to about 1 in 5 (2 of 10), but does not make it deterministic — it
is not reliably reproducible on demand. The contingency alone does not reach significance;
the mechanism and the induced instance are what carry this, not the statistics.

Production exposure is roughly 15x lower than the suite's, since `sealInterval` defaults to
15 min against the test's 1 min.

rodrigopinotii:
One more disclosure, so it is not a surprise if you run the suite.

Two subtests in `logging-fss` fail intermittently. Both are pre-existing, both are test
scaffolding inheriting state from earlier subtests in a long cumulative suite, and neither
touches product code:

- `fss_verification.nix` "Activation live-probe retries through a live-write race" — 2 failures
  in 5 runs. Two distinct signatures seen: a diverged key pair, and corruption in the live
  `system.journal` during the deliberate write race. Not root-caused.
- "Deployed fss-test operator tool runs" — 1 failure in 2 runs since rebasing onto current
  `main`. It reports a corrupted archive left behind by an earlier subtest, so the tool is
  correct and the state is not.

They were invisible until now because the apostrophe bug aborted the suite at its second
subtest, so nothing past that point had ever run. Fixing that exposed them.

Not claiming the suite passes reliably. The evidence for the three review items above is the
red runs — each fix removed, the matching test observed failing, then restored — which does not
depend on the suite being clean end to end.

PR #2259, opened 2026-09-17
fix(orin): restore the default flashing image

vadika:
## Description of Changes

Running `flash-ghaf-host` without `-s` currently fails because fixed partition sizing removed the embedded image dependency. Include the matching image in the public wrapper again so the documented build-and-flash command works without a separate image argument.

Reserve `-s/--signed-sd-image` for an externally signed image and select Secure Boot enrollment through that option. Remove `--secure-boot` and update the instructions. Fixed partition sizing and the underlying firmware flasher stay independent of the image; building `*-flash-script` now also builds or downloads its default image.

### Type of Change

- [x] Bug fix
- [x] Documentation

## Checklist

- [x] Clear summary and a single signed-off commit
- [x] Flashing documentation updated
- [x] Formatting, REUSE and commit hooks passed
- [ ] Full `nix flake check` completed: stopped after a 180-second limit while still evaluating repository-wide configurations
- [ ] Reviewers assigned

### Applicable Targets

- [x] Orin AGX `aarch64`
- [x] Orin NX `aarch64` (shared wrapper; no NX hardware run)

### Installation Method

- [x] Full flash for hardware validation

### Test Steps To Verify

1. Build `nix build .#nvidia-jetson-orin-agx-debug-from-x86_64-flash-script -o flasher`.
2. Put the matching AGX in recovery mode and run `sudo ./flasher/bin/flash-ghaf-host` without `-s`.
3. Confirm boot from eMMC, the expected system closure, running VM services and Secure Boot disabled.
4. For signed flashing, supply the externally signed image output directory with `-s`; no additional Secure Boot flag is needed.

Validation: the no-`-s` path completed a full flash and boot on AGX 32GB, with the expected system closure, four active VM services and EFI `SecureBoot=0`. The pre-existing `ghaf-load-ftpm-module.service` failure remains. The subsequent removal of the redundant flag was rebuilt and checked offline; hardware flashing was not repeated for that removal.

Offline wrapper checks reproduced the original missing-image failure and covered the bundled default, signed overrides and enrollment selection, paths containing spaces, UKI flags, invalid/missing arguments, native option forwarding and the option separator. QSPI-only, LUKS/UKI and verity wrapper evaluations passed before the final flag removal. Signed-image boot and Secure Boot enrollment were not hardware-tested.


kajusnau:
I think it would be good to keep the original PR templates :slightly_smiling_face:  If they need updating to include/remove some stuff, it can be done.

emrahbillur:
Tested on Orin NX and also works on AGX 64Gb.

kajusnau (commented):
Would prefer to split the code changes and doc changes to respective `fix(orin)` and `docs(orin)` commits

TanelDettenborn (changes requested):
Orin NX/AGX pins `appPartitionSizeBytes` to a fixed value, which causes the `*-flash-script`targets to always use the "external" SD image. I also suspect that the NX + NVMe target is currently broken.

I have not done a deep analysis of the issue, but the visible problem appears to be in [partition-template.nix](https://github.com/tiiuae/ghaf/blob/main/modules/reference/hardware/jetpack/nvidia-jetson-orin/partition-template.nix#L250). Code checks whether `appPartitionSizeBytes` is set and, if it is set, then it assigns an empty value to `image_source_root`. This cause the following if-clause fail.

Additionally, the refactoring proposed in this PR fixes the `*-flash-script` target, but it also appears to break secure boot support for AGX and NX because the signed artifacts are no longer used.

I also agree with @kajusnau suggestion that this PR should be split into two/three separate changes: fix(orin). refactor(orin) and docs(orin).

PR #2229, opened 2026-09-08
bump(jetpack-nixos): rebase on upstream

TanelDettenborn:
* Jetpack-nixos input rebased against upstream
  - Includes Jetson BSP from 36.5 to 36.5.2
* Upstream kernel from 6.6.129 to 6.6.155
  - Includes a patch for OP-TEE (optee_notif_wait()-timeout)

### Type of Change

- [ ] New feature
- [x] Bug fix
- [x] Update
- [ ] Improvement / Refactor
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Checklist

- [ ] Clear summary in PR description
- [ ] Detailed and meaningful commit message(s)
- [ ] Commits are logically organized and squashed if appropriate
- [ ] [Contribution guidelines](https://github.com/tiiuae/ghaf/blob/main/CONTRIBUTING.md) followed
- [ ] Ghaf documentation updated with the commit - https://tiiuae.github.io/ghaf/
- [ ] Author has added reviewers and removed PR draft status

### Applicable Targets

- [ ] Orin AGX `aarch64`
- [x] Orin NX `aarch64`
- [ ] Intel Laptop `x86_64`
- [ ] Intel Laptop (low mem) `x86_64`

### Installation Method

- [x] Requires full re-installation
- [ ] Can be updated with `nixos-rebuild ... switch`
- [ ] Other:

### Test Steps To Verify:
Everything works as before, except:
- uefi version is 36.5.2
- kenel version is 6.6.155

PR #2260, opened 2026-09-18
fix(orin): enable fTPM and route EFI variables through OP-TEE

vadika:
## Description of Changes

Orin images on current main load the Linux fTPM driver without enabling the fTPM trusted applications in OP-TEE. Once those applications are enabled, concurrent EFI-variable access can leave fTPM initialization waiting indefinitely for an OP-TEE notification.

Expose `ghaf.hardware.nvidia.orin.optee.ftpm.enable` (default `true`) and `ftpm.unsecureInjectEPS.enable` (default `false`). EPS injection now requires an explicit internal-test opt-in, independently of runtime EK certificate provisioning, including on debug images.

Reuse JetPack's existing driver service without a Ghaf service override. Certificate services request and wait for `ftpm-driver`; disabling fTPM also disables these services.

Move the conditional Linux 6.6 EFI/TEE backports and fTPM kernel configuration to https://github.com/tiiuae/jetpack-nixos/pull/25. Pin its exact commit `0a182aaed7202f456de7a61087df12ce2b28e423`, based directly on the existing JetPack input. This PR includes the complete input update and remains independent of #2229, retaining BSP 36.5.0 / kernel 6.6.129.

### Type of Change

- [x] Bug fix

### Applicable Targets

- [x] Orin AGX `aarch64` — previous revision tested on AGX32
- [x] Orin NX `aarch64` — previous revision tested on NX 16 GB

### Installation Method

- [x] Requires full re-installation, including OP-TEE firmware

### Test Steps To Verify

1. On unfused internal-test boards only, explicitly set `ghaf.hardware.nvidia.orin.optee.ftpm.unsecureInjectEPS.enable = true;`. Build the AGX or NX debug image and its matching flash script, then flash both the system and platform firmware. Production firmware must leave this option disabled and use manufacturing provisioning.
2. Check that `ftpm-driver` succeeds, `/dev/tpmrm0` exists, and the kernel reports `Using TEE-based EFI runtime variable services`.
3. Run `tpm2_getrandom`, `tpm2_pcrread`, and reads of EK certificate NV indices `0x1c00002` and `0x1c0000a`.
4. Read EFI variables concurrently with fTPM driver reloads, then reboot and repeat TPM operations and certificate-hash checks.

### Review-refactor validation

- AGX and NX debug system closures and OP-TEE firmware built successfully from the pinned input.
- Debug/release defaults, explicit test EPS injection, fTPM disabled, OP-TEE disabled, and BSP-kernel backport gating evaluated successfully. Generated certificate units request and wait for JetPack's existing driver.
- The generated kernel configuration matches the previously tested revision. Default firmware omits `CFG_JETSON_FTPM_HELPER_INJECT_EPS`; explicit test opt-in reproduces the previously tested `tos.img` artifact.
- Formatting and REUSE licensing checks passed. The full `nix flake check` was interrupted at the 180-second limit while evaluating unrelated laptop targets; it is not a full-suite pass.

### Hardware baseline before the review refactor

These tests used test EPS injection. The refactored revision has not been reflashed; safe-default production provisioning remains unvalidated.

AGX32 validation passed cold and warm boots, five rounds of 100 concurrent EFI reads and driver reloads, TPM operations, and certificate persistence. TPM remained responsive after 12 hours 36 minutes on the same boot. All four VM services were active. These hardware checks used an unsigned debug image; fused production provisioning and Secure Boot are not qualified by this test.

NX 16 GB validation on BSP 36.5.0 / kernel 6.6.129 passed the full firmware/NVMe flash, cold boot, five rounds of 100 concurrent EFI reads and fTPM reloads, random generation, PCR reads, and both EK certificate reads. A subsequent power cycle preserved all certificate hashes; fTPM and all four VM services were active, with no failed services.

NX warm reboot did not complete and fell through to UEFI network boot; a power cycle restored normal operation. The cause remains undiagnosed.


modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -856,6 +821,10 @@ in
           && config.hardware.nvidia-jetpack.firmware.eksFile == "${firmwareEkbImage}/eks_t234.img"
         )
         "This release image bakes the all-zero development EKB (eks_t234.img); TPM-backed secrets are not protected by fused keys. Set hardware.nvidia-jetpack.firmware.eksFile to production key material.";
+    hardware.nvidia-jetpack.firmware.optee.ftpm = {
```
TanelDettenborn: I would prefer adding these as options to optee.nix, so that downstream projects can toggle ftpm related settings.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -856,6 +821,10 @@ in
...
+      enable = lib.mkDefault true;
+      unsecureInjectEPS.enable = cfg.runtimeEkProvision.enable;
```
TanelDettenborn: This needs to be set to false, or restricted only for internal testing purposes. This basically enables unlimited TPM EPS (SRK) provisioning from user space.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -952,6 +916,42 @@ in
           };
         }
       ]
+      ++
+        lib.optionals
+          (config.hardware.nvidia-jetpack.firmware.optee.ftpm.enable && cfg.kernelVersion == "upstream-6-6")
```
TanelDettenborn: I would move these patches to jetpack-nixos. Imight also consider `ftpm-config` patch.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -1155,29 +1155,13 @@ in
       };
     };
 
-    systemd.services.ghaf-load-ftpm-module = {
-      description = "Load fTPM module after stage-2 OP-TEE readiness";
+    systemd.services.ftpm-driver = mkIf config.hardware.nvidia-jetpack.firmware.optee.ftpm.enable {
```
TanelDettenborn: This systemd service is not needed, as it is already implemented within jetpack-nixos.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -1210,7 +1194,7 @@ in
         "local-fs.target"
         "systemd-modules-load.service"
         "tee-supplicant.service"
```
TanelDettenborn: Redundant. ftpm driver is loaded after tee-supplicant.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
```diff
@@ -1188,7 +1172,7 @@ in
...
```
TanelDettenborn: Redundant. ftpm driver is loaded after tee-supplicant.

TanelDettenborn (changes requested):

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Addressed in 1a21d240. Added `ghaf.hardware.nvidia.orin.optee.ftpm.enable` (default true) and `ftpm.unsecureInjectEPS.enable` (default false) in Ghaf's `optee.nix`, forwarding them to JetPack. Disabling fTPM also removes the certificate services. Enabled/disabled configuration evaluations passed.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Addressed in 1a21d240. EPS injection is now disabled by default in both debug and release configurations, independently of runtime EK certificate provisioning. Internal testing requires the explicit `ghaf.hardware.nvidia.orin.optee.ftpm.unsecureInjectEPS.enable = true;` opt-in. Built firmware confirms that the default omits `CFG_JETSON_FTPM_HELPER_INJECT_EPS`; the opt-in reproduces the previously tested firmware artifact. The PR now distinguishes those earlier hardware tests from validation of this refactor.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Moved both the four conditional EFI/TEE backports and the remaining `ftpm-config` settings into https://github.com/tiiuae/jetpack-nixos/pull/25. Ghaf pins its exact commit `0a182aaed7202f456de7a61087df12ce2b28e423`, directly based on the existing input, so this remains independent of #2229. AGX/NX system builds passed and the generated kernel configuration matches the tested revision.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Removed the Ghaf `ftpm-driver` extension in 1a21d240. The certificate services now use `Wants=ftpm-driver.service` and `After=ftpm-driver.service`, so they start JetPack's existing service without duplicating its definition or ordering. Verified the generated units in the successful AGX/NX builds.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Removed in 1a21d240. The export service requests and waits for `ftpm-driver.service`, which already requires and follows `tee-supplicant.service`. Also removed the transitively covered local-filesystem and module-loading ordering.

modules/reference/hardware/jetpack/nvidia-jetson-orin/jetson-orin.nix
vadika: Removed in 1a21d240. The provisioning service requests and waits for `ftpm-driver.service`, inheriting its supplicant, filesystem, and module-loading ordering.

emrahbillur:
There is a restructuring of the option definitions for fTPM but still is unique. The workaround solves the fTPM issue. Tested on NX, AGX and AGX64 and can be merged. 

But I still agree with Tanel that new kernel with Jetpack bump will solve this issue directly. 

Merging this PR will solve the issue for now.

